PR #932, opened 2024-02-07
preprocessing(specification): Update error schema

theosanderson:
See discussion here: https://loculus.slack.com/archives/C05G172HL6L/p1707326370868269?thread_ts=1707323971.457599&cid=C05G172HL6L

### DIFF
--- a/preprocessing/specification.md
+++ b/preprocessing/specification.md
@@ -97,11 +97,11 @@ The `errors` and `warnings` fields contain an array of objects of the following 
 
 ```js
 {
-    source: [{
+    unprocessedFields: [{
         type: "Metadata" | "NucleotideSequence",
         name: string
     }[],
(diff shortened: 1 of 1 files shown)


corneliusroemer (approved):
I think we can just go with this and adjust during implementation if we find issues.

Field names aren't super important as this is all pretty internal-API stuff, which won't be visible to end users.

theosanderson:
Cornelius Anna and I just discussed and agreed on `unprocessedFields` and `processedFields` and that this is a good idea (though of course anyone is welcome to disagree).

Next task would be to implement it!

anna-parker:
I can try add this later today!

anna-parker:
Closing as this has now been merged in https://github.com/loculus-project/loculus/pull/3348

PR #1080, opened 2024-02-22
dev: recommend extensions that are essentially required for this monorepo

corneliusroemer:
### Summary
Recommend extensions that are useful for this monorepo, saving time for new devs who use vscode.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 !.idea/runConfigurations
 *.iml
 .vscode/*
+!.vscode/extensions.json
 
 # build output
 dist/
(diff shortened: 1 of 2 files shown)


theosanderson (approved):

theosanderson:
This feels like one that can either be merged or closed - doesn't matter to me either way

fhennig:
Comments in JSON? 🤔 

I saw it upstream too, but that isn't really allowed usually, isn't it?

corneliusroemer:
>Comments in JSON? 🤔
>I saw it upstream too, but that isn't really allowed usually, isn't it?

See https://code.visualstudio.com/docs/languages/json#_json-with-comments

fhennig:
Ahh thanks!

PR #1168, opened 2024-02-28
chore: sketch testing concepts in CONTRIBUTING.md

fengelniederhammer:
### Summary
This is a point that came up in our retro. I tried to sketch different test concepts to come to an agreement on how much and what we want to test. Feedback is very welcome!


fengelniederhammer:
There hasn't been any movement for a long time and this is probably outdated anyway.

PR #1634, opened 2024-04-20
feat(website): show a meaningful error if uploaded file is too large

### LINKED ISSUE #1226: Show meaningful error and not "Network error" when uploading too large file
Related to https://github.com/loculus-project/loculus/issues/996

When I try to submit the 7200 mpox sequences mentioned in https://github.com/loculus-project/loculus/issues/996 (1-2GB), I get a non-meaningful error presented by the website: 'Received unexpected message from backend: "Network Error"'

We should present a more precise error, something like 

> Sequence file too large if uncompressed (>1GB), please make multiple smaller submissions

image

chaoran-chen:
resolves #1226
preview URL: https://1226-file-too-big-error.loculus.org

### PR Checklist

**This is blocked by #1635 because the web application cannot see the response status due to CORS errors.**

### DIFF
--- a/backend/src/main/resources/application.properties
+++ b/backend/src/main/resources/application.properties
@@ -6,8 +6,8 @@ springdoc.default-consumes-media-type=application/json
 springdoc.api-docs.path=/api-docs
 springdoc.swagger-ui.operationsSorter=alpha
 server.forward-headers-strategy=framework
-spring.servlet.multipart.max-file-size=800MB
-spring.servlet.multipart.max-request-size=800MB
+spring.servlet.multipart.max-file-size=200MB
+spring.servlet.multipart.max-request-size=200MB
 
 
 server.compression.enabled=true


website/src/components/Submission/DataUploadForm.tsx
```diff
@@ -37,6 +37,9 @@ type DataUploadFormProps = {
 
 const logger = getClientLogger('DataUploadForm');
 
+// This should be the same value as defined in the backend.
+const MAX_UPLOAD_SIZE = '800 MB';
+
```
fhennig, line 42: Can you be more specific; is it this property? `spring.servlet.multipart.max-file-size`?

Also, it looks like the PR sets this property to 200MB, so they are not the same.

fhennig (changes requested):
Nice small fix! Just the docs and numbers are a bit off.

The duplicated number is a bit brittle; maybe we can also add a doc-string in the application.properties to update the UI?

Or could we even have this in our values.yaml and have it forwarded to both configs?

PR #2332, opened 2024-07-24
feat(ena-submission): Create ena assembly

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/2397

preview URL: https://create-ena-assembly.loculus.org/

### Summary

Uses same principal as [create_ena_projects](https://github.com/loculus-project/loculus/pull/2293) to keep submission state in DB - please review that PR first :-) 

### Summary

This adds the following rule to the ena-submission snakemake file :
- `create_assembly` rule . This function will continuously (in a loop) scan for new sequences where a sample needs to be created and trigger their creation. It will also update both the `submission_table` and the `assembly_table`. 
- In contrast to create_project and create_sample we need to use the `webin-cli.jar` for assembly submission. This had to be added to the docker image (and its dependencies).
- In contrast to create_project and create_sample we also need to wait for assemblies to be approved and only then we will receive the assembly accessions: https://ena-docs.readthedocs.io/en/latest/submit/assembly/genome.html#assigned-accession-numbers This means there is one additional state in the assembly_table: `WAITING`.
- It also adds some basic unit tests for sub-functions used by create_assembly. 

#### High level overview of `assembly_creation`:
In a loop:
1. Get sequences in `submission_table` in state SUBMITTED_SAMPLE
- if (there exists an entry in the `sample_table` for the corresponding (accession, version)):
-- if (entry is in status SUBMITTED): update `submission_table` to SUBMITTED_ASSEMBLY.
-- else: update `submission_table` to SUBMITTING_ASSEMBLY.
- else: create assembly entry in `assembly_table` for (accession, version).
2. Get sequences in `submission_table` in state SUBMITTING_SAMPLE
- if (corresponding `assembly_table` entry is in state SUBMITTED): update entries to state SUBMITTED_ASSEMBLY.
3. Get sequences in `assembly_table` in state READY, prepare files: we need chromosome_list, fasta files and a manifest file, set status to WAITING
- if (submission succeeds): set status to WAITING and fill in results: ena-internal "erz_accession"
- else: set status to HAS_ERRORS and fill in errors
4. Get sequences in `assembly_table` in state WAITING, every 5minutes (to not overload ENA) check if  ENA has processed the assemblies and assigned them "gca_accession". If so update the table to status SUBMITTED and fill in results
[...]

PR #2417, opened 2024-08-12
feat(ena-submission): Upload ena submission results to Loculus

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/2424

preview URL: https://upload-ena-submission-res.loculus.org/

### Summary
1. This PR encompasses the entire ena-submission process, for details see: [ena-submission/README.md](https://github.com/loculus-project/loculus/pull/2417/files#diff-fbb6131a0af34856a7dfa00148392503ea6c10bcb83279c948c4dab00323e2d3).

2. It adds database connection pooling, otherwise I get this error:
image
caused by Exhaustion of Local Ports: "When your Python application makes connections to the database, it uses a local (ephemeral) port on your machine. If your application opens many connections in a short period and doesn’t properly close them, you might run out of available local ports." - it might be possible to fix this just with better closing of port connections... I checked and the postgresDB can support `SHOW max_connections;` = 100. After the ena-submission-pod jobs were killed there were still `SELECT count(*) FROM pg_stat_activity;` = 20 connections to the db. Probably from the backend. I wasn't closing db updates (this would explain ~20-30 additional connections).

3. It also fixes an issue in the backend when merging external metadata with other metadata fields, an alternative solution could be: 

```sql
create or replace function jsonb_concat(a jsonb, b jsonb) 
returns jsonb
language plpgsql
immutable
parallel safe
as $$
declare
    result jsonb;
begin
    -- Merge the two JSON objects
    result := a || jsonb_object_agg(key, 
                case 
                    when a->key is null or a->key = 'null' then b->key 
                    else a->key 
                end)
                from jsonb_each_text(b) as elem(key, value);
    return result;
end;
$$;
```
and use this like:
```sql
case 
       when all_external_metadata.external_metadata is null then jsonb_build_object('metadata', (sequence_entries_preprocessed_data.processed_data->'metadata'))
        else jsonb_build_object('metadata', (sequence_entries_preprocessed_data.processed_data->'metadata') || all_external_metadata.external_metadata)
    end as joint_metadata
```

### PR Checklist

- [x] Confirm this works locally from uploading new sequences, getting the sequences from the slack notification and uploading to the github repo to appearing on the page of the uploaded sequence: 
image
Note Biosamples was down so I had to artificially add the values in.
- [x] Confirm this works on a preview instance: test full cycle from uploading sequences to the preview instance. Similar to in https://github.com/loculus-project/loculus/pull/2332 I had to log into the database via port forwarding to set the sample results to existing entries: 
```
{
        "ena_sample_accession":  "ERS20170050",
        "biosample_accession": "SAMEA115670243",
        "ena_submission_accession": "ena_submission_accession"
    }
````
(as biosamples is currently down), and then did the same for the erz_accession (this was successful but to not have to wait I set this to an existing erz accession). This finishes the submission cycle and then the results are sent to loculus: 
image
and are seen on the website (with links that work except for sra... it seems like this isn't actually an sra accession??):

image
now without sra_run_accession: image
- [x] Check this works for a single segmented organism

### CODE REVIEW

File: ena-submission/scripts/create_project.py [hunk 1]
```diff
@@ -0,0 +1,380 @@
...
+            group_info = get_group_info(config, row["group_id"])[0]["group"]
+        except Exception as e:
+            logger.error(f"Was unable to get group info for group: {row["group_id"]}, {e}")
+            continue
+
+        project_set = construct_project_set_object(group_info, config, row, test=True)
+        update_values = {
+            "status": Status.SUBMITTING,
+            "started_at": datetime.now(tz=pytz.utc),
+            "center_name": group_info["institution"],
```
chaoran-chen, line 249: Is a SQL injection possible through the `institutions` field? Because it is passed on into `update_db_where_conditions` which uses string concatenation to construct the SQL.

chaoran-chen (dismissed):
I think that there is at least one possible way for an SQL injection.

There are multiple places where a function constructs an SQL query via string concatenation using values passed into the function without escaping and validation. This is very dangerous and even for places where an attack is currently not possible because the input values are all under our control (e.g., the `organism` field), this could potentially lead to future vulnerabilities.

ena-submission/scripts/call_loculus.py
```diff
@@ -77,42 +74,39 @@ def make_request(
...
                 response = requests.post(
-                    url, headers=headers, json=json_body, params=params, data=data
+                    url, headers=headers, json=json_body, params=params, data=data, timeout=60
                 )
         case _:
-            raise ValueError(f"Unsupported HTTP method: {method}")
+            msg = f"Unsupported HTTP method: {method}"
+            raise ValueError(msg)
+
+    if response.status_code != HTTPStatus.OK:
```
chaoran-chen, line 97: Is it intentional to not use `response.ok`?

ena-submission/scripts/call_loculus.py
```diff
@@ -115,11 +119,14 @@ def submit_external_metadata(metadata_file, config: Config, organism: str) -> re
...
-    data = " ".join(pre_ndjson)  # Q: Space not newline?
+    data = json.dumps(external_metadata)
+
+    response = make_request(HTTPMethod.POST, url, config, data=data, headers=headers, params=params)
 
-    return make_request(HTTPMethod.POST, url, config, data=data, headers=headers, params=params)
+    if not response.ok:
+        response.raise_for_status()
+
+    return response
```
chaoran-chen, line 129: Huh? Why this change? I thought that `make_request` will already raise an error if the response is not ok.

ena-submission/scripts/create_assembly.py
```diff
@@ -0,0 +1,527 @@
...
+    logger.info(f"Config: {config}")
+
+    db_config = db_init(config.db_password, config.db_username, config.db_host)
+    slack_config = slack_conn_init(
+        slack_hook_default=config.slack_hook,
+        slack_token_default=config.slack_token,
+        slack_channel_id_default=config.slack_channel_id,
+    )
+
+    while True:
```
chaoran-chen, line 517: Would it make sense to sleep a bit between the iterations?

ena-submission/scripts/ena_submission_helper.py
```diff
@@ -144,10 +159,266 @@ def get_project_xml(project_set):
...
+    https://ena-docs.readthedocs.io/en/latest/submit/fileprep/assembly.html#fasta-file
+    """
+    with tempfile.NamedTemporaryFile(delete=False, suffix=".fasta.gz") as temp:
+        filename = temp.name
+
+    with gzip.GzipFile(filename, "wb") as gz:
+        if len(unaligned_sequences.keys()) == 1:
+            entry = chromosome_list.chromosomes[0]
+            gz.write(f">{entry.object_name}\n".encode())
+            gz.write(f"{unaligned_sequences["main"].replace('-', 'N')}\n".encode())
```
chaoran-chen, line 261: Sorry for bringing this up if it has been discussed already and it was decided to be the right way. I'm just a bit surprised to see a gap being replaced by an `N`. Wouldn't it make more sense to simply remove the `-` as they represent emptiness?

ena-submission/scripts/ena_submission_helper.py
```diff
@@ -144,10 +159,266 @@ def get_project_xml(project_set):
...
+        f.write(f"ASSEMBLY_TYPE\t{manifest.assembly_type!s}\n")
+        f.write(f"COVERAGE\t{manifest.coverage}\n")
+        f.write(f"PROGRAM\t{manifest.program}\n")
+        f.write(f"PLATFORM\t{manifest.platform}\n")
+        f.write(f"FASTA\t{manifest.fasta}\n")
+        f.write(f"CHROMOSOME_LIST\t{manifest.chromosome_list}\n")
+        if manifest.description:
+            f.write(f"DESCRIPTION\t{manifest.description}\n")
+        if manifest.moleculetype:
+            f.write(f"MOLECULETYPE\t{manifest.moleculetype!s}\n")
```
chaoran-chen, line 294: Is this safe? It looks like a potential source for an injection. Should we perform some escaping or validations (e.g., check that user-specified fields such as platform/sequencingProtocol do not contain forbidden/problematic characters)?

ena-submission/scripts/create_assembly.py
```diff
@@ -0,0 +1,527 @@
...
+
+def create_manifest_object(
+    config: Config,
+    sample_table_entry: dict[str, str],
+    project_table_entry: dict[str, str],
+    submission_table_entry: dict[str, str],
+    seq_key: dict[str, str],
+    group_key: dict[str, str],
+    test=False,
+) -> str:
```
chaoran-chen, line 108: I think the result type is wrong, this function is returning a `AssemblyManifest`, isn't it?

ena-submission/scripts/ena_submission_helper.py
theosanderson: Or to handle this outside ENA submission - IMO Loculus should never contain raw sequences with `-`s

chaoran-chen:
Thanks @anna-parker for this huge piece of work!! I read (rather briefly) through the code and commented whenever I noticed something. I found a few vulnerabilities that should be fixed. Otherwise, it looks great for now!

backend/src/main/resources/db/migration/V1.2__update_view.sql
```diff
@@ -0,0 +1,46 @@
+drop view if exists  external_metadata_view cascade;
```
theosanderson, line 1:
```suggestion
drop view if exists external_metadata_view cascade;
```

ena-submission/README.md
```diff
@@ -14,34 +16,80 @@ The ENA submission pod creates a new schema in the loculus DB, this is managed b
    postgres:latest
 ```
 
-In our kubernetes pod we run flyway in a docker container, however when running locally it is best to [download the flyway CLI](https://documentation.red-gate.com/fd/command-line-184127404.html).
+### Install and run flyway
 
-You can then run flyway using the
+In our kubernetes pod we run flyway in a docker container, however when running locally it is [download the flyway CLI](https://documentation.red-gate.com/fd/command-line-184127404.html) (or `brew install flyway` on macOS).
```
theosanderson, line 21:
```suggestion
In our kubernetes pod we run flyway in a docker container, however when running locally it is best to [download the flyway CLI](https://documentation.red-gate.com/fd/command-line-184127404.html) (or `brew install flyway` on macOS).
```

ena-submission/README.md
```diff
@@ -14,34 +16,80 @@ The ENA submission pod creates a new schema in the loculus DB, this is managed b
...
+
+It is then the reviewer's turn to review these sequences. [TODO: define review criteria] If these sequences meet our criteria they should be uploaded to [pathoplexus/ena-submission](https://github.com/pathoplexus/ena-submission/blob/main/approved/approved_ena_submission_list.json) (currently we read data from the [test folder](https://github.com/pathoplexus/ena-submission/blob/main/test/approved_ena_submission_list.json) - but this will be changed to the `approved` folder in production). The `trigger_submission_to_ena` rule is constantly checking this folder for new sequences and adding them to the submission_table if they do not exist. Note we cannot yet handle revisions so these should not be added to the approved list [TODO: do not allow submission of revised sequences in `trigger_submission_to_ena`]- revisions will still have to be performed manually.
```
theosanderson, line 84:
```suggestion
It is then the reviewer's turn to review these sequences. [TODO: define review criteria] If these sequences meet our criteria they should be uploaded to [pathoplexus/ena-submission](https://github.com/pathoplexus/ena-submission/blob/main/approved/approved_ena_submission_list.json) (currently we read data from the [test folder](https://github.com/pathoplexus/ena-submission/blob/main/test/approved_ena_submission_list.json) - but this will be changed to the `approved` folder in production). The `trigger_submission_to_ena` rule is constantly checking this folder for new sequences and adding them to the submission_table if they are not already there. Note we cannot yet handle revisions so these should not be added to the approved list [TODO: do not allow submission of revised sequences in `trigger_submission_to_ena`]- revisions will still have to be performed manually.
```

theosanderson: Can we raise an Issue for the second TODO before merging this

ena-submission/scripts/create_project.py
anna-parker: I have modified the function to now always use `%s` for passing inputs and validating the name.

ena-submission/scripts/create_project.py
```diff
@@ -0,0 +1,384 @@
...
+                    conditions=group_key,
+                    update_values=update_values,
+                )
+                tries += 1
+            if number_rows_updated == 1:
+                logger.info(f"Project creation for group_id {row["group_id"]} succeeded!")
+        else:
+            update_values = {
+                "status": Status.HAS_ERRORS,
+                "errors": json.dumps(project_creation_results.errors),
```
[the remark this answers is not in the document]
anna-parker: I think this should now be resolved with the `%s`

ena-submission/scripts/submission_db_helper.py
```diff
@@ -122,10 +135,14 @@ def find_conditions_in_db(db_conn_pool, table_name, conditions):
     con = db_conn_pool.getconn()
     try:
         with con, con.cursor(cursor_factory=RealDictCursor) as cur:
-            query = f"SELECT * FROM {table_name} WHERE "
-            query += " AND ".join([f"{key}='{str(value)}'" for key, value in conditions.items()])
+            # Prevent sql-injection with table_name validation
+            TableName.validate(table_name)
+            query = f"SELECT * FROM {table_name}"
+
+            where_clause = " AND ".join([f"{key}=%s" for key in conditions])
```
chaoran-chen, line 142: I think that this implementation is already entirely safe for now, but to make the function even more future-proof, it would be good to validate the keys as well. A simple implementation, where you wouldn't need to list all possible keys, and that should be enough for this case, could be to check that the key only consists of letters (A-Za-z), numbers, and underscores.

ena-submission/scripts/submission_db_helper.py
```diff
@@ -139,12 +156,16 @@ def find_errors_in_db(db_conn_pool, table_name, time_threshold=15):
...
-            query += f"(status='HAS_ERRORS' AND started_at < timestamp '{min_start_time}')"
-            query += f" OR (status='SUBMITTING' AND started_at < timestamp '{min_start_time}')"
+            query = f"""
+                SELECT * FROM {table_name} 
+                WHERE (status = %s AND started_at < %s)
+                OR (status = %s AND started_at < %s)
+            """
 
-            cur.execute(query)
+            cur.execute(query, ("HAS_ERRORS", min_start_time, "SUBMITTING", min_start_time))
```
chaoran-chen, line 168: To improve readability, the constants such as `"HAS_ERRORS"` can be moved directly into the SQL. Otherwise, this looks great and safe!

ena-submission/ENA_submission.md
```diff
@@ -35,7 +35,7 @@ We require the following components:
 
 - Analysis: An analysis contains secondary analysis results derived from sequence reads (e.g. a genome assembly).
 
-At the time of writing (October 2023), in contrast to ENA, Pathoplexus has no hierarchy of study/sample/sequence: every sequence is its own study and sample. Therefore we need to figure out how to map sequences to projects, each submitter could have exactly _one_ study pre organism (this is the approach we are currently taking), or each sequence could be associated with its own study.
+At the time of writing (October 2023), in contrast to ENA, Pathoplexus has no hierarchy of study/sample/sequence: every sequence is its own study and sample. Thus, each sequence will have to be submitted to ENA as its own study and sample. Alternatively, each submitter could have exactly _one_ study pre organism (this is the approach we are currently taking).
```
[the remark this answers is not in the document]
anna-parker: I have updated with current state

ena-submission/scripts/ena_submission_helper.py
anna-parker: I removed '-' from the example data: https://github.com/pathoplexus/example_data/pull/4 - I will add a prepro check before removing this code.

ena-submission/scripts/call_loculus.py
anna-parker: Yes we can just use that - I changed now to be more consistent

ena-submission/README.md
anna-parker: I added a check for this in the code to prevent us from accidentally submitting revisions :-) there is already an issue on figuring out how to handle revisions to ENA: https://github.com/loculus-project/loculus/issues/2428

ena-submission/scripts/submission_db_helper.py
anna-parker: I added a function to perform an explicit check that all keys in the conditions dictionary are actual columns.

ena-submission/scripts/ena_submission_helper.py
anna-parker: I am not quite sure what you mean - this function produces a manifest file and isn't added to the postgres database. The file is sent to ENA using their webin-cli so I don't think there is any injection risk here? But maybe I don't understand?

corneliusroemer (approved):
I'll review as we go - let's try this out

ena-submission/scripts/ena_submission_helper.py
anna-parker: created a follow up issue: https://github.com/loculus-project/loculus/issues/2831

ena-submission/scripts/create_assembly.py
anna-parker: created a follow up issue: https://github.com/loculus-project/loculus/issues/2831

ena-submission/scripts/create_project.py
```diff
@@ -0,0 +1,384 @@
...
+    ena_submission_username: str
+    slack_hook: str
+    slack_token: str
+    slack_channel_id: str
+
+
+def construct_project_set_object(
+    group_info: dict[str, str],
+    config: Config,
+    entry: dict[str, str],
```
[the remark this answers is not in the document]
anna-parker: created a follow up issue: https://github.com/loculus-project/loculus/issues/2831

anna-parker:
@chaoran-chen sorry I had to dismiss your review to merge this (but I did address and fix the issues you found), from discussion with @rneher and @corneliusroemer we will merge this now and work on https://github.com/loculus-project/loculus/issues/2832 in a separate PR.

PR #2513, opened 2024-08-22
wip: group fields in selector

theosanderson:
resolves #

preview URL: https://groupfieldsinselecto.loculus.org

### PR Checklist


anna-parker:
@theosanderson sorry removing the preview flag as we are low on argocd resources

fhennig:
Ohh cool! I like it. But _maybe_ it's a bit annoying to search for a field? I can't just scan through the fields, I have to unfold all the groups. This makes me think that maybe a collapse/open all button would be cool, or maybe they are all open by default (maybe also annoying?). So maybe the collapse/open all button would be good.

Further (but probably too much for this PR) it would be nice to have a filter field for the names. I could just type in 'auth' and would get the author fields. But maybe that's for another time.

PR #2634, opened 2024-08-27
fix(website): Do not download unaligned (all N) sequences when segment does not exist

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/2679

preview URL: https://second-hotfix.loculus.org/

### Summary

Example `PP_000PMPM.1` (submissionId `AB069669.1.M`) - does not contain S segment, in the all unaligned sequences of segment S download the accession is not found, but it is in when downloading all aligned segments - there it is all Ns...

### Screenshot

Using this preview example the submissionId now maps to LOC_000PGUG.1 and is no longer in the downloaded aligned sequences. 
I also checked and there are a total of 726 sequences in the file: 
image
as there should be

### PR Checklist


theosanderson:
Thanks @anna-parker. I don't think the current behaviour is too urgently bad in that downloading aligned sequences won't be that common, and the data isn't erroneous. I think we should discuss whether to handle this in Loculus or LAPIS.

fengelniederhammer:
We just discussed this and we want to change SILO: https://github.com/GenSpectrum/LAPIS-SILO/issues/547

This is not an immediate priority in SILO however, so it might take a bit until we get to implement it. Also it's not really clear yet how much effort the change would be.

anna-parker:
I see that https://github.com/GenSpectrum/LAPIS-SILO/issues/547 is in backlog - so do we want to add this until it is fixed?

chaoran-chen (changes requested):
As the `length` fields are created by the pipeline and are instance specific, they don't exist for all Loculus instances. We merged the hotfix for Pathoplexus but for instances that do not have the field, it will break download entirely (and for instances that have the field but use it in a different way, it can lead to undesired behavior). If we want such a hotfix, it has to be configurable - but I think that the issue is not important enough to do it at the moment.

theosanderson:
To elaborate, `NNNNN` is a valid result for these segments. It indicates that we don't know the base at any position. It's not, IMO, the ideal behaviour – and so there's a plan to change it. But it's not a big problem.

anna-parker:
Ok - I don't like having sequences in the aligned and not in the unaligned files - I could see users having issues with this. But I agree its not a major problem and its technically not wrong to have all Ns. I will close this PR - we have the issue to keep track of the bug.

PR #2641, opened 2024-08-27
feat(deploy.py): Allow reconfiguring the live host pointed to

theosanderson:
We will need to start pointing at `demo.pathoplexus.org` instead of `main.loculus.org` for pathoplexus local website dev as Loculus and Pathoplexus configurations diverge a bit. This will make that possible.

Haven't tried it yet.


corneliusroemer (approved):
haven't checked, I trust 😀

PR #2649, opened 2024-08-28
feat(preprocessing): add boolean case handling in processing functions

fengelniederhammer:
preview URL: https://addbooleanpreprocessing.loculus.org/

### Summary

I'd like to be able to get boolean values out of the preprocessing pipeline.
I also had to adjust the website a little so that filtering by and displaying boolean values works correctly.

### PR Checklist


theosanderson:
Cool - code seems good to me! We might potentially want to block on merging this until we've merged https://github.com/loculus-project/loculus/pull/2641 and established that we can work on Pathoplexus locally by pointing at the Pathoplexus demo server. (Atm we would still point at the loculus main server - we need to pull those apart anyway, but until we've done so we may not want to make many further changes that would break the similarity between the Pathoplexus Values file and the default Loculus values file).

fengelniederhammer:
This would also quasi the LAPIS config - and make it incompatible with the existing config. Is this something that the live instance can handle?
(I'm not sure how incompatible it would be - maybe it could still work)

fengelniederhammer:
I'd like to use the boolean preprocessing in our GenSpectrum installation - if it's too risky changing the LAPIS schema, should I split the changes?
* adding the boolean function adaption and the website changes -> should be safe to merge (this PR tested it and it will be unused in Pathoplexus)
* changing isLabHost to type boolean (the schema breaking part - to be merged when we're sure what will happen)

theosanderson:
Pathoplexus now overwrites the `organisms` in the Values file so shouldn't be affected by changes there - so this should be OK - apart from in the context I mentioned. But it's a good instinct to have - there are lots of other places where currently editing them would currently create a change that would break pathoplexus once Loculus was updated.

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -405,18 +405,34 @@ def identity(
...
                 case "int":
                     output_datum = int(input_datum)
                 case "float":
                     output_datum = float(input_datum)
+                case "boolean":
+                    if input_datum.lower() == "true":
+                        output_datum = True
+                    elif input_datum.lower() == "false":
+                        output_datum = False
+                    else:
```
anna-parker, line 421: I would potentially not default to False but default to None (unless the field is required it should be ok it set it to None)

anna-parker (approved):
lgtm! checked the code and the preview and all looks good!

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
fengelniederhammer: SILO anyway always accepts `null`, so I think it's a good idea. I'll change it.

PR #2653, opened 2024-08-28
[don't merge without prior discussion] fix: don't sort author lists, except for segment grouping purposes

### LINKED ISSUE #2650: Author list order is not preserved when ingested
Loculus does not preserve author list. For example:

https://pathoplexus.org/seq/PP_000PLLU.1

```
A. Amuri-Aziza, A. Ayouba, A. Legand, A. Rajan, A. Whitesell, A.J. Porzucek, C. Pratt, C.N. Mores, D. Mukadi-Bamuleka, E. Delaporte, E. Kinganda-Lusamaki, E. Lokilo-Lofiko, E. Pukuta-Simbu, E. Shedroff, E. Talundzic, F. Mambu-Mbika, F. Mbuyi, F. Mulangu, F. Muyembe-Mawete, G. Makaya, G. Tshapenda, G.L.T. Folefack, G.N. Soke, H. Kavunga-Membo, ..., T. Shoemaker
```

https://www.ncbi.nlm.nih.gov/nuccore/OR084897.1
```
Kinganda-Lusamaki,E., Whitmer,S., Lokilo-Lofiko,E., Amuri-Aziza,A.,Muyembe-Mawete,F., Makangara-Cigolo,J.C., Makaya,G., Mbuyi,F.,Whitesell,A., Kallay,R., Choi,M., Pratt,C., Mukadi-Bamuleka,D.,Kavunga-Membo,H., Matondo-Kuamfumu,M., Mambu-Mbika,F.,Ekila-Ifinji,R., Shoemaker,T., Stewart,M., Eng,J., Rajan,A.,Soke,G.N., Fonjungo,P.N., Otshudiema,J.O., Folefack,G.L.T.,Pukuta-Simbu,E., Talundzic,E., Shedroff,E., Bokete,J.L., Legand,A.,Formenty,P., Mores,C.N., Porzucek,A.J., Tritsch,S.R., Kombe,J.,Tshapenda,G., Mulangu,F., Ayouba,A., Delaporte,E., Peeters,M.,Wiley,M.R., Montgomery,J.M., Klena,J.D., Muyembe-Tamfum,J.-J.,Ahuka-Mundeke,S. and Mbala-Kingebeni,P.
```

IMO it's best to preserve the order of authors when ingesting from NCBI.

corneliusroemer:
resolves #2650

preview URL: https://author-order.loculus.org

### Summary

We started sorting author lists when it turned out that different segments of the same isolate could have different author orders. We should have done the sorting only for grouping purposes and not when submitting to Loculus.

This PR fixes that by only sorting for grouping purposes.

Note: this will cause ingest to submit revisions for existing deployments as the ingest hash changes.

### DIFF
--- a/ingest/scripts/call_loculus.py
+++ b/ingest/scripts/call_loculus.py
@@ -483,7 +483,7 @@ def submit_to_loculus(
     if mode == "get-submitted":
         logger.info("Getting submitted sequences")
         response = get_submitted(config)
-        Path(output).write_text(json.dumps(response), encoding="utf-8")
+        Path(output).write_text(json.dumps(response, indent=4, sort_keys=True), encoding="utf-8")
 
 
 if __name__ == "__main__":
--- a/ingest/scripts/group_segments.py
+++ b/ingest/scripts/group_segments.py
@@ -109,11 +109,11 @@ def main(
     # These are the fields that are expected to be identical across all segments for a given isolate
 
     # Dynamically determine the fields that are present in the metadata
-    some_row = next(iter(segment_metadata.values()))
-    if not some_row:
+    first_row = next(iter(segment_metadata.values()))
+    if not first_row:
         msg = "No data found in metadata file"
         raise ValueError(msg)
-    all_fields = sorted(some_row.keys())
+    all_fields = sorted(first_row.keys())
     logger.debug(f"All metadata fields: {all_fields}")
 
     # Metadata fields can vary between segments w/o indicating being from different assemblies
@@ -134,7 +134,7 @@ def main(
 
     equivalence_classes: EquivalenceClasses = defaultdict(lambda: defaultdict(list))
     for accession, values in segment_metadata.items():
-        # Author order sometimes among segments from same isolate
+        # Author order sometimes varies among segments from same isolate
         # Example: JX999734.1 (L) and JX999735.1 (M)
         modified_values = values_with_sorted_authors(values)
         group_key = str(
--- a/ingest/scripts/group_segments.py
+++ b/ingest/scripts/group_segments.py
@@ -134,7 +134,7 @@ def main(
 
     equivalence_classes: EquivalenceClasses = defaultdict(lambda: defaultdict(list))
     for accession, values in segment_metadata.items():
-        # Author order sometimes varies despite segment apparently the same assembly
+        # Author order sometimes among segments from same isolate
         # Example: JX999734.1 (L) and JX999735.1 (M)
         modified_values = values_with_sorted_authors(values)
         group_key = str(
--- a/ingest/scripts/group_segments.py
+++ b/ingest/scripts/group_segments.py
@@ -30,6 +30,19 @@ import click
 import orjsonl
 import yaml
 
+
+def sort_authors(authors: str) -> str:
+    """Sort authors alphabetically"""
+    return ", ".join(sorted(authors.split(", ")))
+
+
+def values_with_sorted_authors(values: dict[str, str]) -> dict[str, str]:
+    """Sort authors values and return modified values"""
+    values_copy = values.copy()
+    values_copy["authors"] = sort_authors(values_copy["authors"])
+    return values_copy
+
+
 logger = logging.getLogger(__name__)
 logging.basicConfig(
     encoding="utf-8",
@@ -48,6 +61,7 @@ class Config:
     segmented: bool
 
 
+# TODO: Better name needed, what is special about these fields?
 SPECIAL_FIELDS: Final = {"segment", "submissionId"}
 
 
@@ -94,16 +108,21 @@ def main(
     # Group segments according to isolate, collection date and isolate specific values
     # These are the fields that are expected to be identical across all segments for a given isolate
 
+    # Dynamically determine the fields that are present in the metadata
     first_row = next(iter(segment_metadata.values()))
     if not first_row:
         msg = "No data found in metadata file"
         raise ValueError(msg)
-    all_fields = first_row.keys()
+    all_fields = sorted(first_row.keys())
+    logger.debug(f"All metadata fields: {all_fields}")
 
+    # Metadata fields can vary between segments w/o indicating being from different assemblies
     insdc_segment_specific_fields = set(config.insdc_segment_specific_fields)
     insdc_segment_specific_fields.add("hash")
 
-    shared_fields = set(all_fields) - insdc_segment_specific_fields - SPECIAL_FIELDS
+    # Fields that in principle should be identical for all segments of the same assembly
+    shared_fields = sorted(set(all_fields) - insdc_segment_specific_fields - SPECIAL_FIELDS)
+    logger.debug(f"Shared metadata fields: {shared_fields}")
 
     # Build equivalence classes based on shared fields
     # Use shared fields as the key to group the data
@@ -111,10 +130,14 @@ def main(
     type Accession = str
     type EquivalenceClasses = dict[tuple[str, str], dict[SegmentName, list[Accession]]]
 
-    # Creating the nested defaultdict with type hints
     equivalence_classes: EquivalenceClasses = defaultdict(lambda: defaultdict(list))
     for accession, values in segment_metadata.items():
-        group_key = tuple((field, values[field]) for field in shared_fields if values[field])
+        # Author order sometimes varies despite segment apparently the same assembly
+        # Example: JX999734.1 (L) and JX999735.1 (M)
+        modified_values = values_with_sorted_authors(values)
+        group_key = str(
+            tuple((field, value) for field in shared_fields if (value := modified_values[field]))
+        )
         segment = values["segment"]
         equivalence_classes[group_key][segment].append(accession)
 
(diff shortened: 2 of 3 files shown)


ingest/scripts/call_loculus.py
```diff
@@ -483,7 +483,7 @@ def record_factory(*args, **kwargs):
     if mode == "get-submitted":
         logger.info("Getting submitted sequences")
         response = get_submitted(config)
-        Path(output).write_text(json.dumps(response), encoding="utf-8")
+        Path(output).write_text(json.dumps(response, indent=4, sort_keys=True), encoding="utf-8")
```
corneliusroemer, line 486: This change is unrelated - could split it into separate PR

Just to make output diffs more stable

ingest/scripts/group_segments.py
```diff
@@ -94,27 +108,36 @@ def main(
...
     # Use shared fields as the key to group the data
     type SegmentName = str
     type Accession = str
     type EquivalenceClasses = dict[tuple[str, str], dict[SegmentName, list[Accession]]]
 
-    # Creating the nested defaultdict with type hints
     equivalence_classes: EquivalenceClasses = defaultdict(lambda: defaultdict(list))
     for accession, values in segment_metadata.items():
-        group_key = tuple((field, values[field]) for field in shared_fields if values[field])
+        # Author order sometimes varies despite segment apparently the same assembly
```
corneliusroemer, line 135:
```suggestion
        # Author order sometimes among segments from same isolate
```

ingest/scripts/group_segments.py
```diff
@@ -48,6 +61,7 @@ class Config:
     segmented: bool
 
 
+# TODO: Better name needed, what is special about these fields?
```
corneliusroemer, line 64: segment is by definition different across segments, so this is unnecessary to make configurable (hard coded special field)

ingest/scripts/prepare_metadata.py
```diff
@@ -145,7 +145,7 @@ def main(
 
     meta_dict = {rec[fasta_id_field]: rec for rec in metadata}
 
-    Path(output).write_text(json.dumps(meta_dict, indent=4), encoding="utf-8")
+    Path(output).write_text(json.dumps(meta_dict, indent=4, sort_keys=True), encoding="utf-8")
```
corneliusroemer, line 148:
```suggestion
    Path(output).write_text(json.dumps(meta_dict, indent=4), encoding="utf-8")
```

ingest/scripts/call_loculus.py
corneliusroemer, line 486:
```suggestion
        Path(output).write_text(json.dumps(response, indent=4), encoding="utf-8")
```

ingest/scripts/group_segments.py
```diff
@@ -94,27 +108,38 @@ def main(
     # Group segments according to isolate, collection date and isolate specific values
     # These are the fields that are expected to be identical across all segments for a given isolate
 
-    first_row = next(iter(segment_metadata.values()))
-    if not first_row:
+    # Dynamically determine the fields that are present in the metadata
+    some_row = next(iter(segment_metadata.values()))
```
theosanderson, line 112: Is this _not_ the first row? If it is still that then I find the original version clearer fwiw (though I found your changes above very helpful)

ingest/scripts/group_segments.py
```diff
@@ -94,27 +108,38 @@ def main(
...
     # Use shared fields as the key to group the data
     type SegmentName = str
     type Accession = str
     type EquivalenceClasses = dict[tuple[str, str], dict[SegmentName, list[Accession]]]
 
-    # Creating the nested defaultdict with type hints
     equivalence_classes: EquivalenceClasses = defaultdict(lambda: defaultdict(list))
     for accession, values in segment_metadata.items():
-        group_key = tuple((field, values[field]) for field in shared_fields if values[field])
+        # Author order sometimes among segments from same isolate
```
theosanderson, line 137: Missing word?

ingest/scripts/group_segments.py
```diff
@@ -189,11 +214,15 @@ def main(
...
-            row[field] = deduplicated_values.pop()
+            deduplicated_values = sorted(set(values.values()))
+            if len(deduplicated_values) > 1:
+                if field == "authors":
+                    # For authors, we accept different orders
+                    logger.info(f"Author orders differ for group {joint_key}: {values}")
+                else:
+                    msg = f"Assertion failed: values for group must be identical: {values}"
+                    raise ValueError(msg)
+            row[field] = deduplicated_values[0]
```
theosanderson, line 225: Just to be sure: this is sorted such that `0` will always be the same for the same data (even if the data comes up in different orders)? (I'm not doubting it, just checking)

theosanderson: Yes it is :)! Just above

theosanderson (commented):
Very good work - looks good to me. I don't think we should merge this until tomorrow at the point that we're ready to start looking into bringing this to prod, because once we do we block on bringing in other new things (if something even more urgent were to come up in the intervening time).

ingest/scripts/group_segments.py
corneliusroemer: Since we're using Python >=3.7 there _is_ a first element, so you're right!

https://mail.python.org/pipermail/python-dev/2017-December/151283.html

ingest/scripts/group_segments.py
corneliusroemer: true!

corneliusroemer:
Thanks for the review @theosanderson, I've addressed your points - I won't merge then, let's coordinate tomorrow on how to phase this into production!

anna-parker (approved):
Thanks so much for doing this @corneliusroemer!

anna-parker:
One suggestion
```
BEGIN TRANSACTION;

SELECT accession
INTO TEMPORARY TABLE temp_accessions
FROM sequence_entries
WHERE submitter = 'insdc_ingest_user' AND version = 2;

DELETE FROM sequence_entries
WHERE accession IN (SELECT accession FROM temp_accessions) AND version = 1;

UPDATE sequence_entries
SET version = 1
WHERE submitter = 'insdc_ingest_user' AND version = 2;

DROP TABLE temp_accessions;


SELECT accession
INTO TEMPORARY TABLE temp_accessions_preprocessed
FROM sequence_entries_preprocessed_data
WHERE submitter = 'insdc_ingest_user' AND version = 2;

DELETE FROM sequence_entries_preprocessed_data
WHERE accession IN (SELECT accession FROM temp_accessions_preprocessed) AND version = 1;

UPDATE sequence_entries_preprocessed_data
SET version = 1
WHERE submitter = 'insdc_ingest_user' AND version = 2;

DROP TABLE temp_accessions_preprocessed;

COMMIT;
```

anna-parker:
DO $$ 
DECLARE 
    rec RECORD; 
BEGIN 
    -- Loop over each accession where submitter is 'insdc_ingest_user' and version is 2
    FOR rec IN 
        SELECT accession 
        FROM sequence_entries 
        WHERE submitter = 'insdc_ingest_user' AND version = 2 
    LOOP 
        -- Start a new transaction for each accession
        BEGIN 
            -- Delete entries where accession is found and version is 1
            DELETE FROM sequence_entries 
            WHERE accession = rec.accession AND version = 1;

            -- Update the version to 1 for the current accession and version 2
            UPDATE sequence_entries 
            SET version = 1 
            WHERE accession = rec.accession AND version = 2;

            -- Commit the transaction
            COMMIT; 
        EXCEPTION 
            -- If there's an error, rollback the transaction and continue with the next accession
            WHEN OTHERS THEN 
                ROLLBACK; 
        END; 
    END LOOP; 
END $$;

anna-parker:
Weird I didn't think we could have duplicate ingest anymore: 
image

theosanderson:
Yes, that is mysterious

corneliusroemer:
Yeah it's odd, but I can't see why author order would have this effect?

anna-parker:
uggh... https://github.com/loculus-project/loculus/pull/2837 is now at the same state as this, I just rebased but it had the exact same bug in west nile and only in west nile ... so I do think it is this PR...

theosanderson:
It doesn't seem to have the duplication right now? https://ing-upd.loculus.org/ did we change sthng?

anna-parker:
I rebased to see if that had an impact - now https://ing-upd.loculus.org/  has a duplication in cchf

anna-parker:
But maybe this is more linked to our ingest deployment? Is there any chance ingest pods can run concurrently? 

I'm uncertain why this would only show up in this PR... maybe now ingest takes longer?

anna-parker:
Lets maybe not sort on keys? e.g. https://github.com/loculus-project/loculus/pull/2837/commits/1d833f9ac5eb94e9bbfcf7d2ef5581851f97a966 - adding the sort makes ingest slower and we want to minimize time between the get-original-metadata call and the submit calls.

corneliusroemer:
@anna-parker have you measured how much this changes? I'd be _very_ surprised if the sort takes more than a second. Sorting is fast, this is a small dataset.

anna-parker:
Yeah its not a big difference, e.g. for west nile
```
10:37:12    INFO ( prepare_metadata.py: 151) - Unsorted write took: 0.32383300000000004 
10:37:13    INFO ( prepare_metadata.py: 154) - Sorted write took: 0.356398 
```
But found it suspicious we saw this for the organisms with the most data

PR #2657, opened 2024-08-29
feat(k8s): allow addition of response header "X-Robots-Tag: noindex, nofollow"

corneliusroemer:
### Summary

Adding custom response headers is a convenient way to tell search engines what to index

X-Robots-Tag is supported wherever robots.txt is: https://developers.google.com/search/docs/crawling-indexing/robots-meta-tag#xrobotstag

By default, indexing is now _off_, unless .Values.noindex=false is set.

For relevant traefik docs, see https://doc.traefik.io/traefik/middlewares/http/headers/#adding-and-removing-headers

This PR allows pathoplexus/pathoplexus#48 to easily set noindex on all previews. This is shown/done in pathoplexus/pathoplexus#54

We probably _never_ want to index authentication pages, but this can be left for another PR.

preview URL: https://noproduction-noindex.loculus.org

### Screenshot

image

Google confirms that it found the noindex tag 🎉 

image

### Followup

- [x] Test that `noindex: false` in values.yaml actually works: https://github.com/loculus-project/loculus/pull/2659
- [ ] ~Merge pathoplexus/pathoplexus#54 to keep pathoplexus.org indexed~
- [ ] Add config value to non-prod instances in pathoplexus/pathoplexus
- [ ] ~Decide whether we want to keep loculus indexed (it would be no-index which would make sense, even for main.loculus.org I think) - by default, once merged, no loculus preview build would be indexed anymore.~
- [ ] Add no-index config to all non-main deployments of loculus?


kubernetes/loculus/values.yaml
```diff
@@ -1,4 +1,5 @@
 environment: server
+noindex: true
```
theosanderson, line 2: Suggest moving this to the preview server yaml

theosanderson: And maybe it should be camel case?

theosanderson (approved):
LGTM

kubernetes/loculus/values.yaml
theosanderson: And lastly - do we actually want to disable this for the main branch? I was imagining it just for everything else

corneliusroemer: Up to you, I think hiding by default is ok, if people want indexing it's a one line change. But I don't mind it the other way round. I also admit I don't know how to make loculus config specific to the main preview but not on other previews.

What do you mean by preview server yaml? Do we have an explicit config key "preview server" already? Feel free to push your suggestion or make a PR into this PR.

It's camel case, it's a single word 🤓 we can also call it `noindexRobotsHeader` to make it self explanatory.

I don't feel strongly about any of this. Happy for you to make a call if you feel like it.

chaoran-chen: I think that it's fine to disable it for the main branch as well because I don't really see valuable content that should be found by a search engine. The useful content are on loculus.org.

chaoran-chen (approved):

kubernetes/loculus/values.yaml
corneliusroemer: Yep that was my thinking - and Google would still get confused by us having very similar content on main.loculus and pathoplexus - that's one reason to disable on main. We can just change that later as well though.

corneliusroemer: Do we have a way to make `main` config different from _default_? The default is now _not_ to add the noindex, so indexing on. We can do this later, just wondering how we'd do it

chaoran-chen (approved):
Thanks, looks all good to me!

kubernetes/loculus/values.yaml
theosanderson: To do this we would probably:
- update the [appset](https://github.com/loculus-project/loculus/blob/main/kubernetes/appset.yaml) to expect a `noIndex` value that it would pass on as a helm value
- update https://github.com/loculus-project/argocd_metadata/blob/main/config.json to pass on that value

We don't need to do that now. IMO it would be nice in the fullness of time, potentially with some more tweaks to make it even more clear that it's a demo instance.

theosanderson (approved):

corneliusroemer:
I need to update the pathoplexus PR to use the new value name!

PR #2661, opened 2024-08-29
feat(doc): add simple manual sitemap to docs

corneliusroemer:
resolves #

preview URL:

### PR Checklist


theosanderson:
Personally my instinct would be to wait and see if indexing is actually an issue before adding something else to maintain

chaoran-chen (commented):
There is an Astro integration for generating sitemaps: https://docs.astro.build/en/guides/integrations-guide/sitemap/, so unless we encounter issues with it, I don't think that we should add a manual one. But it's a good idea and might be helpful.

PR #2663, opened 2024-08-29
[For Anya] Edits while PR reviewing create ena projects

corneliusroemer:
Just putting these out there to prevent merge conflicts in case you keep editing the base @anna-parker 

Probably best to merge these sooner. Also, it's fine with me to pile all the PRs up into one, to some extent it feels like some bits and pieces are missing (maybe part of the other PRs).

### DIFF
--- a/ena-submission/Snakefile
+++ b/ena-submission/Snakefile
@@ -18,20 +18,22 @@ with open("results/config.yaml", "w") as f:
     f.write(yaml.dump(config))
 
 LOG_LEVEL = config.get("log_level", "INFO")
-ORGANISMS = config['organisms'].keys()
+ORGANISMS = config["organisms"].keys()
 
 
 rule submit_all_external_metadata:
     input:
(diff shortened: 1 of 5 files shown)


ena-submission/Snakefile
```diff
@@ -18,20 +18,22 @@ with open("results/config.yaml", "w") as f:
...
 
 
 rule submit_all_external_metadata:
     input:
-        expand("results/submitted_{organism}.json", organism=ORGANISMS)
-    
+        expand("results/submitted_{organism}.json", organism=ORGANISMS),
+
+
 rule submit_external_metadata:
```
anna-parker, line 29: This will be added in a later PR - for the moment you can test it out using data in the test folder

anna-parker (approved):
Lgtm! Thanks for these improvements!

ena-submission/Snakefile
corneliusroemer: I realized that after I left the comment, thanks!

PR #2667, opened 2024-08-29
fix(website): revert hotfix

fengelniederhammer:
preview URL: https://reverthotfix.loculus.org/

### Summary

reverting the commits:
9eed2bc54ee55047d9834c8824c3971485879b48
565cc79f264fed135513f5b8a9281b1389bb39fd
Now that SILO 0.2.16 fixed the issues

### PR Checklist

### DIFF
--- a/website/src/components/SearchPage/DownloadDialog/generateDownloadUrl.ts
+++ b/website/src/components/SearchPage/DownloadDialog/generateDownloadUrl.ts
@@ -64,22 +64,6 @@ export const generateDownloadUrl = (
         }
     });
 
-    // Hotfix for LAPIS segment bug
-    // https://loculus.slack.com/archives/C05G172HL6L/p1724767046331529
-    if (
-        option.dataType.type === 'unalignedNucleotideSequences' &&
-        option.dataType.segment !== undefined &&
-        option.dataType.segment !== 'main'
-    ) {
-        // Check if the param is already set to something >=1
-        // In that case, no need to lower it
-        const userParam = params.get(`length_${option.dataType.segment.toUpperCase()}From`) ?? '0';
-        const userParamAsInt = parseInt(userParam, 10);
-        const maximumFilterValue = Math.max(1, userParamAsInt).toString();
-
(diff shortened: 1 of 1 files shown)


fengelniederhammer:
```
09:16:12 ~/Downloads ☕ $ curl 'https://lapis-reverthotfix.loculus.org/cchf/sample/unalignedNucleotideSequences/L?versionStatus=LATEST_VERSION&isRevocation=false&dataUseTerms=OPEN' --silent | grep '>' | wc -l
317
09:16:24 ~/Downloads ☕ $ curl 'https://lapis-reverthotfix.loculus.org/cchf/sample/unalignedNucleotideSequences/L?versionStatus=LATEST_VERSION&isRevocation=false&dataUseTerms=OPEN&length_LFrom=1' --silent | grep '>' | wc -l
726
```

When I download all CCHF L-segment sequences:
* without any special filter, it returns 317 sequences
* with length_LFrom=1 set, it returns 726 sequences
  * that also seems to be correct, because for the same filter /aggregated also shows count: 726.

Logically, it doesn't make sense that LAPIS returns more sequences when we add another filter restriction. The good news seems to be that it's correct as long as we have the hotfix in place, but something is odd - we will investigate.

corneliusroemer:
@fengelniederhammer thanks for investigating! The first curl has duplicates?

Don't count, just look at the lines in your pipeline

corneliusroemer (changes requested):
Some duplicates seem to be there - thanks for opening PR to test!

fengelniederhammer:
After analyzing, the bug is supposedly fixed by https://github.com/GenSpectrum/LAPIS-SILO/pull/552 and unrelated to this change.

corneliusroemer:
What I want to know though: how is it possible that the same issue happened in 3 independent organisms - that all run silo preprocessing independently. If it's stochastic, it should only happen in one instance. Why only in this PR and only now?

chaoran-chen (approved):
I think that this is now also safe to merge. Thanks! (There is no huge rush to merge this but we also shouldn't wait for too long as this is breaking download for all Loculus instances with multi-segmented genomes that do not have the `length` fields)

PR #2674, opened 2024-08-30
feat(website): Use sandwich menu even at slightly larger screen sizes

theosanderson:
Preview: https://use-sandwich-menu-even-at.loculus.org/

This is required due to the number of pathoplexus menu items. In time we should adapt our nav strategy.

Change from left to right
image


anna-parker (approved):
lgtm

PR #2676, opened 2024-08-30
fix(prepro): Raise warning if not all input segments align

### LINKED ISSUE #2595: Catch is a subset of segments don't align
If one submits data for a segmented virus where one segment aligns and other don't (for example because they are mixed up), the pipeline doesn't error and lets you release those sequences. There is no indication that something is wrong.

image

This doesn't happen if none of the sequences align.

anna-parker:
resolves #2595 

preview URL: https://segments-align-anya.loculus.org/

### Screenshot

1. No warning if 1+ segment is removed but others align: 
image

2. If segments S and L are flipped causing both to not align we see 
image

### DIFF
--- a/preprocessing/nextclade/src/loculus_preprocessing/prepro.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/prepro.py
@@ -383,6 +383,7 @@ def add_input_metadata(
     spec: ProcessingSpec,
     unprocessed: UnprocessedAfterNextclade,
     errors: list[ProcessingAnnotation],
+    warnings: list[ProcessingAnnotation],
     input_path: str,
 ) -> InputMetadata:
     """Returns value of input_path in unprocessed metadata"""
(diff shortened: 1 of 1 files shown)


preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -118,10 +118,16 @@ def parse_nextclade_json(
...
 ) -> defaultdict[AccessionVersion, defaultdict[SegmentName, dict[str, Any]]]:
     """
-    Update nextclade_metadata object with the results of the nextclade analysis
+    Update nextclade_metadata object with the results of the nextclade analysis.
+    If the segment existed in the input (unaligned_nucleotide_sequences) but did not align
+    nextclade_metadata[segment]=None.
     """
+    for id, seg_dict in unaligned_nucleotide_sequences.items():
+        if segment in seg_dict and seg_dict[segment] is not None:
+            nextclade_metadata[id][segment] = None
```
theosanderson, line 130:
```suggestion
    for id, segment_sequences in unaligned_nucleotide_sequences.items():
        if segment in segment_sequences and segment_sequences[segment] is not None:
            nextclade_metadata[id][segment] = None
```
If this dict is indeed like { "S": "ATTTTGG"...}

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
theosanderson, line 1: Worth rebasing before merge to check CI - atm it seems to date from before release because I got a password on preview :)

theosanderson:
Tested this out, and looks great! One suggested change for readability.

theosanderson (approved):

PR #2680, opened 2024-08-31
fix(prepro): Label a sequence with error if not all input segments align

### LINKED ISSUE #2595: Catch is a subset of segments don't align

anna-parker:
Preview: https://revert-2678-revert-2676-s.loculus.org/

Resolves #2595

#### Motivation

Currently when a user submits multi-segmented entries where at least one segment aligns, then they will not be warned if other segments of the same entry do not align (Use case: someone submits a proper CCHF L segment, but the M segment isn't the M segment, it's actually S).

For non-segmented viruses, we guarantee, that every sequence that's released actually aligns.

The bug here currently is, that this is not true for segmented viruses.

#### Desired (?)

All segments that are submitted _must_ align, otherwise the user cannot release the entry.

#### Pathoplexus changes by this PR (if any)

There should be no change to pathoplexus production DB state with this PR, as ingest already filters out all sequences that do not align.

It only impacts future user submissions of segmented viruses by blocking submissions of entries with non-aligning segments.

### PR History
In https://github.com/loculus-project/loculus/pull/2676 (reverted quickly before making it to Pathoplexus) I changed the code to label a sequence with a warning if a segment does not align.

This had the unintended side effect that for single-segment organisms, sequences that did not align made it through ingest and onto main (ingest auto approves everything that doesn't error).

It was thus reverted very quickly before it reached pathoplexus prod (in #2678).

In this PR, we now more carefully redo the original intent of the original PR but avoiding the side effect that made us revert by emitting an _error_ in case of non-alignment not merely a warning.

### Screenshots

1. No unaligned sequences on website (check that total number of sequences is the same as on main): 
image

2. If 1+ segment is missing (as in not provided by submitter) from input do not throw an error:
image

3. If any submitted segment does not align (including single segment case) throw an error - here I introduced an error by switching the L and S segment:
image

### DIFF
--- a/preprocessing/nextclade/src/loculus_preprocessing/prepro.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/prepro.py
@@ -391,7 +391,6 @@ def add_input_metadata(
     spec: ProcessingSpec,
     unprocessed: UnprocessedAfterNextclade,
     errors: list[ProcessingAnnotation],
-    warnings: list[ProcessingAnnotation],
     input_path: str,
 ) -> InputMetadata:
     """Returns value of input_path in unprocessed metadata"""
@@ -399,23 +398,15 @@ def add_input_metadata(
     nextclade_prefix = "nextclade."
     if input_path.startswith(nextclade_prefix):
         segment = spec.args.get("segment", "main")
-        if not unprocessed.nextcladeMetadata:
-            errors.append(
-                ProcessingAnnotation(
-                    source=[
-                        AnnotationSource(
-                            name="main",
-                            type=AnnotationSourceType.NUCLEOTIDE_SEQUENCE,
-                        )
-                    ],
-                    message="Nucleotide sequence failed to align",
-                )
-            )
-            return None
         sub_path = input_path[len(nextclade_prefix) :]
         if segment in unprocessed.nextcladeMetadata:
             if not unprocessed.nextcladeMetadata[segment]:
-                warnings.append(
+                message = (
+                    "Nucleotide sequence failed to align"
+                    if segment == "main"
+                    else f"Nucleotide sequence for {segment} failed to align"
+                )
+                errors.append(
                     ProcessingAnnotation(
                         source=[
                             AnnotationSource(
@@ -423,7 +414,7 @@ def add_input_metadata(
                                 type=AnnotationSourceType.NUCLEOTIDE_SEQUENCE,
                             )
                         ],
-                        message=f"Nucleotide sequence for {segment} failed to align",
+                        message=message,
                     )
                 )
                 return None
@@ -476,9 +467,7 @@ def get_metadata(
         args["submitter"] = unprocessed.submitter
     else:
         for arg_name, input_path in spec.inputs.items():
-            input_data[arg_name] = add_input_metadata(
-                spec, unprocessed, errors, warnings, input_path
-            )
+            input_data[arg_name] = add_input_metadata(spec, unprocessed, errors, input_path)
         args = spec.args
         args["submitter"] = unprocessed.inputMetadata["submitter"]
 
--- a/preprocessing/nextclade/src/loculus_preprocessing/prepro.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/prepro.py
@@ -83,7 +83,7 @@ def parse_nextclade_tsv(
     ],
     result_dir: str,
     config: Config,
-    segment: str,
+    segment: SegmentName,
 ) -> tuple[
     defaultdict[AccessionVersion, defaultdict[GeneName, list[AminoAcidInsertion]]],
     defaultdict[AccessionVersion, defaultdict[SegmentName, list[NucleotideInsertion]]],
@@ -116,8 +116,10 @@ def parse_nextclade_tsv(
 def parse_nextclade_json(
     result_dir,
     nextclade_metadata: defaultdict[AccessionVersion, defaultdict[SegmentName, dict[str, Any]]],
-    segment,
-    unaligned_nucleotide_sequences,
+    segment: SegmentName,
+    unaligned_nucleotide_sequences: dict[
+        AccessionVersion, dict[SegmentName, NucleotideSequence | None]
+    ],
 ) -> defaultdict[AccessionVersion, defaultdict[SegmentName, dict[str, Any]]]:
     """
     Update nextclade_metadata object with the results of the nextclade analysis.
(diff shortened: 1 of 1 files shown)


preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -390,21 +398,26 @@ def add_input_metadata(
     nextclade_prefix = "nextclade."
     if input_path.startswith(nextclade_prefix):
         segment = spec.args.get("segment", "main")
-        if not unprocessed.nextcladeMetadata:
```
anna-parker, line 393: this case will now never occur as nextcladeMetadata always includes a dictionary, potentially though a dictionary with empty values

theosanderson:
Thanks - this took a bit of time to get my head around but I can see why you came to this conclusion. It may be something where we should ask the wider group what we think the desired behaviour should be. One downside to this behaviour I can see would be for ingest, for which the better behaviour might e.g. to be to keep the sequences that do align and entirely lose those that don't. (That might work for users too).

A practical concern I have is around pathoplexus prod, and upgrades that would change whether previously ingested sequences were valid (which we'll have to deal with at some point but afaik we don't have a strategy for yet). But it seems that the number of CCHF sequences is the same so I guess this isn't a real issue yet?

corneliusroemer:
I'll review properly later, just to share a misunderstanding when I read the PR title, "raising an Error" sounds like the Python script would raise an exception (at least that was my first thought). "Error a sequence" or "Label a sequence with error".

anna-parker:
> Thanks - this took a bit of time to get my head around but I can see why you came to this conclusion. It may be something where we should ask the wider group what we think the desired behaviour should be. One downside to this behaviour I can see would be for ingest, for which the better behaviour might e.g. to be to keep the sequences that do align and entirely lose those that don't. (That might work for users too - with a warning).
> 
> A practical concern I have is around pathoplexus prod, and upgrades that would change whether previously ingested sequences were valid (which we'll have to deal with at some point but afaik we don't have a strategy for yet). But it seems that the number of CCHF sequences is the same so I guess this isn't a real issue yet?

I was also initially surprised that CCHF was not effected - but this is actually a nice side effect that we align CCHF in ingest (otherwise we didn't know which segment was which) - this means ingest will only send sequences that actually align.

theosanderson:
Thanks for clarifying the ingest point: that makes sense!

> Removing the segments that dont align in prepro would make the most sense to not confuse users but as you mentioned this would actually break the hash (ingest would not know that the segment does not align and would hash including the segment that we would remove in prepro).

I didn't fully follow this actually (unless by users you meant admins) - as you pointed out in the case of ingest we are already guaranteeing that everything submitted aligns? Or is it if we were to change that?

anna-parker:
> Removing the segments that dont align in prepro would make the most sense to not confuse users but as you mentioned this would actually break the hash (ingest would not know that the segment does not align and would hash including the segment that we would remove in prepro).

A sorry - you are correct ingest at the moment does not send any (segment)-sequences to loculus from multi-segment organisms that do not align. So it should not be an issue to have prepro remove the segments that do not align as only sequences submitted by users could have segments that do not align. 

However, if we ever change ingest to not align - or add another multi-segmented organism where we do not align re-ingest will very likely fail because of this- and I think it will be confusing to debug if anyone else decides to use Loculus. 

Assume ingest sends 3 segments in a group. One segment does not align and prepro removes it from the unaligned sequences. Now reingest will ask loculus for all sequences and calculate a hash from the metadata and unaligned segments. It will compare to the data it got from NCBI. In this case 3 segments it wants to group. But this 3rd segment is not in Loculus. re-ingest thinks it needs to revoke the old group of two segments in Loculus and create a new one with three segments. This will get into a loop as every time we reingest reingest will try to do this again.

theosanderson:
Sorry that I dropped off this thread. IMO we mostly shouldn't make decisions that will affect user-submitted sequences on the basis of what happens in ingest: we have control over ingest and our major priority should be user experience for submitters. That said, I'm not 100% sure what the best user experience _is_, whether it's to tolerate sequences that don't align with a warning or to raise an error because someones pipeline might be wrong.

corneliusroemer:
In principle I think I understand what's going on here - just checking on a few things that remain unclear (there's lots of reverts and reverts of reverts).

I edited the PR description to make it less confusing (reversion of reversion is hard to wrap one's head around, now it's a bit more linear, also the extra paragraphs help skimming in my experience).

I actually think it could be edited further to move the whole reversion history to a "History" section. The history doesn't actually matter to understand this PR? IIUC, you made a change, we merged, it had bad side effect so we immediately reverted. Now we're back to status quo. So it'd be best to argue this PR from the status quo, rather than from the state after the initial PR was merged. Let's forget about the reverted PR (mostly, other than for history section).

Could you please answer the following questions (I have vague gut feelings about them but would be good to get your more definitive thoughts @anna-parker):
1. What if anything will change on production pathoplexus with this PR?
2. What is the reason for this PR (ignoring the history), a 1-2 sentence summary would be great.

corneliusroemer:
IIUC, the high level behaviour change is:
- For segmented viruses
- Error on entries where not all submitted segments align

I'm not 100% what the status quo is. We probably let things through as long as at least one segment aligns?

corneliusroemer:
I confirmed with these test sequences that we now error when at least one segment doesn't align (e.g. due to segment typo): [cchf_test.zip](https://github.com/user-attachments/files/16908216/cchf_test.zip)

Brave Browser 2024-09-06 14 49 04

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
corneliusroemer: An entirely empty dictionary will be trigger this branch afaik

corneliusroemer:
Interestingly my commit seems to break things 👀  now we have 23k WNV again

corneliusroemer (approved):

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
anna-parker: That is true but we should never have an empty dictionary and get to this function, as if there existed an unaligned sequence for that segment the dictionary will now have segmentName as a key. Otherwise if there were no unaligned sequences even process_single will break early and we will get to this function. However, I can keep this case and just mark in the warning that this case should never happen and that users should contact the administrator.

PR #2685, opened 2024-09-01
fix(website): fix resizing issue on DocsMenu

### LINKED ISSUE #2662: Dynamics resizing of pages with collapsible content
On pages with a lot of collapsible content (in particular, Pathoplexus FAQ), the page resizes from when all is collapsed and then when 1 or more things are expanded. This provides a bit of a weird user experience as the page moves around a bit.

While collapsed:
image

Expanded:
image

(Subtle difference but page 'jumps' when you do it - try it out [here](https://pathoplexus.org/about/faq))

Would be nice (but not urgent) to have page size anticipate the expanded size somehow so the page stays the same ratios. 

This could also be being exacerbated by 'long' menu items like 'Frequently Asked Questions (FAQ)' and 'Open Data Terms of Use (Summary)' that wrap and unwrap with this resizing.

theosanderson:
This should resolve #2662 by fixing the width of the menu

I've tried to test it just by editing live CSS and it seemed to work

(Cornelius update): https://fix-resizing-issue.loculus.org

### DIFF
--- a/website/src/components/Navigation/DocsMenu.tsx
+++ b/website/src/components/Navigation/DocsMenu.tsx
@@ -143,7 +143,7 @@ const DocsMenu: React.FC<DocsMenuProps> = ({ docsPages, currentPageUrl, title })
 
                     <DisclosurePanel className='sm:hidden'>{menuContent}</DisclosurePanel>
 
-                    <div className='hidden sm:block'>{menuContent}</div>
+                    <div className='hidden sm:block sm:w-64'>{menuContent}</div>
                 </>
             )}
         </Disclosure>


corneliusroemer (commented):
Ah damn we can't actually test this until it's used in pathoplexus preview? Could you maybe make a pathoplexus pr using this commit hash here?

Maybe you think this is unnecessary but I: a) haven't been able to understand the bug fully from the screenshots Emma shared (@emmahodcroft I can show you how to record screencast videos very quickly, just like screenshots) b) afaict the preview here doesn't allow repro because we don't actually show the pathoplexus style docs on loculus.

I don't want to be annoying, just genuinely trying to check things I approve - and I don't feel I can here based on info I've got :)

theosanderson:
The quickest way to test is to make the corresponding change in Chrome dev tools - that's what I did

corneliusroemer:
> The quickest way to test is to make the corresponding change in Chrome dev tools - that's what I did.

On the pathoplexus page right?

corneliusroemer:
I tried to find the right div in dev console but it wasn't trivial for me - so made a quick pp preview pr - that's faster and safer, so overall better I think :)

https://github.com/pathoplexus/pathoplexus/pull/87

theosanderson:
Yes:

https://tailwindcss.com/docs/width

```
w-64 | width: 16rem; /* 256px */


```

theosanderson:
Crossed replies. Cool - of course always feel free to make any previews you like!

corneliusroemer (approved):
Confirmed fixed in pathoplexus preview

I hope I didn't annoy you too much with my insistence on testing for real

theosanderson:
Thanks! No prob - hope I didn't annoy you to much by making you do the work to test (in this particular context)

corneliusroemer:
Oh no, I enjoyed figuring out how to test it - in future I would appreciate of course if you made a quick test pr like this :D and screencast the verification etc, but it's also fine for me to do it - has the advantage of independent set of eyes!

PR #2707, opened 2024-09-04
fix(deployment): Only try to include SILO-specific config in the LAPIS/SILO configmap if it is actually defined

### LINKED ISSUE #2668: It should be possible to run Loculus without `dateToSortBy`
We established today that we when we no longer set `dateToSortBy` in Loculus values, the LAPIS instances crash. The intended behaviour is that it is possible to run Silo without this (and maybe it is, with the right configuration) and we would like it to be possible to deploy Loculus with no date fields. We should investigate what the problem was and either raise a bug in Silo or adjust Loculus configuration.

theosanderson:
resolves #2668

Previously failing to set `silo:` resulted in 
image

which led LAPIS to crash

https://theosanderson-patch-7.loculus.org

### DIFF
--- a/kubernetes/loculus/templates/_siloDatabaseConfig.tpl
+++ b/kubernetes/loculus/templates/_siloDatabaseConfig.tpl
@@ -32,5 +32,7 @@ schema:
   {{- end }}
   {{- end }}
   primaryKey: accessionVersion
+{{ if .silo}}
   {{- .silo | toYaml | nindent 2 }}
+{{ endif }}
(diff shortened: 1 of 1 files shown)


theosanderson:
I'm pretty sure this works, but even if it didn't it would be an improvement over the current state (it shouldn't be possible to get `null` in the yaml as occurs here) so it would be great to merge it

corneliusroemer (approved):

Taepper (approved):

PR #2709, opened 2024-09-04
deps(backend): bump gradle to 8.8 from 8.5

corneliusroemer:
Last bump was 9 months ago. We should check this whenever kotlin version is bumped by dependabot.

8.8 is the max version of gradle supported by kotlin 2.0.20
See https://kotlinlang.org/docs/gradle-configure-project.html#apply-the-plugin

preview URL: https://bump-gradle.loculus.org

### Summary
Bump gradle version to 8.8 from 8.5


chaoran-chen (approved):
Thanks!

Issue #2708, opened 2024-09-04
Flaky backend test: `DeleteSequencesEndpointTest > WHEN superuser deletes entries of other user THEN is successfully deleted() FAILED`


corneliusroemer:
It sometimes fails, sometimes doesn't, strange. Maybe race condition in test?

```
> Task :test

DeleteSequencesEndpointTest > WHEN superuser deletes entries of other user THEN is successfully deleted() FAILED
    java.lang.AssertionError: JSON path "$[0].accession" expected:<LOC_000001Y> but was:<LOC_00000AE>
        at org.springframework.test.util.AssertionErrors.fail(AssertionErrors.java:59)
        at org.springframework.test.util.AssertionErrors.assertEquals(AssertionErrors.java:122)
        at org.springframework.test.util.JsonPathExpectationsHelper.assertValue(JsonPathExpectationsHelper.java:123)
        at org.springframework.test.web.servlet.result.JsonPathResultMatchers.lambda$value$2(JsonPathResultMatchers.java:111)
        at org.springframework.test.web.servlet.MockMvc$1.andExpect(MockMvc.java:214)
        at org.loculus.backend.controller.submission.DeleteSequencesEndpointTest.WHEN superuser deletes entries of other user THEN is successfully deleted(DeleteSequencesEndpointTest.kt:308)
```

corneliusroemer:
No this is not caused by parallelism, this happens without parallelism, before I started adding it.

corneliusroemer:
Closing as I can't reproduce right now - so maybe fixed/related to #2710 - will reopen if I see it again.

PR #1914, opened 2024-05-15
feat(backend): don't require authentication for /get-released-data

corneliusroemer:
resolves #1913
resolves #1625

preview URL: https://get-no-auth.loculus.org

### Summary

- Don't require authn/authz for `get-released-data` endpoint
- Fix lints
- Run backend tests in paralell

PR #2710, opened 2024-09-05
chore(backend): disable parallel test execution to avoid flaky tests

chaoran-chen:
Maybe helps with #2708 (hopefully, let's see)

This reverts the parallel test execution that was introduced in #1914. Some tests expect an empty database, so parallel execution can very likely be a source of flakiness.

I will run backend tests five times: https://github.com/loculus-project/loculus/actions/runs/10716872022/job/29715313279?pr=2710

### DIFF
--- a/backend/build.gradle
+++ b/backend/build.gradle
@@ -79,7 +79,6 @@ tasks.named('test') {
         exceptionFormat TestExceptionFormat.FULL
         showExceptions true
     }
-    maxParallelForks = Runtime.runtime.availableProcessors().intdiv(2) ?: 1
 }
 
 tasks.named('bootBuildImage') {
--- a/backend/gradle.properties
+++ b/backend/gradle.properties
@@ -1,1 +1,0 @@
-org.gradle.parallel=true


corneliusroemer:
Ah sorry, makes sense to revert if there's a chance for flakiness. Not that it has happened in practice I think.

Isn't it dangerous to work on one database? Do you assume that every test cleans up after itself?

corneliusroemer (approved):

fengelniederhammer:
> Do you assume that every test cleans up after itself?

Yes, there is a clean up step after (or before) every test and thus every test assumes that it's running on a clean state.

fengelniederhammer:
> speeding up tests isn't completely without benefit!

You have to weigh it against initial setup and long-term maintenance. My feeling is that it's not worth the effort yet.

corneliusroemer:
Sure! I just don't know how hard it is to give each runner its own db.

corneliusroemer:
Sure! I just don't know how hard it is to give each runner its own db.

Note I've reworded the PR to not auto close the flakiness because I think the flakiness predates parallelity.

corneliusroemer:
I'm merging this now already because I introduced the change and shouldn't have. No need to discuss further at this point and waste our time ;)

PR #2711, opened 2024-09-05
k8s: Remote dateToSortBy SILO config flag that contributed to bug

corneliusroemer:
preview URL: https://try-minimal-silo-config.loculus.org

### Summary

- Remove the config flag that contributed to the SILO bug. By using the default, we're more confident about the test coverage.

### Followup

- [ ] Remove the config flag also from pathoplexus so loculus and pathoplexus are in sync

### DIFF
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -1033,8 +1033,6 @@ defaultOrganismConfig: &defaultOrganismConfig
         - length
       defaultOrderBy: sampleCollectionDate
       defaultOrder: descending
-    silo:
-      dateToSortBy: sampleCollectionDate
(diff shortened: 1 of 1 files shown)


Taepper:
Similar to what Theo mentioned on Slack, I would not remove this, without a good reason to. We have more e2e tests *with* this flag as opposed to without this flag. It was just that the tests that were added in https://github.com/GenSpectrum/LAPIS-SILO/pull/542 did not contain any e2e tests, but only unitTests (which test without the flag)

fengelniederhammer:
`dateToSortBy` was implemented for a reason: it speeds up queries. I also think that we should actually use the feature.

PR #2712, opened 2024-09-05
docs: Remove date from "my first loculus" tutorial

theosanderson:
This tutorial is intended to be the most basic configuration of Loculus possible, without requiring people to think about preprocessing functions, etc. but we previously had to include a date due to the fact that our configuration essentially required it for SILO. We've now fixed that, so we can remove it.


corneliusroemer (approved):
Haven't run this tutorial ever - probably should at some point, so trusting you here this is a good change.

Issue #702, opened 2023-12-21
Rework user sequence page


ghost:
This is a stub ticket for the redesign of the user sequence pages. 

With introduction of groups, we need to rework the `get-sequences-of-user` endpoint. It is currently unchanged and returns all sequences they submitted, even when they are not part of that group anymore.

PR #2716, opened 2024-09-05
feat(e2e): Add tests that data submitted as restricted stays restricted state

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/2718, https://github.com/loculus-project/loculus/issues/702

preview URL: https://restricted-data-consisten.loculus.org/

### Summary

1. Adds a test that if data is submitted to the frontend with restricted status it will stay in this restricted status. In order to do this I add the option for the mocked prepro pipeline to process restricted data. **Disclaimer**: With our current e2e set up we cannot test this entire process. So I am actually just testing that if the prepro pipeline sends us restricted data we keep it restricted. The user sets a sequence to state restricted before we send it to the prepro pipeline and the e2e tests mock the entire interaction with the prepro pipeline. So in theory the prepro pipeline could just remove the restricted label and send us back open data and this test would not detect it.
2. Fix error that was flagged when running tests locally: `Warning: Each child in a list should have a unique "key" prop" is a valid error` by adding a key to a map function. 
3. The e2e tests do not actually include the preprocessing pipeline, I will create a new issue to address this. 

### PR Checklist

- [x] Check if it is actually necessary to add a separate group for restricted data. 
- [x] Update docs


website/tests/pages/submission/index.spec.ts
```diff
@@ -48,16 +48,7 @@ test.describe('The submit page', () => {
...
-        const responseHeaderLocator = submitPage.page.getByText(
-            `${restrictedDataUseTermsType} until: ${dateTimeInMonths(6).toFormat('yyyy-MM-dd')}`,
-            {
-                exact: false,
-            },
-        );
-        expect(await responseHeaderLocator.count()).toBe(testSequenceCount);
 
-        // TODO(#702): Redirect to the review page after submission is successful, and check the data use terms there
+        await submitPage.page.waitForURL(`${baseUrl}${routes.userSequenceReviewPage(dummyOrganism.key, groupId)}`);
```
theosanderson, line 52: So this tests that you can go through the UI for submitting restricted data but it doesn't actually test that the UI actually correctly captures that choice and sends it to the server. But maybe that's too hard atm. We should try to full E2E soon which would let us do that.

website/tests/pages/review/index.spec.ts
```diff
@@ -48,4 +49,25 @@ test.describe('The review page', () => {
...
+        await reviewPage.waitForTotalSequenceCountCorrect(total + testSequenceCount);
+
+        await reviewPage.approveAll();
+
+        await reviewPage.waitForTotalSequenceCountCorrect(total);
+
+        await reviewPage.page.goto(
+            `${baseUrl}${routes.mySequencesPage(dummyOrganism.key, groupId)}dataUseTerms=RESTRICTED`,
+        );
+        reviewPage.page.getByText(`Search returned ${testSequenceCount} sequence`);
```
theosanderson, line 71: For safety, it might make sense to click on the top sequence and check that its data use terms are indeed restricted this is because:

if we rename `dataUseTerms` to e.g. `data_use_terms` but forget to change this test then the  `dataUseTerms=RESTRICTED` filter above will cease to have an effect so we could get a test that always passes even if our code stops displaying the restricted data properly

theosanderson (approved):
Thanks a lot for taking this on - I think it could be good to click through to the sequence page and check the sequence is restricted

website/tests/pages/review/index.spec.ts
theosanderson: We could do this by looking for selectors regex matching `LOC_`

website/tests/pages/submission/index.spec.ts
anna-parker: Exactly. It is not possible with our current e2e test set up to test the full process. The user sets a sequence to state restricted before we send it to the prepro pipeline and the e2e tests mock the entire interaction with the prepro pipeline. So in theory the prepro pipeline could just remove the restricted label and send us back open data and this test would not detect it.

website/tests/pages/review/index.spec.ts
anna-parker: Done :-) I click the first on the first accession link in the table :-)

theosanderson: Nice!

PR #2728, opened 2024-09-09
Remove '-' from list of allowed symbols in unaligned sequences.

anna-parker:
resolves #

preview URL: https://no-gaps-in-unaligned.loculus.org/

### Summary

'-' only makes sense in the context of aligned sequences. It is not accepted by ENA and is not included in official IUPAC lists: https://genome.ucsc.edu/goldenPath/help/iupac.html#:~:text=The%20International%20Union%20of%20Pure,for%20either%20G%20or%20A).

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/service/submission/ProcessedSequenceEntryValidator.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/service/submission/ProcessedSequenceEntryValidator.kt
@@ -266,12 +266,12 @@ class ProcessedSequenceEntryValidator(private val schema: Schema, private val re
             "nucleotideInsertions",
         )
 
-        validateNoUnknownAlignedNucleotideSymbol(
+        validateNoUnknownNucleotideSymbol<AlignedNucleotideSymbols>(
             processedData.alignedNucleotideSequences,
             "alignedNucleotideSequences",
         )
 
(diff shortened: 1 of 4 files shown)


anna-parker:
@fengelniederhammer I think (correct me if I am wrong) the backend uses the nucleotide type for both unaligned and aligned sequences - it is a correct symbol for aligned sequences so I think it is ok to keep the backend as is.

fengelniederhammer:
Isn't it the same issue? In aligned sequences `-` should be allowed, in unaligned it is not allowed. We could easily also introduce a new list in the backend that splits the validation.

backend/src/main/kotlin/org/loculus/backend/service/submission/ProcessedSequenceEntryValidator.kt
```diff
@@ -293,18 +311,18 @@ class ProcessedSequenceEntryValidator(private val schema: Schema, private val re
         }
     }
 
-    private fun validateNoUnknownNucleotideSymbol(
+    private inline fun <reified ValidSymbols> validateNoUnknownNucleotideSymbol(
```
fengelniederhammer, line 314: I generalized the existing method and pushed a commit, since I figured it would be easier to do it rather than explain it.

Issue #2092, opened 2024-06-03
Improve efficiency of frequently polled backend endpoints


corneliusroemer:
When I showed our kubernetes resource usage to Richard, he raised the good point that it is surprising that without any user interaction/load, we're using around 500 millicpus per instance.

I had previously looked into this: the main cpu load (when not preprocessing, which only happens right after startup) is backend and database.

My gut feeling for what keeps them at around 250 mCPU each is that for each organism (which are 5 by default right now), we:
- a) preprocessing asks whether there are any new sequences every few seconds (to keep processing start delay low)
- b) silo preprocessing downloads _all_ sequences every 30 seconds

It would be simple to find out what the impact of each is by simply tuning the request frequency and plotting load against the frequency (this should give a straight line with the slope showing the average load per request per second).

I'm sure we could easily reduce the background usage significantly without sacrificing silo/preprocessing responsiveness. It's not a big priority right now but we should eventually tackle it.

### Ideas

- Silo load can be reduced by having SILO preprocessing submit a timestamp in the header saying when it last received data from the backend. The backend would keep track of the last time of modification of the "data_for_release" view, and can decide whether there's been new data. If no, it can reply with "no change" to SILO preprocessing. There's a bit of care required for making sure that _every_ relevant change causes an update of the "last_change" timestamp in the backend, but this shouldn't be too hard. Simple logic (though with false positives): "whenever the database makes changes to a subset of the tables `preprocessing` and `submission`, it updates the last "release_data_timestamp" _after_ the change transaction has completed.
- There's nothing to preprocess if: since the last request to preprocess, there haven't been any changes to the preprocessing and submission tables. Looks like the same timestamp might be usable here. In this case though, the backend would keep track of that timestamp of the last request, not the silo preprocessing client.

Maybe there are other polling requests that create background load, but the 2 above are the most likely candidates to me.

corneliusroemer:
One more thing that keeps polling: ingest's approve could benefit from etag approach as well

theosanderson-agent:
This appears to have been substantially addressed by PRs #2731, #2764, and #2768, which introduced etag-based caching to reduce database load on frequently polled backend endpoints. Proposing for closure — please verify the polling efficiency improvements are sufficient.

PR #2731, opened 2024-09-09
feat: use etags for caching to avoid backend/db load when idle

corneliusroemer:
co-authored with @anna-parker and Claude Sonnet

related to: https://github.com/loculus-project/loculus/issues/2092

preview URL: https://table-update-tracker.loculus.org

### Summary

1. Adds UpdateTrackerTable to database with last time a table in the database was modified.
iTerm2 2024-09-09 19 46 36

2. Adds optional `IF_NONE_MATCH` header to `get-released-data`, this accepts an etag and returns a lastDatabaseWriteETag corresponding to the lastDatabaseWrite. If the etag and the current lastDatabaseWriteETag differ `get-released-data` will respond with all released data and the current lastDatabaseWriteETag. If the atg is equal to current lastDatabaseWriteETag  then `get-released-data` will respond with 304: NOT_MODIFIED.
3. Updates `silo_import_job.sh` and `silo_import_wrapper.sh` to use the `IF_NONE_MATCH` header. This significantly reduces the load on the database and backend, times are from after preprocessing has finished (CPU is in mCPU): 

```
kubectl top pod --sum --containers --all-namespaces | egrep "backend|s-database" | egrep "main|tracker-anya" | sort -hk4                                           bs-mbpas-0092: Wed Sep 11 19:51:30 2024

prev-table-update-tracker-anya   loculus-backend-c859b6f4b-gj5k5                                   backend                              19m          421Mi
prev-table-update-tracker-anya   loculus-database-67f7d88557-9mf6v                                 database                             101m         326Mi
prev-main                        loculus-database-645647c886-kh5vf                                 database                             113m         234Mi
prev-main                        loculus-backend-55f5cdb6cb-p7p4f                                  backend                              300m         483Mi
```

### PR Checklist


anna-parker:
Ideas from pair programming: We can use this with the http header with `if-modified-since` for the get-released-data endpoints and also for the preprocessing pipeline.

corneliusroemer:
One issue we have right now is that "current_preprocessing_pipeline" table seems to be updated without any real changes. It still shows this:

```
loculus=# select * from table_update_tracker;
 groups_table                       | 2024-09-09 17:14:34.637079
 user_groups_table                  | 2024-09-09 17:14:34.637079
 current_processing_pipeline        | 2024-09-09 17:46:29.561886
 sequence_entries_preprocessed_data | 2024-09-09 17:46:13.487679
 sequence_entries                   | 2024-09-09 17:23:21.609788
 sequence_upload_aux_table          | 2024-09-09 17:15:51.885098
 metadata_upload_aux_table          | 2024-09-09 17:15:51.885098

loculus=# select * from current_processing_pipeline;
       1 | 2024-09-09 17:13:09.196022
```

We either have to make the trigger function smarter to only trigger on actual changes (not just update commands that don't cause changes) - or we use that timestamp from that table directly instead of the "last update" table.

corneliusroemer:
Same with the `sequence_entries_preprocessed_data` - it also gets an "update" issued without any real underlying change it seems.

Probably for now, in the version table we can look up the timestamp directly. And we might be able to ignore the preprocessed data table entirely.

backend/src/main/resources/db/migration/V1.2__add_table_update_tracker.sql
```diff
@@ -0,0 +1,42 @@
...
+            EXECUTE FUNCTION update_table_tracker()', table_name);
+    END IF;
+END;
+$$ LANGUAGE plpgsql;
+
+SELECT create_update_trigger_for_table('sequence_entries');
+SELECT create_update_trigger_for_table('sequence_entries_preprocessed_data');
+SELECT create_update_trigger_for_table('external_metadata');
+SELECT create_update_trigger_for_table('groups_table');
+SELECT create_update_trigger_for_table('current_processing_pipeline');
```
chaoran-chen, line 39: Given, as you noted, that the tracking doesn't work reliably for `current_preprocessing_pipeline` and `sequence_entries_preprocessed_data`, why would it still be useful to add the trigger for them? Wouldn't it be rather misleading?

backend/src/main/resources/db/migration/V1.2__add_table_update_tracker.sql
chaoran-chen, line 1: Could you please implement a use case (e.g., for get-release-data) as part of this PR or as a PR that merges into this PR? This code otherwise wouldn't be useful and we should refrain from adding "dead code".

chaoran-chen: I'm curious about how much overhead this approach would add to the backend code base and it would be very helpful to see it implemented for `get-release-data` to decide whether it is a good idea.

backend/src/main/resources/db/migration/V1.2__add_table_update_tracker.sql
```diff
@@ -0,0 +1,42 @@
...
+        EXECUTE format('
+            CREATE OR REPLACE TRIGGER update_tracker_trigger
+            AFTER INSERT OR UPDATE OR DELETE OR TRUNCATE ON %I
+            FOR EACH STATEMENT
+            EXECUTE FUNCTION update_table_tracker()', table_name);
+    END IF;
+END;
+$$ LANGUAGE plpgsql;
+
+SELECT create_update_trigger_for_table('sequence_entries');
```
chaoran-chen, line 35: Most tables (when reading) use the `sequence_entries_view` (which depends on which depends on `sequence_entries`, `sequence_entries_preprocessed_data`, `current_processing_pipeline` and `all_external_metadata`), so would it be useful to directly track changes to the view instead? Otherwise, the backend would need to calculate it itself which doesn't seem great because so far, it just used the view without the need to know on which tables it is based on.

corneliusroemer: As views aren't written I don't think that works - or how would you attach a trigger to a view?

>  Otherwise, the backend would need to calculate it itself which doesn't seem great because so far, 

The calculation is essentially (pseudocode)
```
max(table_update_tracker.last_update)
```

where one excludes maybe `sequence_entries_preprocessed_data` (or fixes that it doesn't update all the time).

Do you mean you'd construct a new view `last_update_to_released_data` that simply contains the result of the calculation to put the code into sql as opposed to in the backend? I guess one could do that.

chaoran-chen: Either a view as you suggested, or let any change to one of the four tables trigger an update of the entry of the view.

chaoran-chen: The goal is simply to keep the view-related logic together in one place and not divide it up between the SQL scripts and the backend.

anna-parker:
Hmm I wonder if the fact that `current_processing_pipeline` and `sequence_entries_preprocessed_data` are continuously getting updated is actually a sign of a bug in the backend - every update call will mean a database transaction that is not required

anna-parker:
> sequence_entries_preprocessed_data

Just looked into this with @chaoran-chen - thanks! `useNewerProcessingPipelineIfPossible` is constantly performing an update transaction - this is where a change would be required. I just checked with chatgpt and even better than splitting the query into a check and update we can do this with a conditional update: 
```
UPDATE users
SET name = 'John Doe', email = 'john.doe@example.com'
WHERE (name != 'John Doe' OR email != 'john.doe@example.com')
AND user_id = 123;
```

corneliusroemer:
Great! Much easier to get rid of these spurious updates than I feared!

What about the continuous updates to `sequence_entries_preprocessed_data`? Is that also covered by your commit? I'll merge it into this one after checks.

corneliusroemer:
Woops @anna-parker backend tests now fail: https://github.com/loculus-project/loculus/actions/runs/10790922697/job/29927194807?pr=2731

anna-parker:
After further investigation the conditional update that chatgpt suggested does not seem to work, additionally the cleanUpStaleSequencesInProcessing will always perform a delete action on the database even when there is nothing to delete.

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -984,18 +987,34 @@ class SubmissionDatabaseService(
             Clock.System.now().toEpochMilliseconds() - timeToStaleInSeconds * 1000,
         ).toLocalDateTime(TimeZone.UTC)
 
-        val numberDeleted = SequenceEntriesPreprocessedDataTable.deleteWhere {
-            statusIs(IN_PROCESSING) and startedProcessingAtColumn.less(staleDateTime)
+        transaction {
```
fengelniederhammer, line 990: Why do we need this transaction here? The whole service is `@Transactional` already.

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -984,18 +987,34 @@ class SubmissionDatabaseService(
...
+            // Check if there are any stale sequences before attempting to delete
+            val staleSequencesExist = SequenceEntriesPreprocessedDataTable
+                .selectAll()
```
fengelniederhammer, line 993: Blocking: This is equivalent to a `select *`. I guess you don't want to fetch the processed data, including the sequences. Please restrict this to the columns that you actually need.

(The unit tests log the executed SQL statements)

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -1028,16 +1047,37 @@ class SubmissionDatabaseService(
...
+            } else {
+                false
+            }
+            if (pipelineNeedsUpdate) {
+                log.info { "Updating current processing pipeline to newer version: $newVersion" }
+                val now = Clock.System.now().toLocalDateTime(TimeZone.UTC)
+                CurrentProcessingPipelineTable.update(
+                    where = {
+                        CurrentProcessingPipelineTable.versionColumn neq
+                            newVersion!!
```
fengelniederhammer, line 1076:
```suggestion
            if (newVersion == null) {
                return newVersion
            } 
            
            
             val pipelineNeedsUpdate = CurrentProcessingPipelineTable
                    .selectAll().where { CurrentProcessingPipelineTable.versionColumn neq newVersion!! }
                    .limit(1)
                    .empty()
                    .not()
                    
            if (pipelineNeedsUpdate) {
                log.info { "Updating current processing pipeline to newer version: $newVersion" }
                val now = Clock.System.now().toLocalDateTime(TimeZone.UTC)
                CurrentProcessingPipelineTable.update(
                    where = {
                        CurrentProcessingPipelineTable.versionColumn neq
                            newVersion
```

fengelniederhammer: And you can also get rid of the `?: 0L` below.

fengelniederhammer (changes requested):
Is it possible to add a unit test that checks that it works? The new SQL file seems at least medium complex.

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
corneliusroemer: Thanks, didn't know this!

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
corneliusroemer: As this is a single SQL statement (I would hope) doesn't it optimize away the unnecessary columns?

corneliusroemer: What you are suggesting would result in duplication of the needed columns which might be necessary but only if the ORM exposed is bad 🙃

fengelniederhammer: No, how should it know the necessary columns? After `.limit(1)`, you have an iterator over the results. `.empty` is not part of the exposed code anymore (AFAIK), so the generated SQL statement has to be general enough to support anything else, too.

corneliusroemer: Ah I see, I would have hoped/expected the empty is part of the ORM. We need to look at emitted SQL to see!

fengelniederhammer: Maybe there is an alternative that is part of Exposed? Counting?

backend/src/main/resources/db/migration/V1.2__add_table_update_tracker.sql
corneliusroemer: it's massively reducing backend load in our benchmarks, and hardly adds much code

backend/src/main/resources/db/migration/V1.2__add_table_update_tracker.sql
corneliusroemer: It does now - we've fixed it

corneliusroemer:
ETag works on client side as well:

Brave Browser 2024-09-12 00 35 24

backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
```diff
@@ -42,6 +44,16 @@ class ReleasedDataModel(
             .map { computeAdditionalMetadataFields(it, latestVersions, latestRevocationVersions) }
     }
 
+    @Transactional(readOnly = true)
+    open fun getLastDatabaseWriteETag(): String {
+        val lastUpdateTime = UpdateTrackerTable.selectAll()
+            .mapNotNull { it[UpdateTrackerTable.lastTimeUpdatedDbColumn] }
```
fengelniederhammer, line 50: Isn't this missing a `where table_name in (some subset of tables)`?

backend/src/main/kotlin/org/loculus/backend/service/submission/UpdateTrackerTable.kt
```diff
@@ -0,0 +1,8 @@
+import org.jetbrains.exposed.sql.Table
+
+const val UPDATE_TRACKER_TABLE_NAME = "table_update_tracker"
+
+object UpdateTrackerTable : Table(UPDATE_TRACKER_TABLE_NAME) {
+    val lastTimeUpdatedDbColumn = varchar("last_time_updated", 255).nullable()
+    val tableNameColumn = varchar("table_name", 255)
```
fengelniederhammer, line 7: This doesn't match the declaration in the SQL file.
```suggestion
    val tableNameColumn = text("table_name")
    val lastTimeUpdatedDbColumn = datetime("last_time_updated")

    override val primaryKey = PrimaryKey(tableNameColumn)
```

backend/src/main/kotlin/org/loculus/backend/service/submission/UpdateTrackerTable.kt
```diff
@@ -0,0 +1,8 @@
+import org.jetbrains.exposed.sql.Table
```
fengelniederhammer, line 1: This is missing it's package declaration. (Annoying, but we're still in the Java world)
```suggestion
package org.loculus.backend.service.submission

import org.jetbrains.exposed.sql.Table
```

backend/src/main/resources/db/migration/V1.2__add_table_update_tracker.sql
```diff
@@ -0,0 +1,42 @@
...
+SELECT create_update_trigger_for_table('sequence_entries_preprocessed_data');
+SELECT create_update_trigger_for_table('external_metadata');
+SELECT create_update_trigger_for_table('groups_table');
+SELECT create_update_trigger_for_table('current_processing_pipeline');
+SELECT create_update_trigger_for_table('metadata_upload_aux_table');
+SELECT create_update_trigger_for_table('sequence_upload_aux_table');
+SELECT create_update_trigger_for_table('user_groups_table');
```
fengelniederhammer, line 42: Let's add a terminal new line
```suggestion
SELECT create_update_trigger_for_table('user_groups_table');

```

backend/src/test/kotlin/org/loculus/backend/controller/submission/GetReleasedDataEndpointTest.kt
```diff
@@ -53,11 +56,17 @@ class GetReleasedDataEndpointTest(
...
         val response = submissionControllerClient.getReleasedData()
 
         val responseBody = response.expectNdjsonAndGetContent<ProcessedData<GeneticSequence>>()
         assertThat(responseBody, `is`(emptyList()))
+        response.andExpect(status().isOk)
+            .andExpect(header().exists(ETAG))
+            .andExpect { result ->
+                val etag = result.response.getHeader(ETAG)
+                assertThat(etag, `is`(notNullValue()))
+            }
```
fengelniederhammer, line 69: That would be a more intuitive order of calls:

```suggestion
        val response = submissionControllerClient.getReleasedData()
            .andExpect(status().isOk)
            .andExpect(header().string(ETAG, notNullValue()))

        val responseBody = response.expectNdjsonAndGetContent<ProcessedData<GeneticSequence>>()
        assertThat(responseBody, `is`(emptyList()))
```

backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
```diff
@@ -42,6 +44,16 @@ class ReleasedDataModel(
...
+            .maxOrNull()
+            ?.replace(" ", "Z")
```
fengelniederhammer, line 52: Why is this necessary? Wouldn't it be better to use some sort of datetime object and format that to the desired format?

backend/src/test/kotlin/org/loculus/backend/controller/submission/GetReleasedDataEndpointTest.kt
```diff
@@ -113,6 +122,39 @@ class GetReleasedDataEndpointTest(
...
+            ifNoneMatch = initialEtag,
+        )
+
+        responseAfterMoreDataAdded.andExpect(status().isOk)
+            .andExpect(header().exists(ETAG))
+            .andExpect { result ->
+                val newEtag = result.response.getHeader(ETAG)
+                assertThat(newEtag, `is`(notNullValue()))
+                assertThat(newEtag, greaterThan(initialEtag))
+            }
```
fengelniederhammer, line 155:
```suggestion
            .andExpect(header().string(ETAG, notNullValue()))
            .andExpect(header().string(ETAG, greaterThan(initialEtag)))
```

kubernetes/loculus/silo_import_job.sh
fengelniederhammer, line 1: To be honest - this script is horrible and IMO close to being unmaintainable.

Let's maybe first migrate it to a proper language (Python?) and then do these changes? My impression is that especially these latest changes would be a lot easier, since you don't need to store everything in a file, but you can have variables that hold state and functions that return something.

kubernetes/loculus/silo_import_wrapper.sh
fengelniederhammer, line 1: Same here - this is supposed to be a simple cronjob, now it's escalating to something where it's easy to fall into one of the many bash pitfalls.

fengelniederhammer (changes requested):

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
```diff
@@ -252,23 +242,23 @@ class SubmissionController(
     fun getReleasedData(
         @PathVariable @Valid organism: Organism,
         @RequestParam compression: CompressionFormat?,
+        @RequestHeader(value = HttpHeaders.IF_NONE_MATCH, required = false) ifNoneMatch: String?,
     ): ResponseEntity<StreamingResponseBody> {
+        val lastDatabaseWriteETag = releasedDataModel.getLastDatabaseWriteETag()
+        if (ifNoneMatch == lastDatabaseWriteETag) return ResponseEntity.status(HttpStatus.NOT_MODIFIED).build()
```
fengelniederhammer, line 248: It would also be good to add the HttpStatus.NOT_MODIFIED response possibility to Swagger.

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
```diff
@@ -252,23 +242,23 @@ class SubmissionController(
     fun getReleasedData(
         @PathVariable @Valid organism: Organism,
         @RequestParam compression: CompressionFormat?,
+        @RequestHeader(value = HttpHeaders.IF_NONE_MATCH, required = false) ifNoneMatch: String?,
```
anna-parker, line 245: Can we change this name? `if_changed_since` for example makes it clear that the tag is a timestamp.

corneliusroemer: it doesn't matter whether it's a timestamp or not - if it's not equal, it needs to be refreshed. this keeps it close to etag spec/as vanilla as possible

kubernetes/loculus/silo_import_job.sh
corneliusroemer: Yes, but:
1. We considered this and decided we'll refactor separately
2. The script currently runs in silo docker image which doesn't have Python, so we'd need to add a whole new docker build to allow Python
3. Rewrite likely introduces new errors, the current thing works, it's not that bad once one wraps one's head round. I've been maintaining it for a while (it was originally written by Tobias and I took it over to add hash checks etc)
4. We've tested it and it does the job, we'll happily make a rewrite issue (if we haven't done so already) @anna-parker

kubernetes/loculus/silo_import_wrapper.sh
corneliusroemer: Which bash pitfalls is it falling into?

backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
corneliusroemer: No, we're using whatever is in the database - not converting from/to timestamp avoids the risks of timezone change etc.

We're leaving all logic in the database here, that's safest. It doesn't mater what it is. The reason I replace `" "` with `"Z"` is so that the bash script doesn't choke - otherwise no reason, it doesn't really matter what this etag is as long as it's ascending.

backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
```diff
@@ -27,13 +29,13 @@ import org.springframework.transaction.annotation.Transactional
 private val log = KotlinLogging.logger { }
 
 @Service
-class ReleasedDataModel(
+open class ReleasedDataModel(
```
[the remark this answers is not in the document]
corneliusroemer: I don't know, I fixed the lints wherever I saw them. Maybe make an issue and have a look at new kotlin lints? I don't think this needs to be part of this PR.

backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
corneliusroemer: We only track the tables that matter at the moment, see the migration - we could later subset this, that's true.

backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
fengelniederhammer: The let's use numeric timestamps? They are guaranteed to be comparable.

backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
fengelniederhammer: I'd do it now. If someone else later adds other tables to the tracking, adapting this here is very likely to be forgotten.

kubernetes/loculus/silo_import_job.sh
anna-parker: Issue made here: https://github.com/loculus-project/loculus/issues/2765

backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
anna-parker: I think we can still convert to timestamp as we specify that we are using utc time in the sql tables and we can test that the tags are less than utc time now in our tests: 
```
lessThanOrEqualTo(Clock.System.now().toLocalDateTime(TimeZone.UTC).toTimestamp())
```
As Fabian says then we also know for certain that the value is numeric (the bash scripts require numeric values for the comparison)

corneliusroemer: Don't we not have microsecond precision with this? There's really no reason to do any conversion here - every conversion adds a chance of introducing an error.

We treat whatever the db outputs as an etag - that's it. No need for _any_ processing.

fengelniederhammer: I still find the `Z` weird. Then let's simply remove whitespaces?

kubernetes/loculus/silo_import_wrapper.sh
fengelniederhammer: Nothing specific - there are many in general.

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
anna-parker: Ah I see this is standard when using e-tags: https://www.geeksforgeeks.org/http-headers-if-none-match/

corneliusroemer:
This decreases backend/database load (and hence costs) by a lot! As I expected when making #2092 🎉 

iTerm2 2024-09-12 10 03 36

backend/src/main/kotlin/org/loculus/backend/service/submission/UpdateTrackerTable.kt
corneliusroemer: That's on purpose, whatever comes out of the db is a blackbox ordered string (ascending) - that's all we need, no conversion necessary, any conversion may introduce bugs

fengelniederhammer: Then let's make it a `text`.

fengelniederhammer: But I'm not so sure about it. The Etag should be a comparable, but otherwise black-box string. At this point, we can still be specific about the type. Maybe this table will be used in another context later, where it's actually better to have the actual type.

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionControllerDescriptions.kt
```diff
@@ -145,7 +145,9 @@ and roll back the whole transaction.
 
 const val GET_RELEASED_DATA_DESCRIPTION = """
 Get released data as a stream of NDJSON.
-This returns all accession versions that have the status 'APPROVED_FOR_RELEASE' 
+This returns all accession versions that have the status 'APPROVED_FOR_RELEASE'. 
+Optionally add HttpHeader If-Modified-Since in unix timestamp (in seconds), 
+to only retrieve all released data if the database has changed since If-Modified-Since.
```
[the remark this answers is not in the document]
corneliusroemer: @anna-parker could you possibly fix this? 🙏

backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
corneliusroemer: @anna-parker could you subset the tables her? Fabian makes a good point that it doesn't harm to do the subsetting explicitly here

backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
corneliusroemer, line 52:
```suggestion
            // Replace not strictly necessary but does no harm and a) shows UTC, b) simplifies silo import script logic
            ?.replace(" ", "Z")
```

corneliusroemer: Adding comment

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
corneliusroemer: @anna-parker

kubernetes/loculus/silo_import_wrapper.sh
corneliusroemer: Using a linter for the worst ones ;)

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
corneliusroemer: We've done this now, please rereview

backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
corneliusroemer: Comment added

backend/src/test/kotlin/org/loculus/backend/controller/submission/GetReleasedDataEndpointTest.kt
corneliusroemer: @anna-parker

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -984,18 +987,34 @@ class SubmissionDatabaseService(
...
+            // Check if there are any stale sequences before attempting to delete
+            val staleSequencesExist = SequenceEntriesPreprocessedDataTable
+                .selectAll()
+                .where {
+                    SequenceEntriesPreprocessedDataTable.statusIs(PreprocessingStatus.IN_PROCESSING) and
+                        (SequenceEntriesPreprocessedDataTable.startedProcessingAtColumn.less(staleDateTime))
+                }
+                .limit(1)
+                .empty()
+                .not()
```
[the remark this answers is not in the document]
corneliusroemer: Seems to be outdated, but maybe @anna-parker you can still have a quick look?

backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
```diff
@@ -1,5 +1,6 @@
 package org.loculus.backend.model
 
+import UpdateTrackerTable
```
fengelniederhammer, line 3: Let's fix this import

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
```diff
@@ -260,30 +250,34 @@ open class SubmissionController(
...
-        headers.contentType = MediaType.parseMediaType(MediaType.APPLICATION_NDJSON_VALUE)
-        if (compression != null) {
-            headers.add(HttpHeaders.CONTENT_ENCODING, compression.compressionName)
+        val lastDatabaseWriteETag = releasedDataModel.getLastDatabaseWriteETag()
+        if (ifNoneMatch == lastDatabaseWriteETag) {
+            return ResponseEntity.status(HttpStatus.NOT_MODIFIED).build()
         }
 
+        val headers = HttpHeaders()
+        headers.eTag = lastDatabaseWriteETag
```
fengelniederhammer, line 261: Let's add the header to the Swagger Docs.

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -994,64 +996,109 @@ open class SubmissionDatabaseService(
...
-            exec(sql, explicitStatementType = StatementType.SELECT) { rs ->
-                if (rs.next()) {
-                    newVersion = rs.getLong("version")
+        log.info("Checking for newer processing pipeline versions")
+        return transaction {
+            val newVersion = findNewPreprocessingPipelineVersion()
+
+            if (newVersion == null) {
+                return@transaction null
+            }
```
fengelniederhammer, line 1028: My IDE suggests
```suggestion
            val newVersion = findNewPreprocessingPipelineVersion() 
                ?: return@transaction null
```

backend/src/test/kotlin/org/loculus/backend/controller/submission/GetReleasedDataEndpointTest.kt
```diff
@@ -53,13 +56,17 @@ class GetReleasedDataEndpointTest(
...
 
         val responseBody = response.expectNdjsonAndGetContent<ProcessedData<GeneticSequence>>()
         assertThat(responseBody, `is`(emptyList()))
-
         response.andExpect(status().isOk)
+            .andExpect(header().exists(ETAG))
+            .andExpect { result ->
+                val etag = result.response.getHeader(ETAG)
+                assertThat(etag, `is`(notNullValue()))
+            }
```
fengelniederhammer, line 69:
```suggestion
            .andExpect(header().string(ETAG, notNullValue()))
```
does the same in a single line.

backend/src/test/kotlin/org/loculus/backend/controller/submission/GetReleasedDataEndpointTest.kt
```diff
@@ -118,6 +125,35 @@ class GetReleasedDataEndpointTest(
...
+
+        prepareRevokedAndRevocationAndRevisedVersions()
+
+        val responseAfterMoreDataAdded = submissionControllerClient.getReleasedData(
+            ifNoneMatch = initialEtag,
+        )
+
+        responseAfterMoreDataAdded.andExpect(status().isOk)
+            .andExpect(header().string(ETAG, notNullValue()))
+            .andExpect(header().string(ETAG, greaterThan(initialEtag)))
```
fengelniederhammer, line 154: Let's maybe add an assertion that the response body contains something?

fengelniederhammer (changes requested):
A few minor things left, but looks good :+1:

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
corneliusroemer: It should be?

fengelniederhammer: Yes, the return header. You can simply add it to the `@ApiResponse` above.

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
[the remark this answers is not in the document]
anna-parker: outdated - resolving comment

backend/src/test/kotlin/org/loculus/backend/controller/submission/GetReleasedDataEndpointTest.kt
anna-parker: this is also outdated - resolving

backend/src/test/kotlin/org/loculus/backend/controller/submission/GetReleasedDataEndpointTest.kt
anna-parker: done :-)

backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
anna-parker: outdated - resolving

backend/src/test/kotlin/org/loculus/backend/controller/submission/GetReleasedDataEndpointTest.kt
anna-parker: Done :-)

backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
anna-parker: @corneliusroemer I just added this option - I was now actually wondering if for get-released-data we can actually already take a subset, I guess we don't really care about group-related changes so we could alert on changes in tables: 
['sequence_entries', 'sequence_entries_preprocessed_data', 'external_metadata', 'current_processing_pipeline', 'metadata_upload_aux_table', 'sequence_upload_aux_table']

anna-parker: at the moment we still return new data when the groups_table and user_groups_table change

anna-parker: I added this now - please check its ok.

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
```diff
@@ -254,36 +246,53 @@ open class SubmissionController(
...
-            headers.add(HttpHeaders.CONTENT_ENCODING, compression.compressionName)
+        val lastDatabaseWriteETag = releasedDataModel.getLastDatabaseWriteETag(
+            RELEASED_DATA_RELATED_TABLES,
+        )
+        if (ifNoneMatch == lastDatabaseWriteETag) {
+            return ResponseEntity.status(HttpStatus.NOT_MODIFIED).build()
         }
 
+        val headers = HttpHeaders()
+        headers.eTag = lastDatabaseWriteETag
```
fengelniederhammer, line 276: Btw - this has the same issue as the `x-total-records`. In some edge cases it probably possible to:
* compute the etag
* change the data in the DB (changing the actual etag!)
* streaming the updated data with the old etag.

fengelniederhammer (approved):

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
anna-parker: Although not optimal I don't think this edge case is too bad as this means that we stream a more recent database state than would be expected by the etag. In the subsequent request the client will ask for new data using an etag which is older than the data it has. Potentially it will get the same data that it already obtained again but this is an edge case and again still much better than the current state. 

Of course if the order was reverted (i.e. stream data obtained before header calculated) this would be an enormous bug and would make this unusable.

fengelniederhammer: I also don't think that it is too bad.

corneliusroemer: yes that's possible and not a problem - because it just causes one extra request in the future which is inconsequential.

backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
```diff
@@ -26,14 +27,24 @@ import org.springframework.transaction.annotation.Transactional
 
 private val log = KotlinLogging.logger { }
 
+val RELEASED_DATA_RELATED_TABLES: List<String> =
+    listOf(
+        "sequence_entries",
+        "sequence_entries_preprocessed_data",
+        "external_metadata",
+        "current_processing_pipeline",
+        "metadata_upload_aux_table",
```
corneliusroemer, line 36: I actually think we can get rid of those two aux tables from this list

corneliusroemer: but doesn't matter much will merge

PR #2740, opened 2024-09-09
Remove potentially unused website dependencies

theosanderson:
resolves #

preview URL:

### PR Checklist


fengelniederhammer:
There is tooling that can detect that:
https://github.com/GenSpectrum/dashboard-components/blob/979b4b15e16a912aa2183a7337b54ad4775fa3be/components/package.json#L100
It will detect a bunch of false-positives that you can manually ignore:
https://github.com/GenSpectrum/dashboard-components/blob/main/components/.depcheckrc.json

theosanderson:
Yes, thanks - I found these with depcheck :) (after spotting one myself). I didn't add it to the PR only because I did some guessing about which would be false positives and which not so there still may be unnecessary deps left.

corneliusroemer:
I'll make a PR to run depcheck in CI so we keep the clean state :)

corneliusroemer (approved):

fengelniederhammer:
The advantage of using depcheck is that you document which are false-positives.

PR #2742, opened 2024-09-10
chore(deps): separate minor and patch updates in npm dependabot

corneliusroemer:
It happens fairly frequently that npm dependabot updates fail, separating minor and patch will reduce time spent debugging

Inspired by the latest dependabot PR that wasn't quite trivial (multiple unrelated failures, grouping made it harder to figure out what's wrong)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,9 +9,14 @@ updates:
     schedule:
       interval: weekly
     groups:
-      minorAndPatch:
+      minor:
         update-types:
           - "minor"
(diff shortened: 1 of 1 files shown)


theosanderson (approved):

corneliusroemer:
I'm missing a dependabot bisect command - there's been a feature proposal but nothing concrete yet. Also, I'm surprised there isn't yet an "npm bisect" command that let's one split updates into half recursively: https://github.com/dependabot/dependabot-core/issues/8450

Issue #2752, opened 2024-09-10
E2E in Github actions running out of storage? Stalling at "installing system deps"


corneliusroemer:
I noticed taht E2E seems to stall a lot right now.

We're not out of RAM, so maybe storage?

<img width="2219" alt="image" src="https://github.com/user-attachments/assets/44c12ae2-1613-4448-9520-5bb49b9bad84">

<img width="2150" alt="Brave Browser 2024-09-10 14 26 01" src="https://github.com/user-attachments/assets/4c9f0e9e-4e1a-4712-ba43-c3c30edabadf">

<img width="2252" alt="Brave Browser 2024-09-10 14 26 04" src="https://github.com/user-attachments/assets/5527e116-505a-4a4f-af91-c2634ee1d6c7">

corneliusroemer:
Seems to have fixed itself. Also @theosanderson's investigation suggests we have enough storage somehow.

PR #2753, opened 2024-09-10
chore(ci): Use 2core 'larger' runner, with 75GB disk space

corneliusroemer:
Related to #2752 

Using cheapest "larger" runner that has 2 cores and 75GB storage instead of 14GB storage. This might solve the issue with stalling.

### DIFF
--- a/.github/workflows/e2e-k3d.yml
+++ b/.github/workflows/e2e-k3d.yml
@@ -34,7 +34,8 @@ jobs:
     env:
       ALL_BROWSERS: ${{ github.ref == 'refs/heads/main' || github.event.inputs.all_browsers && 'true' || 'false' }}
       sha: ${{ github.event.pull_request.head.sha || github.sha }}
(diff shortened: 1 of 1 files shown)


theosanderson:
Given that E2E failed for this specific run and that the cost implications could be significant I think it would be good to confirm that we are running out of storage before merging this

theosanderson:
It doesn't seem like this is it?: https://github.com/loculus-project/loculus/actions/runs/10793315258

corneliusroemer:
Yeah - not sure, but it wasn't just one run that failed but 3 on main

image

corneliusroemer:
Here it was just timeout due to slower setup, will increase timeout

theosanderson:
My test was running on the current system! I think it's more complicated than it appears - I presume it has 14 GB of storage to use for custom stuff and 50 GB or whatever of baked in standard tools

theosanderson:
I suggest closing this issue. We still have 14 GB "free" with the current set up. Problems on main were possibly some external thing. If they recur we can re-open an issue but it doesn't seem to be about storage.

Issue #2724, opened 2024-09-08
If an error occurs while streaming released data from the backend to SILO, the incomplete streamed data should not be used


theosanderson:
We briefly encountered an issue in production where increased database load led to errors in the backend while streaming `get_released_data` leading to truncated files being received by SILO ingest. SILO ingest used these files normally with the result that our query engine had a reduced dataset. Instead we should in some way detect the problem and not use these data.

I haven't established whether this is possible simply through some HTTP behaviour (that might already exist on the backend) or if we need to do something special. Ultimately if needed we could just have a last line of NDSON that would be { "file_complete": true } or similar.

fengelniederhammer:
We could also set a header for starting to stream the data (e.g. for the expected line count).

corneliusroemer:
@fengelniederhammer good point, even better would be to add a checksum (otherwise last line truncation would be invisible)

Easiest might be to first write the response to file in the backend, meanwhile calculate the checksum, then send that result to client with checksum in header. If we compress with zstd, performance overhead of writing to file should be minimal, and storage not too bad either, e.g. all of SARS-CoV-2 should still not be more than say 5-10GB.

By the way, this might be a good time to reconsider how we can avoid having to redownload unchanged data in the SILO preprocessing script. Right now we stream all of the data every 30s even though it is almost never changed.

corneliusroemer:
I think one reason we are in this pickle is that we use ndjson. If we streamed json, for example, we would naturally get incomplete json if the streaming aborts early.

We now need to invent our own mini protocol to indicate successful end of ndjson stream.

I suppose multiple empty lines works ok for now.

fengelniederhammer:
https://www.rfc-editor.org/rfc/rfc2616#section-3.6.1

If I understand this correctly, then chunked encoding already provides something to determine whether the stream was successfully closed by the server. Is that something that we could use? Maybe we just need to adapt something on client-side when fetching the data?

PR #2763, opened 2024-09-10
WIP: Append two empty JSON objects to mark the end of the NDJSON stream

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/2724

preview URL: https://add-explicit-end-to-strea.loculus.org/

### Summary

Currently we stream sequences using NDJSON for multiple endpoints, including get-released-data which sends clients all data. However, if the stream breaks off the client will not know that the server did not send all sequences. This lead to not all sequences being shown for west nile when we had high database load because the silo import script was calling get-released-data. 

### PR Checklist


chaoran-chen:
Instead of two empty objects, would two empty lines also work?

anna-parker:
Closing as we are going another way and this did not work

PR #2764, opened 2024-09-11
Table update tracker anya

anna-parker:
related to: https://github.com/loculus-project/loculus/issues/2092

preview URL: https://table-update-tracker-anya.loculus.org/

### Summary

1. Adds UpdateTrackerTable to database with last time a table in the database was modified.
2. Adds optional `last-modified-if` header to `get-released-data`, this accepts a Long = unix timestamp. If the database was modified since the value in `last-modified-if` `get-released-data` will respond with all released data and `last-modified` header, where `last-modified` is the time of the last database update. If the value in `last-modified-if` is equal or greater to the last time the database was modified `get-released-data` will respond with 304: NOT_MODIFIED and no `last-modified` header.
3. Updates `silo_import_job.sh` and `silo_import_wrapper.sh` to use the `last-modified-if` header. This significantly reduces the load on the database and backend, times are from after preprocessing has finished (CPU is in mCPU): 

```
kubectl top pod --sum --containers --all-namespaces | egrep "backend|s-database" | egrep "main|tracker-anya" | sort -hk4                                           bs-mbpas-0092: Wed Sep 11 19:51:30 2024

prev-table-update-tracker-anya   loculus-backend-c859b6f4b-gj5k5                                   backend                              19m          421Mi
prev-table-update-tracker-anya   loculus-database-67f7d88557-9mf6v                                 database                             101m         326Mi
prev-main                        loculus-database-645647c886-kh5vf                                 database                             113m         234Mi
prev-main                        loculus-backend-55f5cdb6cb-p7p4f                                  backend                              300m         483Mi
```

### PR Checklist

- [ ] Now or in a later PR ensure that we are compliant https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since#syntax,


backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
```diff
@@ -252,16 +256,30 @@ class SubmissionController(
     fun getReleasedData(
         @PathVariable @Valid organism: Organism,
         @RequestParam compression: CompressionFormat?,
+        @RequestHeader(value = HttpHeaders.IF_MODIFIED_SINCE, required = false) ifModifiedSince: Long?,
```
corneliusroemer, line 259: If we pass a long we are kind of violating HTTP specs which have a specific (annoying) format for this particular header. We might want to choose our own header field if we are not spec compliant to not confuse people.

We probably want extra millisecond precision here just to make sure we don't miss an update.

To make a potential cache race condition less impactful, we could always request all data after time X, say 1hr, then we are at most that much out of date, but this is best handled in client I think

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
```diff
@@ -252,16 +256,30 @@ class SubmissionController(
...
         if (compression != null) {
             headers.add(HttpHeaders.CONTENT_ENCODING, compression.compressionName)
         }
 
-        val streamBody = streamTransactioned(compression) { releasedDataModel.getReleasedData(organism) }
+        var lastTime: Long? = null
 
-        return ResponseEntity(streamBody, headers, HttpStatus.OK)
+        transaction {
+            lastTime = UpdateTrackerTable
```
corneliusroemer, line 270: No longer using the view right? That's ok I guess.

kubernetes/loculus/silo_import_job.sh
```diff
@@ -90,63 +123,64 @@ preprocessing() {
...
-    echo "SiloApi command succeeded"
-    echo "Removing touchfile $new_input_touchfile to indicate successful processing"
+    echo "SiloApi command succeeded" >&2
+    echo "Removing touchfile $new_input_touchfile to indicate successful processing" >&2
     rm -f "$new_input_touchfile"
   fi
 
-  echo "preprocessing for $current_timestamp done"
-  echo
+  echo "preprocessing for $current_timestamp done" >&2
```
corneliusroemer, line 135: Why are you adding all these >&2?

kubernetes/loculus/silo_import_wrapper.sh
```diff
@@ -1,7 +1,8 @@
 #!/bin/bash
 
+lastSnapshot=0
 while true
 do
-    bash /silo_import_job.sh
+    lastSnapshot=$(bash /silo_import_job.sh --lastSnapshot=$lastSnapshot)
```
corneliusroemer, line 6: Ah I see to use the script output. We can also just write the snapshot value to a file. And read from the file.

kubernetes/loculus/silo_import_job.sh
```diff
@@ -10,73 +30,86 @@ new_input_data_dir="$input_data_dir/$current_timestamp"
...
   if [ $exit_code -ne 0 ]; then
-    echo "Curl command failed with exit code $exit_code, cleaning up and exiting."
+    echo "Curl command failed with exit code $exit_code, cleaning up and exiting." >&2
     rm -rf "$new_input_data_dir"
+    echo "$lastSnapshot"
     exit $exit_code
   fi
+  
+  last_modified=$(grep '^last-modified:' "$new_input_header" | awk '{print $2}')
+  echo "$last_modified"
```
corneliusroemer, line 76: Should just use a variable here I think rather than using echo and subshell

corneliusroemer (changes requested):
Cool stuff, amazing this works, just some refactoring comments 😀

corneliusroemer:
Maybe we should turn the bash script into a Python script given the complexity. There are now multiple places where we return things or not. It's hard to wrap head round in bash I think.

We can add a rewrite to the todos though in case this here works ok for now.

kubernetes/loculus/silo_import_wrapper.sh
anna-parker: Yes actually lets do that - much easier to handle edge cases

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
anna-parker: Ah true - I see that the expected input is: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since#syntax, so we could check if there is a way to convert a string to a sql.Timestamp object and then compare.

anna-parker: Or yeah we just make our own header field with unix timestamp - that is easiest

kubernetes/loculus/silo_import_job.sh
fengelniederhammer, line 1: Maybe we should migrate this to a proper language? It's running in an Ubuntu base image now, Python would probably work.
(not necessarily a task for this PR)

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
```diff
@@ -252,16 +256,30 @@ class SubmissionController(
...
+                .selectAll() // Select all rows
+                .mapNotNull { it[UpdateTrackerTable.lastTimeUpdatedDbColumn] } // Extract non-null datetime values
+                .maxOrNull()?.toTimestamp() // Find the maximum value
+        }
```
fengelniederhammer, line 274: Actually we had the agreement that we don't put much logic in controller methods (especially database access). I would prefer if this was in a separate model/service class.

kubernetes/loculus/silo_import_job.sh
anna-parker: Issue raised in https://github.com/loculus-project/loculus/issues/2765

PR #2768, opened 2024-09-11
feat(backend,prepro): use etag to reduce database load

corneliusroemer:
relates to #2092 

preview URL: https://etag-prepro.loculus.org

### Summary
Send etag to /extract-unprocessed-sequences to reduce db/backend load

### Screenshot
Big reduction in db and backend load, approximately to 1/4 to 1/3 of the previous base load

Brave Browser 2024-09-12 03 18 16

### PR Checklist

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
@@ -133,8 +133,20 @@ open class SubmissionController(
                 schema = Schema(implementation = UnprocessedData::class),
             ),
         ],
+        headers = [
+            Header(
+                name = "eTag",
+                description = "Last database write Etag",
+                schema = Schema(type = "integer"),
+            ),
+        ],
     )
-    @ApiResponse(responseCode = "304", description = "Not Modified")
+    @ApiResponse(
+        responseCode = "304",
+        description =
+        "No database changes since last request " +
+            "(Etag in HttpHeaders.IF_NONE_MATCH matches lastDatabaseWriteETag)",
+    )
     @ApiResponse(responseCode = "422", description = EXTRACT_UNPROCESSED_DATA_ERROR_RESPONSE)
     @PostMapping("/extract-unprocessed-data", produces = [MediaType.APPLICATION_NDJSON_VALUE])
     fun extractUnprocessedData(
--- a/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
@@ -134,6 +134,7 @@ open class SubmissionController(
             ),
         ],
     )
+    @ApiResponse(responseCode = "304", description = "Not Modified")
     @ApiResponse(responseCode = "422", description = EXTRACT_UNPROCESSED_DATA_ERROR_RESPONSE)
     @PostMapping("/extract-unprocessed-data", produces = [MediaType.APPLICATION_NDJSON_VALUE])
     fun extractUnprocessedData(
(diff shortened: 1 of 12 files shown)


preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -725,16 +703,25 @@ def run(config: Config) -> None:
...
             logging.debug("Fetching unprocessed sequences")
-            unprocessed = parse_ndjson(fetch_unprocessed_sequences(config.batch_size, config))
-            if len(unprocessed) == 0:
+            etag, unprocessed = fetch_unprocessed_sequences(etag, config)
+            if not unprocessed:
                 # sleep 1 sec and try again
                 logging.debug("No unprocessed sequences found. Sleeping for 1 second.")
                 time.sleep(1)
                 continue
+            # Reset etag if we have new data, only use if we have no new data
```
anna-parker, line 715: I am not quite sure why you reset this here?
The new etag will be set in 
```
etag, unprocessed = fetch_unprocessed_sequences(etag, config)
```
when there is a change in data. If anything maybe we should reset the etag to "" when there has been no change in e.g. 10min?

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
```diff
@@ -141,6 +141,7 @@ class SubmissionController(
             message = "You can extract at max $MAX_EXTRACTED_SEQUENCE_ENTRIES sequence entries at once.",
```
anna-parker, line 141: again we should probably add the HttpStatus.NOT_MODIFIED status as an ApiResponse code in the swaggerUI

anna-parker (changes requested):
Looks good! Only two places I am a bit unsure about

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
corneliusroemer: This is a good question, I'm pretty sure I wrapped my head around this and it made sense at the time but I clearly need to improve the comments then if even I can't immediately answer your question.

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -725,17 +703,29 @@ def run(config: Config) -> None:
...
+            etag, unprocessed = fetch_unprocessed_sequences(etag, config)
+            if not unprocessed:
                 # sleep 1 sec and try again
                 logging.debug("No unprocessed sequences found. Sleeping for 1 second.")
                 time.sleep(1)
                 continue
-            # Process the sequences, get result as dictionary
-            processed = process_all(unprocessed, dataset_dir, config)
-            # Submit the result
+            # Don't use etag if we got data, as there might be more
```
anna-parker, line 720:
```suggestion
            # Don't use etag if we just got data, preprocessing only asks for 100 sequences to process at a time, so there might be more
```

anna-parker (approved):

backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
```diff
@@ -1,5 +1,6 @@
 package org.loculus.backend.model
 
+import UpdateTrackerTable
```
corneliusroemer, line 3: No clue where that is from

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
```diff
@@ -135,6 +134,7 @@ open class SubmissionController(
             ),
         ],
     )
+    @ApiResponse(responseCode = "304", description = "Not Modified")
```
anna-parker, line 137: lets maybe add a test for this - we can basically just copy the test we had for get-released-data?

anna-parker: done

PR #2772, opened 2024-09-12
Add latitude and longitude to default metadata values.

anna-parker:
resolves #

preview URL: https://add-lat-long.loculus.org/

### Screenshot

ENA accepts latitude and longitude values - we do not, as some submitters have these values it would be great to also add this option for compatibility. 

I used the decimal degree format as this is expected by ENA: https://www.ebi.ac.uk/ena/browser/view/ERC000033 
### PR Checklist

- [x] Checked that latitude and longitude turn up in submission templates

### DIFF
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -100,12 +100,12 @@ defaultOrganismConfig: &defaultOrganismConfig
         displayName: Latitude
         header: Sample details
         type: float
-        guidance: Geo-coordinate latitude in decimal degree (WGS84) format, i.e. values in range -90 to 90.
+        guidance: Geo-coordinate latitude in decimal degree (WGS84) format, i.e. values in range -90 to 90, where positive values are north of the Equator.
       - name: geoLocLongitude
         displayName: Longitude
         header: Sample details
         type: float
-        guidance: Geo-coordinate longitude in decimal degree (WGS84) format, i.e. values in range -180 to 180.
+        guidance: Geo-coordinate longitude in decimal degree (WGS84) format, i.e. values in range -180 to 180, where positive values are east of the Prime Meridian.
       - name: geoLocCountry
         displayName: Collection country
         generateIndex: true
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -100,12 +100,12 @@ defaultOrganismConfig: &defaultOrganismConfig
         displayName: Latitude
         header: Sample details
         type: float
-        guidance: Geo-coordinate latitude in decimal degree format, i.e. values in range -90 to 90.
+        guidance: Geo-coordinate latitude in decimal degree (WGS84) format, i.e. values in range -90 to 90.
       - name: geoLocLongitude
         displayName: Longitude
         header: Sample details
         type: float
-        guidance: Geo-coordinate longitude in decimal degree format, i.e. values in range -180 to 180.
+        guidance: Geo-coordinate longitude in decimal degree (WGS84) format, i.e. values in range -180 to 180.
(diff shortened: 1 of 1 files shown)


corneliusroemer (changes requested):
Longitude needs 360 degrees 🫣

kubernetes/loculus/values.yaml
```diff
@@ -96,6 +96,16 @@ defaultOrganismConfig: &defaultOrganismConfig
         noInput: true
         perSegment: true
         oneHeader: true
+      - name: latitude
+        displayName: Latitude
+        header: Sample details
+        type: float
+        guidance: Decimal degree format of latitude, i.e. values in range -90 to 90.
```
fengelniederhammer, line 103: Out of interest: Latitude and longitude of what?

corneliusroemer:
We might also ask for the coordinate reference system used as these can matter on small scales
https://en.m.wikipedia.org/wiki/Spatial_reference_system

kubernetes/loculus/values.yaml
corneliusroemer: Geocoordinates

anna-parker:
> We might also ask for the coordinate reference system used as these can matter on small scales https://en.m.wikipedia.org/wiki/Spatial_reference_system

I would suggest we keep this simple for now - INSDC doesn't have a field for this either

corneliusroemer:
Ok, I guess then we could specify that if people are aware and have a
choice they should use WGS84 which is the one that gps gives you and Google
maps etc uses

anna-parker:
> Ok, I guess then we could specify that if people are aware and have a
choice they should use WGS84 which is the one that gps gives you and Google
maps etc uses

Done

kubernetes/loculus/values.yaml
```diff
@@ -96,6 +96,16 @@ defaultOrganismConfig: &defaultOrganismConfig
...
+      - name: geoLocLatitude
+        displayName: Latitude
+        header: Sample details
+        type: float
+        guidance: Geo-coordinate latitude in decimal degree (WGS84) format, i.e. values in range -90 to 90.
+      - name: geoLocLongitude
+        displayName: Longitude
+        header: Sample details
+        type: float
+        guidance: Geo-coordinate longitude in decimal degree (WGS84) format, i.e. values in range -180 to 180.
```
corneliusroemer, line 108: Maybe just for sanity checks, we could add that West of Greenwich is Negative, so US is negative, Germany positive

kubernetes/loculus/values.yaml
```diff
@@ -96,6 +96,16 @@ defaultOrganismConfig: &defaultOrganismConfig
         noInput: true
         perSegment: true
         oneHeader: true
+      - name: geoLocLatitude
+        displayName: Latitude
+        header: Sample details
+        type: float
+        guidance: Geo-coordinate latitude in decimal degree (WGS84) format, i.e. values in range -90 to 90.
```
corneliusroemer, line 103: I think positive is north, maybe worth adding again so people don't need to look up (I'm not sure we use the guidance yet but I hope we will soon)

corneliusroemer (approved):

kubernetes/loculus/values.yaml
anna-parker: I really don't think that we need to add this: the standard definition of latitude and longitude is
- longitude: positive numbers are east of the Prime Meridian and negative numbers are west of it
- latitude positive values are north of the equator, negative values are south of the equator. 

So we can add this but if someone is adding data in this field they will be using a standard tool for calculating coordinates anyways - as long as we say we want decimal degree it should be clear.

PR #2773, opened 2024-09-12
fix(backend, silo-preprocessing):  add new header x-total-records to get-released-data to enable streaming result verification

anna-parker:
partially resolves https://github.com/loculus-project/loculus/issues/2724

preview URL: http://add-count-in-header.loculus.org/

### Summary
As we send data in ndjson format, clients (such as the silo_import_job) have no way of knowing if the data they have received is complete or if the stream was interrupted early.

To fix this, /get-released-data adds header x-rowcount to the response with the total number of expected records in the accompanying stream.

Clients can then compare this number with the responseBody to see if they have received all records and discard/re-request in case the stream was aborted early.

### Small caveat

The actual streaming is executed after `getReleased` is defined. This means the count calculated for the header and the results of the stream are not inside the same database transaction. This means that if the table was updated in between calculating the record and streaming there could be more records in the stream than expected from the header. 

`silo_import_job` will not use the data received from `get-released-data` if `x-total-records` header and the actual record count do not match. Meaning it could take slightly longer for new sequences to appear in the database but this should be negligible. 

This only causes a problem if the stream is cut off after exactly as many records as expected from the header. (e.g. count is 20, add one new sequence, stream should include 21 records, stream is cut off after record 20, client thinks stream completed successfully - but we lost an arbitrary sequence).

### PR Checklist

- [x] Add zstd ~(and probably jq to the silo-preprocessing docker image)~ - jq is already there: https://github.com/GenSpectrum/LAPIS-SILO/pull/574, update: I can install zstd after the container is set up, not very clean but this works
- [x] Break stream locally and check that this still works as intended.
Test locally by adding a Thread.sleep(1000) between stream iterations in `streamAsNdjson` and killing the backend mid-stream, the import script responds as expected:
```
Response should contain a total of : 2141 records
jq: parse error: Invalid numeric literal at line 30, column 3
Response contained a total of : 29 records
Expected and actual number of records are not the same
```

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
@@ -66,7 +66,7 @@ private val log = KotlinLogging.logger { }
 @RequestMapping("/{organism}")
 @Validated
 @SecurityRequirement(name = "bearerAuth")
-class SubmissionController(
+open class SubmissionController(
     private val submitModel: SubmitModel,
     private val releasedDataModel: ReleasedDataModel,
     private val submissionDatabaseService: SubmissionDatabaseService,
@@ -260,7 +260,7 @@ class SubmissionController(
     )
     @GetMapping("/get-released-data", produces = [MediaType.APPLICATION_NDJSON_VALUE])
     @Transactional(isolation = Isolation.REPEATABLE_READ) // All operations will be performed on the same snapshot
-    fun getReleasedData(
+    open fun getReleasedData(
         @PathVariable @Valid organism: Organism,
         @RequestParam compression: CompressionFormat?,
     ): ResponseEntity<StreamingResponseBody> {
@@ -269,9 +269,13 @@ class SubmissionController(
         if (compression != null) {
             headers.add(HttpHeaders.CONTENT_ENCODING, compression.compressionName)
         }
+
         val totalRecords = submissionDatabaseService.countReleasedSubmissions(organism)
         headers.add("x-total-records", totalRecords.toString())
-
+        // There's a possibility that the totalRecords change between the count and the actual query
+        // this is not too bad, if the client ends up with a few more records than expected
+        // We just need to make sure the etag used is from before the count
+        // Alternatively, we could read once to file while counting and then stream the file
         val streamBody = streamTransactioned(compression) { releasedDataModel.getReleasedData(organism) }
 
         return ResponseEntity(streamBody, headers, HttpStatus.OK)
--- a/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
@@ -43,8 +43,6 @@ import org.springframework.http.HttpHeaders
 import org.springframework.http.HttpStatus
 import org.springframework.http.MediaType
 import org.springframework.http.ResponseEntity
-import org.springframework.transaction.annotation.Isolation
-import org.springframework.transaction.annotation.Transactional
 import org.springframework.validation.annotation.Validated
 import org.springframework.web.bind.annotation.DeleteMapping
 import org.springframework.web.bind.annotation.GetMapping
--- a/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
@@ -259,7 +259,6 @@ open class SubmissionController(
         ],
     )
     @GetMapping("/get-released-data", produces = [MediaType.APPLICATION_NDJSON_VALUE])
-    @Transactional(isolation = Isolation.REPEATABLE_READ) // All operations will be performed on the same snapshot
     open fun getReleasedData(
         @PathVariable @Valid organism: Organism,
         @RequestParam compression: CompressionFormat?,
--- a/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
@@ -257,7 +257,7 @@ open class SubmissionController(
         ],
     )
     @GetMapping("/get-released-data", produces = [MediaType.APPLICATION_NDJSON_VALUE])
-    open fun getReleasedData(
+    fun getReleasedData(
         @PathVariable @Valid organism: Organism,
         @RequestParam compression: CompressionFormat?,
     ): ResponseEntity<StreamingResponseBody> {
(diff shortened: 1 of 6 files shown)


backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
```diff
@@ -247,8 +250,16 @@ class SubmissionController(
...
+        headers = [
+            Header(
+                name = "x-total-records",
+                description = "The total number of records sent in responseBody",
+                schema = Schema(type = "integer"),
+            ),
+        ],
     )
     @GetMapping("/get-released-data", produces = [MediaType.APPLICATION_NDJSON_VALUE])
+    @Transactional(isolation = Isolation.REPEATABLE_READ) // All operations will be performed on the same snapshot
```
fengelniederhammer, line 262: I feat that this won't work - the transaction where the stream happens is executed later. We only pass a callback to the output stream which will be executed somewhen later by Spring.

Taepper: Why do we actually not set our Isolation level to Serializable?

kubernetes/loculus/silo_import_job.sh
```diff
@@ -47,16 +49,29 @@ download_data() {
   ls -l "$new_input_data_dir"
   echo
 
+  expected_record_count=$(grep -i '^x-total-records:' "$new_input_header_path" | awk '{print $2}' | tr -d '[:space:]')
+  echo "Response should contain a total of : $expected_record_count records"
+
+  true_record_count=$(zstd -d -c "$new_input_data_path" | jq -c . | wc -l | tr -d '[:space:]')
```
fengelniederhammer, line 55: Why do we need the jq there? To only count JSON lines?

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
```diff
@@ -247,9 +248,16 @@ class SubmissionController(
...
+            Header(
+                name = "x-total-records",
+                description = "The total number of records sent in responseBody",
+                schema = Schema(type = "integer"),
+            ),
+        ],
     )
     @GetMapping("/get-released-data", produces = [MediaType.APPLICATION_NDJSON_VALUE])
-    fun getReleasedData(
+    open fun getReleasedData(
```
fengelniederhammer, line 260:
```suggestion
    fun getReleasedData(
```

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
fengelniederhammer: What would be the benefit of that?

fengelniederhammer (approved):

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
```diff
@@ -259,6 +267,12 @@ class SubmissionController(
             headers.add(HttpHeaders.CONTENT_ENCODING, compression.compressionName)
         }
 
+        val totalRecords = submissionDatabaseService.countReleasedSubmissions(organism)
+        headers.add("x-total-records", totalRecords.toString())
+        // There's a possibility that the totalRecords change between the count and the actual query
+        // this is not too bad, if the client ends up with a few more records than expected
+        // We just need to make sure the etag used is from before the count
+        // Alternatively, we could read once to file while counting and then stream the file
```
chaoran-chen, line 275: Could you please create an issue so that we can fix this at some point? :)

chaoran-chen (approved):
Thanks!

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
fengelniederhammer: And mention the issue in the comment.

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
Taepper: Some information is also [here](https://en.wikipedia.org/wiki/Isolation_(database_systems)#Repeatable_reads) and [here](https://stackoverflow.com/questions/48417632/why-write-skew-can-happen-in-repeatable-reads).

One example:

> You are writing an application for doctors to manage their on-call shifts at a hospital. The hospital usually tries to have several doctors on call at any one time, but it absolutely must have at least one doctor on call. Doctors can give up their shifts (e.g., if they are sick themselves), provided that at least one colleague remains on call in that shift

```
BEGIN TRANSACTION;
    SELECT * FROM doctors
        WHERE on_call = true
        AND shift_id = 1234;
    if (current_on_call >= 2) {
        UPDATE doctors
        SET on_call = false WHERE name = 'Alice' AND shift_id = 1234;
    }
COMMIT;  
```

In repeatable read there is a race condition possible. I.e. the result will not always be the same as if the transactions were run one after the other, which is exactly what serializable guarantees

Taepper: I actually looked it up and Transactional uses the database's default, which is read-commited (one level lower than repeatable read) for Postgres

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
corneliusroemer: Sure good idea, in the future you can also immediately create the issue yourself

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -550,6 +550,14 @@ class SubmissionDatabaseService(
             .associate { it[SequenceEntriesView.accessionColumn] to it[maxVersionExpression]!! }
     }
 
+    fun countReleasedSubmissions(organism: Organism): Long = SequenceEntriesView.select(
```
corneliusroemer, line 553:
```suggestion
    // Make sure to keep in sync with streamReleasedSubmissions query
    fun countReleasedSubmissions(organism: Organism): Long = SequenceEntriesView.select(
```

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -550,6 +550,14 @@ class SubmissionDatabaseService(
...
+        SequenceEntriesView.accessionColumn,
+    ).where {
+        SequenceEntriesView.statusIs(Status.APPROVED_FOR_RELEASE) and SequenceEntriesView.organismIs(
+            organism,
+        )
+    }.count()
+
```
corneliusroemer, line 560:
```suggestion
// Make sure to keep in sync with countReleasedSubmissions query
```

kubernetes/loculus/silo_import_job.sh
corneliusroemer, line 55:
```suggestion
  # jq validates each individual json object, to catch truncated lines
  true_record_count=$(zstd -d -c "$new_input_data_path" | jq -c . | wc -l | tr -d '[:space:]')
```

Issue #2567, opened 2024-08-25
Allow customizing where github button points to


theosanderson:
Currently the GitHub icon at the bottom right points to Loculus which may not always be what we want

PR #2797, opened 2024-09-15
feat(website): Add custom gitHub url setting

theosanderson:
Relates to: https://github.com/pathoplexus/pathoplexus/issues/131
Resolves: https://github.com/loculus-project/loculus/issues/2567

This feature allows the GitHub icon at the bottom of the page to be customized to point to a different URL.

At some point we should also consider whether consensus is in favour of adding a "Powered by Loculus" message (discussion in https://github.com/pathoplexus/pathoplexus/issues/131)


website/src/layouts/BaseLayout.astro
chaoran-chen, line 1: Thanks, Theo, for implementing this! As indicated in the discussion, I'm in favor of this (only) in combination with adding a "Powered by Loculus" note.

theosanderson: (thanks for confirming!)

theosanderson: Hi Chaoran, per the poll I just carried out (and adding your, and Cornelius's, views which I think I already know from prior discussion), I think there is majority support for this change. If it's OK I'd like to proceed with that, and we can still add anything else in the future.

chaoran-chen: Hi Theo, sorry for the delay with drafting something with the "powered by loculus" note (I promised doing that last week)! Can I quickly do it now and then we can see whether we can merge both at the same time?

theosanderson: Np on the delay. It's not a big deal for me to wait another few hours, but as long as we do have a fair consensus on this first change I don't really see a reason for coupling these.

chaoran-chen: Here it is: #2901 :) I'm happy to proceed with this already (and hopefully we will come to a consensus w.r.t. #2901 soon)

website/src/layouts/BaseLayout.astro
```diff
@@ -72,7 +72,7 @@ const lastTimeBannerWasClosed = Astro.cookies.get('lastTimeBannerWasClosed')?.va
                             <OrganismSelector implicitOrganism={implicitOrganism} />
                         </div>
                     </div>
-                    <Navigation implicitOrganism={implicitOrganism} />
+                    <Navigation implicitOrganism={implicitOrganism} gitHubMainUrl={gitHubMainUrl} />
```
corneliusroemer, line 75: We could just pass down the whole config instead of unpacking in each component, this reduces number of changes to add extra config items in the future. But not necessary to do now, just a thought when seeing this PR

corneliusroemer (approved):

website/src/layouts/BaseLayout.astro
theosanderson: Yeah - sometimes on the website this isn't desirable as it can pass heavy stuff to the client side that doesn't need to be passed. Oh and actually maybe that does apply here to the sandwich menu (but I can't remember the details, maybe its not too heavy). I'll leave it for now.

PR #2802, opened 2024-09-15
feat(deployment): Allow detailed customizable resources

### LINKED ISSUE #2642: Allow fine-grained control of resources for kubernetes charts
Relates to https://github.com/pathoplexus/pathoplexus/issues/43

I think a nice way to do this might be to have:

Values.yaml:

```
resources:
    [deployment-name]:
         requests:
            memory: "256Mi"
            cpu: "500m"
         limits:
            memory: "512Mi"
            cpu: "1000m"
```

And for any of those set the setting would be taken account of, if set.

theosanderson:
https://resources.loculus.org/

Resolves #2642

Moves specification of resources to `values.yaml`, allowing them to be customised


corneliusroemer (approved):
Great stuff, really cool to be able to customize so flexibly now, this allows us to protect production much better!

I tested with `helm template kubernetes/loculus` and it looks perfect.

PR #2810, opened 2024-09-16
[do not merge] Temporary pop up that cross ref is down when user clicks to create DOI

anna-parker:
resolves #

preview URL:

### Summary

Uglier back up option: 
image
When user clicks on generate DOI button pop up tells user that crossref is temporarily down

### PR Checklist


theosanderson (approved):
LGTM - thanks a lot!

Issue #2636, opened 2024-08-27
South Sudan missing in accepted countries due to missing from ENA checklist


corneliusroemer:
Loculus' default list of accepted countries misses South Sudan (independent since 2011) and possibly some more recently independent countries.

We use the ENA checklist https://www.ebi.ac.uk/ena/browser/view/ERC000011 - so that list might have to be updates as well.

South Sudan _is_ in the official INSDC country list, it's just missing from _that_ checklist.

https://www.ncbi.nlm.nih.gov/genbank/collab/country/

We might also want to add historical countries (in case they haven't been added) - as it is customary to describe geolocations by the names at the time of collection, at least during modern times.

anna-parker:
I did not enforced the accepted countries list for samples from INSDC so I don't think that this is high priority.

chaoran-chen:
Current question: We don't know whether ENA will accept sequences with "South Sudan" as country. We will test it.

PR #2813, opened 2024-09-17
Add countries in NIH list which are missing from ENA list

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/2636

preview URL: https://add-nih-countries.loculus.org/

### Summary

As calculated in https://github.com/loculus-project/loculus/pull/2335 these are the countries on the NIH list yet missing from the ENA list, adding them here for consistency. 

For reference here is:
- The ENA checklist: https://www.ebi.ac.uk/ena/browser/api/xml/ERC000011
- The NIH checklist https://www.ncbi.nlm.nih.gov/genbank/collab/country/.

I added these countries to the end of the list. Currently we do not use the ENA checklist for verification - if this changes these countries will need to be properly mapped.

### PR Checklist

- [ ] Check if there are further countries in https://www.insdc.org/submitting-standards/geo_loc_name-qualifier-vocabulary/ that we should add

### DIFF
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -395,6 +395,15 @@ defaultOrganismConfig: &defaultOrganismConfig
               - Yemen
               - Zambia
               - Zimbabwe
+              - Eswatini # Add countries in NIH list which are missing from ENA list
+              - Line Islands
+              - Micronesia, Federated States of
+              - North Macedonia
+              - Saint Barthelemy
+              - Saint Martin
+              - South Sudan
+              - State of Palestine
+              - Timor-Leste
       - name: geoLocAdmin1
         displayName: Collection subdivision level 1
         generateIndex: true
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -140,8 +140,8 @@ defaultOrganismConfig: &defaultOrganismConfig
               - Azerbaijan
               - Bahamas
               - Bahrain
-              - Baker Island
               - Baltic Sea
+              - Baker Island
               - Bangladesh
               - Barbados
               - Bassas da India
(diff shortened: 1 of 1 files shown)


kubernetes/loculus/values.yaml
```diff
@@ -395,6 +395,15 @@ defaultOrganismConfig: &defaultOrganismConfig
               - Yemen
               - Zambia
               - Zimbabwe
+              - Eswatini # Add countries in NIH list which are missing from ENA list
```
corneliusroemer, line 398: I'd probably add them alphabetically?

corneliusroemer (approved):

anna-parker:
We should probably also add the historical names in case someone submits older data: 
```
Belgian Congo
British Guiana
Burma
Czechoslovakia
Czech Republic
East Timor
Korea
Macedonia
Micronesia
Netherlands Antilles
Serbia and Montenegro
Siam
Swaziland
The former Yugoslav Republic of Macedonia
USSR
Yugoslavia
Zaire
```

corneliusroemer:
How is Czech Republic a historic name? 👀

corneliusroemer:
They now call it Czechia. I see. It would be great if there was a cutoff date for each historical name. One shouldn't use Soviet Union past 1990 etc. We could come up with our own cut offs. Czechia/Czech Republic is weird though.

What about Turkey/Türkiye? It still says Turkey but I bet Turkey would prefer Türkiye.

kubernetes/loculus/values.yaml
```diff
@@ -1557,9 +1571,9 @@ resources:
...
       memory: "3Gi"
 defaultResources:
-    requests:
-      memory: "200Mi"
-      cpu: "20m"
-    limits:
-       memory: "1Gi"
-       cpu: "20m"
\ No newline at end of file
+  requests:
```
corneliusroemer, line 1574: thanks that's correct indentation now

corneliusroemer (approved):

kubernetes/loculus/values.yaml
anna-parker: done :-)

PR #2814, opened 2024-09-17
feat(website): Only count latest versions for stats on main page

theosanderson:
https://only-count-latest-version.loculus.org/

### DIFF
--- a/website/src/components/IndexPage/getOrganismStatistics.ts
+++ b/website/src/components/IndexPage/getOrganismStatistics.ts
@@ -1,5 +1,4 @@
 import { DateTime, FixedOffsetZone } from 'luxon';
-
 import { LapisClient } from '../../services/lapisClient.ts';
 import { RELEASED_AT_FIELD, VERSION_STATUS_FIELD } from '../../settings.ts';
 import { siloVersionStatuses } from '../../types/lapis';
@@ -68,12 +67,25 @@ const getTotalAndLastUpdatedAt = async (
 const getRecent = async (organism: string, numberDaysAgo: number): Promise<number> => {
     const recentTimestamp = Math.floor(Date.now() / 1000 - numberDaysAgo * 24 * 60 * 60);
     const client = LapisClient.createForOrganism(organism);
-    return (
+    const recentTotal =  (
         await client.call('aggregated', {
             [`${RELEASED_AT_FIELD}From`]: recentTimestamp,
             version: 1,
+            
         })
     )
         .map((x) => x.data[0].count)
         .unwrapOr(0);
-};
+        const recentRevoked =  (
+            await client.call('aggregated', {
+                [`${RELEASED_AT_FIELD}From`]: recentTimestamp,
+                version: 1,
+                [VERSION_STATUS_FIELD]: siloVersionStatuses.revoked
+
+                
+            })
+        )
+            .map((x) => x.data[0].count)
+            .unwrapOr(0);
+            return recentTotal-recentRevoked;
+    };
--- a/website/src/components/IndexPage/getOrganismStatistics.ts
+++ b/website/src/components/IndexPage/getOrganismStatistics.ts
@@ -1,4 +1,5 @@
 import { DateTime, FixedOffsetZone } from 'luxon';
+
 import { LapisClient } from '../../services/lapisClient.ts';
 import { RELEASED_AT_FIELD, VERSION_STATUS_FIELD } from '../../settings.ts';
 import { siloVersionStatuses } from '../../types/lapis';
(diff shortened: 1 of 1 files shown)


website/src/components/IndexPage/getOrganismStatistics.ts
```diff
@@ -49,7 +50,7 @@ const getTotalAndLastUpdatedAt = async (
     const client = LapisClient.createForOrganism(organism);
     return (
         await client.call('aggregated', {
-            version: 1,
+            [VERSION_STATUS_FIELD]: siloVersionStatuses.latestVersion,
```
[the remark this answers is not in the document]
theosanderson: 🤔  good point. I also can't come up with a good way. IMO this change would still be preferable to no change:
- with this change there is a potential edge case, most likely in the first 30 days, where as you say - this could look inconsisent (if we wrote `X submissions in the past month` it would be OK, but the `+` would look odd)
- _without_ this change, as soon as there is a revocation there is a permanent inconsistency in which when you click on an organism the search page shows a different number of total sequences to the index page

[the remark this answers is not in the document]
corneliusroemer: All we'd have to do is subtract the recently added sequences that are revoked here?
https://github.com/loculus-project/loculus/blob/e5ad697d2572fa14849efe41d43538a169bdc77f/website/src/components/IndexPage/getOrganismStatistics.ts#L72-L75

Agreed the small chance of inconsistency does not make the big improvement not woth it.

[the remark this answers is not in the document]
theosanderson: > All we'd have to do is subtract here the recently added sequences that are revoked?

I forget what happens if a sequence is revised and then revoked

Taepper:
How about changing the '+ count' to the following:
Number of sequences that had any new version added within the last 30 days. -> this would mean that we do not count multiple versions for the same sequence multiple times

website/src/components/IndexPage/getOrganismStatistics.ts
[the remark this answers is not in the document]
corneliusroemer: we just:
1. get all sequences that have latest version being revoked
2. filter on release data of their _first_ version being within 1 month
3. count

website/src/components/IndexPage/getOrganismStatistics.ts
```diff
@@ -68,12 +67,25 @@ const getTotalAndLastUpdatedAt = async (
...
         })
     )
         .map((x) => x.data[0].count)
         .unwrapOr(0);
-};
+        const recentRevoked =  (
+            await client.call('aggregated', {
+                [`${RELEASED_AT_FIELD}From`]: recentTimestamp,
+                version: 1,
+                [VERSION_STATUS_FIELD]: siloVersionStatuses.revoked
```
corneliusroemer: Not sure this works? As this filters for sequences that are version 1 AND revoked? Which will be 0?

corneliusroemer: I think we need a 2 step process where we get accessions as intermediate step? Unless I misunderstand SILO capabilities @Taepper

theosanderson: My understanding is that the `versionStatus` field is always either "LATEST_VERSION" or "REVISED" or "REVOKED". It relates to how the version has been affected by other newer versions.

Taepper: I suspect you are correct.. let me think for a few more minutes

Taepper: So what will happen to older version if the newest version is revoked?
If this is possible:
`PP_01.1` `REVISED`
`PP_01.2` `REVOKED`
I think we would need the intermediate step mentioned by Cornelius with current LAPIS/SILO limitations

chaoran-chen: Sounds good to me! Would you like to create an issue to document the edge cases that still need to be solved? Otherwise, I'm happy with the solution.

chaoran-chen (approved):

website/src/components/IndexPage/getOrganismStatistics.ts
corneliusroemer: What does that `versionStatus` field do? I'm confused now.

Taepper: > I'm confused now

Me too 😅

theosanderson: (Sorry, when I resolved this I hadn't seen the comments)

theosanderson: So here is the code that computes SILO versions https://github.com/loculus-project/loculus/blob/cca00bbc9174e3080e7507802ce838f186babc22/backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt#L105

Here is a Claude description:
>First, it checks if the given version is the latest. If so, it returns LATEST_VERSION.
>If not the latest, it checks if there's a revocation version for this accession that's newer than the given version. If so, it returns REVOKED.
>If neither of the above conditions are met, it returns REVISED, indicating that newer versions exist but this version hasn't been revoked.

Thus I think actually there are no edge cases?

chaoran-chen: aah, right, okay, forgot about the exact implementation myself but it makes sense! Then, this should indeed fully cover all cases.

theosanderson:
I tested that this is working and I now believe there are no edge cases 🥳

website/src/components/IndexPage/getOrganismStatistics.ts
theosanderson: (sequences which were submitted, revoked, then reinstated won't be counted as new submissions but that doesn't seem worth an issue)

website/src/components/IndexPage/getOrganismStatistics.ts
```diff
@@ -49,7 +50,8 @@ const getTotalAndLastUpdatedAt = async (
     const client = LapisClient.createForOrganism(organism);
     return (
         await client.call('aggregated', {
-            version: 1,
+            [VERSION_STATUS_FIELD]: siloVersionStatuses.latestVersion,
```
corneliusroemer, line 53: Maybe we should rename the function to `getTotalUnrevokedAndLastUpdatedAt`

theosanderson: Personally I think this is ok as is (but I can live with the other option too)

PR #2817, opened 2024-09-17
refactor(backend): rename enum `SiloVersionStatus` to `VersionStatus` and document the enum

corneliusroemer:
Spawned by this Slack discussion: https://loculus.slack.com/archives/C05G172HL6L/p1726586424473839

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
@@ -295,10 +295,10 @@ enum class PreprocessingStatus {
     FINISHED,
 }
 
-enum class SiloVersionStatus {
-    REVOKED,
-    REVISED,
-    LATEST_VERSION,
+enum class VersionStatus {
+    REVOKED, // Not the highest version, and a higher version is a revocation
+    REVISED, // Not the highest, and there is no higher revocation
+    LATEST_VERSION, // Highest version
 }
 
 enum class CompressionFormat(val compressionName: String) {
--- a/backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
@@ -296,9 +296,9 @@ enum class PreprocessingStatus {
 }
 
 enum class VersionStatus {
-    REVOKED, // Not the highest version, and a higher version is a revocation
-    REVISED, // Not the highest, and there is no higher revocation
-    LATEST_VERSION, // Highest version
+    REVOKED, // This is not the highest version of the sequence entry, and a higher version is a revocation
+    REVISED, // This is not the highest version of the sequence entry, and no higher version is a revocation
+    LATEST_VERSION, // This is the highest version of the sequence entry
 }
 
 enum class CompressionFormat(val compressionName: String) {
--- a/backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
@@ -13,7 +13,7 @@ import org.loculus.backend.api.DataUseTerms
 import org.loculus.backend.api.GeneticSequence
 import org.loculus.backend.api.Organism
 import org.loculus.backend.api.ProcessedData
-import org.loculus.backend.api.SiloVersionStatus
+import org.loculus.backend.api.VersionStatus
 import org.loculus.backend.config.BackendConfig
 import org.loculus.backend.service.submission.RawProcessedData
 import org.loculus.backend.service.submission.SubmissionDatabaseService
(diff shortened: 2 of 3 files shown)


backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
```diff
@@ -295,10 +295,10 @@ enum class PreprocessingStatus {
     FINISHED,
 }
 
-enum class SiloVersionStatus {
-    REVOKED,
-    REVISED,
-    LATEST_VERSION,
+enum class VersionStatus {
+    REVOKED, // Not the highest version, and a higher version is a revocation
+    REVISED, // Not the highest, and there is no higher revocation
```
theosanderson, line 300:
```suggestion
    REVISED, // This is not the highest version, and no higher version is a revocation
```

backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
```diff
@@ -295,10 +295,10 @@ enum class PreprocessingStatus {
...
+    LATEST_VERSION, // Highest version
```
theosanderson, line 301:
```suggestion
    LATEST_VERSION, // This is the highest version of the sequence entry
```

backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
```diff
@@ -102,22 +102,26 @@ class ReleasedDataModel(
         DataUseTerms.Open
     }
 
-    private fun computeSiloVersionStatus(
+    // LATEST_VERSION: Highest version
```
theosanderson, line 105: IMO my suggestions above are ideal

backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
theosanderson, line 1: `siloVersionStatus` exists in the website too

corneliusroemer: Not really blocking IIUC because these are internal changes only? the field is actually called "versionStatus" in export

But yeah we should get rid of the silo there too

theosanderson: Blocking because the way that I found my way to this definition in the backend was by searching with the name used in the website

theosanderson: (so IMO this would be a regression in code understandability if the consistency was lost)

corneliusroemer: See: https://github.com/loculus-project/loculus/pull/2818 :)

corneliusroemer: I'm not at all against renaming website, in fact that was on my list as well, just meant to say that it's not breaking anything - which might have been how blocking could be understood.

backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
```diff
@@ -102,22 +102,26 @@ class ReleasedDataModel(
...
+    // REVOKED: Not highest version, higher version is a revocation
```
corneliusroemer, line 106:
```suggestion
    // REVOKED: This is not the highest version of the sequence entry, and a higher version is a revocation
```

backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
```diff
@@ -102,22 +102,26 @@ class ReleasedDataModel(
...
+    // REVISED: Not the highest version, and there's no higher version that is a revocation
```
corneliusroemer, line 107:
```suggestion
    // REVISED: This is not the highest version of the sequence entry, and no higher version is a revocation
```

backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
corneliusroemer, line 105:
```suggestion
    // LATEST_VERSION: This is the highest version of the sequence entry
```

backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
corneliusroemer: Resolved via #2818

theosanderson: Ah, I think we also often use "blocking" to mean "in the reviewer's view this needs to be resolved before merging" [or in this case resolved in a PR alongside which is fine too]

theosanderson (approved):

PR #2818, opened 2024-09-17
refactor(website): rename `siloVersionStatus` to `versionStatus`

corneliusroemer:
Also some slight tweaks to function names in organism statistics, follow up to #2814

Backend rename (not coupled as versionStatus usage is only internal) is done in: #2817

### DIFF
--- a/website/src/components/IndexPage/getOrganismStatistics.ts
+++ b/website/src/components/IndexPage/getOrganismStatistics.ts
@@ -4,7 +4,6 @@ import { LapisClient } from '../../services/lapisClient.ts';
 import { RELEASED_AT_FIELD, VERSION_STATUS_FIELD, IS_REVOCATION_FIELD } from '../../settings.ts';
 import { versionStatuses } from '../../types/lapis';
 
-// Excluding sequences whose latest version is revoked
 export type OrganismStatistics = {
     totalSequences: number;
     recentSequences: number;
@@ -67,10 +66,16 @@ const getTotalAndLastUpdatedAt = async (
         });
 };
 
+/**
+ * Note: This method undercounts in cases where recently released sequences
+ * are later revoked and then unrevoked (revised), all within the "recency window".
+ * This trade-off allows for a simpler, more efficient query
+ * without needing to fetch individual accession lists.
+ */
 const getRecent = async (organism: string, numberDaysAgo: number): Promise<number> => {
     const recentTimestamp = Math.floor(Date.now() / 1000 - numberDaysAgo * 24 * 60 * 60);
     const client = LapisClient.createForOrganism(organism);
-    const recentTotalIncludingRevoked = (
+    const recentlyReleasedTotal = (
         await client.call('aggregated', {
             [`${RELEASED_AT_FIELD}From`]: recentTimestamp,
             version: 1,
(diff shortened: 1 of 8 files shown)


website/src/components/IndexPage/getOrganismStatistics.ts
```diff
@@ -2,8 +2,9 @@ import { DateTime, FixedOffsetZone } from 'luxon';
 
 import { LapisClient } from '../../services/lapisClient.ts';
 import { RELEASED_AT_FIELD, VERSION_STATUS_FIELD, IS_REVOCATION_FIELD } from '../../settings.ts';
-import { siloVersionStatuses } from '../../types/lapis';
+import { versionStatuses } from '../../types/lapis';
 
+// Excluding sequences whose latest version is revoked
```
theosanderson, line 7: If we want this clarification it would be clearer next to `totalSequences` (for recent sequences its actually a bit more complicated). I would maybe just leave this to the code.

theosanderson: And actually calling a latest version "revoked" is maybe confusing - I think you mean a "latest version is a revocation"

theosanderson (approved):

website/src/components/IndexPage/getOrganismStatistics.ts
corneliusroemer: happy with this - just curious what you mean by "for recent sequences it's more complicated"

theosanderson: For recent sequences we only count version 1 to only count truly "new" sequences. Then we exclude version 1s which have since been revoked. If any of those have since been reinstated, then their "latest version" is not a revocation. But we still exclude them.

website/src/components/IndexPage/getOrganismStatistics.ts
```diff
@@ -69,22 +70,22 @@ const getTotalAndLastUpdatedAt = async (
...
     )
         .map((x) => x.data[0].count)
         .unwrapOr(0);
-    const recentRevoked = (
+    const recentTotalRevoked = (
         await client.call('aggregated', {
             [`${RELEASED_AT_FIELD}From`]: recentTimestamp,
             version: 1,
-            [VERSION_STATUS_FIELD]: siloVersionStatuses.revoked,
+            [VERSION_STATUS_FIELD]: versionStatuses.revoked,
```
corneliusroemer, line 85: Ah - I see, technically things could be resubmitted and then we _should_ count them, but we aren't.

Because it would require a more complicated query that would require fetching accessions - correct?

We should clarify this in a comment.

website/src/components/IndexPage/getOrganismStatistics.ts
corneliusroemer: Yeah going through the code carefully I realized this - we should document that we are aware of this issue in the code and state why we chose this (to avoid complicated 2-step query with accession lists)

corneliusroemer: Thanks for making me realize the complexity, adding a comment later to explain the tradeoff and potential for edge cases

corneliusroemer:
I made slight tweaks based on your comments @theosanderson 
https://github.com/loculus-project/loculus/pull/2818/commits/bfba0b76d542437a9cbae13d2afce40aecf6a1db

PR #2825, opened 2024-09-17
Revert "refactor(website): rename `siloVersionStatus` to `versionStatus`"

anna-parker:
Just checking if this works while investigating

Reverts loculus-project/loculus#2818

Doesn't appear to be the cause: https://revert-2818-version-statu.loculus.org/


anna-parker:
This was a test and was not required

corneliusroemer:
Cause of what?

Issue #2828, opened 2024-09-18
NCBI datasets error causing ingest to fail for west nile and cchf


anna-parker:
ncbi datasets cli does not recognize the data we downloaded using that same cli in the previous rule
```
localrule extract_ncbi_dataset_sequences:
    input: results/ncbi_dataset.zip
    output: results/sequences.fasta
    jobid: 8
    reason: Missing output files: results/sequences.fasta; Input files updated by another job: results/ncbi_dataset.zip
    resources: tmpdir=/tmp
        unzip -jp results/ncbi_dataset.zip         ncbi_dataset/data/genomic.fna         | seqkit seq -w0 -i         > results/sequences.fasta
dataformat doesn't recognize this input
For best results
1. Make sure to use --as-json-lines with the datasets command
2. Make sure that you're using the latest version of the datasets command line tool
Use --force to remove this warning.
Download the latest version of the datasets command line tool: https://www.ncbi.nlm.nih.gov/datasets/docs/v2/download-and-install
Error:  unknown field "usaState"
Usage
  dataformat tsv virus-genome [flags]
Examples
  dataformat tsv virus-genome --inputfile sars2_package/ncbi_dataset/data/data_report.jsonl
  dataformat tsv virus-genome --package virus-sars2-refseq.zip
Flags
      --fields strings     Comma-separated list of fields
```

PR #2829, opened 2024-09-18
feat(ingest): Update datasets package in ingest

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/2828

preview URL: https://fix-nci-datasets.loculus.org/

### Summary

Apply error log suggestions (see below) and update the datasets package.
```
For best results
1. Make sure to use --as-json-lines with the datasets command
2. Make sure that you're using the latest version of the datasets command line tool
```

### Screenshot

Sequences are showing up again for cchf and west nile: 
image

### DIFF
--- a/ingest/environment.yml
+++ b/ingest/environment.yml
@@ -10,7 +10,7 @@ dependencies:
   - click
   - ijson
   - jsonlines
-  - ncbi-datasets-cli >=16.21.0
+  - ncbi-datasets-cli >=16.29.0
   - nextclade >=3.7.0
   - orjsonl
   - pandas


ingest/environment.yml
```diff
@@ -10,7 +10,7 @@ dependencies:
   - click
   - ijson
   - jsonlines
-  - ncbi-datasets-cli >=16.21.0
+  - ncbi-datasets-cli >=16.29.0
```
anna-parker, line 13: Just checked and replacing with ncbi-datasets-cli would also work and might be better - @corneliusroemer I assume you fixed this for a reason - are you ok with unfixing it?

corneliusroemer: I'm not sure I understand what you mean here

corneliusroemer (approved):

Issue #2788, opened 2024-09-15
NCBI Ingest: Correct geolocation metadata mappings


anna-parker:
- [ ] #3105
![image](https://github.com/user-attachments/assets/ce55f719-bad7-46ab-b5ac-c631ec71003d)

- [x] Additionally we map ncbiGeoRegion we map to geoLocAdmin2, which is not correct as ncbiGeoRegion corresponds to continent and geoLocAdmin2 should be "A local administrative region from which the sample was collected (ex: county or municipality)": 
![image](https://github.com/user-attachments/assets/f9cc6ec4-09ed-42a2-aa16-9988a3945aaa)

- [ ] Can we reconsider the displayName for these fields? I think it will be confusing to submitters that geoLocAdmin1 gets mapped to collection subdivision etc.

corneliusroemer:
Yeah this was on my list of things to fix/improve eventually. Thanks for making an issue.

PR #2833, opened 2024-09-18
fix: don't ingest region as geoLocAdmin2 (it's continent)

corneliusroemer:
Relates to #2788

Might require pp values.yaml changes - need to check

https://fix-continent.loculus.org

### DIFF
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -407,7 +407,6 @@ defaultOrganismConfig: &defaultOrganismConfig
         generateIndex: true
         autocomplete: true
         header: Sample details
-        ingest: ncbiGeoRegion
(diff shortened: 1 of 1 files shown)


theosanderson:
This will indeed have no effect on PP without changing its values. I guess it will also trigger a bunch of new versions? (Which we've made peace with)

theosanderson:
Maybe we roll out name reordering and this together on PP?

anna-parker (approved):

corneliusroemer:
Good point @theosanderson about whether we want to ingest as something else.

The answer is no I think. We will want to eventually have a "continent/region" field - but we should _derive_ this in prepro from the countries through a mapping, rather than rely on users settings this (extra work and they might get it wrong). I'll make an issue for that.

PR #2839, opened 2024-09-18
fix: make filename changes change file hash

### LINKED ISSUE #2838: Renaming file is not enough to change the docker image cache hash in our CI
@anna-parker hit an interesting bug in our docker image hash cache strategy.

She changed a file name only, and to our current hashing strategy file names do not seem to matter.

This is a bug - I should fix it, for now workaround is to make a trivial change in a readme for example.

corneliusroemer:
partially resolves #2838

resolves #

preview URL:

### PR Checklist


corneliusroemer:
Alright, now this worked - so file names are now taken into account in the hash for backend

We should roll this out to the other cache builds as well.

theosanderson (approved):

PR #2846, opened 2024-09-19
Check-aux-table

anna-parker:
resolves #

preview URL:

### PR Checklist

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
@@ -37,6 +37,7 @@ import org.loculus.backend.model.RELEASED_DATA_RELATED_TABLES
 import org.loculus.backend.model.ReleasedDataModel
 import org.loculus.backend.model.SubmissionParams
 import org.loculus.backend.model.SubmitModel
+import org.loculus.backend.service.submission.MetadataUploadAuxTable
 import org.loculus.backend.service.submission.SubmissionDatabaseService
 import org.loculus.backend.utils.Accession
 import org.loculus.backend.utils.IteratorStreamer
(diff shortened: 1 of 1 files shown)


anna-parker:
I tried to add a test, e.g. but this fails as the auxtable is empty:
```
@Test
    fun `GIVEN still processing submitted data THEN return 503`() {
        convenienceClient.submitDefaultFiles(organism = organism)
        val response = submissionControllerClient.getOriginalMetadata()
        response.andExpect(status().isNotAvailable)
    }
```

anna-parker:
CCHF ingest was stopped by this new 503 response:
```
14:46:45    ERROR call_loculus.py:get-submitted - Request failed:
URL: https://backend-check-aux-table.loculus.org/cchf/get-original-metadata
Method: GET
Status Code: 503
Response Content:
```
But sadly it doesn't seem like the ingest deployment is restarting

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
```diff
@@ -66,7 +66,7 @@ private val log = KotlinLogging.logger { }
 @RequestMapping("/{organism}")
 @Validated
 @SecurityRequirement(name = "bearerAuth")
-open class SubmissionController(
+class SubmissionController(
```
corneliusroemer, line 69:
```suggestion
open class SubmissionController(
```

corneliusroemer (approved):

fengelniederhammer:
Testing it would have probably been possible by "manually" adding something to the aux tables. I.e. calling a function from the test that inserts a row to simulate being in the middle of an upload.

corneliusroemer:
@fengelniederhammer makes sense. I'm just not familiar with the framework we're using here. Would be great if you could add such a test in a PR!

Issue #2832, opened 2024-09-18
Ingest: get nucleotide accessions to ignore from extended metadata table


corneliusroemer:
We need to find out the nucleotide accessions that correspond to ENA deposited sequences so ingest can ignore these (otherwise we end up with infinite loop).

This might require some changes to ena submission so we find out the nucleotide accessions from the GCA assembly accessions.

For reference:

<img width="1030" alt="image" src="https://github.com/user-attachments/assets/7e49d6c8-5a60-40d7-a766-43b310e0e284">

<img width="1030" alt="Brave Browser 2024-09-18 14 40 44" src="https://github.com/user-attachments/assets/18e96c10-aa9f-433e-8170-9176b1116b12">

@anna-parker (this is a test)

theosanderson:
We plan to annotate our ENA depositions with PP metadata right (the PP accession, as some cross-reference?) Can't we use that?

corneliusroemer:
It probably doesn't end up in NCBI virus export so can't do that easily afaict

@anna-parker

theosanderson:
Hmm does this mean we could be failing to ingest some other quite important data? (Not a criticism - just for understanding: I guess I previously thought we were capturing 100% of data)

PR #2848, opened 2024-09-19
Get list of accessions to not ingest from ena-submission schema.

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/2832

preview URL: https://prevent-ingest-ingesting.loculus.org/

### Summary

0. Give ingest db access
1. Adds two new snakemake rules to ingest:
- `get_loculus_submissions`: calls the db and gets a list of all biosample and insdc accessions of sequences we have submitted (I add biosample to prevent the race condition if ncbi virus adds loculus submissions to their list of results before we can get/add the corresponding accession to the db)
- `filter_out_loculus_submissions`: filters out any metadata entries with these fields
2. Updates ingest tests to test that if an entry has an insdc accession in this list it will not be included in results.
3. Confirmed by test on preview that adding the known ebola zaire accession `MZ424862.1` to this list means it does not get ingested: 
image
image

### PR Checklist

- [x] Update: Make sure this also works when enaSubmission is disabled!!!


theosanderson:
(It's kind of a shame that this breaks the modularity of Depostion vs Ingest - we should try to bring that back later when we move to ingest data beyond that from NCBI datasets)

anna-parker:
Totally, this was a fast way to add the check but ideally ingest wouldn't also have access to the db. The preferable option would probably be:

1. Add an `UPLOADED_INITIAL_RESULTS_TO_LOCULUS` state (or sth to that same affect) in the submission_table and code to upload the biosample and bioproject accessions to Loculus after they have been created in ENA. Currently we only upload results to ENA once we also have the assembly accessions (this can take a couple days).
2. Add a `get_external_metadata` endpoint in the backend. Ingest should be able to call this endpoint to get all metadata submitted by the ena-submission pod, any biosample_accession or insdc_accession in this table corresponds to data we submitted and should not be ingested.
3. Update ingest to not ingest any data with a biosample or insdc_accession in this list. Using biosample is good as it should protect us from the edge case of the sequence being added to the insdc results before we get the corresponding new accessions from ENA or send them to Loculus.

theosanderson:
An alternative is potentially that (at some stage in the future), Ingest could just use the metadata on INSDC to understand whether data is originally from this Loculus instance (i.e. a cross-reference to the loculus-instance-accession)

ingest/scripts/get_loculus_submissions.py
```diff
@@ -0,0 +1,182 @@
...
+            # Result is a jsonb column
+            query = "SELECT accession, result FROM sample_table WHERE STATUS = 'SUBMITTED'"
+
+            cur.execute(query)
+
+            results = cur.fetchall()
+    finally:
+        db_conn_pool.putconn(con)
+
+    return {result["accession"]: result["result"]["biosample_accession"] for result in results}
```
theosanderson, line 103:
```suggestion
def get_bio_sample_accessions(db_conn_pool: SimpleConnectionPool) -> dict[str, str]:
    with db_conn_pool.getconn() as con:
        with con.cursor(cursor_factory=RealDictCursor) as cur:
            query = "SELECT accession, result FROM sample_table WHERE STATUS = 'SUBMITTED'"
            cur.execute(query)
            results = cur.fetchall()
    return {result["accession"]: result["result"]["biosample_accession"] for result in results}
```

ingest/scripts/get_loculus_submissions.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+def get_insdc_accessions(db_conn_pool: SimpleConnectionPool) -> dict[str, str]:
+    con = db_conn_pool.getconn()
+    try:
```
theosanderson, line 108: simplification as above?

ingest/scripts/get_loculus_submissions.py
```diff
@@ -0,0 +1,182 @@
...
+    with open(config_file, encoding="utf-8") as file:
+        full_config = yaml.safe_load(file)
+        relevant_config = {key: full_config.get(key, []) for key in Config.__annotations__}
+        config = Config(**relevant_config)
+    logger.info(f"Config: {config}")
+
+    db_config = db_init(config.db_password, config.db_username, config.db_host)
+    insdc_accessions_submitted_by_loculus = get_insdc_accessions(db_config)
+    all_insdc_accessions_submitted_by_loculus = [
+        item for sublist in insdc_accessions_submitted_by_loculus.values() for item in sublist
```
theosanderson, line 165: I remember at Google the linter didn't allow nested list comprehensions because they are too hard to understand - IMO there's a reasonable argument to avoid them but not blocking.

ingest/scripts/get_loculus_submissions.py
anna-parker: Ah so from my research we need `db_conn_pool.putconn(con)` - if the connection breaks off or somehow the execution fails and we don't do this connections will not be added back to the pool and we will end up with not enough ephermeral ports for connecting to the db.

Normally this is not required with pyscopy (as soon as you connect with the `with ... as con:` statement it should handle closing the connection if this fails) but as I now use a connection pool this is required.

anna-parker: ah but maybe we just update and in psycopg3 they do it for you: https://www.psycopg.org/psycopg3/docs/advanced/pool.html#basic-connection-pool-usage

theosanderson: Thanks - sorry for bad advice! It seemed to me like this might be possible and then Claude agreed with me and I wrongly trusted it :) (But further research, plus ChatGPT, agrees with you!)

ingest/Snakefile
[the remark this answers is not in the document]
anna-parker: Makes sense! I have renamed now!

ingest/scripts/get_loculus_submissions.py
anna-parker: Added an issue to update: https://github.com/loculus-project/loculus/issues/2879

corneliusroemer:
Per microservice architecture, ingest would get it's own DB and could just keep track of things there, using get released data to look up the accessions to avoid.

It would indeed be really cool if we could figure out what's from Pathoplexus via some NCBI Virus metadata. But it might be risky to rely on that as there's a bunch of things that could go wrong there that's outside of our control.

anna-parker:
@corneliusroemer yeah totally agree long term the backend db should be the source of truth and ingest should not have to connect to the db. 

However, at the moment the backend will only receive accessions (biosample, bioproject, gca and chromosome) of samples we deposited in ENA after we have received all these accessions from ENA. I do want to change the submission pod to send every accession to the backend as soon as that accession is known but until we make that change I really don't want ingest to wait until the backend knows all accessions. 

1. There is probably some delay between NIH creating the insdc accession and giving it back to ENA
2. There is probably some delay in ENA returning this accession to us.
3. There is definitely some delay in the ingest pod getting the accession and sending it to the backend.
4. There is definitely some delay in the backend getting the accession and returning it in the get-released-data response.

There is probably also some delay in the samples we deposited showing up in the ncbi datasets download, however I would prefer not to risk this and filter out sequences based on the biosample and chromosome accession we have in the ena-submission schema. 

But yes as soon as possible I will try to change the pod to return every accession to Loculus as soon as it is known, and then we can remove these rights from the ingest pods. If you would like we could maybe create a read only db user for the ingest pod.

ingest/scripts/get_loculus_depositions.py
```diff
@@ -0,0 +1,161 @@
...
+
+    return {result["accession"]: result["result"]["biosample_accession"] for result in results}
+
+
+def get_insdc_accessions(db_conn_pool: SimpleConnectionPool) -> dict[str, str]:
+    con = db_conn_pool.getconn()
+    try:
+        with con, con.cursor(cursor_factory=RealDictCursor) as cur:
+            # Result is a jsonb column
+            query = "SELECT accession, result FROM assembly_table WHERE STATUS = 'SUBMITTED'"
```
corneliusroemer, line 90: query = "SELECT accession, result FROM assembly_table WHERE STATUS IN ('SUBMITTED', 'WAITING')"
```suggestion
            query = "SELECT accession, result FROM assembly_table WHERE STATUS IN ('SUBMITTED', 'WAITING')"
```

corneliusroemer (approved):

PR #2849, opened 2024-09-19
fix(deploy): hotfix mistake in my changes to `deploy.py` to handle subdomains appropriately

theosanderson:
I made a mistake in what was meant to be a helpful feature to avoid confusion. It should allow `main.loculus.org` but not `loculus.org` but I used `<=` not `<` so it currently doesn't allow `main.loculus.org`. Apologies.


emmahodcroft (approved):
makes sense!

PR #2851, opened 2024-09-19
feat(website): One tab for all protein sequences

theosanderson:
preview URL: https://seq-changes.loculus.org/ebola-zaire/search?

- Fixes the height of the sequence container avoiding layout shifts when moving between tabs
- Uses a single tab for all proteins - selected with a dropdown- , rather than one tab per protein, which would be important for users working on some organisms
- Adjusts the names of the raw and aligned nucleotide tabs

image

### DIFF
--- a/website/src/components/SequenceDetailsPage/SequencesContainer.tsx
+++ b/website/src/components/SequenceDetailsPage/SequencesContainer.tsx
@@ -113,7 +113,7 @@ const SequenceTabs: FC<SequenceTabsProps> = ({
                 />
                 <BoxWithTabsTab
                     isActive={activeTab === 'gene'}
-                    label='Amino acid sequences'
+                    label='Aligned amino acid sequences'
                     onClick={() => setActiveTab('gene')}
                 />
             </BoxWithTabsTabBar>
--- a/website/src/components/SequenceDetailsPage/SequencesContainer.tsx
+++ b/website/src/components/SequenceDetailsPage/SequencesContainer.tsx
@@ -119,13 +119,17 @@ const SequenceTabs: FC<SequenceTabsProps> = ({
             </BoxWithTabsTabBar>
             <BoxWithTabsBox>
                 {activeTab === 'gene' && <GeneDropdown genes={genes} sequenceType={sequenceType} setType={setType} />}
-                <SequencesViewer
-                    organism={organism}
-                    accessionVersion={accessionVersion}
-                    clientConfig={clientConfig}
-                    sequenceType={sequenceType}
-                    isMultiSegmented={isMultiSegmented(nucleotideSegmentNames)}
-                />
+                {activeTab !== 'gene' || isGeneSequence(sequenceType.name, sequenceType) ? (
+                    <SequencesViewer
+                        organism={organism}
+                        accessionVersion={accessionVersion}
+                        clientConfig={clientConfig}
+                        sequenceType={sequenceType}
+                        isMultiSegmented={isMultiSegmented(nucleotideSegmentNames)}
+                    />
+                ) : (
+                    <div className='h-80'></div>
+                )}
             </BoxWithTabsBox>
         </>
     );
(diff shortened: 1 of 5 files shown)


corneliusroemer (approved):
Really cool! Amazing that you improve the layout shift, has been something I'd noticed in the past.

There's still shift when switching to the aligned seqs because of the drop-down showing up.

In particular for multi-segment viruses the picker would make sense for all tabs. Not just aligned.

That would fix it. For non-segmented not sure.

chaoran-chen (changes requested):
Thanks Theo, this looks mostly really good to me!

One bug (I think it's a bug?) I noticed is that when you switch to the "Amino acid sequences" tab, you first still see the sequence from the tab that you have been previously on and no protein is selected. Only after selecting a protein, it shows an amino acid sequece.

image

website/src/components/SequenceDetailsPage/SequencesContainer.tsx
```diff
@@ -154,15 +205,47 @@ const AlignmentSequenceTabs: FC<NucleotideSequenceTabsProps> = ({ nucleotideSegm
...
+
+    return (
+        <div className='mb-4'>
+            <select
+                className='select select-bordered w-full max-w-xs'
+                value={selectedGene}
+                onChange={(e) => setType(geneSequence(e.target.value))}
+            >
+                <option value='' disabled>
+                    Select a protein
```
chaoran-chen, line 237: I think that we should avoid using the term "protein" here as not necessarily each AA seqeunce represents a different protein? Everywhere else, I believe that we have also rather just used "amino acid sequence" as term.

```suggestion
                    Select a sequence
```

corneliusroemer:
> One bug (I think it's a bug?)

Yeah you're right, I can now also reproduce and agree it's a bug

website/src/components/SequenceDetailsPage/SequencesContainer.tsx
theosanderson: Hmm - you mean in that one protein can have multiple chains? In code we call these identifiers `genes` and I'd be OK with that here. I find `Select a sequence` a bit unexpected - here I'd expect the drop downs to be accessions. Would `Select a gene` work for you?

theosanderson:
Thanks - I fixed the bug by not displaying that old, wrong, sequence. I think the chances that users happen to want the top protein in the list are relatively low so I think this behaviour is OK for now - though I'm not averse to future PRs that change it. I adjusted `Select a protein` to `Select a gene`.

website/src/components/SequenceDetailsPage/SequencesContainer.tsx
```diff
@@ -49,75 +49,118 @@ export const InnerSequencesContainer: FC<SequenceContainerProps> = ({
...
+                <AlignmentSequenceTabs
+                    nucleotideSegmentNames={nucleotideSegmentNames}
+                    sequenceType={sequenceType}
+                    setType={setType}
+                    isActive={activeTab === 'aligned'}
+                    setActiveTab={setActiveTab}
+                />
+                <BoxWithTabsTab
+                    isActive={activeTab === 'gene'}
+                    label='Amino acid sequences'
```
chaoran-chen, line 116: Shoud we rather name it "Aligned amino acid sequences" which would be more accurate? It might be worth emphasizing this so that users know that insertions are not included.

chaoran-chen (approved):
Thanks!

website/src/components/SequenceDetailsPage/SequencesContainer.tsx
chaoran-chen: For now, I think it's fine. When we look at more [schema designs](https://loculus.org/for-administrators/schema-designs/#multiple-references-for-an-organism) such as the multiple references for an organism approach, different AA sequences could be from the same gene aligned to different references.

theosanderson: Sounds good to discuss then (I think in such a case I'd favour an approach where the UX involves selecting a reference and a gene(

PR #2852, opened 2024-09-20
test(backend): test `/get-original-metadata` locked during new submissions

fengelniederhammer:
preview URL: https://testMtadataLock.loculus.org

### Summary

This is a follow-up to #2846

### PR Checklist

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/model/SubmitModel.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/model/SubmitModel.kt
@@ -17,11 +17,14 @@ import org.loculus.backend.controller.UnprocessableEntityException
 import org.loculus.backend.service.datauseterms.DataUseTermsPreconditionValidator
 import org.loculus.backend.service.groupmanagement.GroupManagementPreconditionValidator
 import org.loculus.backend.service.submission.CompressionAlgorithm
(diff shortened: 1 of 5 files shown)


backend/src/main/kotlin/org/loculus/backend/model/SubmitModel.kt
```diff
@@ -313,4 +316,13 @@ class SubmitModel(
             throw UnprocessableEntityException(metadataNotPresentErrorText + sequenceNotPresentErrorText)
         }
     }
+
+    @Transactional(readOnly = true)
+    fun checkIfStillProcessingSubmittedData(): Boolean {
```
corneliusroemer, line 321: We could eventually allow filtering on the submitting group rather than on any entries as only the same group matters for ingest purposes, but this shouldn't be an issue unless we always have something in aux table which is unlikely to ever be the case.

corneliusroemer (approved):
Thanks, that's great!

backend/src/main/kotlin/org/loculus/backend/model/SubmitModel.kt
fengelniederhammer: If you think it's necessary, create an issue. :) But it's definitely out of scope for this PR.

PR #2854, opened 2024-09-20
feat: Edit group info

### LINKED ISSUE #2258: Allow editing group contact information through website
As a group member I want to update group information to keep group information up to date and correct in case of changes.

## Acceptance Criteria

There is an edit group button that leads to an edit UI where all the group info can be edited and there is 'save changes' button to save the changes.
Group info can only be edited by members of the group.

There are unit tests for the backend.

## Tasks

- [ ] Update the backend to include a new `PUT /groups/<id>` endpoint to update group information.
- [ ] Add tests for the backend.
- [ ] ensure that only group members can edit the group.
- [ ] Update the UI to include an _edit_ button on a group and a new `/group/<id>/edit` endpoint with a new edit form based on the _new group_ form.

fhennig:
resolves #2258 

preview URL: https://feat-2258-edit-group-info.loculus.org

### Summary

This PR adds a new backend API endpoint `PUT /group/<groupId>` as well as a group edit button to the group details view in the UI. Only group members can edit the group. All group info can be edited (except the group ID).

### Screenshot
Screen Recording 2024-10-01 at 17 47 52

### PR Checklist

  - [ ] Warn users that if they change group name they will change ENA center name
- [ ] The implemented feature is covered by an appropriate test:
  - Manual tests:
    - [ ] 1. make an account, 2. create group, 3. submit sequences, 4. change everything in group, 5. see if we can still revise, etc.
    - [ ] In staging, change everything for grubaugh lab and see if it causes issues, e.g. also adding new sequences and seeing if ENA is happy: @anna-parker and @corneliusroemer should do this
  - Automatic tests:
    - [ ] Add end-to-end tests for the manual scenario
- [ ] Thought about potential security implications:
  - [ ] Ensure id can't be changed
  - [ ] We never use group name anywhere for anything security relevant
  - [ ] Ensure only group members can change group info


corneliusroemer (commented):
Looks great. Good start!

chaoran-chen (commented):
Nice code outlines and great to see the small docs and script improvements! If you create small PRs, I think that we can already merge them.

backend/src/main/kotlin/org/loculus/backend/service/groupmanagement/GroupManagementDatabaseService.kt
```diff
@@ -88,6 +88,15 @@ class GroupManagementDatabaseService(
         )
     }
 
+    fun updateGroup(groupId: Int, group: NewGroup, authenticatedUser: AuthenticatedUser) {
+        groupManagementPreconditionValidator.validateThatUserExists(authenticatedUser.username)
+
+        groupManagementPreconditionValidator.validateUserIsAllowedToModifyGroup(groupId, authenticatedUser)
+
+        // TODO do the editing
+        // Update the groupId group with the info in NewGroup
```
fengelniederhammer, line 97: Btw Kotlin has a `TODO()` function that usually makes your code compile, which is quite nice. (Not necessary here, because you don't need to return anything)

That reminds me of: should the new endpoint maybe return the new (updated) group information?

corneliusroemer: The point of returning the new information would be as a sanity check or would it serve any other purpose?

fengelniederhammer: Yes, sanity check / convenience for users and easier testing

fhennig: thanks for the tip regarding TODO! I've also updated the function to return the group now 9ec8bda

fhennig:
I'm done with the API part of the change, next is the UI. 

I've also made a small change to the gradle build; next time I'll split it into a separate PR (just like the script changes).

backend/src/main/kotlin/org/loculus/backend/service/groupmanagement/GroupManagementDatabaseService.kt
```diff
@@ -172,4 +152,31 @@ class GroupManagementDatabaseService(
                 contactEmail = it.contactEmail,
             )
         }
+
+    private fun GroupEntity.updateWith(group: NewGroup) {
```
fengelniederhammer, line 156: Why not move those functions to the file where `GroupEntity` is defined? Extension functions are nice, but in general I think it's better to have all implementations in one place. I can imagine that at least `toGroup` might also be useful in other places.

fengelniederhammer, line 156: This method looks like one can only update all values at once. Is that what we want? Or do we want to offer users the possibility to just update a single field (and not writing all the other fields to the request, too)?

fhennig: I wasn't sure whether it would be okay to reference the `NewGroup` object from the DB class since it's an API object. I thought maybe those should be kept separate. But I'm happy to move the functions to the object directly.

fhennig: @chaoran-chen Do you wanna weigh in on this?

chaoran-chen: The PUT specification [here](https://httpwg.org/specs/rfc9110.html#PUT) says

> A successful PUT of a given representation would suggest that a subsequent GET on that same target resource will result in an equivalent representation being sent in a [200 (OK)](https://httpwg.org/specs/rfc9110.html#status.200) response.

So I think that it's better if the endpoint user always has to provide the whole object. Otherwise, we also would have to deal with distinguishing whether the user wants to not change something or set it to `null`.

fengelniederhammer: Ah good point. Maybe it's actually better to keep it separated :+1:

website/src/components/Group/GroupForm.tsx
```diff
@@ -0,0 +1,156 @@
...
+                        <CountryInput defaultValue={defaultGroupData?.address.country} />
+                    </div>
+
+                    <div className='flex justify-end py-8 gap-4 '>
+                        <button
+                            type='submit'
+                            className='btn btn-primary px-4 py-2 loculusColor text-white rounded'
+                            disabled={!isClient}
+                        >
+                            {buttonText}
```
corneliusroemer, line 110: Can we not use some automatic mapping here from field name to id? we can rename html IDs to match field names and use camel case to kebab case. Then this reduces to a few lines instead of 20

website/src/components/Group/GroupForm.tsx
```diff
@@ -0,0 +1,156 @@
...
+    };
+
+    const isClient = useClientFlag();
+
+    return (
+        <div className='p-4 max-w-6xl mx-auto'>
+            <h2 className='title'>{title}</h2>
+
+            {errorMessage !== undefined && (
+                <ErrorFeedback message={errorMessage} onClose={() => setErrorMessage(undefined)} />
```
corneliusroemer, line 83: Same thing here, we should be able to do this automatically instead of verbosely

corneliusroemer (commented):
Not blocking but there seems to be quite some probably unnecessary repetition of the group fields - already from the legacy code, but now the duplication becomes more noticeable

We could make an issue to cleanup later, it's not necessarily pressing, just a bit of code smell (not your fault by the way, legacy)

corneliusroemer:
Cool! I tested and it works, you could record a quick Screencast to showcase :)

website/src/components/Group/GroupForm.tsx
fhennig: You mean line 45 to 53 right?

website/src/components/Group/GroupForm.tsx
fhennig: I'm assuming you mean line 93 to 101 here? Sounds reasonable, I'll look into it!

Now that we're touching the code anyways might be a good time to tidy it up a bit

website/src/components/Group/GroupForm.tsx
```diff
@@ -0,0 +1,156 @@
...
+                    <div className='mt-5 grid grid-cols-1 gap-x-6 gap-y-4 sm:grid-cols-6'>
+                        <GroupNameInput defaultValue={defaultGroupData?.groupName} />
+                        <EmailContactInput defaultValue={defaultGroupData?.contactEmail} />
+                        <InstitutionNameInput defaultValue={defaultGroupData?.institution} />
+                        <AddressLineOneInput defaultValue={defaultGroupData?.address.line1} />
+                        <AddressLineTwoInput defaultValue={defaultGroupData?.address.line2} />
+                        <CityInput defaultValue={defaultGroupData?.address.city} />
+                        <StateInput defaultValue={defaultGroupData?.address.state} />
+                        <PostalCodeInput defaultValue={defaultGroupData?.address.postalCode} />
+                        <CountryInput defaultValue={defaultGroupData?.address.country} />
```
corneliusroemer, line 101: Yes and also this range here

website/src/components/Group/GroupForm.tsx
```diff
@@ -0,0 +1,156 @@
...
+    },
+    state: {
+        id: 'state',
+        required: false,
+    },
+    postalCode: {
+        id: 'postal-code',
+        required: true,
+    },
+} as const;
```
corneliusroemer, line 156: And this range

website/src/components/Group/Inputs.tsx
```diff
@@ -0,0 +1,236 @@
...
+    },
+    state: {
+        id: 'state',
+        required: false,
+    },
+    postalCode: {
+        id: 'postal-code',
+        required: true,
+    },
+} as const;
```
corneliusroemer, line 44: and this range

website/src/components/Group/Inputs.tsx
```diff
@@ -0,0 +1,236 @@
...
+export const PostalCodeInput: FC<PostalCodeInputProps> = ({ defaultValue }) => (
+    <TextInput
+        className='sm:col-span-2'
+        type='text'
+        label='ZIP / Postal code'
+        name='postal-code'
+        fieldMappingKey='postalCode'
+        defaultValue={defaultValue}
+    />
+);
```
corneliusroemer, line 236: And this one is also quite repetitive I think, but not quite straightforward to condense maybe as there are some more things that differ between fields. But DRYer seems definitely possible nonetheless.

website/src/components/Group/GroupForm.tsx
corneliusroemer: ah yes no clue why it's wrongly annotated, my bad

website/src/components/Group/GroupForm.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+        const groupName = formData.get(fieldMapping.groupName.id) as string;
+        const institution = formData.get(fieldMapping.institution.id) as string;
+        const contactEmail = formData.get(fieldMapping.contactEmail.id) as string;
+        const country = formData.get(fieldMapping.country.id) as string;
+        const line1 = formData.get(fieldMapping.line1.id) as string;
+        const line2 = formData.get(fieldMapping.line2.id) as string;
+        const city = formData.get(fieldMapping.city.id) as string;
+        const state = formData.get(fieldMapping.state.id) as string;
+        const postalCode = formData.get(fieldMapping.postalCode.id) as string;
```
corneliusroemer, line 53: indeed here this is part of the verbosity

fhennig:
I've refactored the field mapping a bit, but I don't have an idea for how to completely get rid of a mapping function from form data to `NewGroup` -- my Typescript skills aren't good enough for that yet and ChatGPT didn't have a good idea either 😄 

For the repetitive form inputs, I'd say it's fine for now. It's repetitive but simple to deal with. I am of course happy to change it if you have a concrete suggestion what to do.

website/src/components/Group/GroupForm.tsx
fhennig: yes for this one i'm not sure if there's a clever way to condense this

website/src/components/Group/Inputs.tsx
```diff
@@ -1,46 +1,20 @@
...
-    state: {
-        id: 'state',
-        required: false,
-    },
-    postalCode: {
-        id: 'postal-code',
-        required: true,
-    },
+    groupName: 'group-name',
+    institution: 'institution-name',
```
corneliusroemer, line 10: We can also change the id here - there's no reason the html id needs to be `institution-name` rather than `institution`
```suggestion
    institution: 'institution',
```
this can then be calculated from a kebabCase() js function

website/src/components/Group/Inputs.tsx
```diff
@@ -229,8 +198,27 @@ export const PostalCodeInput: FC<PostalCodeInputProps> = ({ defaultValue }) => (
...
         label='ZIP / Postal code'
-        name='postal-code'
-        fieldMappingKey='postalCode'
+        name={fieldMapping.postalCode}
+        required
         defaultValue={defaultValue}
     />
 );
+
+export const groupFromFormData = (formData: FormData) => {
```
corneliusroemer, line 207:
```suggestion
export const groupFromFormData = (formData: FormData): NewGroup => {
    const getField = (key: keyof typeof fieldMapping) => formData.get(fieldMapping[key]) as string;

    return {
        groupName: getField('groupName'),
        institution: getField('institution'),
        contactEmail: getField('contactEmail'),
        address: {
            line1: getField('line1'),
            line2: getField('line2'),
            city: getField('city'),
            postalCode: getField('postalCode'),
            state: getField('state'),
            country: getField('country')
        }
    };
};
```

Copilot suggests this for example

corneliusroemer: Or even dryer:
`
``
export const groupFromFormData = (formData: FormData): NewGroup => {
    const group: Partial<NewGroup> = {
        address: {} as NewGroup['address']
    };

    Object.keys(fieldMapping).forEach((key) => {
        const value = formData.get(fieldMapping[key as keyof typeof fieldMapping]) as string;
        if (key in group.address) {
            group.address[key as keyof NewGroup['address']] = value;
        } else {
            group[key as keyof NewGroup] = value;
        }
    });

    return group as NewGroup;
};
```

Comment (fhennig) [reply website/src/components/Group/Inputs.tsx:207]: I think having the IDs be in line with the names of the properties of the object makes sense, I'll go do that. but that next step introduces a lot of complexity and is difficult to read, I wouldn't want to add that.

Comment (fhennig) [reply website/src/components/Group/Inputs.tsx]: https://github.com/loculus-project/loculus/pull/2854/commits/6408ce2571b2a0bdc79408b6f6d40828df9d8ee3

File: website/src/components/Group/Inputs.tsx
Comment (corneliusroemer) [inline website/src/components/Group/Inputs.tsx:1]: Copilot is suggesting this, quite nice, much less repetition, can likely still be tweaked

To make the input components even dryer by using a configuration file and deducing defaults, we can follow these steps:

Step 1: Create a Configuration File with Inferred Defaults
Create a new file, inputConfig.ts, to store the configuration for the input fields with inferred defaults:

// inputConfig.ts
```
export const defaultConfig = {
    type: "text",
    required: true,
    className: "default-css-class", // Placeholder for actual default CSS class
    defaultValue: "",
};

export const inputConfig = [
    { name: "groupName", className: "sm:col-span-4" },
    { name: "institution", className: "sm:col-span-4" },
    { name: "contactEmail", type: "email", className: "sm:col-span-4" },
    { name: "country", type: "select", className: "sm:col-span-3" },
    { name: "line1", className: "col-span-full" },
    { name: "line2", required: false, className: "col-span-full" },
    { name: "city", className: "sm:col-span-2 sm:col-start-1" },
    { name: "state", required: false, className: "sm:col-span-2" },
    { name: "postalCode", className: "sm:col-span-2" },
];
```
Step 2: Modify Inputs.tsx to Use the Configuration File
Update the Inputs.tsx file to dynamically generate the input fields based on the configuration and inferred defaults:

```
import { type ComponentProps, type FC } from 'react';
import { listOfCountries } from './listOfCountries.ts';
import type { NewGroup } from '../../types/backend.ts';
import { inputConfig, defaultConfig } from './inputConfig.ts';

export const CountryInputNoOptionChosen = 'Choose a country...';

const fieldMapping = {
    groupName: 'group-name',
    institution: 'institution-name',
    contactEmail: 'email',
    country: 'country',
    line1: 'address-line-1',
    line2: 'address-line-2',
    city: 'city',
    state: 'state',
    postalCode: 'postal-code',
} as const;

const groupCreationCssClass = 'some-css-class'; // Placeholder for actual CSS class

const toSentenceCase = (str: string) =>
    str.replace(/([A-Z])/g, ' $1').replace(/^./, (char) => char.toUpperCase());

type InputFieldProps = {
    label?: string;
    name: keyof typeof fieldMapping;
    type?: ComponentProps<'input'>['type'] | 'select';
    defaultValue?: string;
    required?: boolean;
    className?: string;
};

const InputField: FC<InputFieldProps> = ({
    label,
    name,
    type = defaultConfig.type,
    defaultValue = defaultConfig.defaultValue,
    required = defaultConfig.required,
    className = defaultConfig.className,
}) => (
    <LabelledInputContainer
        className={className}
        label={label || toSentenceCase(name)}
        htmlFor={name + '-input'}
        required={required}
    >
        {type === 'select' ? (
            <select
                id={name + '-input'}
                name={fieldMapping[name]}
                required={required}
                autoComplete="country-name"
                className={groupCreationCssClass}
                defaultValue={defaultValue}
            >
                <option>{CountryInputNoOptionChosen}</option>
                {listOfCountries.map((country) => (
                    <option key={country} value={country}>
                        {country}
                    </option>
                ))}
            </select>
        ) : (
            <input
                type={type}
                name={fieldMapping[name]}
                required={required}
                id={name + '-input'}
                className={groupCreationCssClass}
                autoComplete={type === 'email' ? 'email' : undefined}
                defaultValue={defaultValue}
            />
        )}
    </LabelledInputContainer>
);

export const groupFromFormData = (formData: FormData): NewGroup => {
    const group: Partial<NewGroup> = { address: {} as NewGroup['address'] };

    Object.keys(fieldMapping).forEach((key) => {
        const value = formData.get(fieldMapping[key as keyof typeof fieldMapping]) as string;
        if (key in group.address) {
            group.address[key as keyof NewGroup['address']] = value;
        } else {
            group[key as keyof NewGroup] = value;
        }
    });

    return group as NewGroup;
};

// Dynamically generate input fields based on configuration
const InputFields = ({ defaultValues }: { defaultValues?: Record<string, string> }) => (
    <>
        {inputConfig.map((config) => (
            <InputField
                key={config.name}
                label={config.label}
                name={config.name as keyof typeof fieldMapping}
                type={config.type}
                defaultValue={defaultValues ? defaultValues[config.name] : defaultConfig.defaultValue}
                required={config.required ?? defaultConfig.required}
                className={config.className ?? defaultConfig.className}
            />
        ))}
    </>
);

export default InputFields;
```

Comment (fhennig) [reply website/src/components/Group/Inputs.tsx:1]: I don't find the current input definitions problematic and I'm not a fan of the added complexity your suggested solution brings -- I'd rather have the slightly more verbose but very obivous current implementation. But if you want to make the changes and commit them to the branch here, feel free to do so, I'm fine with it to (just don't wanna do it myself).

Otherwise I'd consider this out of scope for the original task if that would be fine with you.

Comment (fhennig) [reply website/src/components/Group/Inputs.tsx:1]: there's not just DRY, there's also KISS ;)

File: website/src/services/groupManagementApi.ts [hunk 12]
```diff
@@ -18,6 +18,21 @@ const createGroupEndpoint = makeEndpoint({
...
+    parameters: [
+        authorizationHeader,
+        {
+            name: 'data',
+            type: 'Body',
+            schema: newGroup,
+        },
+    ],
+    response: group,
+    errors: [notAuthorizedError], // TODO not sure which errors the API can throw
```
fengelniederhammer, line 34: We are not fully utilizing this `errors` feature. It's ok if this list is not exhaustive. It's good to include those that you care about though.

website/src/components/Group/Inputs.tsx
fhennig: I have changed the ids here to match the property names of the objects.

website/src/services/groupManagementApi.ts
fhennig: Ok good to know thanks!

website/src/components/User/GroupPage.tsx
```diff
@@ -94,22 +95,32 @@ const InnerGroupPage: FC<GroupPageProps> = ({
...
-                        </button>
+                        <>
+                            <button
+                                className='object-right p-2 loculusColor text-white rounded px-4 mr-2'
+                                onClick={() => {
+                                    window.location.href = routes.editGroupPage(groupId);
+                                }}
+                            >
+                                Edit group
+                            </button>
```
fengelniederhammer, line 106: Why don't we make this an `<a>` tag? It's essentially only a link. Browsers don't need JS for that.

fengelniederhammer (commented):
I quickly looked over the code. Looks good from my side. But it would be good if someone has a closer look.

website/src/components/Group/Inputs.tsx
fhennig: Marking this as resolved as it's a duplicate of below

website/src/components/User/GroupPage.tsx
fhennig: good point, I had just copied this way of doing it from the button below which also has a confirmation dialog.

I've changed it now: https://github.com/loculus-project/loculus/pull/2854/commits/2f2ce7eadf1c703a1946e17de5f19be15501be6e

fhennig:
@corneliusroemer I think this is a good point: "Warn users that if they change group name they will change ENA center name" -- How should we do this? Maybe we can even add a little note to the edit UI if the group name is edited (some text in yellow next to the edit field maybe?)

Regarding "group ID can't be changed" -- This is the definition of the endpoint:

    @Operation(description = "Edit a group. Only users part of the group can edit it. The updated group is returned.")
    @ResponseStatus(HttpStatus.OK)
    @PutMapping("/groups/{groupId}", produces = [MediaType.APPLICATION_JSON_VALUE])
    fun editGroup(
        @HiddenParam authenticatedUser: AuthenticatedUser,
        @Parameter(
            description = "The id of the group to edit.",
        ) @PathVariable groupId: Int,
        @Parameter(description = "Updated group properties.")
        @RequestBody
        group: NewGroup,
    ): Group = groupManagementDatabaseService.updateGroup(groupId, group, authenticatedUser)
    
The ID is part of the `path` and not part of the `NewGroup`, so I would say that's sufficient? By definition the ID can't be changed.


For "Ensure only group members can change group info" I will add another backend unit test.

corneliusroemer:
This is where we use group info in cached ways:
- ENA submission: group info used for project creation, currently not updated, see #2939
- Ingest: Only sets group name of ingest group, never uses it
- Prepro: Never uses any group info apparently, even though group id is in data returned from prepro AFAICT

Backend tests:
-


Website tests:
- 

Manual tests:
- 

Loculus docs updates:
- [ ] Mention that groups can be edited somewhere appropriate

Pathoplexus docs updates:
- [ ] Mention where appropriate

backend/src/test/kotlin/org/loculus/backend/controller/groupmanagement/GroupManagementControllerTest.kt
```diff
@@ -104,6 +106,50 @@ class GroupManagementControllerTest(@Autowired private val client: GroupManageme
             .andExpect(jsonPath("\$.[0].contactEmail").value(NEW_GROUP.contactEmail))
     }
 
+    @Test
+    fun `GIVEN a group is created WHEN I edit the group THEN the group information is updated`() {
```
corneliusroemer, line 110:
```suggestion
    fun `GIVEN I'm a member of a group WHEN I edit the group THEN the group information is updated`() {
```

backend/src/test/kotlin/org/loculus/backend/controller/groupmanagement/GroupManagementControllerTest.kt
corneliusroemer, line 1: - [ ] Add test that one can't edit a group that one isn't a member of
- [ ] Add test that superuser can (or can't, just do one for now) edit group info of all groups

backend/src/test/kotlin/org/loculus/backend/controller/groupmanagement/GroupManagementControllerTest.kt
```diff
@@ -104,6 +106,50 @@ class GroupManagementControllerTest(@Autowired private val client: GroupManageme
...
+            .andExpect(jsonPath("\$.group.institution").value(newInfo.institution))
+            .andExpect(jsonPath("\$.group.address.line1").value(newInfo.address.line1))
+            .andExpect(jsonPath("\$.group.address.line2").value(newInfo.address.line2))
+            .andExpect(jsonPath("\$.group.address.city").value(newInfo.address.city))
+            .andExpect(jsonPath("\$.group.address.state").value(newInfo.address.state))
+            .andExpect(jsonPath("\$.group.address.postalCode").value(newInfo.address.postalCode))
+            .andExpect(jsonPath("\$.group.address.country").value(newInfo.address.country))
+            .andExpect(jsonPath("\$.group.contactEmail").value(newInfo.contactEmail))
+    }
+
```
corneliusroemer, line 152:
```suggestion
    val groupId = client.createNewGroup(group = DEFAULT_GROUP, jwt = jwtForDefaultUser)
        .andExpect(status().isOk)
        .andGetGroupId()
    
    val newInfo = NewGroup(
        groupName = "Updated group name",
        institution = "Updated institution",
        address = Address(
            line1 = "Updated address line 1",
            line2 = "Updated address line 2",
            postalCode = "Updated post code",
            city = "Updated city",
            state = "Updated state",
            country = "Updated country",
        ),
        contactEmail = "Updated email",
    )

    fun ResultActions.verifyGroupInfo(prefix: String = "$") {
        andExpect(status().isOk)
            .andExpect(content().contentType(MediaType.APPLICATION_JSON_VALUE))
            .andExpect(jsonPath("$prefix.groupName").value(newInfo.groupName))
            .andExpect(jsonPath("$prefix.institution").value(newInfo.institution))
            .andExpect(jsonPath("$prefix.address.line1").value(newInfo.address.line1))
            .andExpect(jsonPath("$prefix.address.line2").value(newInfo.address.line2))
            .andExpect(jsonPath("$prefix.address.city").value(newInfo.address.city))
            .andExpect(jsonPath("$prefix.address.state").value(newInfo.address.state))
            .andExpect(jsonPath("$prefix.address.postalCode").value(newInfo.address.postalCode))
            .andExpect(jsonPath("$prefix.address.country").value(newInfo.address.country))
            .andExpect(jsonPath("$prefix.contactEmail").value(newInfo.contactEmail))
    }

    client.updateGroup(groupId = groupId, group = newInfo, jwt = jwtForDefaultUser)
        .verifyGroupInfo()

    client.getDetailsOfGroup(groupId = groupId, jwt = jwtForDefaultUser)
        .verifyGroupInfo(prefix = "$.group")
}
```

backend/src/test/kotlin/org/loculus/backend/controller/groupmanagement/GroupManagementControllerTest.kt
fhennig: https://github.com/loculus-project/loculus/pull/2854/commits/cf5d7df8ae96e1f64e0dd4d531fd8ac89e5192b2

https://github.com/loculus-project/loculus/pull/2854/commits/00e5f3a28c5acb98358a9070cdeb8f32746ea6a2

fhennig:
hey @corneliusroemer do you want to have a look at the playwright test I wrote?

fhennig:
I also have a PR ready for the Pathoplexus doc changes but I don't have commit permissions there yet

website/tests/pages/user/group/index.spec.ts
```diff
@@ -25,4 +25,48 @@ test.describe('The group page', () => {
...
+
+        await groupPage.goToUserPage();
+        await groupPage.goToGroupCreationPage();
+
+        const groupName = v4();
+        await groupPage.createGroup(groupName);
+        const newName = v4();
+        const newInstitution = v4();
+        const newEmail = `${v4()}@example.com`;
+        const newCountry = 2;
```
corneliusroemer, line 40: There must be a better way to link list item 2 and Albania

corneliusroemer, line 40:
```suggestion
        const newCountry = 2;
        const newCountryString = listOfCountries[2] // pseudocode
```

docs/src/content/docs/for-users/create-manage-groups.md
```diff
@@ -28,3 +28,16 @@ To remove a user:
 2. Within your groups, select the name of the group that you'd like to remove a user from
 3. Using the 'user' interface at the bottom of the page, click the red button next to the user you wish to remove
 4. The user is now removed
+
+## Edit group information
+
+Anyone in the group can edit group information such as the group name, contact email and adress.
```
corneliusroemer, line 34:
```suggestion
Any group member can edit all group information such as the group name, contact email and address, except for group id which is unchangeable.
```

docs/src/content/docs/for-users/create-manage-groups.md
```diff
@@ -28,3 +28,16 @@ To remove a user:
...
+
+1. Log into your account and navigate to your account page ('My account' in the top-right of the website)
+2. Within your groups, select the name of the group that you would like to edit
+3. In the group detail view, click the 'Edit group' button at the top
+4. Update any information in the edit group form that you would like to change and then click the 'Update group' button
+
+:::caution
+If you change the name of a group the center name in corresponding ENA submissions will also change.
+:::
```
corneliusroemer, line 43:
```suggestion
```

corneliusroemer: Very pathoplexus specific at the moment so better to deduplicate and only keep in pathoplexus docs

corneliusroemer:
We checked that group updates end up in LAPIS (that etag gets refreshed): 
image

website/tests/pages/user/group/index.spec.ts
```diff
@@ -25,4 +25,48 @@ test.describe('The group page', () => {
...
+        const tableLocator = groupPage.page.locator('table');
+
+        await expect(tableLocator).toContainText(newInstitution);
+        await expect(tableLocator).toContainText(newEmail);
+        await expect(tableLocator).toContainText(newLine1);
+        await expect(tableLocator).toContainText(newLine2);
+        await expect(tableLocator).toContainText(newCity);
+        await expect(tableLocator).toContainText(newState);
+        await expect(tableLocator).toContainText(newPostalCode);
+        await expect(tableLocator).toContainText('Albania');
```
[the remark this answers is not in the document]
fhennig: https://github.com/loculus-project/loculus/pull/2854/commits/e52ee6f1fede21ca4193024792176cbc8057f631

website/tests/pages/user/group/index.spec.ts
fhennig: https://github.com/loculus-project/loculus/pull/2854/commits/e52ee6f1fede21ca4193024792176cbc8057f631

fhennig: https://github.com/loculus-project/loculus/pull/2854/commits/e52ee6f1fede21ca4193024792176cbc8057f631

corneliusroemer (approved):
Good stuff, I think all the checkboxes are ticked now!

corneliusroemer:
Woops - E2E still fails it seems

corneliusroemer (changes requested):
E2E test no longer works it seems
https://github.com/loculus-project/loculus/actions/runs/11384438057/job/31672159380?pr=2854#step:21:41
Brave Browser 2024-10-17 14 07 47

fhennig:
damn!

website/tests/pages/user/group/index.spec.ts
```diff
@@ -25,4 +26,49 @@ test.describe('The group page', () => {
...
+        await groupPage.goToUserPage();
+        await groupPage.goToGroupCreationPage();
+
+        const groupName = v4();
+        await groupPage.createGroup(groupName);
+        const newName = v4();
+        const newInstitution = v4();
+        const newEmail = `${v4()}@example.com`;
+        const newCountryIndex = 2;
+        const newCountry = listOfCountries[newCountryIndex];
```
corneliusroemer, line 42:
```suggestion
        const newCountry = listOfCountries[newCountryIndex-1];
```
Seems to be off-by-one

corneliusroemer (approved):
Now tests look good!

PR #2855, opened 2024-09-20
fix(silo-preprocessing): Fix interpolation syntax error

tsibley:
Accidentally introduced in "feat(silo-prepro): use hash instead of line count to test for equality (#1488)" (b0523ad9) when changing the interpolation.

### DIFF
--- a/kubernetes/loculus/silo_import_job.sh
+++ b/kubernetes/loculus/silo_import_job.sh
@@ -140,7 +140,7 @@ download_data() {
         exit 0
       else
         echo "Hashes are unequal, deleting old input data dir"
-        rm -rf "$old_input_data_dir:?}"
+        rm -rf "$old_input_data_dir"
       fi
     fi
   else


corneliusroemer:
Thanks @tsibley!

corneliusroemer (approved):

theosanderson (approved):

theosanderson:
I transferred this commit to https://github.com/loculus-project/loculus/pull/2856 as I'm not sure we have a way to run our CI here

corneliusroemer:
@tsibley I'm curious how you came across this bug - were you reading through the code out of curiosity or did you notice something misbehaving in a test environment you'd set up?

It looks like if we hit this branch previously bash would throw a syntax error instead of removing the directory as intended so I'm wondering why we hadn't caught the bug earlier.

tsibley:
@corneliusroemer I figured you'd be curious. :-) I was reading thru parts of the codebase (on behalf of the Blab Nextstrain team) to understand if/how Loculus could be useful for Nextstrain's curated pathogen data (e.g. to present that data more approachably).

I believe if you hit this branch previously it wouldn't actually throw any error—bash just sees an interpolation (`$old_input_data_dir`) followed by a few literal characters (`:?}`) and the `-f` suppresses `rm`'s complaints via exit status about the path not existing—but instead it would just quietly fail to remove the directory as intended. This kind of error isn't something ShellCheck catches (I checked), but it'd be nice if it did!

tsibley:
(In retrospect, I probably should have written that second paragraph for the commit message as the analysis of a bug's ramification is useful commit context.)

corneliusroemer:
Strange thing is I would have thought we'd catch this kind of failure to remove a directory - but apparently we didn't? Anyways, good that you caught it. We've been discussing rewriting that shell script in a proper language with better linters but haven't got there yet.

@fengelniederhammer here we have an example of bash causing a bug as you predicted.

tsibley:
@chaoran-chen FYI, this same bug exists in the [copy of `silo_import_job.sh` in GenSpectrum/loculus-config](https://github.com/GenSpectrum/loculus-config/blob/660e5f6720362166c0ae7ac2dc5be99c5b2501f6/config/silo-lapis/silo_import_job.sh#L102).

chaoran-chen:
@tsibley, thanks for pointing it out!

PR #2857, opened 2024-09-22
feat(ingest/backend): add and use 'x-total-record' header for /get-original-metadata endpoint

### LINKED ISSUE #2847: Add x-record-total to get-original-metadata

corneliusroemer:
resolves #2847 

preview URL: https://original-x-count.loculus.org

### Summary
Add and use 'x-total-record' header for /get-original-metadata endpoint in ingest and backend.

### PR Checklist

### DIFF
--- a/ingest/scripts/call_loculus.py
+++ b/ingest/scripts/call_loculus.py
@@ -315,6 +315,7 @@ def get_submitted(config: Config):
     logger.info("Getting previously submitted sequences")
 
     response = make_request(HTTPMethod.GET, url, config, params=params)
+    expected_record_count = int(response.headers["x-total-records"])
 
     entries: list[dict[str, Any]] = []
     try:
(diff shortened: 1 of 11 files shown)


ingest/scripts/call_loculus.py
```diff
@@ -327,6 +328,10 @@ def get_submitted(config: Config):
         logger.error(f"Error decoding JSON from /get-original-metadata: {response_summary}")
         raise ValueError from err
 
+    if len(entries) != expected_record_count:
```
anna-parker, line 331: Only wonder if we should retry here instead of raising an error. The case `len(entries) != expected_record_count` is most likely to happen at the start of ingest and we then have to wait 30min until the approve rule finishes for the pods to restart.

anna-parker (approved):
One thought but overall LGTM :-)

ingest/scripts/call_loculus.py
corneliusroemer: How about this now?

corneliusroemer: https://github.com/loculus-project/loculus/pull/2857/commits/b75314da6ef0f90d5ee6cea0339792a3bbc05366

anna-parker: Looks good! Lets wait till ingest succeeds again incase I overlooked sth in the code then feel free to merge :-)

corneliusroemer: Looks good, but let's just wait a bit more because why not :)

anna-parker: Reingest went well, logs look good - marking this as resolved

PR #2858, opened 2024-09-23
feat(deploy): verify loculus version for config file before using it for preprocessing

theosanderson:
We encountered an issue during a rollout in which the config map was changed and triggered a reload of the pod before the pod itself was upgraded to the next version, causing double version bumps in staging. This avoids that by making the config map contain the loculus version that it targets and making an initContainer in the deployment that checks that this matches that expected by the deployment.

https://verify-loculus-ver.loculus.org/


anna-parker:
This might not work, but chatgpt suggested removing the `reloader.stakater.com/auto: "true"` and just annotating the deployment with a config related hash:
```
spec:
  template:
    metadata:
      annotations:
        configHash: "{{ .Values.configmap | sha256sum }}"
```
thus triggering a new deployment when the config changes and/or a sync happens. This seems a bit easier to maintain in future if it would work.

theosanderson:
I don't really see how that can work here @anna-parker (the config map isn't in values)

theosanderson:
As @anna-parker pointed out to me - this should be ingest!!

Issue #2425, opened 2024-08-13
After ENA dev testing complete let main talk to ENA production instance


anna-parker:

anna-parker:
True! We need to update the config in Loculus tho to allow this via a flag change in Pathoplexus

PR #2874, opened 2024-09-24
Update ena-submission config

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/2425

preview URL: https://update-config.loculus.org/

### Summary

1. Adds the field submitToEnaProduction to the values.yaml config (when set to true the ena submission pod will send requests to ena prod)
2. Updates the deploy script to include the ena-submission.yaml config for easier debugging
3. Adds github actions that will block the creation of an ena-submission-image with this flag for loculus (preventing previews from sending requests to ena prod, but still allowing previews to send requests to ena dev using the test user credentials we have in sealed secrets).

### Screenshot

The check works! It blocks E2E tests and ena-submission-image builds (this in turn should block update-argocd-metadata.yml)
image
Just confirmed, argocd can start but ena-submission pod will not as they are unable to pull the new image (as this has not been created)
image
`container "ena-submission" in pod "loculus-ena-submission-5485589f45-t85fm" is waiting to start: trying and failing to pull image`

However, if the preview is already running and then the flag is changed to submitToEnaProduction: true the reloader will push the changes to the ena-submission pod, to prevent this add an additional list of `allowed_submission_hosts` in the defaults. Only hosts in this list will be able to talk to ena production (even if the submitToEnaProduction flag is set to true)

### PR Checklist


.github/workflows/e2e-k3d.yml
```diff
@@ -47,6 +47,13 @@ jobs:
           curl -s https://raw.githubusercontent.com/rancher/k3d/main/install.sh | bash
           k3d version
       - uses: azure/setup-helm@v4
+      - name: Validate submitToEnaProduction is not true in values.yaml
```
theosanderson, line 50: Do we need this check? As far as I can imagine we should never even check secrets that would let us do this in?

anna-parker: So I have secrets checked in for the ena accounts and I do like to be able to test submission to ena dev on the previews (you use the same credentials for prod and dev sadly). Previously I hard coded which ena instance we talked to make sure it could never happen that we talked to ena prod on a preview but now I need to add a flag for pathoplexus.

.github/workflows/e2e-k3d.yml
```diff
@@ -47,6 +47,13 @@ jobs:
           curl -s https://raw.githubusercontent.com/rancher/k3d/main/install.sh | bash
           k3d version
       - uses: azure/setup-helm@v4
+      - name: Wait for submitToEnaProduction check
+        uses: lewagon/wait-on-check-action@v1.3.4
+        with:
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
+          check-name: Validate submitToEnaProduction is not true in values.yaml
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          wait-interval: 10
```
corneliusroemer, line 56:
```suggestion
          wait-interval: 1
```

I don't like this wait-on-check-action, it's stupidly slow to setup, large overhead, but I guess we can use for now

corneliusroemer: Why do we need to fail in e2e-k3d? Ingest is not used in e2e so we don't need to wait for it?

.github/workflows/validate-ena-flags.yaml
```diff
@@ -0,0 +1,31 @@
...
+    env:
+      sha: ${{ github.event.pull_request.head.sha || github.sha }}
+      wait_timeout: ${{ github.ref == 'refs/heads/main' && 900 || 240 }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Validate submitToEnaProduction is not true in values.yaml
+        run: |
+          SUBMIT_TO_ENA_PROD=$(grep 'submitToEnaProduction:' kubernetes/loculus/values.yaml | awk '{print $2}')
+          if [[ "$SUBMIT_TO_ENA_PROD" == "true" ]]; then
```
corneliusroemer, line 28:
```suggestion
          if [[ "$SUBMIT_TO_ENA_PROD" == "true" ]]; then
```

I think this isn't great, you should use `yq` to check the truthiness here rather than look for an exact string match as many values can evaluate to true in yaml, more things than `true` for sure - e.g. `True` is likely also true

corneliusroemer: ChatGPT suggests the following, which is better as it uses the same library to validate the truthyness as the eventual program that reads that config value:

```
      - name: Validate submitToEnaProduction is not true in values.yaml
        run: |
          python -c "
import yaml
with open('kubernetes/loculus/values.yaml', 'r') as file:
    values = yaml.safe_load(file)
    submit_to_ena_prod = values.get('submitToEnaProduction', False)
    if submit_to_ena_prod:
        print('Error: The flag \'submitToEnaProduction\' is set to true - this will submit data to ENA production. Please set it to false in values.yaml')
        exit(1)
"

File: .github/workflows/ena-submission-image.yaml [hunk 4]
```diff
@@ -36,6 +36,13 @@ jobs:
       packages: write
       checks: read
     steps:
+      - name: Wait for submitToEnaProduction check
```
corneliusroemer, line 39: I would say we just make the config check part of the image workflow - it's kind of directly related to it, and would save on the need to use this wait on check action in yet one more place.

corneliusroemer (commented):
I'm wondering whether this is necessary in the loculus code base: if we just don't store the prod credentials in loculus then people can set that value to true and it's not an issue for us.

If there _is_ a place we should check it, it's in pathoplexus values for main/staging/demo.

What do you think @anna-parker?

anna-parker:
So we store the credentials for the ena test account, the ena test account can still submit sequences to ena production it is a normal account it just isn't a broker (and it will submit seqeunces to ena production if we let it talk to ena production, see https://github.com/loculus-project/loculus/blob/main/ena-submission/README.md#testing the warning for details). If you would like we can remove these credentials from loculus but as I mentioned to Theo I like being able to test ena submission to dev on a preview instance

.github/workflows/ena-submission-image.yaml
anna-parker: From my understanding the ena-submission-image workflow runs in the docker image, so it does not have access to the files in the kubernetes folder (I tried this before).

.github/workflows/e2e-k3d.yml
anna-parker: we can remove it in e2e - its most important in the ena-submission-image

.github/workflows/ena-submission-image.yaml
theosanderson: The workflow is just a GitHub action that runs in a GitHub workflow runner. After the checkout step below it will have all the files in the repository.

.github/workflows/validate-ena-flags.yaml
anna-parker: cool - I couldn't install `yq` for the check but maybe this works -> lets try!

.github/workflows/ena-submission-image.yaml
anna-parker: oh awesome! then I can change that! sorry thought the env.DOCKER_IMAGE set the runtime env

anna-parker:
Shit, if the ena-submission pod is already running the reloader will update the config: leading to `ena_submission_url='https://www.ebi.ac.uk/ena/submit/drop-box/submit'`... thankfully this is now looking at an empty folder so nothing has been uploaded but this doesn't work

ena-submission/Snakefile
```diff
@@ -15,6 +15,9 @@ for key, value in defaults.items():
 
 LOG_LEVEL = config.get("log_level", "INFO")
 SUBMIT_TO_ENA_PROD = config.get("submit_to_ena_prod", False)
+if config.get("backend_url", "") not in config.get("allowed_submission_hosts", []):
```
theosanderson, line 18: My expectation would have been that this wouldn't be the external URL but I may be wrong

anna-parker: On this preview the backend_url that is passed here is https://backend-update-config.loculus.org

anna-parker: So I think this is correct?

theosanderson: Ah interesting; @corneliusroemer this potentially complicates your password plans

theosanderson: So basically we should probably change the backend URL to the internal version eventually. I guess the concern is we might do that and still rely on this check which would no longer work

anna-parker: A bit tedious but we could still make that change by first setting this list to include the external and internal url, changing the urls to internal and then removing the external url.

theosanderson: Well I think the internal URL will be the same across deployments. But yes, you're right. We wouldn't accidentally make the check pass – I was thinking of my approach of banning "loculus.org" not your allowList approach, so this is aace

theosanderson (approved):

ena-submission/Snakefile
anna-parker: ah wupps - if they are all the same we might have to add a better check! But yeah we can do that in future - thanks for all the help and feedback!

PR #2876, opened 2024-09-25
cherypicked: remove reloader from cronjob on 4de75675b18e737f59f00eaa2469d2b5403df111

theosanderson:
This is to use temporarily for staging and production on pathoplexus


kubernetes/loculus/templates/ingest-deployment.yaml
corneliusroemer, line 1: We should eventually change the file name if it also contains a cronjob not just a deployment :)

PR #2877, opened 2024-09-25
fix(ingest): Update metadata hash calculation, do not include empty fields in hash

anna-parker:
resolves #

preview URL: https://no-hash-of-empty-fields.loculus.org/ (shows state on a persistent db for easier testing)

### Summary

Desired behavior: Hash of all metadata fields should be the same if
  1. field is not in keys_to_keep and
  2. field is in keys_to_keep but is ""
 
Testing here that in future a config update that removes/adds metadata fields to the list of `fields_to_keep` in ingest will only cause an update of fields that have changed. 

To do this I set to a persistent db with the mapping to geolocadmin2 and then after ingest has completed I remove the mapping to geolocadmin2. I then confirm that this only triggers an update for sequences that previously had a value in the geolocadmin2 field. 

From pathoplexus.org:
- cchf: 2066/2141 sequences have a value for geolocadmin2 (logs show Sequences to revise: 2066, Unchanged sequences: 75)
- ebola_zaire: 3581/3689 (logs show 3581 revised, 274 unchanged)
- west nile: 6216/6940 (logs show 6216 sequences were revised, 16946 are unchanged)

### PR Checklist

### DIFF
--- a/ingest/scripts/group_segments.py
+++ b/ingest/scripts/group_segments.py
@@ -234,8 +234,8 @@ def main(
 
         # Hash of all metadata fields should be the same if
         # 1. field is not in keys_to_keep and
-        # 2. field is in keys_to_keep but is ""
-        filtered_record = {k: v for k, v in row.items() if v}
+        # 2. field is in keys_to_keep but is "" or None
+        filtered_record = {k: str(v) for k, v in row.items() if v is not None and str(v)}
 
         row["hash"] = hashlib.md5(
             json.dumps(filtered_record, sort_keys=True).encode(), usedforsecurity=False
--- a/ingest/scripts/group_segments.py
+++ b/ingest/scripts/group_segments.py
@@ -232,8 +232,13 @@ def main(
 
         row["submissionId"] = joint_key
 
+        # Hash of all metadata fields should be the same if
+        # 1. field is not in keys_to_keep and
+        # 2. field is in keys_to_keep but is ""
+        filtered_record = {k: v for k, v in row.items() if v}
+
         row["hash"] = hashlib.md5(
-            json.dumps(row, sort_keys=True).encode(), usedforsecurity=False
+            json.dumps(filtered_record, sort_keys=True).encode(), usedforsecurity=False
         ).hexdigest()
 
         metadata[joint_key] = row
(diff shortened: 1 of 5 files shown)


theosanderson:
A universal bump is what I expected and I think is ok in the context of our rollout

anna-parker:
> A universal bump is what I expected and I think is ok in the context of our rollout
Ah I am just checking that this has desired behavior in future when we remove/add fields and using geolocadmin as an example (I am not starting from the current state as this will mean a universal version bump and doesn't let me check future changes well)

theosanderson:
👍  Sorry, yeah, bad reading on my part

anna-parker:
> 👍 Sorry, yeah, bad reading on my part

No worries! My description was pretty empty!

ingest/scripts/group_segments.py
```diff
@@ -232,8 +232,13 @@ def main(
 
         row["submissionId"] = joint_key
 
+        # Hash of all metadata fields should be the same if
+        # 1. field is not in keys_to_keep and
+        # 2. field is in keys_to_keep but is ""
```
corneliusroemer, line 237: Or None or other falsy values. This is fine if things are always strings, but would be problematic if we sometimes submit integers or booleans and don't coerce to string immediately

ingest/scripts/group_segments.py
```diff
@@ -232,8 +232,13 @@ def main(
...
+        filtered_record = {k: v for k, v in row.items() if v}
```
corneliusroemer, line 238: We should probably coerce values to string here, so that a 0 integer still gets hashed

corneliusroemer (changes requested):
From the limited context without reading the whole code it's safer to coerce to string before filtering. With that change looks great to me! Thanks a lot!

ingest/scripts/prepare_metadata.py
```diff
@@ -138,7 +138,12 @@ def main(
             msg = f"No hash found for {record[config.fasta_id_field]}"
             raise ValueError(msg)
 
-        metadata_dump = json.dumps(record, sort_keys=True)
+        # Hash of all metadata fields should be the same if
```
anna-parker, line 141: to be fair record is of type dict[str, str] so I am not sure this is needed?

ingest/scripts/group_segments.py
anna-parker: row is produced from segment_metadata: dict[ str, dict[str, str]] so I'm not sure if this is that risky

corneliusroemer (approved):

PR #2888, opened 2024-09-27
feat(prepro): Add tests for metadata processing functions

anna-parker:
resolves #

preview URL:

### Summary

This adds tests to the preprocessing pipeline, using the no-alignment option for the prepro nextclade pipeline. It is basically a high level unit test, where a `prepro_config.yaml` and corresponding `test_metadata.tsv` are generated in the expected user input format. The test then mocks the backend and returns the entries of the `test_metadata.tsv` as `UnprocessedEntries`.

The test then calls the `process_all` function on this input and checks the output is as expected. This is given in the `expected_output.tsv` - this is a simplified version of the `ProcessedEntry` without sequence-related fields and only including the error and warning messages. 

Each entry in the `test_metadata.tsv` corresponds to an error or warning case that the preprocessing pipeline should handle. If you add new preprocessing functions you should add examples that test the edge cases of these functions. 

### PR Checklist


preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -196,23 +200,6 @@ def process_date(
         warnings = []
         errors = []
 
-        # TODO: required check is also in process_single - check if can be removed here
```
anna-parker, line 199: I confirmed with the `missing_one_required_field` test that this error will still be raised

preprocessing/nextclade/tests/test.py
```diff
@@ -0,0 +1,83 @@
...
+        map[("LOC_" + str(pos))] = metadata["submissionId"]
+    return map
+
+
+class PreprocessingTests(unittest.TestCase):
+    def test_process_all(self) -> None:
+        unprocessed = read_in_test_metadata(test_metadata_file)
+        map_accessions_to_submissions = map_accession_to_submission_id(test_metadata_file)
+        config = get_config(test_config_file)
+        processed: list[ProcessedEntry] = process_all(unprocessed, "temp", config)
```
fengelniederhammer, line 54: Is the "temp" value important? What does it do? Will there be something that gets written to that directory?

preprocessing/nextclade/tests/test.py
```diff
@@ -0,0 +1,83 @@
...
+            submission_id = map_accessions_to_submissions[entry.accession]
+            expected_output_entry = expected_output[submission_id]
+            error_messages = {error.message for error in entry.errors}
+            expected_error_messages = set(expected_output_entry["error_messages"])
+            if error_messages != expected_error_messages:
+                message = (
+                    f"{submission_id}: Error messages: {error_messages} do not match expected "
+                    f"error messages: {expected_error_messages}."
+                )
+                raise AssertionError(message)
```
fengelniederhammer, line 65: Is this how one asserts in Python? From a quick glimpse into the Pytest docs I got the impression that one should use `assert`?

fengelniederhammer (commented):
Thanks for implementing this!

I'd like to propose a different format for the input data though (beware - horrible pseudocode):

```python
# Array of dicts { input: UnprocessedEntry, expected_output: ProcessedEntry }
test_cases = [
    {
        "input": 
            UnprocessedEntry(
                accessionVersion= "accessionVersion",
                data=UnprocessedData(
                    submitter="test_submitter",
                    metadata={"field1": "value1", "field2": "value2"},
                    unalignedNucleotideSequences={"main": ""},
                ),
            ),
        "expected_output": ProcessedEntry(
            "fields": {
                "continent": null,
                "collection_date": null,
                "sequenced_timestamp": null,
                "age_int": null,
                "percentage_float": null,
                "name_required": null,
                "other_date": null,
                "is_lab_host_bool": null,
                "required_collection_date": null,
                "concatenated_string": "LOC_0.1"
            },
            "error_messages": [
                "Metadata field name_required is required.",
                "Metadata field required_collection_date is required."
            ],
            "warning_messages": []
        ),
    },
    {
        "input": ...,
        "expected_output": ...,
    },
    ...
]
``` 

That would have the advantage that you don't need to
* read files
* parse the CSV file into `UnprocessedEntry`s
* think of accession versions
* match the input with its expected output (because they are naturally attached!)

I imagine this to be executed like
```
for test_case in test_cases:
    result = process_all(test_case.input, "temp", config)
    processed_entry = result[0]
    assert processed_entry == test_case.expected_output
```

I think it would simplify the code a lot. And I think as next steps, it would easily possible to
* also alignment, computed metadata and other things that pop out of Nextclade
* group test cases thematically:
```
test_cases = data_test_cases + int_test_cases + alignment_test_cases + ...
```
where the individual test case lists could live in different files (thus keeping the individual file lengths within a certain limit).

preprocessing/nextclade/tests/test.py
anna-parker: It is actually not used here as it is only used when we use nextclade for alignment - and here we do not align. So "temp" essentially does nothing but is a required field for the function

preprocessing/nextclade/tests/test.py
anna-parker: You can do either - we are using ruff as a formatter and ruff doesn't like assert (see https://docs.astral.sh/ruff/rules/assert/), additionally this way I can write a custom message so I switched to AssertionError because of that.

anna-parker:
Hi @fengelniederhammer, `ProcessedEntry` is actually quite a complicated data structure - which is also why I simplified the output in the files. See https://github.com/loculus-project/loculus/blob/main/preprocessing/nextclade/src/loculus_preprocessing/datatypes.py#L105. 

In order for your suggestion to work I could create a new dataclass for a simplified ProcessedEntry and then construct a ProcessedEntry from that dataclass for comparison with the true output.

I can try this out - but it might make the test code actually more complex - I will create a separate branch for this and then let you know :-)

fengelniederhammer:
> `ProcessedEntry` is actually quite a complicated data structure

Two thoughts on this:
* If it's complicated, then it's probably worth testing it in its full extent?
* You can write factory methods that create `ProcessedEntry`s that fill up fields with reasonable default values, if they are not interesting for the test case that one currently writes.

anna-parker:
Closing in favor of https://github.com/loculus-project/loculus/pull/2971,

Issue #2895, opened 2024-09-28
Correct mapping of chromosome/gca accession fields in check_ena function.


anna-parker:
My current function does not expect:
1. That we can get either chromosome or gca accession before the other
2. That for a single chromosome the returned accession is of the format `x1-x1`

```
07:40:24  WARNING (ena_submission_helper.py: 453) - ENA Check returned errors or is in unexpected format. Request: <PreparedRequest [GET]>, Response: [{"report":{"id":"{erz_accession}","analysisType":"SEQUENCE_ASSEMBLY","acc":"chromosomes:OZ185635-OZ185635","processingStatus":"COMPLETED","processingStart":"27-09-2024 06:06:40","processingEnd":"27-09-2024 06:07:06","processingError":null},"links":[]}]
07:45:24    ERROR (ena_submission_helper.py: 407) - Unexpected response format: genome not in acc_dict
```
This means that the ena-submission pipeline will keep sending a get request until it returns the gca and insdc accession and will not update the database with the results until it has received both

Additionally, for single segmented viruses the code with add "_main" to the insdcAccession currently.

anna-parker:
Ah I also wonder, the status is "processingStatus":"COMPLETED" - so maybe we actually only get the chromosome accession without the gca when the sequences are private?

PR #2896, opened 2024-09-28
fix(ena-submission): Parse get analysis process response better for single segment case, return partial results and submit projects with holdUntilDate

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/2895

preview URL: https://patch-create-assembly.loculus.org/

### Summary

Note that chromosome accessions are the accessions of the nucleotides in each submitted segment, in PP we call this insdcAccession_{segment}.

1. Rename check_ena to get_ena_analysis_process, refactor so that get_chromsome_accessions is a separate sub-function of get_ena_analysis_process, and results are returned when either the gca accession OR the chromosome accessions are returned (in contrast to now only returning results when both are returned)
2. Add tests that get_chromsome_accessions works for single and multi-segmented viruses now that we know the expected response format for both. I added a test where I mock the response using a known response text for a 1-segment case and confirm this works,
3. Update assembly_table.results column response results as soon as gca or chromosome accessions are known, but only change from WAITING to SUBMITTED state when both are known. Do not update table when there are no results or results are the same as already in table. 

This PR additionally modifies the project submission request to by default submit projects with a `hold_until_date` parameter which is set to the day of submission (as we anyways only submit data to ENA when it is OPEN we should make the project public). 

### Testing

Tested on preview by
- setting github link to https://github.com/pathoplexus/ena-submission/pull/5
- connecting to the db using portforwarding and altering results to known erz accessions
- for the 2-segment case set to `ERZ24784470`, confirm this leads to state SUBMITTED:
```
17:36:33     INFO (  create_assembly.py: 513) - Assembly submission for accession LOC_000RBTK succeeded and accession returned!
```
- Sadly the erz accessions are private so I could not fully confirm this works for the single segment case
- I additionally test that when only the chromosome accession is returned the table (by editing out code to add in the gca accession) the table stays in state WAITING:
```
17:52:59     INFO (  create_assembly.py: 489) - Partial results of assembly submission for accession LOC_000RBTK returned!
```

### DIFF
--- a/ena-submission/Snakefile
+++ b/ena-submission/Snakefile
@@ -28,7 +28,7 @@ if SUBMIT_TO_ENA_DEV:
     print("Submitting to ENA dev environment")
     config["ena_submission_url"] = "https://wwwdev.ebi.ac.uk/ena/submit/drop-box/submit"
     config["github_url"] = (
-        "https://raw.githubusercontent.com/pathoplexus/ena-submission/main/test/approved_ena_submission_list.json"
+        "https://raw.githubusercontent.com/pathoplexus/ena-submission/loculus_test/test/approved_ena_submission_list.json"
     )
     config["ena_reports_service_url"] = "https://wwwdev.ebi.ac.uk/ena/submit/report"
 
--- a/ena-submission/Snakefile
+++ b/ena-submission/Snakefile
@@ -28,7 +28,7 @@ if SUBMIT_TO_ENA_DEV:
     print("Submitting to ENA dev environment")
     config["ena_submission_url"] = "https://wwwdev.ebi.ac.uk/ena/submit/drop-box/submit"
     config["github_url"] = (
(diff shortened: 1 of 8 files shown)


ena-submission/Snakefile
```diff
@@ -28,7 +28,7 @@ if SUBMIT_TO_ENA_DEV:
     print("Submitting to ENA dev environment")
     config["ena_submission_url"] = "https://wwwdev.ebi.ac.uk/ena/submit/drop-box/submit"
     config["github_url"] = (
-        "https://raw.githubusercontent.com/pathoplexus/ena-submission/main/test/approved_ena_submission_list.json"
+        "https://raw.githubusercontent.com/pathoplexus/ena-submission/loculus_test/test/approved_ena_submission_list.json"
```
anna-parker, line 31: Revert before merging!
```suggestion
        "https://raw.githubusercontent.com/pathoplexus/ena-submission/main/test/approved_ena_submission_list.json"
```

corneliusroemer (approved):

Issue #2675, opened 2024-08-30
website: Submitting group shows up twice in the search


anna-parker:
<img width="315" alt="image" src="https://github.com/user-attachments/assets/d9aa75eb-ed90-49d0-9376-149747222def">

theosanderson:
This is an unforeseen (though known in the last few days) result of the changes made to combine these (ID and text versions) on the details page. Should be very fixable. Thanks for the issue.

PR #2899, opened 2024-09-29
fix(website): Change submitting group ID displayName 

### LINKED ISSUE #2633: Duplicated fields in search for 'Submitting group'
Submitting group appears twice: one is the submitter ID and one is the submitting group name, but this is not clear from the search view:
image

Link to try: 
https://demo.pathoplexus.org/west-nile/search?visibility_dataUseTerms=false&visibility_submissionId=false&visibility_submitter=false&visibility_groupId=true&visibility_groupName=true

theosanderson:
Resolves #2633, resolves https://github.com/loculus-project/loculus/issues/2675

Preview: https://groupchange.loculus.org/

image

There are some technical constraints here: the first listed name is the one that will appear in the combined field on the sequence page, which is why I've gone for this exact solution. Happy for a fuller change in the future.

image

### DIFF
--- a/kubernetes/loculus/templates/_common-metadata.tpl
+++ b/kubernetes/loculus/templates/_common-metadata.tpl
@@ -39,21 +39,21 @@ fields:
     generateIndex: true
     autocomplete: true
     header: Submission details
-  - name: groupId
-    displayName: Group ID
-    type: int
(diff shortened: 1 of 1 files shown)


anna-parker (approved):
LGTM!

PR #2900, opened 2024-09-30
fix(ena-submission): fix ena request error handling

anna-parker:
resolves #

preview URL: https://fix-ena-request-error-han.loculus.org/

### Summary

Currently one of the ena-submission pod rules has failed with an exception in PP prod:
```
File "/opt/conda/lib/python3.12/site-packages/requests/adapters.py", line 713, in send
    raise ReadTimeout(e, request=request)
requests.exceptions.ReadTimeout: HTTPSConnectionPool(host='raw.githubusercontent.com', port=443): Read timed out. (read timeout=10)
[Sun Sep 29 14:30:14 2024]
Error in rule trigger_submission_to_ena:
```

Python requests (using `requests.get` and `requests.post` library) can fail in two different ways: 
1. The server responds with a non 200-status code, `requests.get` and `requests.post` will return with a Response object that has this non-200 status code. No exception will be raised unless `response.raise_for_status()` is used, in this case if the Response object has a non-200 status code it will raise a requests.exceptions.RequestException.
2. Another failure occurs and the response from the server is not returned (e.g. timeout, ssl issue etc.) , in this case a requests.exceptions.RequestException is raised and the request does not return a Response object.

Here I add exception handling for the second case.

I additionally add the option to set the logging level to debug via the config. 

### PR Checklist


ena-submission/Snakefile
```diff
@@ -28,7 +28,7 @@ if SUBMIT_TO_ENA_DEV:
     print("Submitting to ENA dev environment")
     config["ena_submission_url"] = "https://wwwdev.ebi.ac.uk/ena/submit/drop-box/submit"
     config["github_url"] = (
-        "https://raw.githubusercontent.com/pathoplexus/ena-submission/main/test/approved_ena_submission_list.json"
+        "https://raw.githubusercontent.com/pathoplexus/ena-submission/loculus_test/test/approved_ena_submission_list.json"
```
anna-parker, line 31: Revert this change before merging!
```suggestion
        "https://raw.githubusercontent.com/pathoplexus/ena-submission/main/test/approved_ena_submission_list.json"
```

ena-submission/scripts/trigger_submission_to_ena.py
```diff
@@ -103,19 +103,25 @@ def trigger_submission_to_ena(
...
+                config.github_url,
+                timeout=10,
+            )
+            response.raise_for_status()
+        except requests.exceptions.RequestException as e:
+            logger.error(f"Failed to retrieve file due to requests exception: {e}")
+            time.sleep(min_between_github_requests * 60)
+            continue
+        try:
             sequences_to_upload = response.json()
```
corneliusroemer, line 119:
```suggestion
            sequences_to_upload = response.json()
```

I think you could move the sequences_to_upload outside of the try block?

ena-submission/scripts/trigger_submission_to_ena.py
```diff
@@ -103,19 +103,25 @@ def trigger_submission_to_ena(
...
-        else:
-            error_msg = f"Failed to retrieve file: {response.status_code}"
-            logger.error(error_msg)
-        upload_sequences(db_config, sequences_to_upload)
-        time.sleep(min_between_github_requests * 60)  # Sleep for x min to not overwhelm github
+            upload_sequences(db_config, sequences_to_upload)
+        except Exception as upload_error:
+            logger.error(f"Failed to upload sequences: {upload_error}")
+        finally:
```
corneliusroemer, line 123: I think the finally is unnecessary because if an exception is thrown we won't need to sleep anyways

corneliusroemer (approved):
Minor non-blocking comments, good stuff!

ena-submission/scripts/trigger_submission_to_ena.py
anna-parker: I think its good to keep it in, if we for what ever weird reason can't upload to the db we should probably rate limit a bit not to overwhelm github or the database even more.

kubernetes/loculus/templates/ena-submission-config.yaml
```diff
@@ -4,6 +4,7 @@
 {{- $submitToEnaProduction := .Values.submitToEnaProduction | default false }}
 {{- $enaDbName := .Values.enaDbName | default false }}
 {{- $enaUniqueSuffix := .Values.enaUniqueSuffix | default false }}
+{{- $enaLogLevel := .Values.enaLogLevel | default "INFO" }}
```
anna-parker, line 7: I just realized these aren't required - @corneliusroemer you showed me how to set this in the values.yaml already so I will remove before merging.

ena-submission/scripts/trigger_submission_to_ena.py
corneliusroemer: In my understanding, if we throw an uncaught exception, it will exit the loop anyways - but yeah maybe it restarts then? Doesn't really matter, so happy to let go, just noticed when I read the code and was wondering.

corneliusroemer:
I also approve this change:
image

Thanks!

PR #2901, opened 2024-09-30
feat(website): add "powered by Loculus" note

chaoran-chen:
preview URL:

### Summary

This adds a note about Loculus to the footer.

I added a ❤️ emoji to show our love :)

### Screenshot

image

### PR Checklist

### DIFF
--- a/website/src/layouts/BaseLayout.astro
+++ b/website/src/layouts/BaseLayout.astro
@@ -95,6 +95,10 @@ const lastTimeBannerWasClosed = Astro.cookies.get('lastTimeBannerWasClosed')?.va
                     }
                 </div>
 
+                <div>
+                    Powered by <a href='https://loculus.org' class='underline'>Loculus</a> ❤️
+                </div>
+
                 <a href='https://github.com/loculus-project' class='h-full py-6'>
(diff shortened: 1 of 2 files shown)


chaoran-chen:
@theosanderson, as mentioned before, I personally think that "Powered by" is a good (because it is quite common in this context) wording but happy to hear other suggestions!

website/src/layouts/BaseLayout.astro
theosanderson, line 1: Thanks Chaoran - as previously discussed I'm personally against hardcoding a link to Loculus into every page of deployments. I think that it doesn't provide user-value, and as a result looks a little tacky - which is I think why SaaS companies are able to charge extra for the removal of such banners. It feels particularly unexpected to me on pages like the Governance page, as I mentioned. I'm of course happy to be outvoted.

tanja819: I have no strong opinions. Generally Powered by Loculus may be useful, and also highlight that Loculus \neq Pathoplexus. I'd go without the heart.

rneher:
I don't have strong feelings. My preference would probably be to have it on pages that are database related, not others. But a small footer text isn't too obstructive and I'd be ok with having it everywhere. I'd prefer it without emoji.

PR #2903, opened 2024-09-30
Move version comment to common metadata

fengelniederhammer:
preview URL: https://versioncommenttocommonmet.loculus.org/

### Summary

Because the backend always emits the field - so it should go with the other "Loculus internal metadata fields".

### PR Checklist
- [x] Make pathoplexus PR as this is a breaking change: https://github.com/pathoplexus/pathoplexus/pull/190
- [ ] Synchronize merge of pathoplexus PR https://github.com/pathoplexus/pathoplexus/pull/190


theosanderson:
This looks great - we should check things fairly carefully and then we should make sure to make the corresponding removal from the pathoplexus values as we roll it out

fengelniederhammer:
I tested it - it's working when revoking a sequence:

grafik

fengelniederhammer:
Feel free to merge this PR once you think it's good to go.

theosanderson (approved):
Thank you!

theosanderson:
(I'm away for a bit so personally not gonna merge now - happy for anyone to. Probably best not to until we're literally ready to bump everything)

corneliusroemer:
@anna-parker and I are just rolling something out but that's underway, so let's do it later

corneliusroemer:
All done

corneliusroemer:
This PR made versionComments no longer be treated as things users input, because things in commonMetadata are interpreted by some Helm functions as being stuff that's auto-generated. I didn't make this explicit, so it's not obvious that this was going to happen here. We noticed this just now.

Why does the versionComment get always emitted by backend? Is it treated specially for anything by the backend?

theosanderson:
It's at minimum treated specially by the frontend - the revocation form asks for it

theosanderson:
I actually am not sure if this PR had the effect you say - because adding it broke a test re the generation of the template - and this PR didn't affect any tests

corneliusroemer:
> I actually am not sure if this PR had the effect you say - because adding it broke a test re the generation of the template - and this PR didn't affect any tests

If we don't have a test that checks that versionComment can be input/edited in the "edit" form (and we probably don't?) then I don't see how it would have necessarily broken tests? I.e. tests not breaking doesn't mean that it didn't have the effect.

All metadata fields defined in values.yaml are by default inputtable, unless they have `noInput: true`. `versionComment` didn't have that field so it should have been inputtable (unless I miss some special treatment of versionComment somewhere else)

theosanderson:
Here is the broken test on adding it back:
https://github.com/loculus-project/loculus/actions/runs/11612964061/job/32337627087
image

theosanderson:
@corneliusroemer pointed out the test was added after!

corneliusroemer:
Indeed :) The test seems to have been added afterwards (not so easy to find but git blame helped!)
https://github.com/loculus-project/loculus/commit/2a1501d1890079441dc65c08f1ca91ddc790cf1d

It was stupid to use submissionId in `extraInputFields` in my original yaml, anyways, your approach looks like decent hack that I can fix properly in a slightly bigger helm refactor

PR #2905, opened 2024-09-30
fix(config): helm default removed

corneliusroemer:
resolves #

preview URL:

### PR Checklist

### DIFF
--- a/kubernetes/loculus/templates/ingest-config.yaml
+++ b/kubernetes/loculus/templates/ingest-config.yaml
@@ -1,5 +1,4 @@
 {{- $dockerTag := include "loculus.dockerTag" .Values }}
-{{- $disableEnaSubmission := .Values.disableEnaSubmission | default true }}
 {{- $testconfig := .Values.testconfig | default false }}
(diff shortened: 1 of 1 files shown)


theosanderson:
Whether in this PR or elsewhere, could we at some stage disable ENA submission by default? (I just checked Values.yaml and realised we don't)

I realise it's not a real problem as one needs secrets, etc. - but I think default Loculus behaviour should not be (and should not appear to be) to submit to ENA

anna-parker (approved):

corneliusroemer:
The change here is because somehow helm default seems to not work in our tests - this is a workaround - we're happy to make an issue to change defaults later.

anna-parker:
@theosanderson if its ok as we are currently rolling this out to PP and testing it by default in Loculus is helpful, but yes in a week or so I am very happy to set this to disabled by default!

theosanderson:
There's a way to achieve your desired behaviour now :) I'll make a PR

corneliusroemer:
> There's a way to achieve your desired behaviour now :) I'll make a PR

I'm curious, we just lost quite a few hairs trying to figure the behaviour out. Would have been 1min with cdk8s 🙃

theosanderson:
> I'm curious, we just lost quite a few hairs trying to figure the behaviour out. Would have been 1min with cdk8s 🙃

I think we're probably at cross-purposes, I was just replying to Anna's last message: https://github.com/loculus-project/loculus/pull/2907 (which I know won't be news to you).

Issue #2906, opened 2024-09-30
Disable ENA submission by default, rename disableEnaSubmission to enableEnaSubmission


anna-parker:
Whether in this PR or elsewhere, could we at some stage disable ENA submission by default? (I just checked Values.yaml and realised we don't)

I realise it's not a real problem as one needs secrets, etc. - but I think default Loculus behaviour should not be (and should not appear to be) to submit to ENA

_Originally posted by @theosanderson in https://github.com/loculus-project/loculus/issues/2905#issuecomment-2383649583_

PR #2907, opened 2024-09-30
feat(deployment)disable ENA submission by default but re-enable it for preview and e2…

theosanderson:
…e/dev

relates to #2906

https://disable-ena.loculus.org/

### DIFF
--- a/kubernetes/loculus/values_e2e_and_dev.yaml
+++ b/kubernetes/loculus/values_e2e_and_dev.yaml
@@ -7,5 +7,6 @@ secrets:
       externalMetadataUpdaterPassword: "external_metadata_updater"
       backendUserPassword: "backend"
 createTestAccounts: true
+disableEnaSubmission: false
 backendExtraArgs:
   - "--loculus.debug-mode=true"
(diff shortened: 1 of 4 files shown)


kubernetes/loculus/values_e2e_and_dev.yaml
```diff
@@ -7,5 +7,6 @@ secrets:
       externalMetadataUpdaterPassword: "external_metadata_updater"
       backendUserPassword: "backend"
 createTestAccounts: true
+disableEnaSubmission: false
```
anna-parker, line 10: maybe we can disable this here as well, enabling submission for the e2e tests seems wrong, for local testing it can be changed manually. 
```suggestion

```

anna-parker (approved):

kubernetes/loculus/values_e2e_and_dev.yaml
theosanderson: Sorry I forgot about this PR then almost created a new one but found the branch name was taken and found it again!

PR #2909, opened 2024-09-30
feat(deployment): Add basic-auth to keycloak

theosanderson:
Adds basic auth for keycloak for use on staging with a sealed secret

https://basic-auth-keycloak.loculus.org/

loculus / widetailpotato

### DIFF
--- a/kubernetes/loculus/values_preview_server.yaml
+++ b/kubernetes/loculus/values_preview_server.yaml
@@ -1,10 +1,5 @@
 createTestAccounts: true
 secrets:
-  basicauth:
-    type: rawhtpasswd
-    data:
-      username: loculus
-      password: widetailpotato
   smtp-password:
     type: sealedsecret
     clusterWide: "true"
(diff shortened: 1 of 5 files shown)


kubernetes/loculus/values_preview_server.yaml
```diff
@@ -1,5 +1,10 @@
 createTestAccounts: true
 secrets:
+  basicauth:
```
theosanderson, line 3: revert before merging

corneliusroemer:
Ingest and/or prepro seem to fail? Don't see any sequences. We might have to add basic auth support to the pipeline's keycloak key requests?

corneliusroemer:
Ah, were using public DNS instead of kubernetes internal one for keycloak url. Switching out the URL to use kubernetes internal one should be enough to fix prepro and ingest.

corneliusroemer:
@theosanderson Anya and I fixed the issue by using k8s-internal DNS now, tested and works. I've also reverted the basicauth, so will merge this and make PP PR so we can protect staging soonish. Anya and I will coordinate this

PR #2914, opened 2024-10-01
fix(website): don't display data as "restricted" when the last restricted data use terms expired

fengelniederhammer:
preview URL: http://fixdatausetermsdisplay.loculus.org

### Bug

Website seq details showed restricted even after expiration of restriction.

See https://loculus.slack.com/archives/C05G172HL6L/p1727769592602569

### Summary

Root cause was misunderstanding in website code of what the dataUseTermsHistory contains.

The website code assumed there would be a new entry when something becomes open, but in fact the history is just a log of changes.

The backend only returns the history - which does not include "OPEN" as a last entry when the restricted period expired. LAPIS has the correct values - let's simply rely on that.

This assumes the following as correct:
* The backend just stores the history (not the state) of data use terms. 
* getting the history from the backend for a sequence entry that has expired restricted use terms will still only show "restricted" entries

We now simply take the use terms from LAPIS, that means the calculation is done in a single place for search and display.

### PR Checklist

### DIFF
--- a/website/src/components/SequenceDetailsPage/SequenceDataUI.tsx
+++ b/website/src/components/SequenceDetailsPage/SequenceDataUI.tsx
@@ -1,13 +1,16 @@
+import type { FC } from 'react';
+
 import DataTable from './DataTable';
 import { RevokeButton } from './RevokeButton';
 import { SequencesContainer } from './SequencesContainer';
 import { getDataTableData } from './getDataTableData';
 import { type TableDataEntry } from './types';
 import { routes } from '../../routes/routes';
-import { type DataUseTermsHistoryEntry, type Group } from '../../types/backend';
+import { DATA_USE_TERMS_FIELD } from '../../settings.ts';
+import { type DataUseTermsHistoryEntry, type Group, type RestrictedDataUseTerms } from '../../types/backend';
 import { type Schema } from '../../types/config';
 import { type ReferenceGenomesSequenceNames } from '../../types/referencesGenomes';
-import { type RuntimeConfig, type ClientConfig } from '../../types/runtimeConfig';
+import { type ClientConfig, type RuntimeConfig } from '../../types/runtimeConfig';
 import { EditDataUseTermsButton } from '../DataUseTerms/EditDataUseTermsButton';
 import ErrorBox from '../common/ErrorBox';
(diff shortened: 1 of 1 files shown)


corneliusroemer:
Testing this here:
- https://github.com/pathoplexus/pathoplexus/pull/197
- https://github.com/pathoplexus/loculus_deployments/pull/90
As it's easiest tested on production-like data

corneliusroemer:
Ok test works! 🎉 
Brave Browser 2024-10-01 13 46 37

I'll add the line to check data use terms are string so as not to produce wrong data.

corneliusroemer (approved):

website/src/components/SequenceDetailsPage/SequenceDataUI.tsx
```diff
@@ -44,12 +47,8 @@ export const SequenceDataUI: React.FC<Props> = ({
     dataUseTermsHistory.sort((a, b) => (a.changeDate > b.changeDate ? -1 : 1));
     const currentDataUseTerms = dataUseTermsHistory[0].dataUseTerms;
 
-    const dataUseTermsIndex = tableData.findIndex((entry) => entry.name === 'dataUseTerms');
-    if (dataUseTermsIndex !== -1) {
-        tableData[dataUseTermsIndex].value = currentDataUseTerms.type;
-    }
-
-    const isRestricted = currentDataUseTerms.type === 'RESTRICTED';
+    const dataUseTerms = tableData.find((entry) => entry.name === DATA_USE_TERMS_FIELD);
```
corneliusroemer, line 50:
```suggestion
    const dataUseTerms = tableData.find((entry) => entry.name === DATA_USE_TERMS_FIELD);
    // Calculation below is based on the assumption that dataUseTerms is a string
    // Hence assert to get loud error instead of quiet false data
    if (typeof dataUseTerms?.value !== 'string') {
        throw new Error('Code assumes dataUseTerms are of type string. If it isn't, then need to change the calculation of isRestricted.');
    }
```

fengelniederhammer: Sorry - already merged. If you still think it's necessary, you can open a separate PR.

PR #2917, opened 2024-10-01
feat: Allow selecting sequences for download by checkbox

### LINKED ISSUE #2247: Allow selection of sequences for download by ticking checkboxes in browse/search view
This comes from our beta testers:

It would be nice to be able to checkbox-select sequences in the browse view to subset sequences for download.

Implementation wise, we could store the list of selected sequences in URL path, similar to how we allow filtering by accession. Just that in this case we only filter at the point of _download_ not as soon as the first accession is entered.

fhennig:
resolves #2247

preview URL: https://feat-2247-checkboxes.loculus.org/

### Summary

- The table now has an additional column of checkboxes.
- When boxes are checked, there is an uncheck-all button in the header of the column
- The number of checked sequences is shown in the little text above the table
- When downloading, the number of selected sequences is shown as a filter. Only this will be shown, as all other filters won't be taken into account. The selected sequence IDs are instead used to specify which sequences to download

### Screenshot
Screenshot 2024-10-02 at 11 46 44

Screenshot 2024-10-02 at 11 46 22

### PR Checklist


website/src/components/SearchPage/SearchFullUI.tsx
```diff
@@ -117,6 +117,23 @@ export const InnerSearchFullUI = ({
         });
     };
 
+    const selectedSeqs = JSON.parse(state.selectedSeqs ?? '[]') as string[];
+
+    const setSelectedSeqs = (newSelectedSeqs: string[]) => {
```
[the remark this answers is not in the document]
fhennig: Yeah true, I remember. I'm still wrapping my head around how stuff works, and to get started I had just put it there with all the other `state`. 😅 

It shouldn't be too difficult to split it out into it's own little state thingie later on, right? My plan was to get _something_ working and iterate from there.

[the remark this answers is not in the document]
fhennig: Thanks for pointing it out though!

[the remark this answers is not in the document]
theosanderson: Yeah it's arguably easier to do it with `useState` than to follow the pattern here :)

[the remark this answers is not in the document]
fhennig: yep, changed it: https://github.com/loculus-project/loculus/pull/2917/commits/f950ac04984807e194a752a20edf5744c59bd597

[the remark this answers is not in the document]
fhennig: Yeah, fiddling with the encoding was tricky 😬

fhennig:
There is a test failure and a couple of errors/warning in the JS console, but I think it's working sort of 😄  I'll get to fixing all that next!

chaoran-chen:
The checkboxes are behaving a bit strangly on my device (Chrome 129 on Mac) 😅

https://github.com/user-attachments/assets/2e094f7c-c48f-4dde-b4c3-d2b3fc68f4ed

theosanderson:
(same as Chaoran - great work though!)

website/src/components/SearchPage/Table.tsx
```diff
@@ -136,6 +157,20 @@ export const Table: FC<TableProps> = ({
                                 onClick={(e) => handleRowClick(e, row[primaryKey] as string)}
                                 onAuxClick={(e) => handleRowClick(e, row[primaryKey] as string)}
                             >
+                                <td
+                                    className='px-2 whitespace-nowrap text-primary-900 md:pl-6'
+                                    onClick={(e) => {
+                                        e.stopPropagation();
+                                        handleRowSelectToggle(e, row[primaryKey] as string);
```
theosanderson, line 164: I think rather than just toggling its nice to really link with the state of the HTML element: https://chatgpt.com/share/66fc61c9-5fa4-8005-8e16-a492aaa0f10f

theosanderson: (that may or may not solve the issue Chaoran and I are seeing)

website/src/components/SearchPage/Table.tsx
```diff
@@ -95,6 +100,14 @@ export const Table: FC<TableProps> = ({
         }
     };
 
+    const handleRowSelectToggle = (e: React.MouseEvent<HTMLTableCellElement>, seqId: string) => {
+        if (selectedSeqs.includes(seqId)) {
+            setSelectedSeqs(selectedSeqs.filter((item) => item !== seqId));
```
theosanderson, line 105:
```suggestion
            setSelectedSeqs((prevSelectedSeqs) => prevSelectedSeqs.filter((item) => item !== seqId));
```

This isn't necessarily essential here - I'm not sure But since you mentioned you were new to react I thought I'd mention this pattern, which I tend to like for things like this, which ensures that the operation happens on the current up to date version of the state. Otherwise you can sometimes get weird behaviour where e.g. if you unclicked two boxes very fast the second box click might restore the first box because its `selectedSeqs` still includes that first box.

theosanderson: (same below)

fhennig: thanks, very good suggestion! I've fixed both instances here: https://github.com/loculus-project/loculus/pull/2917/commits/cf9ab14a549ae09efda5a4fb8712352b1fbf8900

website/src/components/SearchPage/Table.tsx
fhennig: I'm not quite sure what you mean I think. The checkbox `checked` property is linked up with the state.

I trie to be clever and `preventDefault` on the click event in the checkbox and then bubble it up, but that prevented the checkbox from showing checked. Now I changed it to simply call teh handleRowSelectToggle from both elements (the checkbox and the cell). If you have a good idea how to have the call only in a single handler, let me know!

fhennig: https://github.com/loculus-project/loculus/pull/2917/commits/9cbf3a2e8867d9e469f25a4644f82b6595c195ed

theosanderson: Sorry, yes I should have explained better. What I meant was making use of `e.target.checked` is the handler so that when setting you are actually checking the HTML, rather than assuming the HTML matches the React state. I'm not particularly doubting that it will match the React state but it just feels like good practice.

website/src/components/SearchPage/SearchFullUI.tsx
```diff
@@ -174,6 +174,8 @@ export const InnerSearchFullUI = ({
     const aggregatedHook = hooks.useAggregated({}, {});
     const detailsHook = hooks.useDetails({}, {});
 
+    const [selectedSeqs, setSelectedSeqs] = useState<string[]>([]);
```
theosanderson, line 177: Atm I'm often seeing that I tick one thing and the number increments by 2. We could use a `Set` which (A) probably makes sense (B) might fix this, depending on what the bug is.

website/src/components/SearchPage/Table.tsx
fhennig: Ah, I think I understand now. Have a look again?

https://github.com/loculus-project/loculus/pull/2917/commits/4c67c812c29bd1dd04724d3cece578713ae1aec3

I'm still not happy with calling the handler twice (once from the `td` onClick and once from the checkbox onChange) but I think there isn't a clean way around it.

website/src/components/SearchPage/SearchFullUI.tsx
fhennig: The bug should be fixed now, but a Set still makes sense I believe.

website/src/components/SearchPage/Table.tsx
theosanderson: Ah I think I hadn't understood the issue. I'm afk ATM but it sounds like the solution could be stopPropagation instead of prevent default. I'll check back later

website/src/components/SearchPage/SearchFullUI.tsx
fhennig: https://github.com/loculus-project/loculus/pull/2917/commits/4bcc0b0e654be687aadb1bf3a6adc7c9bd884702

website/src/components/SearchPage/Table.tsx
fhennig: The issue I have is that I want the cell to also be clickable, and a cell click should also check the checkbox. But if I call my handler function from the onClick of the cell, as well as from the onChange of the checkbox, if at some point the handlerfunction gets change it's easy to overlook that it is called in two places.

I'd prefer if I could just call it from the checkbox for example, and from the cell simply trigger the change on the checkbox and then have the onChange handler on the checkbox do the rest. But I tried that by manually dispatching a change event to the checkbox, but that doesn't work because the onChange react handler doesn't get triggered by that.

Another way could be to only to the handling in the cell, and from the checkbox bubble up something instead, but that also feels messy. So for now I settled on having the function call duplicated -- it's only 5 lines apart so it should be fine I think.

theosanderson: Ah I hadn't understood this was about making the cell clickable. I guess that does seem nice for UX - cool!

website/src/components/SearchPage/Table.tsx
theosanderson, line 1: If you disable some columns you get this misalignment of checkboxes and clear icon:
image

website/src/components/SearchPage/DownloadDialog/ActiveDownloadFilters.tsx
```diff
@@ -1,33 +1,58 @@
...
+                ));
+            }
+            break;
+        }
+        case 'select': {
+            const count = downloadParameters.selectedSequences.size;
+            if (count > 0) {
+                badges = (
+                    <div className={badgeClasses}>
+                        {count.toLocaleString()} sequence{count === 1 ? '' : 's'} selected
```
theosanderson, line 42: I think the plan was to put this in the Download button? https://loculus.slack.com/archives/C05G172HL6L/p1726830336921539?thread_ts=1726828351.088969&cid=C05G172HL6L

In particular - the selection UI IMO adds confusion about how to download all sequences so adjusting the text of this button would be useful for addressing that too

website/src/components/SearchPage/DownloadDialog/generateDownloadUrl.ts
```diff
@@ -38,31 +39,48 @@ export const generateDownloadUrl = (
...
-    if (lapisSearchParameters.accession !== undefined) {
-        for (const accession of lapisSearchParameters.accession) {
-            params.append('accession', accession);
-        }
-    }
 
-    const mutationKeys = ['nucleotideMutations', 'aminoAcidMutations', 'nucleotideInsertions', 'aminoAcidInsertions'];
+    switch (downloadParameters.type) {
+        case 'filter':
+            const lsps = downloadParameters.lapisSearchParameters;
```
theosanderson, line 45: Our style is mostly not to abbreviate variable names
```suggestion
            const lapisSearchParameters = downloadParameters.lapisSearchParameters;
```

website/src/components/SearchPage/DownloadDialog/ActiveDownloadFilters.tsx
fhennig: Ah yes, sorry for not taking that into account.

But which "Download" button should have the changed text? The one in the modal, right? Or the one that opens the modal?

theosanderson: The one that opens the modal is the main important one IMO

website/src/components/SearchPage/DownloadDialog/generateDownloadUrl.ts
fhennig: https://github.com/loculus-project/loculus/pull/2917/commits/b0431edffde9765810616afeb528103876f2884f

website/src/components/SearchPage/DownloadDialog/ActiveDownloadFilters.tsx
fhennig: I'll add it. But I think I'd keep the "Active filter" bit in there, also the 'showing xx entries, y selected' or should I remove that again? I think it doesn't hurt. Especially in the open modal it's  nice to see that the selection overrides the other filters, and to see how many have been selected.

theosanderson: Keeping the filter sounds great. Yeah I think I agree it doesn't hurt (it will be easier to see in a demo whether it feels odd to see the info twice). I find the comma a bit throwing especially because the numbers can be formatted with commas. I think I would ideally see it as `23,123 sequences (3 selected)`. Potentially with an X in the `3 selected to clear`. But we can do that in a future PR if you prefer.

website/src/components/SearchPage/Table.tsx
fhennig: https://github.com/loculus-project/loculus/pull/2917/commits/6437207540e31ac7178a044974b563bb68bd5971

The button is now centred above the checkboxes consistently.

website/src/components/SearchPage/DownloadDialog/ActiveDownloadFilters.tsx
fhennig: Alright, i've kept the filter but indeed after trying it out it was odd to see the same info twice.

Button text is now in line with what we discussed in Slack!

https://github.com/loculus-project/loculus/pull/2917/commits/b1675cd578809dfb6fd1f8b3e2809e215c7d2c19

fhennig:
... working on test fixes ...

website/src/components/SearchPage/DownloadDialog/ActiveDownloadFilters.spec.tsx
theosanderson, line 1: Adding a comment to ensure we test this thoroughly. Will resolve after the weekend.

theosanderson (approved):

PR #2929, opened 2024-10-01
fix(deployment): Use recreate strategy for stand in  dbs

theosanderson:
This prevents the stand-in databases from using rolling migrations to try to avoid race conditions where the backend runs migrations on the old DB then the new DB has no migrations.
preview URL: https://recreate.loculus.org


corneliusroemer (approved):

corneliusroemer:
IIUC, this prevents rolling migrations, so the db goes down immediately and doesn't wait, right?

PR #2932, opened 2024-10-02
fix(website): fix locales for toLocaleString

theosanderson:
This will avoid hydration problems when client locale is different to server locale


corneliusroemer (approved):
Should we set the locale as a constant somewhere? Not blocking but might be nice if easy

corneliusroemer:
Added format_me to fix the test failure

corneliusroemer:
Ready to merge I think @theosanderson

theosanderson:
Nice - thanks for taking on the change. I was thinking one could also refactor this to a utility function which might be cleaner. I hereby give you permission to merge if you don't want to do that atm. :)

corneliusroemer:
@theosanderson can you check the refactor I just did?

website/src/components/SearchPage/SearchFullUI.tsx
```diff
@@ -303,14 +314,7 @@ export const InnerSearchFullUI = ({
...
                         <div className='mt-auto'>
-                            Search returned{' '}
-                            {totalSequences !== undefined
-                                ? totalSequences.toLocaleString()
-                                : oldCount !== null
-                                  ? oldCount.toLocaleString()
-                                  : initialCount.toLocaleString()}{' '}
-                            sequence
-                            {totalSequences === 1 ? '' : 's'}
+                            {buildSequenceCountText(totalSequences, oldCount, initialCount)}
```
corneliusroemer, line 317: Here I'm fixing a bug where it would show 1 sequence if old count or initial count was 1.

theosanderson:
LGTM - thanks!

Issue #2419, opened 2024-08-13
Check code formatting of the Loculus docs in CI


chaoran-chen:

chaoran-chen:
Oops, I closed this in #2554 by mistake!

PR #2958, opened 2024-10-06
docs: fix formatting

chaoran-chen:
Using prettier to fix formatting of the docs

### DIFF
--- a/docs/src/content/docs/for-administrators/my-first-loculus.md
+++ b/docs/src/content/docs/for-administrators/my-first-loculus.md
@@ -1,11 +1,9 @@
 ---
-title: "Tutorial: My first loculus"
+title: 'Tutorial: My first loculus'
 description: Experimenting with a Loculus interface running in a local mini Kubernetes cluster
 ---
 
-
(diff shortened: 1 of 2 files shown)


docs/src/content/docs/for-administrators/my-first-loculus.md
```diff
@@ -134,50 +126,48 @@ Loculus ships with some default organisms, but you probably want to overwrite th
 Let's edit the `custom_values.yaml` file to the following:
 
 ```yaml
```
theosanderson, line 128: IMO it probably makes sense to keep this consistent with our yaml files which seem to have tabs as two spaces

corneliusroemer: I made it 2 spaces to save some horizontal space as the github actions yamls can get quite nested - maybe prettier doesn't support that type of configuration in md.

corneliusroemer (commented):
We should add prettier formatting to CI if we want this to be maintained as opposed to being a one-off :)

docs/src/content/docs/for-administrators/my-first-loculus.md
chaoran-chen: ok, I added an ignore to avoid formatting this.

chaoran-chen:
@corneliusroemer, yes, I agree! I just moved #2419 to the Prio 1 milestone.

theosanderson (approved):

PR #2960, opened 2024-10-07
feat(deposition): improvements and fixes

### LINKED ISSUE #2955: Improve Taxon ID Handling for Ingestion and Deposition
Currently, the ingestion and deposition processes use identical taxon IDs. This approach can lead to some sequences not appearing in narrow searches.

Problem:

Ingestion and deposition processes use the same taxon ID.
Sequences may not show up in narrow searches due to this, e.g. NCBI virus might use a narrow taxon id for West Nile, so it misses the broad one 

Proposed Solution:

Ingest using as broad taxon IDs as possible 
Deposit using as narrow taxon IDs as possible

anna-parker:
resolves #2955

preview URL: https://add-authors-to-manifest.loculus.org

### Summary
- Deposition gets its own config file so we can submit with a different taxon id than we ingest (see #2955) and otherwise not use the same config as used for ingest
- Add authors to manifest if is_broker is true
- Update taxon_ids to be more broad in ingest and more specific in submission

### PR Checklist


ena-submission/Snakefile
```diff
@@ -28,7 +28,7 @@ if SUBMIT_TO_ENA_DEV:
     print("Submitting to ENA dev environment")
     config["ena_submission_url"] = "https://wwwdev.ebi.ac.uk/ena/submit/drop-box/submit"
     config["github_url"] = (
-        "https://raw.githubusercontent.com/pathoplexus/ena-submission/main/test/approved_ena_submission_list.json"
+        "https://raw.githubusercontent.com/pathoplexus/ena-submission/loculus_test/test/approved_ena_submission_list.json"
```
anna-parker, line 31: Revert before merging
```suggestion
        "https://raw.githubusercontent.com/pathoplexus/ena-submission/main/test/approved_ena_submission_list.json"
```

kubernetes/loculus/templates/_common-metadata.tpl
```diff
@@ -370,12 +370,13 @@ organisms:
   {{ $key }}:
     {{- with $instance.schema }}
     {{- $nucleotideSequences := .nucleotideSequences | default (list "main")}}
-    ingest: {{- $instance.ingest.configFile | toYaml | nindent 8 }}
+    enaDeposition: {{- $instance.enaDeposition.configFile | toYaml | nindent 6 }}
+      segment_names: {{- $nucleotideSequences | toYaml | nindent 8 }}
```
corneliusroemer, line 374: Are we using `segment_names` anywhere? Can't see the change anywhere else in the code, maybe let's revert for now to keep the changeset smaller and do the rename later

corneliusroemer (approved):
LGTM, save 1-2 comments, have a look at them before merging

So this PR:
- adds the is_broker variable to config
- renames `ingest` to `enaDeposition` in some comments
- adds the `AUTHORS` manifest row
- allows setting different ingest and deposition taxon ids

kubernetes/loculus/templates/_common-metadata.tpl
anna-parker: Totally we aren't using segment_names yet - would be good to refactor the code to use this though to not rely on single segmented pathogens having a segment called main - but adding this line atm has absolutely no effect on the code.

anna-parker: I removed but would be good to add in a later PR

corneliusroemer: Thanks for explaining we don't use it right now!

PR #2962, opened 2024-10-07
ix(backend): return open data use terms url in `get-released-data` when data is open

### LINKED ISSUE #2953: dataUseTermsUrl doesn't change to open when restricted sequence becomes open
This is a bug in production:
Brave Browser 2024-10-04 17 37 24

https://pathoplexus.org/seq/PP_000RGTE.1

I noticed it when checking the diff here: https://github.com/pathoplexus/ena-submission/pull/6/files#diff-803d3d887e6e5704cfd0a4da032b606d214193800a5de714f07b2a31dbe07f78R121

fengelniederhammer:
resolves #2953 

preview URL:

### PR Checklist


backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
```diff
@@ -85,31 +84,35 @@ open class ReleasedDataModel(
...
+                ("submittedDate" to TextNode(rawProcessedData.submittedAtTimestamp.toUtcDateString())),
+                ("submittedAtTimestamp" to LongNode(rawProcessedData.submittedAtTimestamp.toTimestamp())),
+                ("releasedAtTimestamp" to LongNode(rawProcessedData.releasedAtTimestamp.toTimestamp())),
+                ("releasedDate" to TextNode(rawProcessedData.releasedAtTimestamp.toUtcDateString())),
+                ("versionStatus" to TextNode(versionStatus.name)),
+                ("dataUseTerms" to TextNode(currentDataUseTerms.type.name)),
+                ("dataUseTermsRestrictedUntil" to restrictedDataUseTermsUntil),
+                ("versionComment" to TextNode(rawProcessedData.versionComment)),
+            ).let {
+                when (backendConfig.dataUseTermsUrls) {
```
corneliusroemer, line 105: This is where the real issue fix is - the rest is mostly refactoring as far as I can tell, oh and extra tests

backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
```diff
@@ -85,31 +84,35 @@ open class ReleasedDataModel(
...
+                    null -> it
+                    else -> {
+                        val url = when (currentDataUseTerms) {
+                            DataUseTerms.Open -> backendConfig.dataUseTermsUrls.open
+                            is DataUseTerms.Restricted -> backendConfig.dataUseTermsUrls.restricted
```
chaoran-chen, line 110: Very naïve question: Is this correct or do we have to write `is DataUseTerms.Open`?

corneliusroemer: GH copilot chat says:
In the provided code, the `currentDataUseTerms` variable is checked in a `when` clause to determine which URL to use for the `dataUseTermsUrl` field. The selection shows:

```kotlin
val url = when (currentDataUseTerms) {
    DataUseTerms.Open -> backendConfig.dataUseTermsUrls.open
    is DataUseTerms.Restricted -> backendConfig.dataUseTermsUrls.restricted
}
```

The use of `is` for `DataUseTerms.Restricted` and not for `DataUseTerms.Open` is correct and necessary. Here's why:

- `DataUseTerms.Open` is likely a specific object or singleton instance of the `DataUseTerms` type, which means it can be directly compared using the equality operator (`==`), or in this context, it can be used directly in the `when` clause.
- `is DataUseTerms.Restricted` is used because `DataUseTerms.Restricted` is probably a class (or a sealed subclass) that can hold additional data (e.g., a restricted until date). The `is` keyword checks if `currentDataUseTerms` is an instance of the `DataUseTerms.Restricted` class (or subclass). 

Therefore, using `is` for `Restricted` and not for `Open` ensures that the correct type (and any associated data) is handled appropriately.

backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCitationsDatabaseService.kt
```diff
@@ -314,8 +312,8 @@ class SeqSetCitationsDatabaseService(
             throw NotFoundException("SeqSet $seqSetId, version $version does not exist")
         }
 
-        val now = Clock.System.now().toLocalDateTime(TimeZone.UTC).toJavaLocalDateTime()
-        val sevenDaysAgo = LocalDateTime.parse(now.minusDays(7).toString())
+        val now = dateProvider.getCurrentInstant()
+        val sevenDaysAgo = now.minus(7, DateTimeUnit.DAY, TimeZone.UTC).toLocalDateTime(TimeZone.UTC)
```
chaoran-chen, line 316: Would it make sense to move `TimeZone.UTC` to the dateProvider as well so that it holds all related constants? i.e., call dateProvider.getTimeZone() here?

backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
fengelniederhammer: Exactly. `DataUseTerms.Open` is an `object` (and thus a singleton). `Restricted` is a class.

chaoran-chen (approved):
Looks great, thanks you!

corneliusroemer:
Test on staging works:
image

PR #2964, opened 2024-10-07
tests(website): skip download tests in webkit

### LINKED ISSUE #2963: Newly added template playwright tests are failing for webkit
```
2 failed
    [webkit] › pages/submission/template.spec.ts:6:5 › The submit page › should download the metadata file template for submission 
    [webkit] › pages/submission/template.spec.ts:17:5 › The submit page › should download the metadata file template for revision 
```

Added in this PR: https://github.com/loculus-project/loculus/pull/2916

@fengelniederhammer - can you have a look at why webkit fails the newly added tests? In PR CI we only run on chromium which is why this will have escaped us. You can make a new PR and add webkit there to debug.

Should we disable the test for now so we don't have failing CI on main?

fengelniederhammer:
resolves #2963

preview URL:

### Summary

webkit seems to ignore the content disposition header. It doesn't download the file, instead it displays it.

### PR Checklist

### DIFF
--- a/website/tests/pages/submission/template.spec.ts
+++ b/website/tests/pages/submission/template.spec.ts
@@ -3,7 +3,13 @@ import type { Download } from '@playwright/test';
 import { expect, test } from '../../e2e.fixture.ts';
 
 test.describe('The submit page', () => {
-    test('should download the metadata file template for submission', async ({ submitPage, loginAsTestUser }) => {
+    test('should download the metadata file template for submission', async ({
+        submitPage,
+        loginAsTestUser,
+        browserName,
+    }) => {
+        skipDownloadTestInWebkit(browserName);
+
         const { groupId } = await loginAsTestUser();
         await submitPage.goto(groupId);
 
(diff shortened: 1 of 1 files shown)


website/tests/pages/submission/template.spec.ts
```diff
@@ -37,4 +46,12 @@ test.describe('The submit page', () => {
             readable.on('end', () => resolve(data));
         });
     }
+
+    function skipDownloadTestInWebkit(browserName: 'chromium' | 'firefox' | 'webkit') {
+        test.skip(
+            browserName === 'webkit',
+            'webkit seems to ignore the content disposition header.\n' +
```
theosanderson, line 53:
```suggestion
            'webkit seems to ignore the content disposition header.\n' +
```
I'm not 100% sure but I think from last time it might be just playwright-webkit rather than Safari.

theosanderson (approved):

corneliusroemer:
Thanks for investigating @fengelniederhammer - is there an upstream issue in playwright for this?

fengelniederhammer:
I didn't find one that's still open. I found some for similar cases, but they are closed.

PR #2971, opened 2024-10-07
feat(prepro): Add metadata processing tests 

anna-parker:
resolves #

preview URL: https://test-prepro-v2.loculus.org/

### Summary

This PR:
- renames `process_date` to `parse_and_assert_past_date` to make function clearer
- adds prepro metadata processing tests to the CI
- changes function checks from `if date is None:` to `if not date` to handle case when input is ""
- Makes error/warning messages clearer by always adding the output_field
- Removes outdated tests
- Switch from raising a warning to raising an error if parse_timestamp cannot parse the input data (this is more in line with the behavior of all other functions, e.g.  identity, parse_and_assert_past_date all raise errors if input data cannot be parsed.

It also uses @fengelniederhammer's suggestion for structuring preprocessing tests: https://github.com/loculus-project/loculus/pull/2888#pullrequestreview-2351168606.

- adds factory methods for generating `UnprocessedEntry` and `ProcessedEntry` objects
- Creates a test structure, with `UnprocessedEntry` as `input` and `ProcessedEntry` as `expected_output` and confirms the output of `process_all` corresponds to `expected_output`.

### PR Checklist

Linked PP refactor PR: https://github.com/pathoplexus/pathoplexus/pull/225


preprocessing/nextclade/src/loculus_preprocessing/datatypes.py
```diff
@@ -115,3 +115,67 @@ class ProcessingResult:
...
+                for error in metadata_errors
+            ],
+            warnings=[
+                ProcessingAnnotation(
+                    source=[AnnotationSource(name=warning[0], type=AnnotationSourceType.METADATA)],
+                    message=warning[1],
+                )
+                for warning in metadata_warnings
+            ],
+        )
```
fengelniederhammer, line 181: This is essentially test code -> Let's move it to the `tests/` directory?

preprocessing/nextclade/tests/test.py
```diff
@@ -0,0 +1,404 @@
...
+            processed_entry = result[0]
+            if (
+                processed_entry.accession != test_case["expected_output"].accession
+                or processed_entry.version != test_case["expected_output"].version
+            ):
+                message = f"{submission_id}: processed entry does not match expected output."
+                raise AssertionError(message)
+            if processed_entry.data != test_case["expected_output"].data:
+                message = f"{submission_id}: processed metadata does not match expected output."
+                raise AssertionError(message)
```
fengelniederhammer, line 387:
```
    raise AssertionError(message)
AssertionError: invalid_float: processed metadata does not match expected output.
``` 
But it doesn't tell me what's wrong.

preprocessing/nextclade/tests/test.py
```diff
@@ -0,0 +1,404 @@
...
+                message = (
+                    f"{submission_id}: processed errors: {processed_entry.errors} does not "
+                    f"match expected output: {test_case["expected_output"].errors}."
+                )
+                raise AssertionError(message)
+            if sort_annotations(processed_entry.warnings) != sort_annotations(
+                test_case["expected_output"].warnings
+            ):
+                message = f"{submission_id}: processed warnings does not match expected output."
+                raise AssertionError(message)
```
fengelniederhammer, line 400: This also doesn't tell me what's wrong
```
    raise AssertionError(message)
AssertionError: date_no_day: processed warnings does not match expected output.
```

preprocessing/nextclade/src/loculus_preprocessing/datatypes.py
```diff
@@ -79,7 +79,7 @@ class UnprocessedAfterNextclade:
     # Derived metadata produced by Nextclade
     nextcladeMetadata: dict[SegmentName, Any] | None
     unalignedNucleotideSequences: dict[SegmentName, NucleotideSequence | None]
-    alignedNucleotideSequences: dict[SegmentName, NucleotideSequence | None]
+    alignedNucleotideSequences: dict[GeneName, NucleotideSequence | None]
```
anna-parker, line 82:
```suggestion
    alignedNucleotideSequences: dict[SegmentName, NucleotideSequence | None]
```

preprocessing/nextclade/tests/test.py
```diff
@@ -0,0 +1,404 @@
...
+
+test_config_file = "tests/test_config.yaml"
+
+test_cases = [
+    {
+        "input": UnprocessedEntryFactory.create_unprocessed_entry(
+            metadata_dict={
+                "submissionId": "missing_required_fields",
+            }
+        ),
```
fengelniederhammer, line 20: Maybe it would even be better to make it a `dict[str, TestCase]` where the keys are the test case descriptions (instead of "hiding" the description in the submissionId).
```suggestion
test_cases = {
    "missing_required_fields": {
        "input": UnprocessedEntryFactory.create_unprocessed_entry(
            metadata_dict={
                "submissionId": "submissionId",
            }
        ),
```

fengelniederhammer (approved):
A couple of minor things and maybe a few missing test cases (all the happy paths). Let's create an issue to add remaining test cases? 

Overall I like this approach!

preprocessing/nextclade/tests/test.py
fengelniederhammer, line 1: I think we should also put the `tests/' on the .dockerignore.

(Btw I usually like the inverse approach there: ignoring everything and then unignoring the things that you need. Like that new stuff doesn't end up accidentally in the Docker images, such as those tests)

preprocessing/nextclade/tests/test.py
anna-parker: done

preprocessing/nextclade/tests/test.py
anna-parker: done

anna-parker:
@fengelniederhammer I don't think happy paths are priority as we essentially test them each time we do a preview (but please feel free to add them if you are feeling like it :-). 

But you are correct tests are definitely still missing. We need to add tests for the segment / alignment processing and edge cases that can happen there. We probably also need to refactor the prepro code a bit. I created https://github.com/loculus-project/loculus/issues/2972 for this.

.github/workflows/preprocessing-tests.yaml
```diff
@@ -0,0 +1,38 @@
+name: preprocessing-tests
+on:
+  # test
```
corneliusroemer, line 3:
```suggestion
```

.github/workflows/preprocessing-tests.yaml
```diff
@@ -0,0 +1,38 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up micromamba
+        uses: mamba-org/setup-micromamba@v1
+        with: 
+            environment-file: preprocessing/nextclade/environment.yml
+            micromamba-version: 'latest'
+            init-shell: >-
+                bash
+                powershell
```
corneliusroemer, line 29:
```suggestion
```

.github/workflows/preprocessing-tests.yaml
corneliusroemer, line 1: IIUC we could theoretically use the docker container here to use the same program versions, but I guess it's worth keeping setup simpler/faster as opposed to equal by not using docker for now?

preprocessing/nextclade/README.md
```diff
@@ -24,23 +24,21 @@ This preprocessing pipeline is still a work in progress. It requests unaligned n
    mamba env create -n loculus-nextclade -f environment.yml
    ```
 
-3. Start backend (see [backend README](../backend/README.md))
-4. Submit sequences to backend
+3. Start backend (see [backend README](../backend/README.md)), run ingest script to submit sequences from INSDC.
```
corneliusroemer, line 27: Maybe easier to just use `./deploy.py` without the `--dev` arg and with ingest enabled but prepro disabled? Then it's a single command versus having to run backend separately in a terminal.

preprocessing/nextclade/README.md
```diff
@@ -24,23 +24,21 @@ This preprocessing pipeline is still a work in progress. It requests unaligned n
...
-       -H 'Content-Type: multipart/form-data'  \
-       -F 'metadataFile=@testdata/metadata.tsv;type=text/tab-separated-values' \
-       -F 'sequenceFile=@testdata/sequences.fasta'
+   mamba activate loculus-nextclade
+   pip install -e .
+   prepro
    ```
-
-5. Run pipeline
+5. Tests can be run from the same directory
```
corneliusroemer, line 36:
```suggestion
### Tests

Tests can be run from the same directory
```

preprocessing/nextclade/README.md
```diff
@@ -92,18 +90,19 @@ If no additional `preprocessing` field is specified we assume that field uses th
 However, the `preprocessing` field can be customized to take an arbitrary number of input metadata fields, perform a function on them and then output the desired metadata field. We have defined the following preprocessing functions but more can be added for your own custom instance.
 
 0. `identity`: Return the input field in the desired type.
-1. `process_date`: Take a date string and return a date field in the "%Y-%m-%d" format
-2. `parse_timestamp`: Take a timestamp e.g. 2022-11-01T00:00:00Z and return that field in the "%Y-%m-%d" format
-3. `concatenate`: Take multiple metadata fields (including the accessionVersion) and concatenate them in the order specified by the `arg.order` parameter, fields will first be processed based on their `arg.type` (the order of the types should correspond to the order of fields specified by the order argument).
-4. `process_options`: Only accept input that is in `args.options`, this check is case-insensitive. If input value is not in options return null.
+1. `parse_and_assert_past_date`: Take a date string and return a date field in the "%Y-%m-%d" format, ensure date is before release_date or today's date.
```
corneliusroemer, line 93:
```suggestion
1. `parse_and_assert_past_date`: Take a date string and return a date field in the "%Y-%m-%d" format, ensure date is before release_date or today's date. Incomplete dates `%Y` or `%Y-%m` default the unspecified part to `1`.
```

preprocessing/nextclade/README.md
```diff
@@ -92,18 +90,19 @@ If no additional `preprocessing` field is specified we assume that field uses th
...
+2. `check_date`: Take a date string and return a date field in the "%Y-%m-%d" format
```
corneliusroemer, line 94:
```suggestion
2. `check_date`: Take a date string and return a date field in the "%Y-%m-%d" format.  Incomplete dates `%Y` or `%Y-%m` default the unspecified part to `1`.
```

preprocessing/nextclade/README.md
```diff
@@ -92,18 +90,19 @@ If no additional `preprocessing` field is specified we assume that field uses th
...
+3. `parse_timestamp`: Take a timestamp e.g. 2022-11-01T00:00:00Z and return that field in the "%Y-%m-%d" format
+4. `concatenate`: Take multiple metadata fields (including the accessionVersion) and concatenate them in the order specified by the `arg.order` parameter, fields will first be processed based on their `arg.type` (the order of the types should correspond to the order of fields specified by the order argument).
+5. `process_options`: Only accept input that is in `args.options`, this check is case-insensitive. If input value is not in options return null.
```
corneliusroemer, line 97:
```suggestion
5. `process_options`: Only accept input literals that are specified in `args.options`, this check is case-insensitive. If input value is not in options return `null`.
```

preprocessing/nextclade/README.md
```diff
@@ -24,23 +24,21 @@ This preprocessing pipeline is still a work in progress. It requests unaligned n
...
+   pip install -e .
+   prepro
    ```
-
-5. Run pipeline
+5. Tests can be run from the same directory
 
    ```bash
    mamba activate loculus-nextclade
    pip install -e .
```
corneliusroemer, line 40:
```suggestion
   pip install -e .[test]
```

If pytest is an optional dependency this is how one installs it

preprocessing/nextclade/src/loculus_preprocessing/config.py
```diff
@@ -91,6 +91,8 @@ def get_config() -> Config:
     config = Config()
 
     # Overwrite config with config in config_file
+    if config_file:
```
corneliusroemer, line 94: I don't quite understand the change here - can you explain? Who passes the config_file arg? As opposed to via `args.config_file`?

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -196,23 +200,6 @@ def process_date(
         warnings = []
         errors = []
 
-        # TODO: required check is also in process_single - check if can be removed here
```
corneliusroemer, line 199: have you checked that we can remove the required logic? I assume so but a quick confirmatory comment would be great :)

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -315,15 +302,15 @@ def parse_timestamp(
             )
             return ProcessingResult(
                 datum=None,
-                warnings=[
+                errors=[
```
corneliusroemer, line 305: Is this behaviour change intentional? It's not mentioned in the PR description so it might be accidental? We should discuss when we change something from error -> warning - maybe make such logic changes in a separate PR that's more targeted and unrelated to testing.

preprocessing/specification.md
```diff
@@ -43,7 +43,7 @@ In the unprocessed NDJSON, each line contains a sequence entry represented as a
 {"accession": 2, "version": 1, "data": {"metadata": {...}, "unalignedNucleotideSequences": {...}}, "submitter": john_smith, ...}
 ```
 
-The `metadata` field contains a flat JSON object in which all values are strings. The fields and values correspond to the columns and values as provided by the submitter.
+The `metadata` field contains a flat JSON object in which all values are strings. The fields and values correspond to the columns and values as provided by the submitter. Empty metadata fields will be returned as "".
```
corneliusroemer, line 46:
```suggestion
The `metadata` field contains a flat JSON object in which all values are strings. The fields and values correspond to the columns and values as provided by the submitter. Fields present in the metadata file header but left empty by the submitter will be returned as "". Only columns present in the submission metadata file will be keys in the JSON object - independent of the metadata schema.
```

corneliusroemer (changes requested):
The tests look great! I have some questions regarding potential behaviour changes. Ideally we'd separate out tests from behaviour changes but that seems a little complicated now so happy to just focus on those for now in my reviews (I trust the tests work, and Fabian seems to have focused on them as well).

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
anna-parker: It is intentional - I should mention it in the PR description. Everywhere else we now raise an error if e.g. the date cannot be parsed or the int/float/bool is invalid - so changing this to an error is more in line with the rest of the code.

.github/workflows/preprocessing-tests.yaml
anna-parker: We set the dockerignore to exclude the tests folder, this can be changed however as we use the environment.yaml in the github actions its probably ok to not use the docker container for now

preprocessing/nextclade/README.md
corneliusroemer: Thanks!

```
(Alternatively you can run `./deploy.py --enablePreprocessing` to start the backend and preprocessing pods in one command.)
```

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
corneliusroemer: Thanks! It's not quite clear yet how this is related to tests though, we're removing some functionality here from the actual code - do you mean that you verified through a test that this code here was no longe required?

preprocessing/nextclade/tests/test.py
anna-parker: done,  created a TestCase with a name field

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
corneliusroemer: Ah good! It's also not used by human input, but by ingest input only, so warning doesn't make sense anyways! Thanks!

https://github.com/pathoplexus/pathoplexus/blob/b7994aec38b4e15c3e956d3dd8ecaae93b02b299/loculus_values/values.yaml#L95-L114

Brave Browser 2024-10-08 10 40 59

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
anna-parker: yes exactly - the code here is not required as this check is performed at another location. Additionally I added a test to confirm this code is indeed not required

.github/workflows/preprocessing-tests.yaml
```diff
@@ -0,0 +1,36 @@
...
+  cancel-in-progress: true
+jobs:
+  unitTests:
+    name: Unit Tests
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up micromamba
+        uses: mamba-org/setup-micromamba@v1
```
corneliusroemer, line 22:
```suggestion
        uses: mamba-org/setup-micromamba@v2
```

preprocessing/nextclade/src/loculus_preprocessing/config.py
anna-parker: config_file will only be passed by the test at this point, I didn't find a smart way to mock `args.config_file`. `args.config_file` is passed as a run time argument and adding this in a test where I am not running the package isn't trivial (maybe you know how to do this I couldn't find a smart way)

So I decided to just add config_file as an optional argument - this makes sure that the config used by the test is parsed in the same way it would be when the package is called via the command line, additionally when run via the command line args.config_file will overwrite config_file making sure this change can't break anything.

corneliusroemer: Thanks for the explanation! I'll quickly think about this, and if I can't figure it out add a comment so at least we know why it exists as an arg :)

preprocessing/nextclade/tests/test.py
```diff
@@ -0,0 +1,414 @@
...
+        "expected_output": ProcessedEntryFactory.create_processed_entry(
+            metadata_dict={
+                "continent": None,
```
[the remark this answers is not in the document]
anna-parker: Hmmm... I would need to pass the list of all output_fields then... thats a bigger change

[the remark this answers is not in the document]
anna-parker: done

preprocessing/nextclade/src/loculus_preprocessing/config.py
corneliusroemer: Ok, this pattern with a test only arg that's optional is not so bad - I've just added a comment and refactored this function in general to clean up a bit (not just related to your addition)

corneliusroemer: See #2973

corneliusroemer:
I'm making progress on moving from unittest to pytest - and using fixtures, by using claude - i'll make a PR in a minute

.github/workflows/preprocessing-tests.yaml
anna-parker: added a line, also btw `pip install -e .[test]` did not work for me locally so I removed it from the docs.

corneliusroemer:
Please make the PPX PR and link to it from here before merging - otherwise sure go ahead and merge!

PR #2973, opened 2024-10-08
Small tweaks to test-prepro-v2

corneliusroemer:
Preview: https://test-prepro-v2-cr.loculus.org/

Use pytest instead of unittest: results in nicer test result view:

iTerm2 2024-10-08 13 29 12

iTerm2 2024-10-08 13 29 16

Just run via `pytest`

### DIFF
--- a/preprocessing/nextclade/src/loculus_preprocessing/backend.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/backend.py
@@ -82,8 +82,7 @@ def parse_ndjson(ndjson_data: str) -> Sequence[UnprocessedEntry]:
             unalignedNucleotideSequences=json_object["data"]["unalignedNucleotideSequences"],
         )
         entry = UnprocessedEntry(
-            accessionVersion=f"{json_object['accession']}.{
-                json_object['version']}",
+            accessionVersion=json_object["accession"] + "." + json_object["version"],
             data=unprocessed_data,
         )
         entries.append(entry)
@@ -140,10 +139,9 @@ def submit_processed_sequences(
     if not response.ok:
         Path("failed_submission.json").write_text(ndjson_string, encoding="utf-8")
         msg = (
-            f"Submitting processed data failed. Status code: {
-                response.status_code}\n"
+            f"Submitting processed data failed. Status code: {response.status_code}\n"
             f"Response: {response.text}\n"
-            f"Data sent in request: {ndjson_string[0:1000]}...\n"
+            f"Data sent: {ndjson_string[:1000]}...\n"
         )
         raise RuntimeError(msg)
     logging.info("Processed data submitted successfully")
--- a/preprocessing/nextclade/src/loculus_preprocessing/backend.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/backend.py
@@ -82,7 +82,8 @@ def parse_ndjson(ndjson_data: str) -> Sequence[UnprocessedEntry]:
             unalignedNucleotideSequences=json_object["data"]["unalignedNucleotideSequences"],
         )
         entry = UnprocessedEntry(
-            accessionVersion=json_object["accession"] + "." + json_object["version"],
+            accessionVersion=f"{json_object['accession']}.{
+                json_object['version']}",
             data=unprocessed_data,
         )
         entries.append(entry)
--- a/preprocessing/nextclade/src/loculus_preprocessing/config.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/config.py
@@ -18,7 +18,7 @@ CLI_TYPES = [str, int, float, bool]
 @dataclass
 class Config:
     organism: str = "mpox"
-    backend_host: str = ""  # populated in get_config if left empty
+    backend_host: str = ""  # populated in get_config if left empty, so we can use organism
     keycloak_host: str = "http://127.0.0.1:8083"
     keycloak_user: str = "preprocessing_pipeline"
     keycloak_password: str = "preprocessing_pipeline"
@@ -37,8 +37,8 @@ class Config:
     pipeline_version: int = 1
 
 
-def load_config_from_yaml(config_file: str, config: Config) -> Config:
-    config = copy.deepcopy(config)
+def load_config_from_yaml(config_file: str, config: Config = None) -> Config:
+    config = Config() if config is None else copy.deepcopy(config)
     with open(config_file, encoding="utf-8") as file:
         yaml_config = yaml.safe_load(file)
         logging.debug(f"Loaded config from {config_file}: {yaml_config}")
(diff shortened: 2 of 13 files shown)


preprocessing/nextclade/src/loculus_preprocessing/config.py
```diff
@@ -78,26 +78,33 @@ def generate_argparse_from_dataclass(config_cls: type[Config]) -> argparse.Argum
 
 
 def get_config(config_file: str | None = None) -> Config:
-    # Config precedence: CLI args > ENV variables > config file > default
+    """
+    Config precedence: Direct function arg > CLI args > ENV variables > config file > default
```
anna-parker, line 82:
```suggestion
    Config precedence: Direct function args > CLI args > ENV variables > config file > default
```

preprocessing/nextclade/src/loculus_preprocessing/config.py
```diff
@@ -78,26 +78,33 @@ def generate_argparse_from_dataclass(config_cls: type[Config]) -> argparse.Argum
...
-    # Load default config
     config = Config()
-
-    # Overwrite config with config in config_file
-    if config_file:
-        config = load_config_from_yaml(config_file, config)
-    if args.config_file:
-        config = load_config_from_yaml(args.config_file, config)
-    if not config.backend_host:  # Check if backend_host wasn't set during initialization
+    if not args.backend_host:  # Set here so we can use organism
```
anna-parker, line 97: Why have you changed from config.backend_host to args.backend_host here? What would happen if you passed backend_host in the config_file and not in CLI arguments? I think then it would also be overwritten?

preprocessing/nextclade/src/loculus_preprocessing/config.py
```diff
@@ -109,4 +116,7 @@ def get_config(config_file: str | None = None) -> Config:
         if value is not None:
             setattr(config, key, value)
 
+    if not args.backend_host:  # Set here so we can use organism
```
anna-parker, line 119: Isn't it enough to just do this check here on config.backend_host and remove the check above?

preprocessing/nextclade/src/loculus_preprocessing/config.py
corneliusroemer: Yeah my bad

preprocessing/nextclade/src/loculus_preprocessing/config.py
```diff
@@ -37,8 +37,8 @@ class Config:
     pipeline_version: int = 1
 
 
-def load_config_from_yaml(config_file: str, config: Config) -> Config:
-    config = copy.deepcopy(config)
+def load_config_from_yaml(config_file: str, config: Config = None) -> Config:
+    config = Config() if config is None else copy.deepcopy(config)
```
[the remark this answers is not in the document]
corneliusroemer: I don't set the default arg, that's the thing, it allows initializing more succinctly, like this:

```
config = load_config_from_yaml(config_file_path) if config_file_path else Config()
```

preprocessing/nextclade/tests/factory_methods.py
```diff
@@ -12,23 +12,31 @@
...
-class TestCase:
+class ProcessingTestCase:
     name: str
     input: UnprocessedEntry
     expected_output: ProcessedEntry
 
+
 @dataclass
 class UnprocessedEntryFactory:
     _counter: int = 0
```
anna-parker, line 23: As we anyways only use this Factory for the tests below let's just get rid of the _counter? As you modified the code to now pass accessions its not really needed anymore. We also don't need reset_counter and to create a fixture for it, so will save some code :-)

preprocessing/nextclade/tests/factory_methods.py
```diff
@@ -12,23 +12,31 @@
...
 
+    @classmethod
+    def reset_counter(cls):
+        cls._counter = 0
+
     @staticmethod
     def create_unprocessed_entry(
         metadata_dict: dict[str, str],
+        accession: str | None = None,
```
anna-parker, line 32: Maybe this would actually be better named accession_id - as here we take the id add the prefix LOC_ to create an accession.

Whereas create_processed_entry actually takes the accession.

preprocessing/nextclade/tests/test.py
```diff
@@ -1,348 +1,354 @@
...
-                    ),
-                ],
+@dataclass
+class Case:
+    name: str
+    metadata: dict[str, str]
+    expected_metadata: dict[str, str]
+    expected_errors: list[tuple[str, str]]
+    expected_warnings: list[tuple[str, str]] = None
+    accession: str = "000999"
```
anna-parker, line 22: And then here we could call this accession_id to make it clearer that it doesn't yet contain the LOC_ prefix.

preprocessing/nextclade/tests/test.py
```diff
@@ -1,348 +1,354 @@
...
+    # Check warnings
+    processed_warnings = sort_annotations(processed_entry.warnings)
+    expected_warnings = sort_annotations(expected_output.warnings)
+    assert processed_warnings == expected_warnings, (
+        f"{test_name}: processed warnings {processed_warnings}"
+        f"does not match expected output {expected_warnings}."
+    )
+
+
+@pytest.mark.parametrize("test_case_def", test_case_definitions, ids=lambda tc: tc.name)
```
anna-parker, line 346: oh this is cool!

anna-parker (changes requested):
Added some more thoughts, somehow the tests aren't running for me with the pytest command - similar to the CI I see `no tests ran in 0.00s`

corneliusroemer:
Heh, pytest has some expectations for filenames, they must start or end with `test_`/`_test` - that's why it worked until I renamed `test_pytest.py` to `test.py`

preprocessing/nextclade/tests/factory_methods.py
anna-parker: I just added the changes in a commit to move this forward

preprocessing/nextclade/tests/factory_methods.py
anna-parker: added

preprocessing/nextclade/tests/test.py
anna-parker: added

anna-parker:
Wupps some issues post rebase: 
image

preprocessing/nextclade/src/loculus_preprocessing/backend.py
```diff
@@ -82,8 +82,7 @@ def parse_ndjson(ndjson_data: str) -> Sequence[UnprocessedEntry]:
             unalignedNucleotideSequences=json_object["data"]["unalignedNucleotideSequences"],
         )
         entry = UnprocessedEntry(
-            accessionVersion=f"{json_object['accession']}.{
-                json_object['version']}",
+            accessionVersion=json_object["accession"] + "." + json_object["version"],
```
anna-parker, line 85: Looks like this change is the culprit - reverting
```suggestion
            accessionVersion=f"{json_object['accession']}.{
                json_object['version']}",
```

anna-parker (approved):
preview looks great! tests are passing (and dont pass when I break them ;-) ) - LGTM!

PR #2975, opened 2024-10-08
fix(kubernetes): don't expect `$instance.enaDeposition` config when ena submission is disabled

fengelniederhammer:
preview URL: https://dontconfigureenadepositio.loculus.org/

### Summary

```
Error: template: loculus/templates/ena-submission-config.yaml:21:8: executing "loculus/templates/ena-submission-config.yaml" at <include "loculus.generateENASubmissionConfig" .>: error calling include: template: loculus/templates/_common-metadata.tpl:373:32: executing "loculus.generateENASubmissionConfig" at <$instance.enaDeposition.configFile>: nil pointer evaluating interface {}.configFile
```
But actually I have `disableEnaSubmission: true` in my config.

The `loculus-ena-submission-config` `ConfigMap` is only used (AFAICS) in the `ena-submission-deployment.yaml`, which also has the condition `{{- if not .Values.disableEnaSubmission }}`.
-> We also don't need the config map in that case.

### PR Checklist


theosanderson (approved):
Thanks!

anna-parker (approved):

PR #2980, opened 2024-10-09
fix(prepro): log and raise if prepro encounters config error to allow early detection

corneliusroemer:
preview URL: https://prepro-raise.loculus.org

### Summary
Under certain circumstances, prepro annotates sequences with entries, but doesn't log nor error - which makes it hard to detect that something is wrong for admins.

This PR makes sure we log and/or raise in those situations.

Previously, there would be nothing in the logs when a preprocessing function is misspelled in config.

Related to https://loculus.slack.com/archives/C05G172HL6L/p1728487952627419?thread_ts=1728487780.666019&cid=C05G172HL6L

Also makes a few small changes by using simple string concatenation when that is more concise than f-strings.

### Screenshot

Now it logs errors when there's a config error:

```
ERROR:root:Processing failed for PPD_000QKNP.1 with error: Processing for spec: ProcessingSpec(inputs={'date': 'sampleCollectionDate'}, function='process_date', requir │
│ ed=True, args={'type': 'date', 'submitter': 'insdc_ingest_user'}) with input data: {'date': '1987'} failed with CRITICAL: No processing function matches: process_date. │
│ This is a configuration error.
```

### DIFF
--- a/.github/workflows/preprocessing-tests.yaml
+++ b/.github/workflows/preprocessing-tests.yaml
@@ -20,17 +20,14 @@ jobs:
       - uses: actions/checkout@v4
       - name: Set up micromamba
         uses: mamba-org/setup-micromamba@v2
-        with: 
-            environment-file: preprocessing/nextclade/environment.yml
-            micromamba-version: 'latest'
-            init-shell: >-
-                bash
-            cache-environment: true
-            post-cleanup: 'all'
+        with:
+          environment-file: preprocessing/nextclade/environment.yml
+          micromamba-version: 'latest'
+          cache-environment: true
+          post-cleanup: 'all'
       - name: Run tests
         run: |
-            micromamba activate loculus-nextclade
-            pip install -e .
-            python3 tests/test.py
+          pip install -e '.[test]'
+          python3 tests/test.py
         shell: micromamba-shell {0}
         working-directory: preprocessing/nextclade/
--- a/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
@@ -113,6 +113,7 @@ class ProcessingFunctions:
     def check_date(
         input_data: InputMetadata,
         output_field: str,
+        args: FunctionArgs = None,  # args is essential - even if Pylance says it's not used
     ) -> ProcessingResult:
         """Check that date is complete YYYY-MM-DD
         If not according to format return error
@@ -164,7 +165,7 @@ class ProcessingFunctions:
     def parse_and_assert_past_date(
         input_data: InputMetadata,
         output_field,
-        args: FunctionArgs = None,
+        args: FunctionArgs = None,  # args is essential - even if Pylance says it's not used
     ) -> ProcessingResult:
         """Parse date string. If it's incomplete, add 01-01, if no year, return null and error
         input_data:
@@ -270,7 +271,7 @@ class ProcessingFunctions:
     def parse_timestamp(
         input_data: InputMetadata,
         output_field: str,
-        args: FunctionArgs = None,
+        args: FunctionArgs = None, # args is essential - even if Pylance says it's not used
     ) -> ProcessingResult:
         """Parse a timestamp string, e.g. 2022-11-01T00:00:00Z and return a YYYY-MM-DD string"""
         timestamp = input_data["timestamp"]
--- a/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
@@ -9,7 +9,6 @@ from datetime import datetime
 
 import dateutil.parser as dateutil
 import pytz
-from networkx import number_connected_components
 
 from .datatypes import (
     AnnotationSource,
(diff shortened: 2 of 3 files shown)


.github/workflows/preprocessing-tests.yaml
```diff
@@ -20,17 +20,14 @@ jobs:
...
+          environment-file: preprocessing/nextclade/environment.yml
+          micromamba-version: 'latest'
+          cache-environment: true
+          post-cleanup: 'all'
       - name: Run tests
         run: |
-            micromamba activate loculus-nextclade
-            pip install -e .
-            python3 tests/test.py
+          pip install -e '.[test]'
```
corneliusroemer, line 30: activate here made ci fail

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
corneliusroemer, line 1: I reduced the level of nesting by using the guard pattern

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -566,18 +564,18 @@ def format_frameshift(result):
     frame_shift_strings = []
     for frame_shift in frame_shifts:
         nuc_abs_list = [
-            f"{nuc["begin"] + 1}-{nuc["end"]}"
+            nuc["begin"] + 1 + "-" + nuc["end"]
```
corneliusroemer, line 567: Here the f strings made things harder to read, so I unnested

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -566,18 +564,18 @@ def format_frameshift(result):
...
             if (nuc["end"] + 1) > nuc["begin"]
-            else f"{nuc["begin"] + 1}"
+            else nuc["begin"] + 1
```
anna-parker, line 569: do we need to convert this to a string now? Both nuc["begin"] and 1 are ints - before the f-string converted this to a string

anna-parker: Ah yes I just checked on the preview and the frameshifts are missing from sequence view - most likely this function raises a warning or otherwise fails: 
image
vs 
image
on main

anna-parker (changes requested):
Looks good! One concern about removing the f-string if you could take a look! Would be cool to add tests for this later as its hard to see how this effects the frameshift field without searching the preview for sequences with total frameshifts>1

.github/workflows/preprocessing-tests.yaml
```diff
@@ -20,17 +20,14 @@ jobs:
...
+          python3 tests/test.py
```
corneliusroemer, line 31:
```suggestion
          pytest
```

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
corneliusroemer: Good catch - I wonder if mypy would catch this, I'll check

corneliusroemer: I've refactored this a little - now should be safely returning strings and accepting `int|str` for positions (depending on what Nextclade outputs)

corneliusroemer:
I (or rather Claude) have added tests for the frameshift and stop formatting and fixed the bug with a small refactor to increase clarity and DRYness

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
anna-parker: oooh! awesome thanks for the tests! I will go over them quickly now to check they make sense to me :-)

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -559,31 +557,29 @@ def format_frameshift(result):
...
+
+    def range_string(_start: str | int, _end: str | int) -> str:
+        """Converts 0-indexed exclusive range to 1-indexed inclusive range string"""
+        start = int(_start) + 1
+        end = int(_end)
+        if end > start:
+            return f"{start}-{end}"
+        return str(start)
+
+    frame_shifts = json.loads(input.replace("'", '"'))  # Why is replace needed?
```
anna-parker, line 571: This is required by json.loads to recognize input as a json object and convert to a dictionary

anna-parker, line 571:
```suggestion
    frame_shifts = json.loads(input.replace("'", '"'))  # Required for json.loads to recognize input as json string and convert to dict
```

preprocessing/nextclade/tests/test.py
```diff
@@ -343,6 +344,44 @@ def test_process_all(self) -> None:
                 )
                 raise AssertionError(message)
 
+    def test_format_frameshift(self):
```
anna-parker, line 347: This is awesome!

anna-parker (approved):
Lgtm! ~Lets maybe still wait for ingest to finish before merging (just to double check the counts are all the same before merging but everything looks great to me!)~ Looks great now - feel free to merge

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
anna-parker: They look good!

Issue #1428, opened 2024-03-21
Use `data_report.jsonl` instead of `dataformat tsv` in ingest pipeline


corneliusroemer:
The JSONL `data_report.jsonl` version is more structured and contains more info than the output of `dataformat tsv`.

For example it has authors as a list, rather than as a concatenated string without whitespace between individuals. It also has a sequence hash field that can be useful (saves us from having to compute the hash ourselves). It also contains the full taxonomic hierarchy.

corneliusroemer:
An extra benefit of doing this ourselves is that we'll avoid this type of bug in the future: https://github.com/ncbi/datasets/issues/404

theosanderson:
https://github.com/loculus-project/loculus/commit/3c20df9f31c6db55a077275b19c0694f09ea2ec1 suggests this has been completed

PR #2986, opened 2024-10-10
feat(prepro, ingest, deposition): Enforce author formatting in prepro, map authors accordingly in ingest and deposition

anna-parker:
resolves #2985, #1428

preview URL: https://format-authors.loculus.org/

##### Preprocessing changes

- From now on, we expect authors to be submitted in the new format: `Doe, John A.; Sanchez Roe, Jane Maria`, where last name(s) is mandatory, a comma is mandatory to separate first names/initials from last name.
- Only ASCII alphabetical characters A-Z are allowed in names
- Failure to do so results in the sequence becoming annotated with `error`, users will receive an error message with expected format
- Warn users if authors list might be in incorrect format.

#### Breaking: Loculus authors in released data/LAPIS

- The same format as used for submission will also be used for programmatic data output, instead of the current format: `FirstNames LastNames, FirstNames LastNames`

##### Ingest changes

- If authors list is entirely capitalized apply title case (leave common last names prefixes such as van etc. lower case)
- Add ingest test by modifying snakemake example files to also have edge cases.
- The tsv generated by the ncbi datasets cli from the raw jsonl does not correctly format authors- leading to the case where the string cannot be correctly parsed. To avoid this use the raw jsonl directly and write a python script to map json fields to a tsv This way we can format authors correctly, and have access to the raw data. 
- The following unused tsv metadata fields are dropped: 

```
ncbiHostBreed
ncbiHostCultivar
ncbiHostEcotype
ncbiHostIsolate
ncbiHostPangolin
ncbiIsVaccineStrain
ncbiMaturePeptideCount
ncbiMolType
ncbiVirusCommonName
ncbiVirusBreed
ncbiVirusCultivar
ncbiVirusEcotype
ncbiVirusIsolate
ncbi_virus (listed as "Virus Infraspecific Names Sex" in column_mapping)
ncbiVirusStrain
ncbiVirusPangolin
```

- The following currently always empty tsv fields are kept but left empty:

```
ncbiHostCommonName
ncbiPurposeOfSampling
```

To be accepted, ingest needs to resubmit all sequences

#### Problematic: Preprocessing version bump no longer possible

Old submissions won't be accepted by the new prepro pipeline, this is a problem for the current preprocessing versioning code which expects all released old sequences to be compatible with a new pipeline. There are a few options but none are obviously the best:
- A) Change author formatting inside the db sequence entries
[...]

PR #2990, opened 2024-10-11
feat(ingest): Do not use processed tsv but raw jsonl when ingesting data from NCBI Virus

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/1428

preview URL: https://change-how-ingest-ingests.loculus.org/

### Summary

Prompted by https://github.com/loculus-project/loculus/pull/2986#issuecomment-2407406053. The tsv generated by the ncbi datasets cli from the raw jsonl does not correctly format authors- leading to the case where the string cannot be correctly parsed. To avoid this I change the ingest code to use the raw jsonl directly. This way we can format authors correctly, and have access to the raw data. 

### Fields that existed in the (renamed) tsv and not in the jsonl

Of these fields only two were actually mapped to Loculus fields. The rest were all dropped by ingest - so not having them in the jsonl should not be a problem

```
ncbiHostBreed
ncbiHostCultivar
ncbiHostEcotype
ncbiHostIsolate
ncbiHostPangolin
ncbiIsVaccineStrain
ncbiMaturePeptideCount
ncbiMolType
ncbiVirusCommonName
ncbiVirusBreed
ncbiVirusCultivar
ncbiVirusEcotype
ncbiVirusIsolate
ncbi_virus (listed as "Virus Infraspecific Names Sex" in column_mapping)
ncbiVirusStrain
ncbiVirusPangolin
ncbiHostCommonName
ncbiPurposeOfSampling
```

### Missing fields that were mapped to Loculus fields

```
ncbiHostCommonName
ncbiPurposeOfSampling
```

### Potentially useful fields (that are currently dropped

```

ncbiIsVaccineStrain
```

### Testing

Checkout https://github.com/pathoplexus/pathoplexus/commit/ce68b92e8b01055a36b5fa60893876078963bf98 to perform regression testing (ignore authors as should be different) by comparing this branch with Loculus main: 

```sh
micromamba activate pp-integrity
cd tests/regression-testing 
snakemake results/ebola-zaire.meta.main.change-how-ingest-ingests.diff 
snakemake results/cchf.meta.main.change-how-ingest-ingests.diff 
snakemake results/west-nile.meta.main.change-how-ingest-ingests.diff 
```

There is no diff in the ebola-zaire, cchf and west nile metadata.

### Additional changes

Remove requirement that authors list ends with ";". However warn users if authors list might be in incorrect format.
Raise error if sequence is not in correct format or not ASCII and give users error message with correct format.
In ingest if authors list is entirely capitalized apply title case (leave common last names prefixes such as van etc. lower case)
Add ingest test by modifying snakemake example files to also have edge cases.

### PR Checklist


ingest/scripts/rename_fields.py
anna-parker, line 1: This will need to be tidied up I just manually mapped the fields of the json dict to the old tsv output using only the CCHF dataset as an example. Then I had chatgpt write me this script to manually map these fields and convert them to a tsv.

ingest/config/defaults.yaml
```diff
@@ -24,52 +24,40 @@ keep:
...
+  - "ncbiIsolateName": "name"
+  - "ncbiIsolateSource": "source"
+  - "ncbiCollectionDate": "collectionDate"
+last_virus_lineage_mappings:
+  - "ncbiVirusName": "name"
+last_host_lineage_mappings:
+  - "ncbiHostTaxId": "taxon_id"
+  - "ncbiHostName": "name"
+  - "ncbiHostSex": "sex"
+unknown_mappings: # I don't know yet where these fields come from
```
anna-parker, line 58: @corneliusroemer I am not sure where these metadata fields would be in the jsonl

corneliusroemer:
Immediate thoughts from reading PR (not looking at code yet):

- I'm confused by the long list of  "Fields that existed in the tsv and not in the jsonl" - are you sure this is correct? I don't think dataformat does much other than reformat the jsonl - but maybe I'm wrong.

anna-parker:
Yes - I am almost 100% sure that the list in "Fields that existed in the tsv and not in the jsonl" is correct - I was constantly comparing the output of my tsv files. I think datasets creates columns for data fields that exist is other datasets even if they do not exist in our own dataset (alternatively it is better at mapping the list of clade names to metadata fields).

theosanderson:
If one wanted to figure out where these come from they'd likely feature in a SC2 dataset

Issue #976, opened 2024-02-10
Allow Keycloak self-service account changes


theosanderson:
atm we are hitting this error
https://keycloak.discourse.group/t/the-account-console-presents-failed-to-initialize-keycloak-init-request-returns-403/8918

at https://authentication-main.loculus.org/realms/loculusRealm/account

theosanderson:
I believe we resolved this long ago

Issue #607, opened 2023-11-29
Implement access for service users


fengelniederhammer:
The preprocessing pipeline and the job that imports the data into SILO need access to the backend. Think of a way to grant those services access and protect the respective endpoints in the backend accordingly.

PR #2993, opened 2024-10-14
chore(kubernetes): update to LAPIS 0.3.5, SILO 0.2.23

fengelniederhammer:
preview URL: https://lapis035silo0223.loculus.org/

# Summary

Both changes are very unlikely to break something for Loculus: 
* In LAPIS we implemented additional checks on the database config. If something changes for Loculus, then LAPIS will crash on startup.
* In SILO we fixed a bug in a code path of the preprocessing that is not used by Loculus.

### Features

* **lapis:** validate that `lapisAllowsRegexSearch`-fields are of type string ([97b3ce9](https://github.com/GenSpectrum/LAPIS/commit/97b3ce912c7b28304aec823fd6c794c9d9a99525))
* **lapis:** validate that metadata fields do not contain the reserved character `.` ([97b3ce9](https://github.com/GenSpectrum/LAPIS/commit/97b3ce912c7b28304aec823fd6c794c9d9a99525)), [#976](https://github.com/GenSpectrum/LAPIS/issues/976)

### Bug Fixes

* **preprocessing:** enforce exact match on file base stem when pairing sequence and metadata when processing from file ([7c9b23d](https://github.com/GenSpectrum/LAPIS-SILO/commit/7c9b23d618a75c4455ed9c7ac4cd76cf1218251a)), [#607](https://github.com/GenSpectrum/LAPIS-SILO/issues/607)

### PR Checklist

### DIFF
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -1451,8 +1451,8 @@ insecureCookies: false
 bannerMessage: "This is a demonstration environment. It may contain non-accurate test data and should not be used for real-world applications. Data will be deleted regularly."
 additionalHeadHTML: '<script defer data-domain="loculus.org" src="https://plausible.io/js/script.js"></script>'
 images:
-  lapisSilo: "ghcr.io/genspectrum/lapis-silo:0.2.22"
-  lapis: "ghcr.io/genspectrum/lapis:0.3.4"
+  lapisSilo: "ghcr.io/genspectrum/lapis-silo:0.2.23"
+  lapis: "ghcr.io/genspectrum/lapis:0.3.5"
 secrets:
   smtp-password:
     type: raw


corneliusroemer:
This is in theory a breaking change due to disallowing `.` from metadata field names (though it probably has been a reserved character for a while, just not validated early?)

fengelniederhammer:
> This is in theory a breaking change

Yes, it is a breaking change.

This is a difficult discussion, since we're not maintaining a library. My point also was that everything breaking should be a new major version. But I think we agreed (for now) that only breaking changes for users will trigger a new major version (user as opposed to instance maintainer). So:
* Renaming a field in the API will cause a new major version, since it affects every user of the API.
* Renaming config values will not cause a new major version, since it only affects maintainers.

I'm not 100% sure whether this was a good decision. If it turns out too bad, we can still revisit it.

theosanderson (approved):
Thanks (my personal take is that we only need to consider breaking changes for known instance maintainers at this stage of our development cycle) – happy to discuss of course

corneliusroemer:
@fengelniederhammer I wasn't saying this should be a major version bump, just that it might be good to beware of the nature of the change - as this seemed the most noteworthy of the changes.

@theosanderson yes we don't need to highlight this, might just be good to note in docs, and if we had a config schema, this would be a good place to validate on loculus end.

PR #2994, opened 2024-10-14
fix(website): show more helpful error when LAPIS returns error when fetching sequences

fengelniederhammer:
preview URL: https://prettierlapissequenceerro.loculus.org/

### Summary

The message before was not helpful and not pretty either. This is at least helpful (to us), but not very pretty for users either.

### Screenshot

After: 
grafik

Before:
grafik

### PR Checklist

### DIFF
--- a/website/src/services/serviceHooks.ts
+++ b/website/src/services/serviceHooks.ts
@@ -1,9 +1,11 @@
 import { Zodios } from '@zodios/core';
 import { ZodiosHooks, type ZodiosHooksInstance } from '@zodios/react';
+import { isAxiosError } from 'axios';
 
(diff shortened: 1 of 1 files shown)


theosanderson (approved):
Thanks!

corneliusroemer:
Nice, thanks!

fengelniederhammer:
As a side note, because I didn't mention it in the PR header: If it fails to parse the LAPIS error then it will fall back to the old, ugly error. If we also want to prevent that, we can add a generic "Oops, we don't know what went wrong" message there.

theosanderson:
👍 my preference is for potentially informative error messages over totally generic ones (where there aren't security implications)

PR #2995, opened 2024-10-14
tests: Add keycloak check to e2e test setup

fhennig:
### Summary

Just a little enhancement to the playwright tests. 
I've added a function that checks if the keycloak server is actually reachable. If it isn't, the setup exits with an error code. Before, it was just silently failing and no error was raised.

### PR Checklist

### DIFF
--- a/website/tests/e2e.fixture.ts
+++ b/website/tests/e2e.fixture.ts
@@ -1,6 +1,8 @@
 import { readFileSync } from 'fs';
+import { exit } from 'process';
 
 import { type Page, test as base } from '@playwright/test';
+import axios from 'axios';
 import { ResultAsync } from 'neverthrow';
 import { Issuer } from 'openid-client';
 import winston from 'winston';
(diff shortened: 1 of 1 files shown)


website/tests/e2e.fixture.ts
```diff
@@ -91,8 +93,36 @@ export const testSequenceCount: number =
...
+        if (response.status === 200) {
+            e2eLogger.debug(`Server reachable at ${issuerUrl}.`);
+        } else {
+            e2eLogger.warn(`Server responded with status ${response.status} at ${issuerUrl}.`);
+        }
+    } catch (error) {
+        /* eslint-disable no-console */
+        e2eLogger.error(`Failed to reach server at ${issuerUrl}: ${error}`);
+        if (axios.isAxiosError(error)) {
+            console.error('Axios error message:', error.message);
```
theosanderson, line 108: Should we just use the e2elogger?

corneliusroemer: this is likely resolved now? @fhennig

website/tests/e2e.fixture.ts
```diff
@@ -91,8 +93,36 @@ export const testSequenceCount: number =
 
 const testUserTokens: Record<string, TokenCookie> = {};
 
+async function ensureKeycloakIsReachable(issuerUrl: string) {
+    try {
+        const response = await axios.get(issuerUrl);
+        if (response.status === 200) {
+            e2eLogger.debug(`Server reachable at ${issuerUrl}.`);
```
corneliusroemer, line 100: Should we have more specific error messages for all the errors here, i.e. mention keycloak? It's not obvious from the message afaict?

corneliusroemer, line 100: Maybe this will be too verbose, can you show some logs before/after? I don't know how often this function here is called, maybe I only a few times then it's fine to log this.

website/tests/e2e.fixture.ts
theosanderson: My suggestion was not to use `console.log`

theosanderson: (one reason is it can be nice to be able to search all files for `console.log` after dev work to find the ones you added)

website/tests/e2e.fixture.ts
fhennig: It is actually called quite often, whenever a user logs in I think. But it's at the 'debug' level so I'd guess it's fine? By default we always run at 'info' anyways. I would leave it, but also happy to drop it if you want to.

website/tests/e2e.fixture.ts
fhennig: yes I have removed the console log https://github.com/loculus-project/loculus/pull/2995/commits/2e17caacd9d52d7427521ab1f914624a23c857db

website/tests/e2e.fixture.ts
fhennig: I have removed it, just in case. 

https://github.com/loculus-project/loculus/pull/2995/commits/2e17caacd9d52d7427521ab1f914624a23c857db

fhennig: mentioned Keycloak now: https://github.com/loculus-project/loculus/pull/2995/commits/2e17caacd9d52d7427521ab1f914624a23c857db

theosanderson (approved):
Thanks!

PR #3003, opened 2024-10-15
feat: Allow downloading metadata in the sequence detail view

### LINKED ISSUE #2019: How to download individual metadata?
I was tying to check our field (column) names and so went to download the info for just one sequence. I went to Ebola Sudan (though don't think this matters) and clicked on the first sequence in the list, bringing it into the 'pop up view'.

However, I was surprised when I clicked the 'download' icon in the top-right and was not given any choice of _what_ to download - it just downloaded a Fasta - when what I actually want is the metadata!

I'd suggest we add possibly two links (one FASTA, one metadata) or some other kind of selector (perhaps also offering GB file that has both?) - it's odd to me one can't just download metadata for one sequence!

fhennig:
resolves #2019

preview URL: https://feat-2019-download-indivi.loculus.org/ [Edited by Theo: fyi this gets truncated so you probably need to look it up in ArgoCD - ping on Slack if you don't have the credentials to access that]

### Summary

Turn the download button into a selection dropdown instead, where you can select to either download FASTA or a Metadata TSV.

### Screenshot

image

### PR Checklist

### DIFF
--- a/website/src/components/SearchPage/SeqPreviewModal.tsx
+++ b/website/src/components/SearchPage/SeqPreviewModal.tsx
@@ -150,7 +150,7 @@ const DownloadButton: React.FC<DownloadButtonProps> = ({ seqId }: { seqId: strin
                         href={routes.sequencesFastaPage(seqId, true)}
                         className="block px-4 py-2 hover:bg-gray-100"
                     >
-                        Download FASTA
+                        FASTA
                     </a>
                 </li>
                 <li>
@@ -158,7 +158,7 @@ const DownloadButton: React.FC<DownloadButtonProps> = ({ seqId }: { seqId: strin
                         href={routes.sequencesTsvPage(seqId, true)}
                         className="block px-4 py-2 hover:bg-gray-100"
                     >
-                        Download Metadata TSV
+                        Metadata TSV
                     </a>
                 </li>
             </ul>
--- a/website/src/components/SearchPage/SeqPreviewModal.tsx
+++ b/website/src/components/SearchPage/SeqPreviewModal.tsx
@@ -102,7 +102,7 @@ export const SeqPreviewModal: React.FC<SeqPreviewModalProps> = ({
                     )}
                 </button>
                 <DownloadButton seqId={seqId} />
-                <a href={routes.sequencesDetailsPage(seqId)} title='Open in full window' className={BUTTONCLASS}>
+                <a href={routes.sequenceEntryDetailsPage(seqId)} title='Open in full window' className={BUTTONCLASS}>
                     <OouiNewWindowLtr className='w-6 h-6' />
                 </a>
                 <button type='button' className={BUTTONCLASS} onClick={onClose} title='Close'>
@@ -146,12 +146,12 @@ const DownloadButton: React.FC<DownloadButtonProps> = ({ seqId }: { seqId: strin
             </button>
             <ul className='dropdown-content z-20 menu p-1 shadow bg-base-100 rounded-btn absolute top-full -left-4'>
                 <li>
-                    <a href={routes.sequencesFastaPage(seqId, true)} className='block px-4 py-2 hover:bg-gray-100'>
+                    <a href={routes.sequenceEntryFastaPage(seqId, true)} className='block px-4 py-2 hover:bg-gray-100'>
                         FASTA
                     </a>
                 </li>
                 <li>
-                    <a href={routes.sequencesTsvPage(seqId, true)} className='block px-4 py-2 hover:bg-gray-100'>
+                    <a href={routes.sequenceEntryTsvPage(seqId, true)} className='block px-4 py-2 hover:bg-gray-100'>
                         Metadata TSV
                     </a>
                 </li>
--- a/website/src/components/SearchPage/SeqPreviewModal.tsx
+++ b/website/src/components/SearchPage/SeqPreviewModal.tsx
@@ -135,33 +135,27 @@ export const SeqPreviewModal: React.FC<SeqPreviewModalProps> = ({
 };
 
 interface DownloadButtonProps {
-    seqId: string
-};
+    seqId: string;
+}
 
 const DownloadButton: React.FC<DownloadButtonProps> = ({ seqId }: { seqId: string }) => {
     return (
-        <div className="dropdown dropdown-hover relative inline-block">
+        <div className='dropdown dropdown-hover relative inline-block'>
             <button className={BUTTONCLASS}>
-                <IcBaselineDownload className="w-6 h-6" />
+                <IcBaselineDownload className='w-6 h-6' />
             </button>
-            <ul className="dropdown-content z-20 menu p-1 shadow bg-base-100 rounded-btn absolute top-full -left-4">
+            <ul className='dropdown-content z-20 menu p-1 shadow bg-base-100 rounded-btn absolute top-full -left-4'>
                 <li>
-                    <a
-                        href={routes.sequencesFastaPage(seqId, true)}
-                        className="block px-4 py-2 hover:bg-gray-100"
-                    >
+                    <a href={routes.sequencesFastaPage(seqId, true)} className='block px-4 py-2 hover:bg-gray-100'>
                         FASTA
                     </a>
                 </li>
                 <li>
-                    <a
-                        href={routes.sequencesTsvPage(seqId, true)}
-                        className="block px-4 py-2 hover:bg-gray-100"
-                    >
+                    <a href={routes.sequencesTsvPage(seqId, true)} className='block px-4 py-2 hover:bg-gray-100'>
                         Metadata TSV
                     </a>
                 </li>
             </ul>
         </div>
     );
-}
+};
--- a/website/src/components/SearchPage/SeqPreviewModal.tsx
+++ b/website/src/components/SearchPage/SeqPreviewModal.tsx
@@ -101,9 +101,7 @@ export const SeqPreviewModal: React.FC<SeqPreviewModalProps> = ({
                         <MdiDockBottom className='w-6 h-6' />
                     )}
                 </button>
-                <a href={routes.sequencesFastaPage(seqId, true)} className={BUTTONCLASS}>
-                    <IcBaselineDownload className='w-6 h-6' />
-                </a>
+                <DownloadButton seqId={seqId} />
                 <a href={routes.sequencesDetailsPage(seqId)} title='Open in full window' className={BUTTONCLASS}>
                     <OouiNewWindowLtr className='w-6 h-6' />
                 </a>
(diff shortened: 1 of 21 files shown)


website/src/services/lapisClient.ts
```diff
@@ -53,6 +53,16 @@ export class LapisClient extends ZodiosWrapperClient<typeof lapisApi> {
...
     }
 
+    public getSequenceEntryVersionDetailsTsv(accessionVersion: string) {
+        return this.call('details', {
+            [this.schema.primaryKey]: accessionVersion,
+            dataFormat: 'TSV',
+            // This type cast isn't pretty, but if the API would be typed correctly, the union type
+            // of the actual details resonse and the potential 'string' would polute the whole API,
+            // so I decided to just do this cast here. We know that the return value is a TSV string.
+        }).then((result) => result.map((data) => data as unknown as string));
```
fhennig, line 63: I would appreciate input on this. I have fiddled with the Zodios setup quite a bit, but reached the conclusion that this would be the most sensible way to handle the TSV data. Alternatives I considered:

- Get the JSON data from the API and turn it into TSV in somewhere in the backend server (not necessarily in this function). Pros: no casting required. Cons: duplicating the TSV functionality.
- Tried to modify the Zodios API definition and adjusted the Type so it can return a string as well. Pro: Typesafety possible. Cons: the 'string' edgecase polutes all the other regular usages of the endpoint and makes the code much more convoluted in many places.

I think with Zodios the _best_ solution would be to distinguish json and TSV via endpoint, i.e. use a different endpoint alltogether like (`sample/details/tsv` - I know, not very RESTy but you get the idea). This way we could have a nicely typed TSV endpoint. But that requires a backend API change.

website/src/components/SequenceDetailsPage/SequencesDataTableTitle.astro
```diff
@@ -29,16 +29,33 @@ const { sequenceEntryHistory, accessionVersion, showFastaDownload } = Astro.prop
...
         }
         {
-            showFastaDownload && (
-                <a
-                    class='sm:outlineButton inline-block'
-                    href={routes.sequencesFastaPage(accessionVersion) + '?download'}
-                >
-                    <span class='hidden sm:inline'>Download FASTA</span>
-                    <span class='sm:hidden inline text-xl'>
+            showDownload && (
```
fhennig, line 32: Also maybe it would be nice to just reuse the react DownloadButton component here, but they handle responsiveness differently and are also styled slightly differently and for now I couldn't be bothered to "merge" the two buttons into a single component. If desired I can put some more work into this refactoring so we end up with a single component.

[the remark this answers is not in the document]
theosanderson: Ah wait - I looked into the existing code now and this isn't about the restricted terms of use. It's about revocations. Revocations are sequence entries with no data (just used to indicate an entry has been revoked) and so there is nothing to download. So having this logic does make sense.

website/src/routes/routes.ts
```diff
@@ -67,3 +65,11 @@ export const routes = {
 function withOrganism(organism: string, path: `/${string}`) {
     return `/${organism}${path}`;
 }
+
+function sequenceDownloadUrl(accessionVersion: AccessionVersion | string, fileType: FileType, download = false) {
```
theosanderson, line 69: nit: In nomenclature terms as a reader I would except this function to only be about downloading _sequences_. Whereas the TSV is actually no sequence, just metadata. I think `sequenceEntryDownloadUrl` would be one option.

website/src/services/lapisClient.ts
theosanderson: I'm not the right person to comment on this - I don't know my zodios. @fengelniederhammer might have ideas

website/src/routes/routes.ts
```diff
@@ -27,13 +28,10 @@ export const routes = {
...
     sequencesVersionsPage: (accessionVersion: AccessionVersion | string) =>
         `/seq/${getAccessionVersionString(accessionVersion)}/versions`,
-    sequencesFastaPage: (accessionVersion: AccessionVersion | string, download = false) => {
-        let url = `${routes.sequencesDetailsPage(accessionVersion)}.fa`;
-        if (download) {
-            url += '?download';
-        }
-        return url;
-    },
+    sequencesFastaPage: (accessionVersion: AccessionVersion | string, download = false) =>
```
theosanderson, line 31: I see this is in existing code but I think we shouldn't put this in the plural (sequences). I would probably change these 4 items to `sequenceEntry[..]`

website/src/components/SequenceDetailsPage/SequencesDataTableTitle.astro
```diff
@@ -29,16 +29,33 @@ const { sequenceEntryHistory, accessionVersion, showFastaDownload } = Astro.prop
...
+                <div class='sm:outlineButton inline-block dropdown dropdown-hover dropdown-end'>
+                    <span tabindex='0' class='hidden sm:inline cursor-pointer'>
+                        Download
```
theosanderson, line 35: I think styling this as a button but it acting as a dropdown isn't great in terms of [affordances](https://en.wikipedia.org/wiki/Affordance). I (genuinely) find myself not expecting things to pop up. A few options:
- have two buttons, one for each action. I actually think this could be quite good as these are really common actions and we may not want to hide them behind a popover
- maybe use the same style as the button in the modal (which is a bit less buttony) - but I think this would be my lower preference

website/src/components/SearchPage/SeqPreviewModal.tsx
```diff
@@ -135,3 +133,29 @@ export const SeqPreviewModal: React.FC<SeqPreviewModalProps> = ({
...
+const DownloadButton: React.FC<DownloadButtonProps> = ({ seqId }: { seqId: string }) => {
+    return (
+        <div className='dropdown dropdown-hover relative inline-block'>
+            <button className={BUTTONCLASS}>
+                <IcBaselineDownload className='w-6 h-6' />
+            </button>
+            <ul className='dropdown-content z-20 menu p-1 shadow bg-base-100 rounded-btn absolute top-full -left-4'>
+                <li>
+                    <a href={routes.sequencesFastaPage(seqId, true)} className='block px-4 py-2 hover:bg-gray-100'>
+                        FASTA
```
theosanderson, line 150:
```suggestion
                        Download FASTA
```

IMO button texts should indicate an action - in the case of actionny buttons (we could make things wider to fit this) 

Same for below

website/src/components/SearchPage/SeqPreviewModal.tsx
```diff
@@ -135,3 +133,29 @@ export const SeqPreviewModal: React.FC<SeqPreviewModalProps> = ({
         </Transition>
     );
 };
+
+interface DownloadButtonProps {
+    seqId: string;
+}
+
+const DownloadButton: React.FC<DownloadButtonProps> = ({ seqId }: { seqId: string }) => {
```
theosanderson, line 141: As discussed below for the other page, I think this is confusing in that some of the items in this are buttons and now we have something that looks like those other buttons but isn't a button but a popover. But I don't really have a good suggestion for what to do about that in this case - and it's not too bad as it's not as "buttony" as the other page so probably let's leave it for now

theosanderson: I had an idea on this - I think if we incorporate three vertical dots at the bottom right of the button that will help to indicate a menu will pop up

website/src/components/SequenceDetailsPage/SequencesDataTableTitle.astro
fhennig: Good feedback on this, thanks. I agree that the button looks like it shouldn't pop up. Maybe a little triangle could also make it more obvious. I don't quite understand the second option, can you clarify? For the 1st option (which I also considered) I didn't see a good way to do it on mobile, because on mobile the text turns into an icon, and then we'd need two icons? or we change the whole control on mobile entirely.

Maybe we can jump on a call to design something together before I fiddle with the code again.

website/src/pages/seq/sequenceDownload.ts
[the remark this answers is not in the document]
fhennig: Hmm, I'm not super happy with that file in general, but I kind of extracted it from the previous FASTA code. it _does_ generate an `APIRoute` which is sort of a page. We could put it in utils? But that doesn't feel like an improvement honestly.

website/src/components/SequenceDetailsPage/SequencesDataTableTitle.astro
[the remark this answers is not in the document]
fhennig: Oops that was my bad 🤦

website/src/routes/routes.ts
fhennig: thanks! I like accurate naming. fixed: https://github.com/loculus-project/loculus/pull/3003/commits/7f34c6fa04d599ccd0845b7c95880e2aced0074c

website/src/routes/routes.ts
fhennig: good point! fixed: https://github.com/loculus-project/loculus/pull/3003/commits/4650614bfa3f4665d0e9ecd5b331fc5369221b17

website/src/components/SearchPage/SeqPreviewModal.tsx
fhennig: I couldn't find a good icon in the icon library, I tried overlaying an ellipsis with CSS like ths: `<span className="custom-stroke absolute bottom-0.5 right-4 text-3xl">…</span>` but it didn't look nice, and I don't want to introduce a custom icon for this right now, so I'd say let's shelve it for now and get back to it again in the future?

website/src/pages/seq/sequenceDownload.ts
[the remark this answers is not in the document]
fhennig: moved the file: [87eea37](https://github.com/loculus-project/loculus/pull/3003/commits/87eea374f673c3ede9fe6a69a3204b85e8cd4f48)

website/src/components/SequenceDetailsPage/SequencesDataTableTitle.astro
fhennig: Next morning, I was quite fond of the idea with the little triangle, what do you think? 

image

fhennig: I've pushed the change for now! [b20b812](https://github.com/loculus-project/loculus/pull/3003/commits/b20b8123ab2e84c6e367d5f650b356342a6f1086)

website/src/components/SearchPage/SeqPreviewModal.tsx
theosanderson: Thanks for trying! I have an idea that I would like to try but I'm hitting unrelated dev problems so I'll leave that for a future PR

website/src/components/SequenceDetailsPage/SequencesDataTableTitle.astro
theosanderson: I agree - this seems nice!

website/src/utils/createDownloadAPIRoute.ts
```diff
@@ -0,0 +1,151 @@
...
+const getSequenceData = async (
+    accessionVersion: string,
+    fileSuffix: string,
+    undefinedVersionRedirectUrl: RedirectRoute,
+    getter: DataDownloader,
+) => {
+    // We don't know which organism the accessionVersion belongs to,
+    // so we just try all of them until we get a success.
+    const organisms = getConfiguredOrganisms();
+    const results = await Promise.all(
```
theosanderson, line 135: Not blocking because it's an existing issue but while reading I noticed that this `Promise.all` isn't ideal. If one LAPIS instance for one organism is unhealthy and takes 10 secs to respond it will make all organisms take 10 seconds to respond. We use better logic here: https://github.com/loculus-project/loculus/blob/a04412f5feca4607961c512968f65ab0be4ace09/website/src/pages/seq/%5BaccessionVersion%5D/findOrganismAndData.ts#L6 . If you are willing to do that while touching this code, great, if not we can make an issue

website/src/pages/seq/sequenceDownload.ts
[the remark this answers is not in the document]
theosanderson: (to explain the rationale more fully: I believe the original location would have attempted to make a page at `main.loculus.org/seq/sequenceDownload`)

theosanderson (approved):
LGTM! Great feature!

website/src/utils/createDownloadAPIRoute.ts
fhennig: Thanks for the pointer, I've made the change: [be67e74](https://github.com/loculus-project/loculus/pull/3003/commits/be67e742bbdb9079a26e8c9fa6b87a0d1e2de066)

website/src/pages/seq/sequenceDownload.ts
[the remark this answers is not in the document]
fhennig: oooohh

fhennig:
@theosanderson if you revisit this and it looks good to you, feel free to mit the 'merge' button

PR #3005, opened 2024-10-15
1072 implement request ids to trace errors

### LINKED ISSUE #1072: Implement request IDs to trace errors
We should implement something that makes tracing requests in log files easier. When there are many concurrent requests, it's impossible to see which message belongs to which request. 

I propose to implement request ids similar to what we did in LAPIS: https://github.com/GenSpectrum/LAPIS/issues/586. This should be done in the website server and in the backend. Whenever there is already a request id, it should be passed on to the next service when issuing a call (e.g. when the website server calls LAPIS).

fengelniederhammer:
resolves #1072 

preview URL: https://1072-implement-request-id.loculus.org/

### Summary

I didn't know how to implement it in the website. Having "request scoped objects" in the Astro server-side code is difficult. But it would be good to also have it there. An idea would be using a middleware to store it in `context.locals`. Then one could retrieve it via `Astro.locals`, but that works only in `.astro` files. Usually we need it in `.ts` files.

### PR Checklist

### DIFF
--- a/preprocessing/nextclade/src/loculus_preprocessing/backend.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/backend.py
@@ -103,6 +103,8 @@ def fetch_unprocessed_sequences(
     }
     logging.debug(f"Requesting data with ETag: {etag}")
     response = requests.post(url, data=params, headers=headers, timeout=10)
+    logging.info(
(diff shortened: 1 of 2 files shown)


anna-parker (approved):
I'm not a kotlin expert but I looked over this with the help of chatGPT and it looks good! I also checked the logs and there is indeed a new UID: 
which can be used for tracking logs from the request: 
image

PR #3011, opened 2024-10-15
Author formatting updates

anna-parker:
resolves #

preview URL: https://add-more-author-warnings.loculus.org/

### Summary

1. Remove requirement that authors list ends with ";". However warn users if authors list might be in incorrect format.
2. Raise error if sequence is not in correct format or not ASCII and give users error message with correct format.
3. In ingest if authors list is entirely capitalized apply title case (leave common last names prefixes such as van etc. lower case)
4. Add ingest test by modifying snakemake example files to also have edge cases. 

### PR Checklist


preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -64,8 +84,11 @@ def format_authors(authors: str) -> bool:
         first_name = " ".join(
             [f"{name}." if len(name) == 1 else name for name in first_name.split(" ")]
         )
-        loculus_authors.append(f"{last_name}, {first_name}")
-    return "; ".join(loculus_authors) + ";"
+        # Capitalize last name and first name
+        loculus_authors.append(
+            f"{convert_to_title_case(last_name)}, {convert_to_title_case(first_name)}"
```
theosanderson, line 89: I don't think we should fiddle with case except in the very specific situation that the entire authors field is in uppercase, suggesting a formatting issue. Here are some cases which would be messed up by the current logic:

`Ronald McDonald`
`The RECOVERY Consortium`

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -47,10 +47,30 @@ def invalid_value_annotation(input_datum, output_field, value_type) -> Processin
...
 
 def valid_authors(authors: str) -> bool:
-    pattern = r"^([a-zA-Z\s\.\-\']+,[a-zA-Z\s\.\-\']*;)+"
+    pattern = r"^([a-zA-Z\s\.\-\']+,[a-zA-Z\s\.\-\']*;)*([a-zA-Z\s\.\-\']+,[a-zA-Z\s\.\-\']*;*)$"
     return re.match(pattern, authors) is not None
 
 
+def warn_potentially_invalid_authors(authors: str) -> bool:
+    authors_split = re.split(r"[,\s]+", authors)
+    return bool(";" not in authors and len(authors_split) > 3)
```
theosanderson, line 56:
```suggestion
def warn_potentially_invalid_authors(authors: str) -> bool:
    if ";" in authors:
        return False
    
    if "," not in authors:
        return False
    
    parts = authors.split(",", 1)
    before_comma = re.split(r"\s+", parts[0].strip())
    after_comma = re.split(r"\s+", parts[1].strip())
    
    return len(before_comma) >= 2 and len(after_comma) >= 2
```

anna-parker: I'm not sure we need this change. If `,` is not in `authors` the string will be invalid and will be caught by the valid_authors check/function. 

I'm fine with changing this but think the function I have is a bit simpler. Potentially we warn a bit more with my function e.g. if someone has three first names or last names - but as this is anyways very rare I think its fine to do that (might also indicate an error) - especially as the warning doesn't block submission.

anna-parker: (sorry for the deleting and adding back I started a review by accident and just wanted to comment)

theosanderson: > If `,` is not in authors the string will be invalid and will be caught by the valid_authors check/function.

I would suggest we may want to remove that check because I think `the RECOVERY Consortium` and similar may be quite common author names.

I don't feel strongly about the other aspects

theosanderson: Hmm this is an interesting Q, thanks! I agree I was not understanding the requirements.

I see that for e.g. https://www.ncbi.nlm.nih.gov/nuccore/OZ013273.1 the submitters have hacked it as you describe: `Danish Covid-19 Genome,Consortium.`

theosanderson: I think I still feel like ideal behaviour for Pathoplexus might be to allow a more natural way of submitting these ,and for us to do some translation for submission to ENA. But if we're not providing a degradation over the current system this feels like something we can push to the future. Thanks for explaining.

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -65,7 +88,7 @@ def format_authors(authors: str) -> bool:
             [f"{name}." if len(name) == 1 else name for name in first_name.split(" ")]
```
anna-parker, line 88: Capitalize initials as well
```suggestion
            [f"{name.upper()}." if len(name) == 1 else name for name in first_name.split(" ")]
```

preprocessing/nextclade/tests/test.py
```diff
@@ -304,23 +353,30 @@ def sort_annotations(annotations: list[ProcessingAnnotation]):
...
+    "Xi,;Yu,X.;": "Xi, ; Yu, X.",
+    "Xi, ;Yu,X.;": "Xi, ; Yu, X.",
+    "Xi, ;Yu,X.": "Xi, ; Yu, X.",
+    "Xi,;": "Xi,",
+    "Xi,": "Xi,",
+    "Smith, Anna Maria; Perez, Jose X.;": "Smith, Anna Maria; Perez, Jose X.",
+    "Smith,Anna Maria;Perez,Jose X;": "Smith, Anna Maria; Perez, Jose X.",
+    "VAN DER MEER, F.; DU PONT, P. J.": "van der Meer, F.; du Pont, P. J.",
+    "DE LA CRUZ, A.; LE BLANC, B.; VON TRAPP, U.;": "de la Cruz, A.; le Blanc, B.; von Trapp, U.",
+    "de souza, a.": "de souza, a.",
```
anna-parker, line 367: Initials should still be capitalized
```suggestion
    "de souza, a.": "de souza, A.",
```

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -47,11 +47,34 @@ def invalid_value_annotation(input_datum, output_field, value_type) -> Processin
...
+    pattern = r"^([a-zA-Z\s\.\-\']+,[a-zA-Z\s\.\-\']*;)*([a-zA-Z\s\.\-\']+,[a-zA-Z\s\.\-\']*;*)$"
     return re.match(pattern, authors) is not None
 
 
+def warn_potentially_invalid_authors(authors: str) -> bool:
+    authors_split = re.split(r"[,\s]+", authors)
+    return bool(";" not in authors and len(authors_split) > 3)
+
+
+def convert_to_title_case(name: str) -> str:
```
anna-parker, line 59: move this to ingest and just warn users if they are submitting a name only in upper case

PR #3015, opened 2024-10-17
Add_country_metadata

anna-parker:
resolves #

preview URL:

### PR Checklist

### DIFF
--- a/ena-submission/.gitignore
+++ b/ena-submission/.gitignore
@@ -5,3 +5,4 @@ project/
 sample/
 __pycache__
 config/config.yaml
+webin-cli.jar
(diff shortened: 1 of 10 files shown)


anna-parker:
Sadly with further testing this no longer seems like a good approach: 
1. It requires a python downgrade and tedious refactor of our code
2. `augur curate` only curates exact cases in the config, e.g. `Conneticut/Hartford` is curated as `Conneticut/xxx` is not unless it is also in the config. It also does not catch cases where the state has been replaced with an abbreviation (unless that exact case is in the config) or the division and location field is not clearly split.

PR #3019, opened 2024-10-17
feat(website): add menu icon to download icon

theosanderson:
tries to resolve confusion that the download button looks like a button but is actually now a dropdown menu

http://kebabdownload.loculus.org

before:
image

after:
image

IMO this is a bit better but I don't feel super strongly.


fhennig (approved):
I like it, better than before!

PR #3020, opened 2024-10-17
fix: prevent download button layout shifts

fhennig:
preview URL:

### Summary

The download button didn't have a fixed with, and so the width dependent on the amount of width of the characters inside. This PR fixes the width to a fixed value so the layout doesn't change anymore when the number of selected sequences changes.

### Screenshot
n/A

### PR Checklist

### DIFF
--- a/website/src/components/SearchPage/DownloadDialog/DowloadDialogButton.tsx
+++ b/website/src/components/SearchPage/DownloadDialog/DowloadDialogButton.tsx
@@ -1,0 +1,36 @@
+import type { FC } from 'react';
+
+import type { DownloadParameters } from './DownloadParameters';
+import { formatNumberWithDefaultLocale } from '../../../utils/formatNumber';
+
+type DownloadDialogButtonProps = {
+    onClick: () => void;
+    downloadParams: DownloadParameters;
+};
+
+/**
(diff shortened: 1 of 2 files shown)


corneliusroemer:
Thanks! I've added the preview label to look at it - it saves the reviewer time if the preview is already up!

Also, this PR is perfect for a screencast, get something like Gifox to make it easy to record GIFs.

corneliusroemer (approved):
Good improvement!

What about the initial layout shift between "all" and "1 selected entry"? That might also be nice to get rid of.

corneliusroemer:
@theosanderson you might have views on the TS code - so just added a review request for you

fhennig:
> What about the initial layout shift between "all" and "1 selected entry"? That might also be nice to get rid of.

I think if the button has a lot of empty space in it (because the label is much shorter) it also looks a bit odd. Especially if the 'selecting sequences' feature is rarely used, the button will look oddly shaped a lot of the time, so I think this intial layout shift is fine.

Now that I think about it, maybe a different solution could be do have the button size change, but have whitespace around the button to not shift other elements around instead!

But I did this as just a filler task and because I was curious, so I won't do anything more on this for now.

Thanks for the review/merge! (Also the gif is great)

PR #3021, opened 2024-10-17
feat: Generate unique and descriptive filenames for download

### LINKED ISSUE #2327: Better file names to avoid `details (18).tsv`, e.g. `loculus_cchf_metadata_202407231323.tsv`
I downloaded some metadata and noticed that the download file name was rather undescriptive: `details.tsv`. It's so generic that it causes duplication issues, e.g. in my case I already have 17 `details.tsv`s, so it ended up as `details (18).tsv`

One approach I find useful in practice is to use timestamped URLs. Anything is better than the current situation, though.

Off the top of my head, this might work well:

```
DOMAIN_ORGANISM_DATATYPE_DATETIME.EXTENSION
```

Where DOMAIN could be something like `loculus`, or `pathoplexus`, maybe this could be configured in values.yaml.

ORGANISM is obvious.

Filetype is good to have, e.g. `metadata` or `sequences` or `aligned`. Datetime could include minutes, or even seconds, to completely avoid duplicates.

Extra things we could possibly embed in file name: 
- number of records included
- open only, vs. restricted included
- ...

I don't think we need to be very stable here, people shouldn't rely on exact filenames - this is more for convenience. We could possibly allow users to set filenames themselves if they are picky, maybe LAPIS supports that already?

fhennig:
resolves #2327

preview URL:

### Summary

Generates filenames like `Loculus_ebola-zaire_unalignedNucleotideSequences_2024-10-17-1645` for the files that are downloaded from the DownloadDialog.

### Screenshot
n/a

### PR Checklist


corneliusroemer:
The preview url doesn't work because we only put the first 25 characters in the URL, if you use shorter branch names it's easier to guess the branch name. But you can look the actual url up in argocd:
Brave Browser 2024-10-18 15 37 17

website/src/components/SearchPage/DownloadDialog/DownloadUrlGenerator.ts
```diff
@@ -0,0 +1,134 @@
...
+        return {
+            url: `${baseUrl}?${params}`,
+            baseUrl,
+            params,
+        };
+    }
+
+    private generateFilename(dataType: string): string {
+        const timestamp = new Date().toISOString().slice(0, 16).replace('T', '-').replace(':', '');
+        return `${this.websiteName}_${this.organism}_${dataType}_${timestamp}`;
```
corneliusroemer, line 112:
```suggestion
        return `${this.websiteName.to_kebap()}_${this.organism.to_kebap()}_${dataType.to_kebap()}_${timestamp}`;
```

I think it's good to only use lower case in the file names, and so turn anything that might be camelCase into kebab-case as well.

website/src/components/SearchPage/DownloadDialog/DownloadDialog.spec.tsx
```diff
@@ -91,21 +102,29 @@ describe('DownloadDialog', () => {
...
         await userEvent.click(screen.getByLabelText(/Yes, include older versions/));
         await userEvent.click(screen.getByLabelText(/Raw nucleotide sequences/));
         await userEvent.click(screen.getByLabelText(/Gzip/));
-        expect(getDownloadHref()).toBe(
-            `${defaultLapisUrl}/sample/unalignedNucleotideSequences?downloadAsFile=true&dataUseTerms=OPEN&compression=gzip&accessionVersion=SEQID1&accessionVersion=SEQID2`,
+
+        [path, query] = getDownloadHref()?.split('?') ?? [];
+        expect(path).toBe(`${defaultLapisUrl}/sample/unalignedNucleotideSequences`);
+        expect(query).toMatch(
+            /downloadAsFile=true&downloadFileBasename=Loculus_ebola_unalignedNucleotideSequences_\d\d\d\d-\d\d-\d\d-\d\d\d\d&dataUseTerms=OPEN&compression=gzip&accessionVersion=SEQID1&accessionVersion=SEQID2/,
```
corneliusroemer, line 118:
```suggestion
            /downloadAsFile=true&downloadFileBasename=loculus_ebola_unaligned-nucleotide-sequences_\d\d\d\d-\d\d-\d\d-\d\d\d\d&dataUseTerms=OPEN&compression=gzip&accessionVersion=SEQID1&accessionVersion=SEQID2/,
```

cleaner file name if there's no casing irregularity

corneliusroemer (changes requested):
Great stuff, works well, just realized that my spec wasn't entirely complete re casing irregularity - I think changing camelCase to kebab-case would be good, and probably using "sequences" instead of "unaligned-nucleotide-sequences".

fhennig:
Thanks for the feedback, I already thought that we might iterate on the exact filename, but I was happy with getting all the components in place. I agree on the casing stuff and the filename being to long, I like all your suggestions!

website/src/components/SearchPage/DownloadDialog/DownloadUrlGenerator.ts
fhennig: https://github.com/loculus-project/loculus/pull/3021/commits/8bebb799b732029c21509f405de80f66362ae5f8

website/src/components/SearchPage/DownloadDialog/DownloadDialog.spec.tsx
fhennig: tests adapted: https://github.com/loculus-project/loculus/pull/3021/commits/4e18a6718aab31b2e885246fa06de4b2ba127f8c

website/src/components/SearchPage/DownloadDialog/DownloadUrlGenerator.ts
```diff
@@ -0,0 +1,114 @@
...
+        return {
+            url: `${baseUrl}?${params}`,
+            baseUrl,
+            params,
+        };
+    }
+
+    private generateFilename(downloadDataType: DownloadDataType): string {
+        const siteName = kebabCase(this.websiteName);
+        const organism = kebabCase(this.organism);
```
theosanderson, line 109: This will already be a "technical name" (if I understand correctly) - I'm not sure we need to do anything to it

website/src/components/SearchPage/DownloadDialog/DownloadUrlGenerator.ts
```diff
@@ -0,0 +1,114 @@
...
+        const dataType = dataTypeForFilename(downloadDataType);
+        const timestamp = new Date().toISOString().slice(0, 16).replace('T', '').replace(':', '').replaceAll('-', '');
```
theosanderson, line 111: IMO we should leave some separators in the datetime or it's not obvious it's a date-time and just seems like a very long weird number

fhennig: I have now opted for this format: `2024-10-22T1430` which is the ISO format with just the colon removed (which I think is not allowed in filenames on windows).

website/src/components/SearchPage/DownloadDialog/DownloadUrlGenerator.ts
corneliusroemer: Might be nice to kebab just in case, but doesn't matter right now

website/src/components/SearchPage/DownloadDialog/DownloadUrlGenerator.ts
corneliusroemer: I probably prefer T over `_` as we use `_` as the highest level separator already and date and time belong together, and aren't separate as for example filetype and organism.

I think most people will recognize the date either way, and the time probably too after they've done a few downloads.

So I'd go for `2024-10-22T1430`

website/src/components/SearchPage/DownloadDialog/DownloadUrlGenerator.ts
fhennig: Yeah I agree so I'd keep it

website/src/components/SearchPage/DownloadDialog/DownloadUrlGenerator.ts
fhennig: agree with Cornelius here, so I'll just leave it at that for now and as Theo said, we can also do this as a follow up change if we maybe gather more feedback and it's an issue.

website/src/components/SearchPage/DownloadDialog/DownloadDataType.ts
```diff
@@ -5,18 +5,18 @@ export type DownloadDataType =
...
- * Get a shortened kebab-case datatype.
+ * Get a shortened kebab-case datatype including the gene and sequence as well.
  */
 export const dataTypeForFilename = (dataType: DownloadDataType): string => {
     switch (dataType.type) {
         case 'metadata':
             return 'metadata';
         case 'unalignedNucleotideSequences':
-            return 'nuc';
+            return dataType.segment !== undefined ? `nuc-${dataType.segment}` : 'nuc';
```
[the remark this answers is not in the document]
fhennig: Ahh that's what you meant in chat. I didn't understand at first, but I checked on ppx and it's undefined, and then I also checked the code:

```tsx
            case 1:
                downloadDataType = {
                    type: 'unalignedNucleotideSequences',
                    segment: isMultiSegmented
                        ? referenceGenomesSequenceNames.nucleotideSequences[unalignedNucleotideSequence]
                        : undefined,
                };
                break;
            case 2:
                downloadDataType = {
                    type: 'alignedNucleotideSequences',
                    segment: isMultiSegmented
                        ? referenceGenomesSequenceNames.nucleotideSequences[alignedNucleotideSequence]
                        : undefined,
                };
                break;
```


-> It's undefined if not multisegmented.

website/package-lock.json
```diff
@@ -24,6 +24,7 @@
                 "chart.js": "^4.4.6",
                 "flowbite-react": "^0.10.2",
                 "jsonwebtoken": "^9.0.2",
+                "just-kebab-case": "^4.2.0",
```
corneliusroemer, line 27: Seems to be decent library 👍

fhennig: It's pretty minimal which is good I think

website/src/components/SearchPage/DownloadDialog/DownloadDataType.ts
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Get a shortened kebab-case datatype including the gene and sequence as well.
+ */
+export const dataTypeForFilename = (dataType: DownloadDataType): string => {
+    switch (dataType.type) {
+        case 'metadata':
+            return 'metadata';
+        case 'unalignedNucleotideSequences':
+            return dataType.segment !== undefined ? `nuc-${dataType.segment}` : 'nuc';
```
corneliusroemer, line 15:
```suggestion
            // segment is undefined in case of single segmented (not e.g. main)
            return dataType.segment !== undefined ? `nuc-${dataType.segment}` : 'nuc';
```

website/src/components/SearchPage/DownloadDialog/DownloadDialog.spec.tsx
```diff
@@ -18,13 +19,14 @@ const defaultReferenceGenome: ReferenceGenomesSequenceNames = {
 };
 
 const defaultLapisUrl = 'https://lapis';
+const defaultOrganism = 'ebola';
```
corneliusroemer, line 22: Should we import `testOrganism` from vitest setup instead?

website/src/components/SearchPage/DownloadDialog/DownloadDialog.spec.tsx
```diff
@@ -66,21 +68,29 @@ describe('DownloadDialog', () => {
         });
         await checkAgreement();
 
-        expect(getDownloadHref()).toBe(
-            `${defaultLapisUrl}/sample/details?downloadAsFile=true&versionStatus=LATEST_VERSION&isRevocation=false&dataUseTerms=OPEN&dataFormat=tsv&accession=accession1&accession=accession2&field1=value1`,
+        let [path, query] = getDownloadHref()?.split('?') ?? [];
+        expect(path).toBe(`${defaultLapisUrl}/sample/details`);
+        expect(query).toMatch(
+            /downloadAsFile=true&downloadFileBasename=ebola_metadata_\d{4}-\d{2}-\d{2}T\d{4}&versionStatus=LATEST_VERSION&isRevocation=false&dataUseTerms=OPEN&dataFormat=tsv&accession=accession1&accession=accession2&field1=value1/,
```
corneliusroemer, line 74: Probably better to use `$defaultOrganism` (or $testOrganism) variable here rather than `ebola` (and in all instances below)

corneliusroemer (approved):
This is great, just some small non-blocking suggested! We can iterate on the rest as we go, nothing blocking!

fhennig:
Thanks for the review!

I didn't make these test changes as I prefer to keep unit tests with rather less than more levels of indirection and templating.

PR #3038, opened 2024-10-21
chore(ci): split gradle dependabot updates into patch and minor

corneliusroemer:
Since we now sometimes get CI failure, easier to have things less grouped

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -35,10 +35,17 @@ updates:
     schedule:
       interval: weekly
     groups:
-      minorAndPatch:
+  - package-ecosystem: gradle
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -35,10 +35,17 @@ updates:
     schedule:
       interval: weekly
     groups:
-      minorAndPatch:
+  - package-ecosystem: gradle
+    directory: backend/
+    schedule:
+      interval: weekly
+    groups:
```
corneliusroemer, line 42:
```suggestion
```

fhennig (approved):

PR #3040, opened 2024-10-23
fix(config): remove unused `secrets.database.data.host` in values.yaml

fengelniederhammer:
preview URL: https://removeunuseddbhost.loculus.org/

### Summary

AFAICS, this was introduced here:
https://github.com/loculus-project/loculus/commit/e76d3ebfcaa46030ed78d32a45529c4ff87b1941#diff-8c45270b76bf36f08f571b509614d6672984b0604844a487a0ad3ba2ad794943R1467 But then later not used anymore:
https://github.com/loculus-project/loculus/commit/43bdbf0876b7d04a20b060522874b7c31a809a50#diff-2770ff9da854df0d7d21f7b41e229f84677a4b0562825d22e84ca421f8a4e3f8R66

Currently, the Loculus docs also don't mention it.

I did not find any usage. But it would be good to have a second pair of eyes checking.

Update: This field does not exist in ppx - so no need to config changes when rolling out to ppx.

### PR Checklist

### DIFF
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -1468,7 +1468,6 @@ secrets:
       url: "jdbc:postgresql://loculus-database-service/loculus"
       username: "postgres"
       password: "password"
-      host: "loculus-database-service"
   keycloak-database:
     type: raw
     data:


fengelniederhammer:
I'm not sure whether we should label this as breaking - previous config will still work as the config value will be even more unused than now.

anna-parker:
Thanks for the catch! Preview looks good! LGTM!

anna-parker (approved):

PR #3041, opened 2024-10-23
fix: add seqset version to row key to make key unique

fhennig:
### Summary
When creating a seqset and then editing it, the same seqset ID will come up twice in the seqset table. The key property should be unique.

This PR adds the version to the key to make it unique, thus fixing the error.

### Screenshot of error msg
image

### PR Checklist


anna-parker (approved):
LGTM!

PR #3042, opened 2024-10-23
chore(dependencies): Bump website dependencies

fhennig:
resolves #

preview URL:

### PR Checklist


anna-parker (approved):
LGTM- I normally just approve/merge astro.js dependency bumps if they are green unless they are in keycloak where I am more careful

corneliusroemer:
@anna-parker The problem here is that they aren't green ;) 

Rather than manually upgrading those deps, you could exclude flowbite from the group updates: https://github.com/loculus-project/loculus/pull/3054
 
Otherwise all future updates will include that again

PR #3062, opened 2024-10-24
feat(website): redesign organism card

fhennig:
preview URL: https://feat-alternative-organism.loculus.org

### Summary
I experimented with redesigning the organism card

### Screenshot
image

### PR Checklist


fhennig:
I didn't adjust the resolution of the pictures yet, if we want this, I'd do that too.

website/src/components/IndexPage/OrganismCard.astro
```diff
@@ -16,18 +16,30 @@ const { key, image, displayName, organismStatistics, numberDaysAgoStatistics } =
...
-        {formatNumberWithDefaultLocale(organismStatistics.totalSequences)} sequences<br />
-        (+{formatNumberWithDefaultLocale(organismStatistics.recentSequences)} in last {numberDaysAgoStatistics} days)<br
-        />
-        <span class='hidden'
-            >{
-                organismStatistics.lastUpdatedAt && <>Last updated {organismStatistics.lastUpdatedAt.toRelative()}</>
-            }</span
-        >
-    </p>
+    {image !== undefined && <img src={image} class='h-32 w-full object-cover rounded-t-[3px]' alt={displayName} />}
```
anna-parker, line 21: Love the direction this is going in!

I would center the organism cards a bit more - they appear quite left-heavy on a standard computer browser, otherwise I mentioned before it would be nice maybe make the cards a big longer - primarily just increasing the size of the image - this will make e.g. ebola look less compressed. Or alternatively zooming out a bit on ebola-zaire. 

But yeah Im really liking this - think with those small changes we can merge this - its a major improvement over the current set up!

website/src/components/IndexPage/OrganismCard.astro
theosanderson, line 1: Something I've had on my mind is that it would be cool to try `text-wrap: balance` for the cards https://simonwillison.net/2024/Oct/20/you-can-use-text-wrap-balance-on-icons/

fhennig: oh yeah, interesting! But would there then be more whitespace on the right-hand side? Or larger gaps inbetween?

website/src/components/IndexPage/OrganismCard.astro
fhennig: Thanks! I just made this alternative design, what do you think?

image

- The images are bigger now (more square)
- I've moved the number bits to be left and right to balance it more

regarding the image: I (but I'm not a virologist/epidemologist) found the images to be more for illustration and didn't think anyone would see the image and be like "Ah, Ebola!" So that's why I'd lean more towards _deemphasizing_ the images. It's not like i.e. a website for cars and you look for a particular shape of car er so -- at least that's how I think? I think the names should instead be more the focus of the card. Also, the images are part of the user config, and so maybe we can't design it to fit all images anyways?

For the left-right thing: I'd have to look at it a few more times, but maybe it's nicer this way too! I didn't think it was overly left heavy before, but I can see where you're coming from!

Overall I'm happy to incorporate the changes (even if maybe not my preference) - I'd agree that either way it's an improvement over what we have now! Maybe others could/want to way in too (@theosanderson ?)

corneliusroemer: Yeah very good, make sure you replace the initial PR message screenshot as you improve things as this is the first thing people see, I'm doing that now for this one

fhennig: Yeah generally I do update, I didn't do it on purpose because I liked the older version better haha. But it's fine, thanks for updating.

theosanderson: I do prefer
image

I think we could in time adjust images or pick new ones to better fit that aspect ratio

chaoran-chen: both look great to me :)

anna-parker: I love the new format - great going @fhennig!

website/src/components/IndexPage/OrganismCard.astro
theosanderson: yeah, too much to think about for now :)

website/src/components/IndexPage/OrganismCard.astro
theosanderson: I agree with @emmahodcroft that I find the new format hard to parse.

Here's a preview of the previous version with Pathoplexus organisms, which I really like: https://org-card-alt2.loculus.org/

(Here is the same centered for interest which I think is less good: https://org-card-alt1.loculus.org/)

fhennig: We had a vote in slack and voted for the "two lines" version, I have updated the PR accordingly. Closing this as resolved.

chaoran-chen (approved):
Thanks, looks good!

(For documentation purposes: We voted for this design in [Slack](https://loculus.slack.com/archives/C05G172HL6L/p1730729381274529).)

PR #3063, opened 2024-10-24
feat(kubernetes): make public runtime config urls of the website configurable

fengelniederhammer:
preview URL: https://makewebsitepublicurlsconf.loculus.org/

### Summary

For GenSpectrum, we need to be able to set the public URLs of the services that the website accesses. Actually this is a generalized case of the "codespaceName" feature, so my approach was to pull out the `$.Values.codespaceName` parameter from the Helm chart into `deploy.py`. The website URLs can be set in values.yaml via
```
website:
  runtetimeConfig:
    public:
      backendUrl: ...
      # where %organism% is a placeholder for the organisms in the config
      lapisUrlTemplate: https://%organism%.my.lapis.org 
      keycloakUrl: ...
```

This is the first step. In a second I will need to do something similar for Keycloak: #3071.

### PR Checklist

### DIFF
--- a/kubernetes/loculus/templates/_common-metadata.tpl
+++ b/kubernetes/loculus/templates/_common-metadata.tpl
@@ -1,16 +1,3 @@
-{{- define "loculus.websiteUrl" -}}
-{{- $websiteHost := "" }}
-{{- if $.Values.codespaceName }}
-  {{- $websiteHost = printf "https://%s-3000.app.github.dev" $.Values.codespaceName }}
-{{- else if eq $.Values.environment "server" }}
-  {{- $websiteHost = printf "https://%s" $.Values.host }}
-{{- else }}
-  {{- $websiteHost = "http://localhost:3000" }}
-{{- end }}
-{{- printf $websiteHost }}
-{{- end }}
-
-
 {{/* Get common metadata fields */}}
 {{- define "loculus.commonMetadata" }}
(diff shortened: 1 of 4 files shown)


kubernetes/loculus/templates/_common-metadata.tpl
```diff
@@ -350,15 +337,29 @@ fields:
 {{- end}}
 
 {{- define "loculus.publicRuntimeConfig" }}
-            {{- if $.Values.codespaceName }}
-            "backendUrl": "https://{{ .Values.codespaceName }}-8079.app.github.dev",
+{{- $publicRuntimeConfig := (($.Values.website).runtimeConfig).public }}
```
theosanderson, line 340: Could you explain the nested parens to me?

fengelniederhammer: https://stackoverflow.com/questions/59795596/how-to-make-nested-variables-optional-in-helm

It makes everything optional. It's the equivalent of what's `?.` in Typescript (there it would be `Values?.website?.runtimeConfig?.public`).

theosanderson: Thanks!

theosanderson (approved):
Seems great! I guess some time we should think of better names for local/server but seems fine for now

PR #3065, opened 2024-10-24
feat(config, prepro): Add dummy check_authors function so that prepro v1 can still run while prepro v2 is being rolled out

anna-parker:
resolves #

preview URL: https://adddummycheckauthorsfunc.loculus.org/

### Summary

See my comment here: https://github.com/pathoplexus/pathoplexus/pull/251/files#r1814729123

This makes calls to the `check_authors` function default to the identity function - which is our current set up.

While updating from prepro pipeline v1 to v2 if anyone submits sequences they will receive an unexpected error (that the `format_authors` function does not exist), adding this will make the prepro v1 backwards compatible - e.g. we will be able to continue to run it while updating to v2 and if we need to go back to v1 we will be able to rollback without prepro breaking due to an "unknown function" warning. 

### PR Checklist

### DIFF
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -452,6 +452,10 @@ defaultOrganismConfig: &defaultOrganismConfig
         header: Authors
         truncateColumnDisplayTo: 15
         ingest: ncbiSubmitterNames
+        preprocessing:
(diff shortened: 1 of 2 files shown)


anna-parker:
We have decided against this - closing for now

PR #3075, opened 2024-10-25
feat(ci, ingest, prepro): Allow custom nextclade datasets

anna-parker:
resolves #

preview URL: https://allow-custom-datasets.loculus.org/

### Summary

This PR allows multi-segmented pathogens to use nextclade datasets from different servers and of different names in the server.

Currently we assume all datasets of a multi-segmented organism are in the same folder on the same server and that that folder has a sub-folder with the name of that respective pathogen - this might not always be the case.

This PR should have no effect on our current pathogens. 

It also allows us to not have to define nucleotide sequences in the ingest and prepro configs separately. 

I removed the preview to save on space - but it has no averse affect: 
image

### PR Checklist


fengelniederhammer (approved):
LGTM

PR #3077, opened 2024-10-25
feat(chore): Update to update to silo 0.3.0, lapis to 0.3.7, fix e2e tests

anna-parker:
resolves #

preview URL: https://fix-e2e.loculus.org/

### Summary

Removee `type: pango_lineage`, also remove `partitionBy` option as this is only a lineage-feature (why?)

### PR Checklist


corneliusroemer:
Do we use that field anywhere? It's only in dummy I think? We could also just drop it there, it doesn't serve much purpose.

fengelniederhammer:
I'm quite sure you can just drop the pango lineage field (or make it a regular field - that's maybe even the better option)

anna-parker:
I tried dropping the type first - but it seems that SILO sees the name as meaning a pango_lineage - so I got the same error. I didn't want to remove the entire field as I think this will mean having to change all e2e tests.

fengelniederhammer:
No, there is no meaning attached to field names in SILO. The SILO database config should look like:
```
    - name: pango_lineage
      type: string
``` 

(`generateIndex: true` is optional).
I'm very sure that this works (from a technical perspective).

fengelniederhammer:
I just read the error message careful enough: 
`partitionBy 'pangoLineage' must be of type STRING and needs 'generateLineageIndex' set`

Simply remove `partitionBy` in that config. That fixes it.

fengelniederhammer:
> @fengelniederhammer what does partitionBy do actually? - it looks like it just splits up the table - so why does it have to be a lineage now?

https://lapis.cov-spectrum.org/open/v2/docs/maintainer-docs/references/database-configuration

It also had to be a lineage before. (Before: `type: pango_lineage`, now: `generateLineageIndex: true`).
It divides the data into partitions.

This provides optimizations:
* It speeds up the queries, because we can process partitions in parallel. 
* We basically sort the data by lineage. Thus similar data is closer together. Thus similar mutations are closer together. That improves compression, because Roaring bitmaps compress better if the they are almost full or almost empty.

kubernetes/loculus/values.yaml
```diff
@@ -1451,7 +1450,7 @@ insecureCookies: false
 bannerMessage: "This is a demonstration environment. It may contain non-accurate test data and should not be used for real-world applications. Data will be deleted regularly."
 additionalHeadHTML: '<script defer data-domain="loculus.org" src="https://plausible.io/js/script.js"></script>'
 images:
-  lapisSilo: "ghcr.io/genspectrum/lapis-silo:0.2.23"
+  lapisSilo: "ghcr.io/genspectrum/lapis-silo:0.3.0"
   lapis: "ghcr.io/genspectrum/lapis:0.3.5"
```
fengelniederhammer, line 1454: You should also update LAPIS (check the README there for which version you need for SILO 0.3 - I think it's 0.3.7)

corneliusroemer:
Progress!

image

corneliusroemer:
image

corneliusroemer:
Ah the problem is that we pass some accept headers and they are used to change the default:

image

The default (if nothing is passed in accept header) is actually still FASTA

corneliusroemer:
Nice, now it works!

image

corneliusroemer (approved):

corneliusroemer (approved):

Issue #3079, opened 2024-10-27
Pagefind for search


corneliusroemer:
Docs and governance pages are hard to find (there are more than 30 or so in Pathoplexus).

We could use [Pagefind](https://pagefind.app/) to implement a site-wide search, as discussed in https://github.com/pathoplexus/pathoplexus/issues/244

I looked into this briefly:
- We can't use the astro-pagefind plugin because we use server side rendering.
- Instead, we can scrape pages with, for instance, `wget` to create a folder tree with render html pages. E.g.:
  ```
  npm run start-server &
  wget --mirror \
       --convert-links \
       --adjust-extension \
       --no-parent \
       --reject "*.js,*.css,*.png,*.jpg,*.jpeg,*.gif,*.ico,*.svg,*.woff,*.woff2,*.ttf" \
       --exclude-directories="/seq" \
       --directory-prefix=./static \
       localhost:3000
  ```
- We then run `npx pagefind --site static/localhost:3000`
- And move the resulting output into the public folder: `mv static/localhost:3000/pagefind public/pagefind`
- We can then add search in the header component of every page by:
  - including this in the html header:
    ```html
    <link href='/pagefind/pagefind-ui.css' rel='stylesheet' />
    <script src='/pagefind/pagefind-ui.js' is:inline></script>
    ```
  - Adding a search div somewhere in the header nav html:
    ```html
    <div id='search' class='px-4'></div>
    ```
  - Initializing the search somewhere:
    ```html
    <script is:inline>
        window.addEventListener('DOMContentLoaded', (event) => {
            new PagefindUI({
                element: '#search',
                showSubResults: true,
            });
        });
    </script>
    ```
- Here's an (ugly) proof-of-concept:
![2024-10-27 02 11 18](https://github.com/user-attachments/assets/0437168a-6b69-4f3c-916c-098fdb6b9e7e)

corneliusroemer:
We could also use Algolia, might be slightly more feature packed, has free plan and if we need more we should qualify for open source: https://www.algolia.com/for-open-source/

theosanderson:
I don't think we should incorporate something into Loculus that requires filling out a form to apply(?)

PR #3081, opened 2024-10-27
wip: Demo of using pagefind search

corneliusroemer:
resolves #

preview URL:

### PR Checklist


website/src/components/Navigation/Navigation.astro
theosanderson, line 1: This is cool! 

I think seeing a search box at the top of a database I might have an initial expectation that it would search _the database_ rather than (just) the docs. I guess one thing we could do would be to detect accession-like search entries and make a special link for them.

corneliusroemer: I was mostly thinking about docs search as the use case here, e.g. locating the minutes - Pagefind might not scale to hundreds of thousands of pages.  But indeed, we could intercept searches for things like `PP_` or similar and don't give them to Pagefind but use our own search engine of kinds.

theosanderson: Yeah basically I think a docs-only search would ideally have a different presentation to make clear that that's what it is.

image

would confuse me if it searched docs :)

theosanderson: It would be cool to include the top level organism pages like https://search.loculus.org/cchf/search? too (I know this is just a demo :) )

corneliusroemer: Ah yeah - it's easy to add extra pages to the index. I created the index as explained here: https://github.com/loculus-project/loculus/issues/3079#issue-2616226087

I'm not sure why the organism search pages weren't indexed, maybe because they are behind a redirect and wget didn't get there as a result. I was just hacking things together, didn't spend much time on any part of it.

PR #3083, opened 2024-10-28
chore: bump msw; switch to happy-dom for testing

### LINKED ISSUE #3043: Investigate MSW update issue
Previously @corneliusroemer reported an issue that we run into when bumping msw: https://github.com/mswjs/msw/issues/2273

The issue was unrelated to msw, but we still have a problem when bumping it. Needs more investigation.

This only affects the tests, so no real prod issues.

The issue appears in msw 2.4.6

fhennig:
resolves #3043

### Summary

Bump to latest MSW version. Switch from `jsdom` to `happy-dom`.

We encountered a problem with the tests. An MSW maintainer helpfully investigated, and suggested to switch to `happy-dom` for the tests (https://github.com/mswjs/msw/issues/2273#issuecomment-2351508014) . I am doing this in this PR, and the tests work.

`happy-dom` is also faster than `jsdom` and we managed to get rid of a few workarounds that were needed with `jsdom`!

### PR Checklist


fhennig:
oops, sorry for merging main and not rebasing

website/src/components/SearchPage/SearchFullUI.spec.tsx
```diff
@@ -234,7 +234,10 @@ describe('SearchFullUI', () => {
         await userEvent.click(field1Checkbox);
         const closeButton = await screen.findByRole('button', { name: 'Close' });
         await userEvent.click(closeButton);
-        await waitForElementToBeRemoved(() => screen.queryByText('Toggle the visibility of search fields'));
+        const toggleText = screen.queryByText('Toggle the visibility of search fields');
+        if (toggleText) {
```
corneliusroemer, line 238: Under which conditions would we not expect the toggleText to be there? Why does it need to be guarded?

website/package.json
```diff
@@ -84,9 +84,10 @@
         "eslint-plugin-prefer-arrow": "^1.2.3",
         "eslint-plugin-react": "^7.35.2",
         "eslint-plugin-react-hooks": "^4.6.2",
+        "happy-dom": "^15.7.4",
```
fengelniederhammer, line 87: What's the benefit of using happy-dom?

website/package.json
```diff
@@ -84,9 +84,10 @@
...
         "jsdom": "^25.0.1",
```
fengelniederhammer, line 88: Could we get rid of the jsdom dependency then, if we decide to use happy-dom?

website/src/components/SearchPage/SearchFullUI.spec.tsx
fhennig: Hmm. The thing didn't work anymore after switching to happy-dom, I'm thinking it might be a timing issue. Given that this is only done to make sure the pop-up is gone (if I understand correctly) I thought it would be fine to do it this way. I didn't investigate further into the reasons

website/package.json
fhennig: It provides a different `File` implementation that works with the newer MSW version. This was suggested upstream here: https://github.com/mswjs/msw/issues/2273#issuecomment-2351508014

website/package.json
fhennig: I thought it might be used throughout in other places, but yeah, might be worth checking 👍 I'll do that.

website/package.json
corneliusroemer: MSW doesn't work with jsdom, see issue

website/package.json
corneliusroemer: I was about to write that @fengelniederhammer, then I got interrupted 😄

fhennig:
I removed `jsdom` and also 2 workarounds are not needed anymore, and I noticed that `happy-dom` is way faster at test execution! 🥳

website/src/components/SearchPage/SearchFullUI.spec.tsx
fhennig: no the await is still there as it is in front of `waitForElementToBeRemoved`. If you modify the strings you can actually check that the test still works.

The new code simply first checks if the element is there before waiting for it to be removed.

fhennig: > The element(s) given to waitForElementToBeRemoved are already removed. waitForElementToBeRemoved requires that the element(s) exist(s) before waiting for removal.

chaoran-chen (approved):
Seems good to me, thanks!

PR #3090, opened 2024-10-28
fix(website): Increase website memory limit to 1024 mb to avoid crashes with very large numbers of deletions

theosanderson:
Increase the memory availability to the website to avoid crashes on sequences with very large numbers of deletions

### DIFF
--- a/website/Dockerfile
+++ b/website/Dockerfile
@@ -24,4 +24,4 @@ EXPOSE 3000
 VOLUME /config
 VOLUME /log
 
-CMD node ./dist/server/entry.mjs
+CMD node --max-old-space-size=1024 ./dist/server/entry.mjs


fhennig:
might be nice to add a little comment as to why this was done

website/Dockerfile
```diff
@@ -24,4 +24,4 @@ EXPOSE 3000
 VOLUME /config
 VOLUME /log
 
-CMD node ./dist/server/entry.mjs
+CMD node --max-old-space-size=1024 ./dist/server/entry.mjs
```
corneliusroemer, line 27: We may want the kubernetes limit to never be smaller than this, ideally we derive this here from the kubernetes limit (passing in a runtime through env variable)

corneliusroemer (approved):

PR #3111, opened 2024-10-29
feat(website): add revise form

theosanderson:
resolves #

preview URL:

### Summary

Not tested atm

### DIFF
--- a/website/src/components/Edit/EditPage.tsx
+++ b/website/src/components/Edit/EditPage.tsx
@@ -6,8 +6,13 @@ import type { Row } from './InputField.tsx';
 import { getClientLogger } from '../../clientLogger.ts';
 import { routes } from '../../routes/routes.ts';
 import { backendClientHooks } from '../../services/serviceHooks.ts';
-import { ACCESSION_FIELD } from '../../settings.ts';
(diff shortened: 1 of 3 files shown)


anna-parker:
@theosanderson this looks amazing - yesterday I tested a bit but the preview went down. I am documenting my testing here now: 

Testing comments: 
- LOC_000PD32.1: revised metadata 
- LOC_000PD24.1: revised sequence
- LOC_000PCWG.1: revised metadata and sequence

anna-parker (approved):
Hi Theo! This looks really great! Also the tests I did look good I was wondering though if you could still verify my understanding a bit.
 So here when you call the editPage when clicking the revise button - am I correct in understanding this is sending a `get-data-to-edit` request to the backend? (e.g. here: https://github.com/loculus-project/loculus/blob/1b7ce069b9b51e8e23d5ec000aa96365318eeab5/website/src/pages/%5Borganism%5D/submission/edit/%5Baccession%5D/%5Bversion%5D.astro#L17C49-L17C62 - which gets the processed and original data from the backend for that sequence) - but we only display the original metadata that is specified in EditableOriginalDataProps.inputFields which comes from the config.

theosanderson:
That sounds about right! (Mostly the behaviour is just like the old edit page, just that it submits differently for the revision case and displays a different header!)

PR #3112, opened 2024-10-29
feat(ingest, backend): Modify ingest to handle curation of ingested sequences

anna-parker:
resolves #

preview URL:

### PR Checklist

### DIFF
--- a/ingest/scripts/compare_hashes.py
+++ b/ingest/scripts/compare_hashes.py
@@ -6,6 +6,7 @@ from dataclasses import dataclass
 from hashlib import md5
 
 import click
+import requests
 import yaml
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


ingest/scripts/compare_hashes.py
```diff
@@ -22,13 +23,177 @@ class Config:
...
+
+def notify(config: Config, text: str):
+    """Send slack notification with blocked revision details"""
+    if config.slack_hook:
+        requests.post(config.slack_hook, data=json.dumps({"text": text}), timeout=10)
+    logger.warn(text)
+
+
+def get_curated_fields(curated_data, pre_curated_data, current_state):
+    curated_fields = {}
```
anna-parker, line 73: we actually need not only the {field: curated_value} map but we should actually have {field: {to: value, from: value}}, then when checking for the re-ingested sequence where we want to change field x from y to z, we can check if the curation is from z to y - in which case we do not revise.

ingest/scripts/compare_hashes.py
```diff
@@ -22,13 +23,177 @@ class Config:
...
+        if submitter != "insdc_ingest_user":
+            curated_data = sorted_versions[index]["original_metadata"]
+            pre_curated_data = sorted_versions[index - 1]["original_metadata"]
+            current_state = pre_curated_data = sorted_versions[-1]["original_metadata"]
+            curated_metadata_fields.update(
+                get_curated_fields(curated_data, pre_curated_data, current_state)
+            )
+    new_changed_metadata_fields = get_changed_fields(
+        record, sorted_versions[-1]["original_metadata"]
+    )
```
anna-parker, line 124: check if for all overlapping fields curation is from-to the opposite of what the current change would be - then keep curated version

ingest/scripts/compare_hashes.py
```diff
@@ -22,13 +23,177 @@ class Config:
...
+    if new_changed_metadata_fields.isdisjoint(curated_metadata_fields.keys()):
+        logger.info("Sequence has been curated before but in different fields - can be revised")
+        update_manager.revise[record["insdcAccessionBase"]] = Revision(
```
anna-parker, line 127: reminder: use hash of pre-curated revised sequence here to avoid getting into a loop, it is possible that the hash will differ but there will be "no change" = INSDC revised sequence in the same way it was curated - submit revision with only hash update to avoid getting into a loop

ingest/scripts/compare_hashes.py
```diff
@@ -22,13 +23,177 @@ class Config:
...
+    update_manager: SequenceUpdateManager,
+):
+    submitter_list = [sorted_version["submitter"] for sorted_version in sorted_versions]
+    curated_metadata_fields = {}
+    for index, submitter in enumerate(submitter_list):
+        if submitter != "insdc_ingest_user":
+            curated_data = sorted_versions[index]["original_metadata"]
+            pre_curated_data = sorted_versions[index - 1]["original_metadata"]
+            current_state = pre_curated_data = sorted_versions[-1]["original_metadata"]
+            curated_metadata_fields.update(
```
anna-parker, line 119: reminder: also check if sequence has been curated - either add a sequence_hash for all entries (this will require db surgery to add to all existing entries) or need to get this information with get-original-metadata (lot of unnecessary sequence versions download from backend) or for every sequence ingest wants to revise call loculus (also a lot of requests)

PR #3120, opened 2024-10-29
chore(website): Replace ManagedErrorFeedback with react-toastify

### LINKED ISSUE #935: Replace ManagedErrorFeedback
Make Error Feedback coherent by replacing the managedErrorfeedback with a red error box we use at other places. 
 
              I'm still not a fan of this. Are we sure that we want to keep using it?

_Originally posted by @fengelniederhammer in https://github.com/loculus-project/loculus/pull/908#discussion_r1481553905_

theosanderson:
Fixes #935

https://theosanderson-replace-err.loculus.org/

Replace `ManagedErrorFeedback` with `react-toastify` for error feedback. (We were already using `react-toastify` in some places).

This means we don't have to store lots of state ourselves.

For now, for consistency with prior behaviour I have disabled `autoClose` for everything. We can probably let some of these autoclose in the future.

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/loculus-project/loculus/issues/935?shareId=1845be21-bfbb-4ed3-bfe8-fcb8b8d99c5c).


fhennig:
How can I provoke an error so I can see it? I think a screenshot would be cool for this. But looks good in principle and also good idea in general!

theosanderson:
Doing anything wrong in the submission process is a good approach

image

corneliusroemer:
This screenshot could be used straight in a new issue for improved punctuation 😀

theosanderson:
Ah yeah, I always read it with the "to agree" that isn't there - that would be the good fix IMO

corneliusroemer (approved):
Looks good, what's the main reason behind replacement? Just curious.

Also, maybe silly comment, feel free to ignore, but maybe you could factor out that repeated lambda into a const function export? Or do you think they'll start a life of their own from now on and diverge? It used to be like this initially, single function.

corneliusroemer:
Copilot not smart enough to do that for me, or maybe not possible, bit slow as well, nice idea but not quite there yet I feel 🙈

website/src/components/User/ListOfGroupsOfUser.tsx
```diff
@@ -15,9 +15,7 @@ const InnerListOfGroupsOfUser: FC<ListOfGroupsOfUserProps> = ({ groupsOfUser })
 
     return (
         <>
-            {errorMessage !== undefined && (
-                <ErrorFeedback message={errorMessage} onClose={() => setErrorMessage(undefined)} />
```
theosanderson, line 19: revert this - as its not managederror feedback

theosanderson:
Yes I think in the future we might want some of these to automatically disappear so a single function might not be ideal. It would be easy to do later anyway if we wanted.

fhennig:
thanks for the picture! cornelius already approved, I have nothing to add.

Issue #3087, opened 2024-10-28
Ingest should not auto-approve submissions that weren't submitted by "insdc_ingest_user"


corneliusroemer:
Relevant for curation of insdc data: https://loculus.slack.com/archives/C05G172HL6L/p1730118798464579?thread_ts=1730118697.621189&cid=C05G172HL6L

chaoran-chen:
Duplicate of #2613 ;) I'll close the old one as this one is more actively used and linked from the current PR.

Issue #3085, opened 2024-10-28
How to maintain curation changes across ingest revisions


corneliusroemer:
If we curate Genbank sequences, we have to deal with the following problem:

- Ingest gets sequence GC1234.1 from genbank, we call it PP_AAAA.1
- Curator curates it, it becomes PP_AAAA.2
- New GC1234.2 becomes available on genbank: ingest submits it as PP_AAAA.3, curation change is lost

One solution is to have a patch file that specifies a sequence (not version) and the curation changes applied to it.

When ingest revises a curated sequence, a curation bot could then reapply the patch file (potentially after curator approval).

anna-parker:
I've started working on handling this programmatically in ingest: https://github.com/loculus-project/loculus/pull/3112. 

I do this by having ingest calculate the diff between versions and keep the diff that is from a curator. However, this is quite inefficient and I realized I also need to do the same to check if the sequence has changed.

Issue #3084, opened 2024-10-28
Ingest should not treat curator revisions as latest and revise


corneliusroemer:
Given curators might change insdc_ingest_user submitted sequences (something that wasn't considered when writing ingest pipeline) we need to avoid the following:

- Ingest gets sequence GC1234.1 from genbank, we call it PP_AAAA.1
- Curator curates it, it becomes PP_AAAA.2
- Ingest sees: oh no, the latest version I got from genbank is different than the latest one on PP, let’s revise: PP_AAAA.3 (with the original)

This may not actually happen with current code, it's just a possibility as it wasn't considered during development. Easy solution would be to filter out non-ingest-user-submitted sequences.

anna-parker:
This is where the comparison is done: https://github.com/loculus-project/loculus/blob/main/ingest/scripts/compare_hashes.py#L153 - so we only compare the hash of the current metadata on NCBI to the hash of the latest version on ingest. So in this case the curator revising and the version being bumped should not cause ingest to update to push out another version unless the curator additionally updated the hash themselves. However, https://github.com/loculus-project/loculus/issues/3085 is a real risk

corneliusroemer:
> So in this case the curator revising and the version being bumped should not cause ingest to update/push out another version u

Unfortunately it does happen: https://main.loculus.org/seq/LOC_00037FR.3

I think your potentially wrong assumption is that `submitted` contains only the insdc_user submitted sequences, have you checked that? I wanted to do that but decided to do a quick experiment first and it turned out to show we are vulnerable to this right now.

anna-parker:
No... my assumption is that submitted should contain a map from insdc accessions to the corresponding loculus version and accession, which is what the general code description is: [get_submitted(config: Config)](https://github.com/loculus-project/loculus/blob/main/ingest/scripts/call_loculus.py#L282C5-L282C34) - so probably there is a bug there

corneliusroemer:
Revisions from superuser just also seem to count in here, doesn't really matter whether it's a bug or not, we assumed that no one else would submit other than insdc_ingest_user and now this requirement is changing.

anna-parker:
I probably am missing sth. I don't understand how revisions from superuser being in the `submitted` list are an issue - in the hash comparison step ingest should only revise if the hash has changed. So if there is a new revision unless the hash changes (compared to the hash calculated on the data in INSDC) I didnt think ingest would revise. But it seems I am missing sth

corneliusroemer:
[...]

Issue #2613, opened 2024-08-27
Revisions from Curators to INSDC-Ingest-Group Are Auto-approved


emmahodcroft:
Ideally, when curating sequences from INSDC, we would like 1 curator to set up the revision, and another (different) to approve it and 'submit' it to become live.

However, at the moment, all sequences in 'Review' are auto-approved for INSDC-ingest-group. So, one reviewer sets this up, and it's auto-approved without having a chance for another reviewer to approve.

One possible fix would be to only auto-approve 'version 1' (`.1`) sequences, and anything higher than this (e.g. a revision) needs to be approved by a curator. 

This would be fairly high-priority post-launch as we have our 'four eyes' principle written into our docs (and it's probably a good idea)

(As a follow up - re-edit docs after fixing this issue
- Curator SOP
- Data submission & processing)

chaoran-chen:
Closing this in favor of the duplicated issue #3087.

PR #3124, opened 2024-10-30
feat(ingest, backend): do not approve curated revisions (by users other than insdc_ingest_user), send notification if ingest wants to revise a curated sequence

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/3087, https://github.com/loculus-project/loculus/issues/3084, https://github.com/loculus-project/loculus/issues/2613

partially: https://github.com/loculus-project/loculus/issues/3085,

preview URL: https://insdc-curation-ingest-qui.loculus.org/

### Summary

1. Ingest no longer approves all revisions in the insdc_ingest_group but only revisions it submitted (e.g. user= insdc_ingest_user) - this allows curators to revise a sequence and not have it be approved directly by ingest. In order to do this the backend now allows filtering by submitter not just submissionGroup
2. Send slack notification if ingest wants to revise a curated sequence - see if revision is in a sequence where any version was submitted by a user other than insdc_ingest_user - in this case send a notification and put the sequence into state blocked. Again the backend must be modified to also return the submitter of the original-metadata.

### PR Checklist

- [x] Requires https://github.com/loculus-project/loculus/pull/3111 to be submitted
- [x] Confirm the state (metadata) is the same as currently using comparison script: Checkout https://github.com/pathoplexus/pathoplexus/commit/ce68b92e8b01055a36b5fa60893876078963bf98
```
micromamba activate pp-integrity
cd tests/regression-testing 
snakemake results/ebola-zaire.meta.main.insdc-curation-ingest-qui.diff 
snakemake results/cchf.meta.main.insdc-curation-ingest-qui.diff 
snakemake results/west-nile.meta.main.insdc-curation-ingest-qui.diff 
````
There is no diff in the ebola-zaire, cchf and west nile metadata.
- [x] Curate three sequences via the revise page as super-user: curate metadata, curate sequence, curate metadata and sequence -> check these are not overwritten by ingest
- [x] Make sure ingest doesn't approve curated sequences
- [x] Connect to preview db - change original data hash of a curated sequence (three cases as above), make sure that we get a notification that ingest doesn't know what to do and will not revise these sequences. 
I see this in logs: 
```
15:24:26  WARNING (   compare_hashes.py:  63) - Sequence MZ424862 has been curated before - do not know how to proceed
```
Somehow the slack hook was removed from the config, I had to add it back. Now we are notified: 
image
- [ ] Double check slack hook is still in pathoplexus ingest config

### CODE REVIEW

**Comment by anna-parker** [comment]
Testing: 
1. LOC_000373F.1: curated metadata
2. LOC_000372H.1: curated sequence
3. LOC_00036WV.1: both

4. LOC_000RSPC.2: both also multi-segmented

File: ingest/scripts/compare_hashes.py [hunk 1]
```diff
@@ -22,13 +23,102 @@ class Config:
     segmented: str
     nucleotide_sequences: list[str]
     debug_hashes: bool = False
+    slack_hook: str = ""
+
+
+InsdcAccession = str
```
anna-parker, line 29: @corneliusroemer this is a small refactor to make the code clearer - especially as in a later PR I started handling merging revisions and curations and this gets complicated.

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -487,12 +488,19 @@ open class SubmissionDatabaseService(
 
         val groupCondition = getGroupCondition(groupIdsFilter, authenticatedUser)
 
+        val submitterCondition = if (submitterNamesFilter !== null) {
```
corneliusroemer, line 491: Maybe we should also include empty list here? Depends what we want empty list to do. Right now empty list means nothing gets approved, which is fine but then what's the point of calling the endpoint, though it does make some client code easier potentially. I'm undecided.

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -487,12 +488,19 @@ open class SubmissionDatabaseService(
...
+            Op.TRUE
+        }
+
         val organismCondition = SequenceEntriesView.organismIs(organism)
 
         val accessionVersionsToUpdate = SequenceEntriesView
             .selectAll()
             .where {
-                statusCondition and accessionCondition and scopeCondition and groupCondition and organismCondition
+                statusCondition and accessionCondition and scopeCondition and groupCondition and organismCondition and
```
corneliusroemer, line 502: As there might be more conditions in the future, would be kind of nice to make them a list like this 

```
statusCondition and 
accessionCondition and 
scopeCondition and 
groupCondition and 
organismCondition and
```

corneliusroemer:
Very cool stuff, this was on my backlog and I'm very grateful you've done already so much more! Still reviewing, but wanted to ask what you think about adding/modifying one or two backend tests to make the new behaviour tested?

ingest/scripts/compare_hashes.py
```diff
@@ -22,13 +23,102 @@ class Config:
...
+        sampled_out.append({fasta_id: insdc_accession_base, "hash": hash_float})
+    return hash_float, sampled_out
+
+
+def process_hashes(
+    insdc_accession_base: str,
+    fasta_id: str,
+    record: dict,
+    submitted: dict,
+    update_manager: SequenceUpdateManager,
```
corneliusroemer, line 86: It's a bit weird that this expects an empty manager (except for config) to be passed in.
Maybe add config as arg instead of update manager and take out config from manager?

ingest/scripts/compare_hashes.py
```diff
@@ -22,13 +23,102 @@ class Config:
...
+):
+    if insdc_accession_base not in submitted:
+        update_manager.submit.append(fasta_id)
+    else:
+        sorted_versions = sorted(
```
corneliusroemer, line 91: You could unnest all of this (get rid of the else) by returning after the submit append above (guard clause pattern is quite nice to deindent stuff: https://medium.com/@niiicolai/design-pattern-guard-clause-c050a26dcf35)

corneliusroemer: Could make the whole logic here easier to follow, I'm struggling a little 🙈

corneliusroemer: I see you just moved this and the code hardly changed, don't feel blocked by this, I can refactor this some other time if you don't want to right now.

corneliusroemer: Great job lightening up the logic of the top level function!

ingest/scripts/compare_hashes.py
```diff
@@ -22,13 +23,102 @@ class Config:
...
+    if not keep:
+        sampled_out.append({fasta_id: insdc_accession_base, "hash": hash_float})
+    return hash_float, sampled_out
+
+
+def process_hashes(
+    insdc_accession_base: str,
+    fasta_id: str,
+    record: dict,
+    submitted: dict,
```
corneliusroemer, line 85: It might not be bad to quickly explain what for example "submitted" is with more typing, instead of just `dict` something like `dict[InsdcAccession, dict[...?]]`

corneliusroemer: Or small example, it's I think:
```
AB1234:
  LoculusAccession: PP_1234
  versions:
  - submitter: insdc_ingest_user
    hash: asdfasd
    status: APPROVED
    version: 1
```

ingest/scripts/compare_hashes.py
```diff
@@ -22,13 +23,102 @@ class Config:
...
+    insdc_accession_base: str, subsample_fraction: float, sampled_out: dict[str, str], fasta_id: str
+) -> tuple[float, list[dict[str, str]]]:
+    hash_float = md5_float(insdc_accession_base)
+    keep = hash_float <= subsample_fraction
+    if not keep:
+        sampled_out.append({fasta_id: insdc_accession_base, "hash": hash_float})
+    return hash_float, sampled_out
+
+
+def process_hashes(
```
corneliusroemer, line 81: Actually rather than passing the whole `submitted` in, you could pass just `submitted[insdc_accession_base]` in (potentially empty)

corneliusroemer (approved):
This looks great, backend changes are very minimal, would just be good to keep things tested in backend.

Python is good, I didn't verify things super closely, but I trust your testing, the changes seem to be fairly small, mostly a refactor - the new logic is very localized I think, simply bailing when a sequence was ever touched by non-insdc_ingest_user.

ingest/scripts/compare_hashes.py
anna-parker: I do need the full submitted dict - I will add a better description tho!

ingest/scripts/compare_hashes.py
anna-parker: Done :-)

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
anna-parker: > Right now empty list means nothing gets approved

I actually like this option so I will leave like this if thats ok

PR #3126, opened 2024-10-30
fix(kubernetes): don't create `external_metadata_updater` user when `disableEnaSubmission`

fengelniederhammer:
preview URL:

### Summary

If `disableEnaSubmission: true`, then I don't want to be forced to configure a `externalMetadataUpdaterPassword`

### PR Checklist


anna-parker:
Just wanting to make sure this is documented here and not only on slack - we currently only use `external_metadata_updater` for the ena submission pod but it was written in a way that it could be used by other updater pods. So now it will require a template change if users want to add other pods like this.

PR #3130, opened 2024-10-31
feat(deployment): Add version comment to input fields

theosanderson:
Adds version comment to input fields so it appears on revision form:
image

There's an argument that it shouldn't necessarily for editing version 1. But:

- I think notes can actually be useful even for version 1
- This lets us quickly make the change and we could always change it later

https://addversioncomment.loculus.org/

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
@@ -91,25 +91,28 @@ open class ReleasedDataModel(
             NullNode.getInstance()
         }
 
-        var metadata = rawProcessedData.processedData.metadata +
-            mapOf(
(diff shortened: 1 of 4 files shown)


anna-parker:
Just tried it out - sadly I dont see the versionComment on the revised sequence page: 
image

theosanderson:
Thanks @anna-parker - I think this PR was maybe based on a total misunderstanding of how things work? https://loculus.slack.com/archives/C05G172HL6L/p1730384655481129

corneliusroemer:
The more I think about it the more I like it, I'm overdoing the generality, no point in this here, better be explicit, it's not so bad. We hopefully won't have many more fields like this, that are treated specially by Loculus but also user-specified.

We need to add versionComment to preprocessing config though, otherwise it won't make it through.

theosanderson:
Closed for #3141

PR #3137, opened 2024-10-31
feat(e2e): test that versionComment gets emitted correctly

corneliusroemer:
Try to see whether we can test #3141


corneliusroemer:
Doesn't fail helpfully, let's give up.

PR #3138, opened 2024-10-31
feat(ci): dump current db schema automatically

### LINKED ISSUE #2791: Add current schema (after all migrations) to backend docs
Right now we only have 2 files in the backend migrations folder, but they will become more numerous.

For reading the code, it doesn't matter really what all the migration files are, what one cares about is the resulting schema after all migrations are applied. This is currently not easily accessible.

We should create an action that dumps the resulting schema into a docs folder after running flyway on all migrations. This action only needs to run on changes in the flyway migration dir.

Some skeleton generated by Claude:

<details>

```
name: Generate and Commit Flyway Schema

on:
  push:
    paths:
      - 'schema/**'
  pull_request:
    paths:
      - 'schema/**'
  workflow_dispatch:

jobs:
  generate-schema:
    runs-on: ubuntu-latest

    services:
      postgres:
        image: postgres:16
        env:
          POSTGRES_PASSWORD: postgres
        options: >-
          --health-cmd pg_isready
          --health-interval 10s
          --health-timeout 5s
          --health-retries 5
        ports:
          - 5432:5432

    steps:
    - name: Checkout repository
      uses: actions/checkout@v4

    - name: Run Flyway migrations
      uses: docker://flyway/flyway:10-alpine
      env:
        FLYWAY_URL: jdbc:postgresql://postgres:5432/postgres
        FLYWAY_USER: postgres
        FLYWAY_PASSWORD: postgres
      with:
        args: -locations=filesystem:/github/workspace/schema migrate

    - name: Generate schema
      run: |
        docker run --network host -v $PWD:/workspace postgres:16 \
        bash -c "PGPASSWORD=postgres pg_dump -h localhost -U postgres -d postgres --schema-only > /workspace/schema.sql"

    - name: Commit and push if changed
      run: |
        git config --global user.name 'GitHub Action'
        git config --global user.email 'action@github.com'
        git add schema.sql
        git diff --quiet && git diff --staged --quiet || (git commit -m "Update schema based on Flyway migrations" && git push)
```

</details>

theosanderson:
resolves #2791 

Adds CI to capture the DB schema after all migrations in `backend/docs/db/schema.sql`. It calculates what this is and if it doesn't match the current captured file the check fails. To make the check pass you can add the `update_db_schema` label which will cause the CI to write over that schema file with the current schema.

Purpose: to make it easy to see diffs in schema, allows seeing the current schema without having to manually add up all migrations in one's head

Limitations:

CI does not re-run after the auto-update. I consider that out of scope for this PR.
I'm not sure if migrations will change order - maybe there is a way to make the order more deterministic?

### DIFF
--- a/.github/workflows/schema-dump.yml
+++ b/.github/workflows/schema-dump.yml
@@ -79,13 +79,13 @@ jobs:
     - name: Handle schema changes
       if: steps.check-changes.outputs.changed == 'true'
       run: |
-        if [[ "${{ contains(github.event.pull_request.labels.*.name, 'update-schema') }}" == "true" ]]; then
+        if [[ "${{ contains(github.event.pull_request.labels.*.name, 'update_db_schema') }}" == "true" ]]; then
           git config --global user.name 'GitHub Action'
           git config --global user.email 'action@github.com'
           git commit -m "Update schema documentation based on migration changes"
           git push
         else
           echo "::error::Schema changes detected but 'update-schema' label is not present on the PR"
-          echo "Please add the 'update-schema' label if these changes are intentional"
+          echo "Please add the 'update_db_schema' label if these changes are intentional"
           exit 1
         fi
--- a/.github/workflows/schema-dump.yml
+++ b/.github/workflows/schema-dump.yml
@@ -60,7 +60,7 @@ jobs:
           > /github/workspace/backend/docs/db/schema.sql"
 
     - name: Ensure schema file is tracked
-      run: git add -f backend/docs/schema.sql
+      run: git add -f backend/docs/db/schema.sql
 
     - name: Cat schema
       run: cat backend/docs/db/schema.sql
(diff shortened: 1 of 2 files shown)


.github/workflows/schema-dump.yml
```diff
@@ -0,0 +1,91 @@
...
+    runs-on: ubuntu-latest
+    
+    # Explicitly set permissions needed for committing
+    permissions:
+      contents: write
+      pull-requests: read
+
+    services:
+      postgres:
+        image: postgres:16
```
corneliusroemer, line 22: we actually are using postgres:15 in production, so should maybe use this? but probably there are no differences.

corneliusroemer (approved):
Excellent stuff! Amazing it took you only like 40min based on commits.

corneliusroemer:
Oh wow, version differences do seem to matter!

image

theosanderson:
Since we support managed DBs its a bit debatable but happy to go with this!

theosanderson:
Nice to see it works :)

PR #3144, opened 2024-10-31
Fix issue with revoking already revoked sequences

### LINKED ISSUE #1946: You can revoke a sequence that's already been revoked
When viewing the page for a revoked sequence, you can see a button to revoke it even _more_

image

You can hit the button, and it will go through the actions - you get taken to the review page where you can release the sequence.

it's then Revoked _twice_.

image

theosanderson:
Fixes #1946

Add a check to prevent revoking an already revoked sequence.

* **SequenceDataUI.tsx**
  - Add a check to verify if the sequence is already revoked before displaying the revoke button.
  - Conditionally render the `RevokeButton` component only if the sequence is not already revoked.

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/loculus-project/loculus/issues/1946?shareId=fd1bd13d-f96f-458d-9075-c1d1dccc4b90).


theosanderson:
(went wrong here - abandoning for now)

PR #3163, opened 2024-11-03
feat!(website): add redirects to docs and add search sequences docs

### LINKED ISSUE #3140: Hide "Help" button in search panel by default
The search panel has a "Help" button on top:

image

This links to `/docs/how-to/search-sequences-website` which has content on Pathoplexus (see [here](https://pathoplexus.org/docs/how-to/search-sequences-website)) but not in the base Loculus, so it should be configurable and be hidden by default.

### LINKED ISSUE #3160: Fix Pathoplexus-specific link in API docs
Currently, on the API documentation page, we write

> For more tips on how to use our API we recommend starting out with our [API Authentication documentation.](https://loculus.genspectrum.org/docs/how-to/authentication-api)

and "API Authentication documentation" points to a page that exists on [Pathoplexus](https://pathoplexus.org/docs/how-to/authentication-api) but not in a base instance. For Loculus, we have the same instructions instead on loculus.org at https://loculus.org/for-users/authenticate-via-api/.

By default, I think that the link should point to the Loculus documentation page. For Pathoplexus, we can either make it somehow configurable in Loculus or overwrite the page on the Pathoplexus-side.

chaoran-chen:
Resolve #3140 and resolve #3160

preview URL: https://redirect-to-docs.loculus.org

### Summary

~This adds a redirect to the Loculus docs from the two locations where Pathoplexus has actual docs pages but not Loculus. This also copies over the mutation search docs from Pathoplexus to the Loculus docs.~

~It unfortunately doesn't seem to possible to have a server-side redirect in .mdx, so I had to create .astro files instead. This means that we have to delete these files on the Pathoplexus-side again because .astro files preceed .mdx files. @theosanderson, do you think that's a reasonable solution?~

The search form help button and the link to the authentication API docs now point to the Loculus docs. The variables are in separate files so that they can be easily overwritten by Pathoplexus. Here is a PR for Pathoplexus: https://github.com/pathoplexus/pathoplexus/pull/281

This also copies over the mutation search docs from Pathoplexus to the Loculus docs.

### PR Checklist

### DIFF
--- a/website/src/pages/api-documentation/index.astro
+++ b/website/src/pages/api-documentation/index.astro
@@ -1,8 +1,10 @@
 ---
+import { authenticationApiDocsUrl } from './authenticationApiDocsUrl';
 import { getRuntimeConfig, getWebsiteConfig } from '../../config';
 import BaseLayout from '../../layouts/BaseLayout.astro';
-const clientConfig = getRuntimeConfig().public;
 import { routes } from '../../routes/routes.ts';
+
+const clientConfig = getRuntimeConfig().public;
 
 const websiteConfig = getWebsiteConfig();
 
@@ -24,17 +26,17 @@ const BUTTON_CLASS =
 
         <div>
             <p class='mt-4 mb-4'>
-                We offer a
+                There is a
                 <a
                     href='https://swagger.io/tools/swagger-ui/'
                     class='underline text-primary-900 hover:text-primary-800 hover:no-underline'
                 >
                     Swagger UI
                 </a>
-                for documentation and direct interaction with our APIs. For more tips on how to use our API we recommend
-                starting out with our
+                for documentation and direct interaction with the APIs. For more tips on how to use the API, it is recommended
+                to start with the
                 <a
-                    href='docs/how-to/authentication-api'
+                    href={authenticationApiDocsUrl}
                     class='underline text-primary-900 hover:text-primary-800 hover:no-underline'
                 >
                     API Authentication documentation.
--- a/docs/astro.config.mjs
+++ b/docs/astro.config.mjs
@@ -37,6 +37,7 @@ export default defineConfig({
                     label: 'For users',
                     items: [
                         { label: 'Introduction', link: '/for-users/introduction/' },
+                        { label: 'Search sequences', link: '/for-users/search-sequences/' },
                         { label: 'Edit account', link: '/for-users/edit-account/' },
                         { label: 'Create and manage groups', link: '/for-users/create-manage-groups/' },
                         { label: 'Submit sequences', link: '/for-users/submit-sequences/' },
--- a/docs/src/content/docs/for-users/search-sequences.md
+++ b/docs/src/content/docs/for-users/search-sequences.md
@@ -1,0 +1,33 @@
+---
+title: Search sequences
+---
+
+## Mutations
+
+### Nucleotide mutations and insertions
+
+A nucleotide mutation has the format `<position><base>` or `<base_ref><position><base>`. A `<base>` can be one of the four nucleotides `A`, `T`, `C`, and `G`. It can also be `-` for deletion and `N` for unknown. For example if the reference sequence is `A` at position 23 both: `23T` and `A23T` will yield the same results.
+
+If your organism is multi-segmented you must append the name of the segment to the start of the mutation, e.g. `S:23T` and `S:A23T` for a mutation in segment `S`.
+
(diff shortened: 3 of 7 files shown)


website/src/pages/api-documentation/index.astro
```diff
@@ -24,15 +24,15 @@ const BUTTON_CLASS =
...
                 <a
                     href='https://swagger.io/tools/swagger-ui/'
                     class='underline text-primary-900 hover:text-primary-800 hover:no-underline'
                 >
                     Swagger UI
                 </a>
-                for documentation and direct interaction with our APIs. For more tips on how to use our API we recommend
-                starting out with our
+                for documentation and direct interaction with the APIs. For more tips on how to use the API, it is recommended
+                to start with the
```
chaoran-chen, line 35: I rephrased this as it seems a bit strange to use "we" if the instance is hosted by someone outside of the Loculus/Pathoplexus team.

theosanderson:
> This means that we have to delete these files on the Pathoplexus-side again because .astro files preceed .mdx files. @theosanderson, do you think that's a reasonable solution?

Given that this is something of a hack anyway (i.e. in some idealised world the links would just point directly to their final destination), my preference would be to just do client side redirects (our website is very dependent on JS anyway). Because at the moment the architecture for Pathoplexus (and anyone else who creates an overlay image, which might be something that happens) is nice and clean with just one structure copied on another and I'm not sure what we would do in this case.

I don't feel super strongly but it would be good to see the proposed Pathoplexus change in parallel if we did go with the route you are proposing (and this PR here would be a breaking change).

theosanderson:
(I would also be open to us just refactoring the code so that these URLs are defined in a special file and overwriting that file in PP)

chaoran-chen:
> Because at the moment the architecture for Pathoplexus (and anyone else who creates an overlay image, which might be something that happens) is nice and clean with just one structure copied on another and I'm not sure what we would do in this case.
> [...]
> (I would also be open to us just refactoring the code so that these URLs are defined in a special file and overwriting that file in PP)

I'm a bit hesitant to introduce too many configs as it will increase complexity and I'm also not sure whether the overlays are something that we would want to officially support. I imagine that the vast majority of future Loculus instances will not create overlays and won't want to write own documentation. Therefore, I think that it makes more sense to have the bit more "hacky" code on the Pathoplexus end which, with the overlay, I think is already the case.

> Given that this is something of a hack anyway (i.e. in some idealised world the links would just point directly to their final destination), my preference would be to just do client side redirects (our website is very dependent on JS anyway).

A JS based client redirect is not ideal for SEO and actually, I wonder whether the current 302 is the right choice or whether a 301 would be better to prevent a search engine to index the page at all. It's not a huge deal but if we can avoid this by adding a line in Pathoplexus, I think that's better.

A very simple solution would be just adding the following line to [sync.sh after line 38](https://github.com/pathoplexus/pathoplexus/blob/ff366647c20fa4ecf5466e60744e94c8fbc9d2b5/sync.sh#L38):

```
rm -f monorepo/website/src/pages/docs/how-to/authentication-api.astro monorepo/website/src/pages/docs/how-to/search-sequences-website.astro
```

chaoran-chen:
(Is `feat!(website)` the right way to highlight a breaking change?)

theosanderson:
> I'm a bit hesitant to introduce too many configs as it will increase complexity

But here we are adding a special file in a particular location, defined by how we set up to Pathoplexus, in order to redirect. To me that seems pretty equal to adding a special file that simply defines where to link to?

I definitely agree on the whole with keeping more hacky stuff on the PP side in general (but obviously we would also like to minimalise hackiness in general). Here what I'm highlighting is one is ending up with a hack on the Loculus side anyway (having a redirect at all), and also one on the PP side.

I would prefer not to have these files specifically mentioned in the PP sync script, at minimum (IMO) one would have some `.files_to_delete` file or something.

theosanderson:
> `(Is feat!(website) the right way to highlight a breaking change?)`
Yes should be (I think there's an option to specify the nature in square brackets but I'm not sure how much we formalised that)

chaoran-chen:
> To me that seems pretty equal in added complexity to adding a special file that simply defines where to link to?

Would you like to outline more or create a PR to show how this would work? Would this affect the config and templating or do you have something much simpler in mind?

theosanderson:
About to go out so made this very quickly but I'm suggesting something like this:

https://github.com/loculus-project/loculus/pull/3164/files

chaoran-chen:
Ah okay, I see, sure, happy with that as well.

chaoran-chen:
@theosanderson, ok, done! Here's the corresponding Pathoplexus PR: https://github.com/pathoplexus/pathoplexus/pull/281

theosanderson (approved):
Nice, thanks!

PR #3168, opened 2024-11-04
fix(ci): Fix CI to dump schema to watch the correct files

theosanderson:
The files to trigger this CI to run were misspecified


anna-parker (approved):
Haven't tested this but confirmed the path is correct :-)

PR #3178, opened 2024-11-05
docs: Fixes to my-first-loculus.md

theosanderson:
Reorder tutorial to try to avoid a problem @DrYak encountered, and use `<pre>` for TSV that was being turned into spaces.

https://docs-docs-update.loculus.org/for-administrators/my-first-loculus/#testing-it-out-with-some-data

(We should do a more substantial change in the future to avoid having to delete the pod at all, which is too complicated for this tutorial)

Thanks to @DrYak and @chaoran-chen


chaoran-chen (approved):
Thanks!

PR #3179, opened 2024-11-05
Add additional pod annotation to deployment templates to enable easy restarts on config changes

### LINKED ISSUE #3170: Figure out how to ensure Pathoplexus restarts appropriately on config update
We recently got confused because a pathoplexus roll out that didn't involve changing the loculus image tag didn't restart the backend (changes that do affect the loculus image tag do restart the backend). This may be expected but we should find a way to ensure restarting (this was traditionally the Reloader).

We should add an an annotation to the template of every deployment which comes from an optional values.yaml setting additionalPodAnnotation, then when we increment that they will restart

theosanderson:
Fixes #3170

### DIFF
--- a/kubernetes/loculus/templates/keycloak-deployment.yaml
+++ b/kubernetes/loculus/templates/keycloak-deployment.yaml
@@ -18,6 +18,8 @@ spec:
       labels:
         app: loculus
         component: keycloak
+      annotations:
+        {{- toYaml .Values.additionalPodAnnotation | nindent 8 }}
     spec:
      {{- include "possiblePriorityClassName" . | nindent 6 }}
       initContainers:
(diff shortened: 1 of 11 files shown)


kubernetes/loculus/templates/keycloak-deployment.yaml
```diff
@@ -18,6 +18,8 @@ spec:
       labels:
         app: loculus
         component: keycloak
+      annotations:
```
anna-parker, line 21: Will this then also cause keycloak to restart when there is a change?

We haven't restarted keycloak in a while - and I recall Cornelius and I noticed that a keycloak restart means temporarily logging all users out - so maybe we don't want to add this for keycloak and really only restart when the pod needs to restart.

PR #3183, opened 2024-11-06
feat(ena-submission, ingest): Submit more geolocation information, filter out all versions of an ENA accession

anna-parker:
resolves #

preview URL: https://small-fixes.loculus.org/

### Summary

1. Submit all additional geoLoc information in the `geographic location (region and locality)` field (see https://www.ebi.ac.uk/ena/browser/view/ERC000033) -> confirm this is successful in tests
2. Filter out all versions of a accession we submitted to ENA - so far I have only been filtering out that explicit version, and as I also filter out the biosample accession we havent re-ingested samples that we submitted - but we should add this check just in case. -> In the zip I added multiple versions of an accession to confirm they are filtered out

### PR Checklist


ena-submission/src/ena_deposition/config.py
```diff
@@ -50,7 +50,7 @@ def secure_ena_connection(config: Config):
         config.test = True
         logging.info("Submitting to ENA dev environment")
         config.ena_submission_url = "https://wwwdev.ebi.ac.uk/ena/submit/drop-box/submit"
-        config.github_url = "https://raw.githubusercontent.com/pathoplexus/ena-submission/main/test/approved_ena_submission_list.json"
+        config.github_url = "https://raw.githubusercontent.com/pathoplexus/ena-submission/loculus_test/test/approved_ena_submission_list.json"
```
anna-parker, line 53: Revert before merge
```suggestion
        config.github_url = "https://raw.githubusercontent.com/pathoplexus/ena-submission/main/test/approved_ena_submission_list.json"
```

chaoran-chen (approved):

Issue #1612, opened 2024-04-16
Review page (and `/get-sequences` endpoint) is slow


fengelniederhammer:
I just noticed that the `/get-sequences` endpoint is slow when there are many sequences in the database:

* Submit 100,000 dummy sequences. 
* Go to the review page
* What I observe (locally and on the current main deployment): The sequences take very long to load, and `https://backend-main.loculus.org/dummy-organism/get-sequences?groupIdsFilter=2&initialStatusesFilter=RECEIVED,IN_PROCESSING,HAS_ERRORS,AWAITING_APPROVAL&statusesFilter=RECEIVED,IN_PROCESSING,HAS_ERRORS,AWAITING_APPROVAL&page=0&size=50` take around 8 seconds to finish.

I added some log statements locally. The method `SubmissionDatabaseService::getSequences` takes the full time. Something in that query must be quite slow.

corneliusroemer:
To distinguish this issue: In contrast to #3269 this is purely about the performance of the backend, not the symptom of the website review page being slow as a result.

PR #3189, opened 2024-11-07
feat: improve review page loading times

### LINKED ISSUE #1000: rework `get-sequences` endpoint to return metadata
The [first approach](https://github.com/loculus-project/loculus/pull/994) shows some shortcomings. 

1) the warning icon should be yellow, but due lazy loading thats not possible when only status infos are available
2) the `get-data-to-edit` endpoint excludes `RECEIVED` and `IN_PROCESSING` statuses. 

Solution:

Send data directly with the `get-sequences` call.

fhennig:
resolves #1000

preview URL: https://improve-review-load.loculus.org

### Summary

This PR changes it so that all data for the whole page is fetched in a single request. Before, every ReviewCard made a separate request for its own data.

### Screenshot

n/A -- no visual changes.

### PR Checklist


fhennig:
cool, the test went through!

backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
```diff
@@ -229,6 +229,15 @@ data class SequenceEntryStatus(
     val isRevocation: Boolean = false,
     val submissionId: String,
     val dataUseTerms: DataUseTerms,
+    val processedData: ProcessedData<GeneticSequence>?,
+    val originalData: OriginalData<GeneticSequence>?,
+    @Schema(description = "The preprocessing will be considered failed if this is not empty")
+    val errors: List<PreprocessingAnnotation>? = null,
+    @Schema(
+        description =
+        "Issues where data is not necessarily wrong, but the user might want to look into those warnings.",
```
theosanderson, line 238:
```suggestion
        "Issues where data is not necessarily wrong, but there are warning messages from the pipeline that the user might want to look into.",
```

(just sounded weird to me before)

website/src/types/backend.ts
theosanderson, line 1: I think that if we were to make the changes in this PR we need to not send sequences in the endpoint used on the review page. This is is because:

- sequences for some viruses are 200 kb, 100 per page = 20 MB, polling constantly = a lot of bandwidth

We may want to delay making any change here for now for simplicity to avoid having to add extra types etc.

theosanderson (changes requested):
Sorry for the slow review, and sorry that this issue wasn't fully specced. Unfortunately I think the PR at present is problematic as it could use 20 MB of bandwidth every few seconds.

fhennig:
Thanks for bringing that up Theo! I didn't realize that the sequences were in there too. In that case you're right, I think it's best to shelve this again for now, as we have something that's currently working and we initially thought that this might be a "quick win".

I'm closing this for now.

corneliusroemer:
It's not really working well though. Load times of 15+ seconds if there are 15k sequences aren't really acceptable.

fhennig:
Maybe worth discussing next monday? pinging @chaoran-chen 

I haven't looked into it that much, but it seems like it might be a good idea to try and redesign it so we can send the metadata withtout the raw sequence data - which is the largest chunk of the data in my understanding. Maybe us 3 (Theo, Cornelius and me) can have a refinement meeting on this ticket and see what would need to get done for this, and then we can decide if the amount of work is something we want to put in now or later.

chaoran-chen:
Sounds good, I added it to the agenda for next Monday.

corneliusroemer:
There might be multiple issues here:
- sending sequence data (totally unnecessary) - but that shouldn't itself cause that much of a slowdown alone especially as many of the sequences loaded for WNV are very short there isn't that much data to be transferred in practice.
- slow performance of backend endpoint even for small number of sequences, maybe due to a sort or something like that in the SQL, see e.g. #1612
- we also send lots of nulls for processed data which is wasteful, we only care about the non-null fields (can assume nulls for the others)

theosanderson:
The issue might be _uncompressing_ the sequence data rather than sending it (we at most send a page's worth of data IIUC)

PR #3193, opened 2024-11-11
refactor getSequences usage in SeqSetCitationsController

fhennig:
resolves #

preview URL:

### Summary

In preparation for this: https://github.com/loculus-project/loculus/issues/1000 where I want to return metadata which requires the `organism` to be known in getSequences, I decided it would make sense to split the usage of the `getSequences` function that _doesn't_ use the `organism` out into its own function.

I added a new function that does exactly what is needed for the usage in seq set citations, which - I think - makes that code path easier to understand and also makes the `getSequences` function (and the changes to come) easier.

### Screenshot
n/a

### PR Checklist

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCitationsDatabaseService.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCitationsDatabaseService.kt
@@ -33,7 +33,6 @@ import org.loculus.backend.controller.UnprocessableEntityException
 import org.loculus.backend.service.crossref.CrossRefService
 import org.loculus.backend.service.crossref.DoiEntry
 import org.loculus.backend.service.submission.AccessionPreconditionValidator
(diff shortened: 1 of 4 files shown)


backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCitationsDatabaseService.kt
```diff
@@ -377,9 +376,9 @@ class SeqSetCitationsDatabaseService(
             val createdAt: Timestamp,
         )
 
-        val userAccessionStrings = userAccessions.flatMap {
-            listOf(it.accession.plus('.').plus(it.version), it.accession)
-        }
+        val userAccessionStrings = accessionVersions
+            .flatMap { listOf(it.accession, "${it.accession}.${it.version}") }
```
fengelniederhammer, line 380:
```suggestion
            .flatMap { listOf(it.accession, it.displayAccessionVersion()) }
```

fengelniederhammer (approved):
LGTM, thanks :)

PR #3212, opened 2024-11-14
feat!: allow disabling of seqsets [CFG]

### LINKED ISSUE #3210: Allow disabling of SeqSets
As a part of #2104, make it possible to disable SeqSets:
* (most important) hide the SeqSets navigation item on the website
* disable SeqSet pages on the website
* disable the `SeqSetCitationsController` in the backend
* make it configurable in the Helm value.yaml

fengelniederhammer:
resolves #3210

preview URL: https://3210-allow-disabling-of-s.loculus.org/

### Summary

I grouped the seqSet config values. The values.yaml section can look like
```
seqSets:
  enabled: false
``` 
or
```
seqSets:
  enabled: false
  crossRef: null
``` 
or
```
seqSets:
  enabled: true
  crossRef:
    DOIPrefix: "placeholder"
    endpoint: "https://test.crossref.org"
``` 

When `seqSets. enabled == false`, 
* the backend endpoint are disabled
* the website SeqSets navigation item is disabled
* the SeqSets pages on the website return 404

BREAKING CONFIG CHANGES:
* `enableCrossRefCredentials` was removed. Use `seqSets.enabled` instead.
* moved `crossRef` to `seqSets.crossRef`

### Screenshot

grafik

grafik

### PR Checklist

### DIFF
--- a/.idea/runConfigurations/BackendApplicationKt.run.xml
+++ b/.idea/runConfigurations/BackendApplicationKt.run.xml
@@ -1,7 +1,7 @@
 <component name="ProjectRunConfigurationManager">
   <configuration default="false" name="BackendApplicationKt" type="SpringBootApplicationConfigurationType" factoryName="Spring Boot" nameIsGenerated="true">
     <module name="backend.main" />
-    <option name="PROGRAM_PARAMETERS" value="--spring.datasource.url=jdbc:postgresql://localhost:5432/loculus --spring.datasource.username=postgres --spring.datasource.password=unsecure --keycloak.user=backend --keycloak.password=backend --keycloak.realm=loculus --keycloak.client=backend-client --keycloak.url=http://localhost:8083 --loculus.config.path=website/tests/config/backend_config.json --loculus.debug-mode=true --spring.security.oauth2.resourceserver.jwt.jwk-set-uri=http://localhost:8083/realms/loculus/protocol/openid-connect/certs" />
+    <option name="PROGRAM_PARAMETERS" value="--spring.datasource.url=jdbc:postgresql://localhost:5432/loculus --spring.datasource.username=postgres --spring.datasource.password=unsecure --keycloak.user=backend --keycloak.password=backend --keycloak.realm=loculus --keycloak.client=backend-client --keycloak.url=http://localhost:8083 --loculus.config.path=website/tests/config/backend_config.json --loculus.debug-mode=true --spring.security.oauth2.resourceserver.jwt.jwk-set-uri=http://localhost:8083/realms/loculus/protocol/openid-connect/certs --loculus.enable-seqset-citations=true" />
     <option name="SPRING_BOOT_MAIN_CLASS" value="org.loculus.backend.BackendApplicationKt" />
     <extension name="net.ashald.envfile">
       <option name="IS_ENABLED" value="false" />
--- a/kubernetes/loculus/templates/loculus-backend.yaml
+++ b/kubernetes/loculus/templates/loculus-backend.yaml
@@ -41,6 +41,8 @@ spec:
           ports:
             - containerPort: 8079
           args:
+            - "--loculus.enable-seqsets={{ $.Values.seqSets.enabled }}"
+            {{- if $.Values.seqSets.crossRef }}
             - "--crossref.doi-prefix=$(CROSSREF_DOI_PREFIX)"
             - "--crossref.endpoint=$(CROSSREF_ENDPOINT)"
             - "--crossref.username=$(CROSSREF_USERNAME)"
@@ -49,6 +51,7 @@ spec:
             - "--crossref.email=$(CROSSREF_EMAIL)"
             - "--crossref.organization=$(CROSSREF_ORGANIZATION)"
             - "--crossref.host-url=$(CROSSREF_HOST_URL)"
+            {{- end }}
             - "--keycloak.password=$(BACKEND_KEYCLOAK_PASSWORD)"
             - "--keycloak.realm=loculus"
             - "--keycloak.client=backend-client"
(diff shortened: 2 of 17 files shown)


kubernetes/loculus/values.yaml
```diff
@@ -1,8 +1,10 @@
 environment: server
 robotsNoindexHeader: false
-crossRef:
-  DOIPrefix: "placeholder"
-  endpoint: "https://test.crossref.org"
+seqSets:
+  enabled: false
```
fengelniederhammer, line 4: Reminder: revert this before merge

theosanderson:
(We eventually coalesced on `feat!` for breaking changes and have some CI that uses that)

kubernetes/loculus/values.yaml
```diff
@@ -1519,7 +1521,6 @@ secrets:
     data:
       username: "dummy"
       password: "dummy"
-enableCrossRefCredentials: true
```
theosanderson, line 1522: (Using this line to create a thread on a general theme)

There are valid use cases for SeqSets without the ability to generate DOIs (i.e. without CrossRef). I'm sure those weren't supported in a super nice way before this (e.g. the DOI button was always shown even if it didn't work) but it seems like this PR probably makes that harder? I think we should maintain the possibility to run SeqSets without  specifying CrossRef credentials (if nothing else, for testing) [that may already be maintained, I'm not clear].

theosanderson:
> Is that actually documented somehwere?

👍 https://github.com/loculus-project/loculus/pull/3213

kubernetes/loculus/values.yaml
fengelniederhammer: Indeed, I didn't have that in mind. Should I change it such that one can have SeqSets without CrossRef?

theosanderson: IMO we shouldn't change it to make it _harder_ to do that unless we have a clear reason to. I don't see one atm. I don't think we have a reason atm to invest in making SeqSets work better without crossref than they do atm (e.g. disabling the generate DOI button) but we could do.

theosanderson: (we can probably still remove the `enableCrossRef credentials` thing and make it directly look for the specification of the secret in the yaml, which I think I did for something else)

fengelniederhammer: Now it's possible to set `seqSets.crossRef = null`: Then you have SeqSets, but no CrossRef integration. Locally, it worked for me.

fengelniederhammer:
There is still something weird with the config... On the preview, I can't enable the CrossRef integration. I need to check tomorrow. Locally everything seems to work as expected.

fengelniederhammer:
Sorry for the noise - I tried to make the CrossRef configuration a bit better (it should warn you when it's not properly configured). But Spring acted weird, I don't know why I couldn't get it working in Kubernetes. I removed those changes how.

.idea/runConfigurations/BackendApplicationKt.run.xml
```diff
@@ -1,7 +1,7 @@
...
     <module name="backend.main" />
-    <option name="PROGRAM_PARAMETERS" value="--spring.datasource.url=jdbc:postgresql://localhost:5432/loculus --spring.datasource.username=postgres --spring.datasource.password=unsecure --keycloak.user=backend --keycloak.password=backend --keycloak.realm=loculus --keycloak.client=backend-client --keycloak.url=http://localhost:8083 --loculus.config.path=website/tests/config/backend_config.json --loculus.debug-mode=true --spring.security.oauth2.resourceserver.jwt.jwk-set-uri=http://localhost:8083/realms/loculus/protocol/openid-connect/certs" />
+    <option name="PROGRAM_PARAMETERS" value="--spring.datasource.url=jdbc:postgresql://localhost:5432/loculus --spring.datasource.username=postgres --spring.datasource.password=unsecure --keycloak.user=backend --keycloak.password=backend --keycloak.realm=loculus --keycloak.client=backend-client --keycloak.url=http://localhost:8083 --loculus.config.path=website/tests/config/backend_config.json --loculus.debug-mode=true --spring.security.oauth2.resourceserver.jwt.jwk-set-uri=http://localhost:8083/realms/loculus/protocol/openid-connect/certs --loculus.enable-seqset-citations=false" />
```
theosanderson, line 4: Q: will this disable seqsets for dev by default? If so I think that's not ideal

kubernetes/loculus/templates/loculus-backend.yaml
```diff
@@ -41,6 +41,8 @@ spec:
           ports:
             - containerPort: 8079
           args:
+            - "--loculus.enable-seqset-citations={{ $.Values.seqSets.enabled }}"
```
theosanderson, line 44: (not blocking) I find this name "enable-seqset-citations" weird - its seqsets themselves that are affected not merely "seqset citations" (which I would think of as about citations of seqsets, tracking of which is a feature yet to come

kubernetes/loculus/values.yaml
theosanderson: Thanks!

website/src/components/Navigation/SandwichMenu.tsx
theosanderson, line 1: thanks for the refactor

.idea/runConfigurations/BackendApplicationKt.run.xml
fengelniederhammer: Indeed. I changed it to enabled

kubernetes/loculus/templates/loculus-backend.yaml
fengelniederhammer: I'll rename to `enable-seqsets'

PR #3223, opened 2024-11-18
make it possible to disable the submit workflow on the website

### LINKED ISSUE #3211: Make it possible to disable the submit workflow on the website
As a part of #2104,
* make it possible to hide the Login button on the website
* make it possible to hide the Submission navigation item on the website
* make it possible to completely disable submission pages on the website.
* configure individually in Helm values.

fengelniederhammer:
resolves #3211

preview URL: https://3211-make-it-possible-to.loculus.org/

### Summary

Adds 3 new config values in the `values.yaml`:
```
website:
  websiteConfig:
    enableLoginNavigationItem: false
    enableSubmissionNavigationItem: false
    enableSubmissionPages: false
``` 

### Screenshot

Locally, when setting all new config values to false:
grafik

### PR Checklist

### DIFF
--- a/docs/src/content/docs/reference/helm-chart-config.mdx
+++ b/docs/src/content/docs/reference/helm-chart-config.mdx
@@ -207,6 +207,30 @@ The configuration for the Helm chart is provided as a YAML file. It has the foll
                 <td>true</td>
                 <td>Whether to completely disable submission related pages. Setting this to false is useful when hosting Loculus for analysis-only purposes.</td>
             </tr>
+            <tr>
+                <td>`website.runtimeConfig.public`</td>
+                <td>Object</td>
+                <td></td>
+                <td>Settings for the `public` section of the `runtime_config.json`</td>
+            </tr>
+            <tr>
+                <td>`website.runtimeConfig.public.backendUrl`</td>
+                <td>String</td>
+                <td>true</td>
+                <td>Overwrite the URL where the client-side website code expects the Loculus backend</td>
+            </tr>
+            <tr>
+                <td>`website.runtimeConfig.public.keycloakUrl`</td>
+                <td>String</td>
+                <td>true</td>
+                <td>Overwrite the URL where the client-side website code expects the Keycloak</td>
+            </tr>
+            <tr>
+                <td>`website.runtimeConfig.public.lapisUrlTemplate`</td>
+                <td>String</td>
+                <td>true</td>
+                <td>Overwrite the URLs where the client-side website code expects the LAPIS instances. Must contain `%organism%` as a placeholder.</td>
+            </tr>
         </tbody>
     </table>
 </div>
--- a/docs/src/content/docs/reference/helm-chart-config.mdx
+++ b/docs/src/content/docs/reference/helm-chart-config.mdx
@@ -164,6 +164,53 @@ The configuration for the Helm chart is provided as a YAML file. It has the foll
     </table>
 </div>
 
+## Website Settings
+
+<div className='overflow-x-scroll'>
+    <table className='min-w-[700px]'>
+        <thead>
+            <tr>
+                <th className='w-56'>Field</th>
+                <th>Type</th>
+                <th className='w-32'>Default</th>
+                <th>Description</th>
+            </tr>
+        </thead>
+        <tbody>
+            <tr>
+                <td>`website`</td>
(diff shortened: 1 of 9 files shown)


kubernetes/loculus/values.yaml
```diff
@@ -10,6 +10,11 @@ disableBackend: false
 disablePreprocessing: false
 disableIngest: false
 disableEnaSubmission: true
+website:
```
anna-parker, line 13: Hi! Just a thought - would it be possible to make this optional and set the default to true?

docs/src/content/docs/reference/helm-chart-config.mdx
```diff
@@ -164,6 +164,77 @@ The configuration for the Helm chart is provided as a YAML file. It has the foll
...
+                <td>Settings for the `website_config.json`</td>
+            </tr>
+            <tr>
+                <td>`website.websiteConfig.enableLoginNavigationItem`</td>
+                <td>Boolean</td>
+                <td>true</td>
+                <td>Whether the website should show the login button.</td>
+            </tr>
+            <tr>
+                <td>`website.websiteConfig.enableSubmissionNavigationItem`</td>
```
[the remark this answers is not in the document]
fengelniederhammer: For GenSpectrum, we want to disable the navigation items, to that users don't see them. But we still want to be able to use the pages to inspect the ingest results.

docs/src/content/docs/reference/helm-chart-config.mdx
```diff
@@ -164,6 +164,77 @@ The configuration for the Helm chart is provided as a YAML file. It has the foll
...
+                <td>`website`</td>
+                <td>Object</td>
+                <td></td>
+                <td>Website specific setting</td>
+            </tr>
+            <tr>
+                <td>`website.websiteConfig`</td>
+                <td>Object</td>
+                <td></td>
+                <td>Settings for the `website_config.json`</td>
```
chaoran-chen, line 190: It seems a bit strange to have a field `websiteConfig` in the config under the key `website` as I'd expect everything under `website` in a config file to be "website config". What about calling it something like `features` as this it defines which features are activated and which not?

fengelniederhammer: I agree that it seems weird - but I still think it's a good approach:
* I tried to cluster the config values a bit (instead of having everything lop-level)
* `features` would also be a good, but a different slicing of the options. (slicing by features instead of by where the feature lives). If we go for that, I would propose to restructure other parts of the config in that direction as well.
  * we already have `seqSets.enabled`. That's why I decided for `website` (and I can imagine that one gets similar things for `backend` - targeting the backend only - and other feature groups such as `seqSets` and `auth`.
  * Also some features (like `seqSets`) might need additional configuration parameters, which makes the `features` section almost obsolete again.
* I named it `website.websiteConfig`, because there is also already `website.runtimeConfig`, to distinguish it from that. We could also argue that it's irrelevant for users in which config file those values go and all move them to `website`, but for us as developers I think it's helpful to keep it as an indicator.

kubernetes/loculus/values.yaml
fengelniederhammer: What do you mean exactly? The defaults of the individual values are `true`?

website/src/pages/organism-selector/[redirectTo].astro
```diff
@@ -1,31 +1,34 @@
...
 };
+if (getWebsiteConfig().enableSubmissionPages) {
+    purposes.submission = 'to access the submission portal';
+}
 
-interface Routes {
-    [key: string]: (organism: string) => string;
+if (!(Astro.params.redirectTo! in purposes)) {
+    return Astro.rewrite('/404');
 }
```
Comment (chaoran-chen) [inline website/src/pages/organism-selector/[redirectTo].astro:21]: Why do we need to redirect here? Wouldn't it be catched by the submissionPagesDisablingMiddleware?

chaoran-chen (approved):
Looks great, thanks! I just left one question.

website/src/pages/organism-selector/[redirectTo].astro
Comment (fengelniederhammer) [reply website/src/pages/organism-selector/[redirectTo].astro:21]: The middleware currently doesn't catch this page (as is separate from `[organism]/submission`). We could include this in the middleware but I found it cleaner to do it directly here. (No strong preference though)

PR #3232, opened 2024-11-18
feat(backend): Don't store null fields of processed data for perf, emit exactly current schema in get-released-data - imputing with null

### LINKED ISSUE #2793: Don't put fields with null values in processed data output
It seems we store a whole lot of nothing (=fields with null values) in the `sequence_entires_preprocessed_data.processed_data` column

We should switch to treating missing fields as having null and not saving them in the table. This would pretty much halve the size of that table.

Example processed data row:

`> LOC_000CV63 |       1 |                1 | {"metadata": {"length": 17, "authors": "J.A. Mcswiggen, L. Blatt", "hostAge": null, "lineage": null, "cellLine": null, "hostRole": null, "cultureId": null, "isLabHost": null, "totalSnps": null, "geoLocCity": null, "geoLocSite": null, "hostAgeBin": null, "hostGender": null, "sampleType": null, "stopCodons": null, "bodyProduct": null, "displayName": "LOC_000CV63.1", "foodProduct": null, "frameShifts": null, "hostDisease": null, "hostTaxonId": null, "ampliconSize": null, "completeness": null, "geoLocAdmin1": null, "geoLocAdmin2": null, "insdcVersion": 1, "ncbiSourceDb": "GenBank", "exposureEvent": null, "geoLocCountry": null, "ncbiVirusName": "West Nile virus", "passageMethod": null, "passageNumber": null, "travelHistory": null, "anatomicalPart": null, "geoLocLatitude": null, "hostNameCommon": null, "ncbiUpdateDate": "2009-05-23", "ncbiVirusTaxId": 11082, "sequencingDate": null, "versionComment": null, "dehostingMethod": null, "depthOfCoverage": null, "exposureDetails":
null, "exposureSetting": null, "geoLocLongitude": null, "hostHealthState": null, "ncbiReleaseDate": "2009-05-23", "sraRunAccession": null, "totalStopCodons": null, "collectionDevice": null, "collectionMethod": null, "signsAndSymptoms": null, "totalDeletedNucs": null, "totalFrameShifts": null, "totalUnknownNucs": null, "breadthOfCoverage": null, "environmentalSite": null, "hostHealthOutcome": null, "hostOriginCountry": null, "purposeOfSampling": null, "totalInsertedNucs": null, "anatomicalMaterial": null, "authorAffiliations": null, "biosampleAccession": null, "hostNameScientific": null, "insdcAccessionBase": "HA068283", "insdcAccessionFull": "HA068283.1", "sampleReceivedDate": null, "sequencingProtocol": null, "specimenProcessing": null, "totalAmbiguousNucs": null, "bioprojectAccession": null, "presamplingActivity": null, "purposeOfSequencing": null, "sequencingAssayType": null, "ncbiSubmitterCountry": null, "qualityControlIssues": null, "sampleCollectionDate": null, "sequencingInstrument": null, "environmentalMaterial": null, "foodProductProperties": null, "hostVaccinationStatus": null, "qualityControlDetails": null, "sequencedByContactName": null, "ampliconPcrPrimerScheme": null, "sequencedByContactEmail": null, "sequencedByOrganization": null, "diagnosticTargetGeneName": null, "diagnosticTargetPresence": null, "previousInfectionDisease": null, "qualityControlMethodName": null, "referenceGenomeAccession": null, "diagnosticMeasurementUnit": null, "previousInfectionOrganism": null, "specimenCollectorSampleId": null, "specimenProcessingDetails": null, "diagnosticMeasurementValue": null, "diagnosticMeasurementMethod": null, "qualityControlDetermination": null, "qualityControlMethodVersion": null, "experimentalSpecimenRoleType": null, "consensusSequenceSoftwareName": null, "rawSequenceDataProcessingMethod": null, "consensusSequenceSoftwareVersion": null}, "aminoAcidInsertions": {}, "nucleotideInsertions": {"main": []}, "alignedAminoAcidSequences": {"2K": null, "NS1": null, "NS3": null, "NS5": null, "env": null, "prM": null, "NS2A": null, "NS2B": null, "NS4A": null, "NS4B": null, "capsid": null}, "alignedNucleotideSequences": {"main": null}, "unalignedNucleotideSequences": {"main": {"compressedSequence": "KLUv/SARPQAAAAEAL5ghEA=="}}} | [{"source": [{"name": "main", "type": "NucleotideSequence"}], "message": "Nucleotide sequence failed to align"}] | []       | HAS_ERRORS        | 2024-09-15 16:34:20.523262 | 2024-09-15 16:34:20.703557`

corneliusroemer:
resolves #2793

preview URL: https://enforce-schema.loculus.org

### What does this PR solve?

This PR resolves 2 issues discussed among others in #2793 but recapitulated here:

1. We were wasting db storage and bandwidth by storing lots of metadata fields with null values in `sequence_entries_preprocessed_data.processed_data` in the db - in Pathoplexus this wastage might even be 50% or more of total row size, reaching a point that's worth optimizing.
2. Whenever we wanted to add a new metadata field, we needed to reprocess the data because the backend would rigidly send exactly what it had in `sequence_entries_preprocessed_data.processed_data` to the `get-released-data` consumer and SILO's preprocessing doesn't like missing (or extra as well, I think) fields in NDJSON. This created a lot of friction in deployment.

### What does this PR do?

This PR introduces the idea of metadata compression, in parallel to the existing sequence compression. This fits very nicely into the existing code base, making the actual code (non-test) changes trivial.

- Processed metadata is _compressed_ before being stored in the db by removing all fields that have `null` values: https://github.com/loculus-project/loculus/blob/67161496adcbbf7d093bda423007ab63ad93124d/backend/src/main/kotlin/org/loculus/backend/service/submission/CompressionService.kt#L121
- Processed metadata is _decompressed_ after being retrieved from the db by filling back null to all the fields that are missing. At the same time, fields that are no longer in the current metadata schema are dropped so as to allow us to remove no-longer-wanted metadata fields: https://github.com/loculus-project/loculus/blob/67161496adcbbf7d093bda423007ab63ad93124d/backend/src/main/kotlin/org/loculus/backend/service/submission/CompressionService.kt#L89-L96

### Aside: should external metadata always be a subset of metadata?

This is kind of a side discussion, only relevant because the decompression code needs to know what all the metadata fields are that should be filled with null.

As part of adapting tests to the new paradigm, I noticed that the backend config is a little unclear regarding external metadata:
- should external metadata always be part of metadata? (this is what we do in practice right now via the helm chart), i.e. should external metadata fields be a subset of metadata
- or should external metadata be allowed to have fields not in metadata? (this is what the backend test config was until now)

For me, it seems most logical to have metadata be the thing that the backend emits in get-released-data. Whether we allow externalMetadataSubmitter to override or not is a property of a metadata field (in my understanding). For now this difference doesn't matter other than in the way I resolved tests. In the backend tests I use my interpretation of enforcing externalMetadata to be a subset of metadata. We could in the future relax this but that would require test changes.

It might make more sense eventually though to get rid of the "externalMetadata" part of the config and add extra optional fields to the "metadata" fields listing whether or not they are allowed to be overridden by external metadata submitter and if so by which user. 

If we want to keep things lose like the tests were right now, we can adapt this PR, it'd just  have to then iterate over both metadata _and_ external metadata schemas to get the full set of expected fields. Also: how would we treat cases where field values are identical but types differ?

### Screenshot

No more nulls in processed data:

pgAdmin 4 2024-11-18 23 42 52

### Testing

I've written a comprehensive unit test for the metadata compression/decompression behavior.

It's a bit tricky to write endpoint test cases for this - the relevant behavior to test would be changes of backend config (before and after addition/removal of a metadata field) which I couldn't figure out how to test well. 

The changes made themselves seen in a couple of existing tests though. I checked thoroughly that we can now add and remove metadata fields on a persistent db and that things work as expected, e.g. https://github.com/loculus-project/loculus/pull/3232/commits/4e884b61d67bde6522ca6a75af304da0cab0d22f and https://github.com/loculus-project/loculus/pull/3232/commits/59833e14c5f6b3340a015c4370f6d2191a3eb590

### PR Checklist


backend/src/main/kotlin/org/loculus/backend/service/submission/CompressionService.kt
```diff
@@ -84,34 +85,40 @@ class CompressionService(private val backendConfig: BackendConfig) {
...
-                }
+    fun decompressProcessedData(processedData: ProcessedData<CompressedSequence>, organism: Organism) = ProcessedData(
+        backendConfig
+            .getInstanceConfig(organism)
+            .schema
+            .metadata
+            .map { it.name }
+            .associateWith { fieldName ->
+                processedData.metadata[fieldName] ?: NullNode.instance
             },
```
corneliusroemer, line 96: Only this part is new, the rest is just due to stupid formatting changes

backend/src/test/resources/backend_config.json
```diff
@@ -75,14 +75,21 @@
...
+                        "name": "other_db_accession",
+                        "type": "string"
                     }
                 ],
                 "externalMetadata": [
                     {
                         "name": "insdcAccessionFull",
                         "type": "string",
-                        "externalMetadataUpdater": "ena",
-                        "required": true
```
corneliusroemer, line 85: What's the point of required external metadata? is that requiredness used in the code anywhere? @anna-parker

backend/src/test/resources/backend_config.json
```diff
@@ -75,14 +75,21 @@
                     {
                         "name": "booleanColumn",
                         "type": "boolean"
+                    },
+                    {
+                        "name": "insdcAccessionFull",
+                        "type": "string"
+                    },
+                    {
+                        "name": "other_db_accession",
```
fengelniederhammer, line 84: Let's keep casing consistent

fengelniederhammer (commented):
How does this play together will data that is already in the database? If I interpret the current code correctly, then it will decompress all processed data. Will this work with uncompressed data in the database?

How much space does this actually save? Metadata isn't that large in terms of memory demand. I wonder whether the overhead (in code complexity and runtime) is worth the trouble?

backend/src/test/kotlin/org/loculus/backend/controller/submission/GetReleasedDataEndpointTest.kt
```diff
@@ -149,14 +149,8 @@ class GetReleasedDataEndpointTest(
                 "releasedDate" to TextNode(currentDate),
                 "submittedDate" to TextNode(currentDate),
                 "dataUseTermsRestrictedUntil" to NullNode.getInstance(),
-                "booleanColumn" to BooleanNode.TRUE,
             )
 
-            assertThat(
```
anna-parker, line 155: Why did you remove this test? It still passes for me locally in this branch

anna-parker:
Thanks @corneliusroemer this looks great to me! I just wanted to ask two follow ups - otherwise I'm happy to approve. 

1. Do understand correctly that only newly processed data will be compressed - unless we increase the preprocessing pipeline version and we will then process all data again and it will be compressed?

2. About the external metadata: Although I wrote external metadata to be potentially disjoint from the metadata now that I understand our code ecosystem better this seems like an unnecessary complication - which we anyways do not use. So I am fine with enforcing external metadata to always be a subset of the full metadata. However, with the current config structure this doesn't have to be the case and we don't check for this. I think we should change the config structure to as you suggested - e.g. as add an additional optional field to the metadata fields which specifies a field is external metadata - even better we could have the backend not expect two separate external metadata and metadata dictionaries but one metadata dictionary with this field. Does that sound alright with you?

backend/src/test/resources/backend_config.json
corneliusroemer: I don't disagree, but a) this predates me, was introduced by @anna-parker and b) is quite tangential to the PR as it's part of external metadata and the changes here are just to regularize test config with what we're doing in practice in Loculus/ppx

fengelniederhammer: Ah ok, I didn't see that it has to match with the external metadata below - but it's weird that one has to configure this field twice.

corneliusroemer: Yeah as I say in the PR description, I think the backend config can be improved. I'm not a big fan of having external metadata being separate from metadata.

We should probably discuss in a separate issue with @anna-parker

It's easy to iterate over both metadata and external metadata here but the question is what's the semantics of the backend config.

backend/src/test/kotlin/org/loculus/backend/controller/submission/GetReleasedDataEndpointTest.kt
corneliusroemer: It's in commit message of https://github.com/loculus-project/loculus/pull/3232/commits/c62496c05e4ee2f4f19717859ce46af5efd1fae3 :) 

Basically why check on size if you check for each field exactly. The latter is much more informative and has the same test strength. Hence removing the weaker, less informative test that cloaks the better information.

corneliusroemer:
Re @fengelniederhammer's comments:

> How does this play together will data that is already in the database? If I interpret the current code correctly, then it will decompress all processed data. Will this work with uncompressed data in the database?

Compression is not really what's happening here, so it seems to give the wrong idea. We're not compressing metadata, we're just not storing null fields. Everything is null unless it's present with non-null value. But there's no harm in having extra nulls in there, the "decompress" code doesn't care, it just adds nulls back when necessary.

As a result, all not previously compressed data already in the db is perfectly valid and can be "uncompressed" without problem or loss of anything.

> How much space does this actually save? Metadata isn't that large in terms of memory demand. 

Saving storage (and amount of data to be streamed by db which is just as much a bottleneck) isn't the main and only reason for this, but it's a nice extra at no cost - well the cost is this line which is negligible I hope you agree: https://github.com/loculus-project/loculus/blob/d78c8b868de59248f23a69a93cac1dd465d6b953/backend/src/main/kotlin/org/loculus/backend/service/submission/CompressionService.kt#L121

I find it hard to argue why we should keep those nulls around if it's so simple to get rid of them.

> I wonder whether the overhead (in code complexity and runtime) is worth the trouble?

If we want to be able to easily add metadata fields without having to reprocess then we need the "decompression" step, whether we keep nulls around or not. So that slight complexity is required: https://github.com/loculus-project/loculus/blob/67161496adcbbf7d093bda423007ab63ad93124d/backend/src/main/kotlin/org/loculus/backend/service/submission/CompressionService.kt#L89-L96

But are we really arguing over 4 lines as complexity - I'm actually quite proud of myself for the _lack_ of complexity of the solution (it's mostly because you've set things up so nicely with the compression service, so full credit to you @fengelniederhammer)? For runtime, I'm very sure that this is almost entirely negligible. And we agreed that we would do this in #2793 and related PRs (where you said that you were happy for the backend to always emit what the schema defines ;) )

corneliusroemer:
Re @anna-parker's comments:

> Do understand correctly that only newly processed data will be compressed - unless we increase the preprocessing pipeline version and we will then process all data again and it will be compressed?

That's correct - compression is really not the right word though for metadata. It's more dehydration, throwing out things that are not required (but also don't do harm). Compression is entirely optional, has slight benefits of saving space and is otherwise irrelevant. But it just takes one line of code, so why not.

> Even better we could have the backend not expect two separate external metadata and metadata dictionaries but one metadata dictionary with this field.

Excellent, that's exactly what I was thinking of! Make externality an optional extra of a metadata fields. We should probably do this in a separate PR - it should be pretty straightforward!

fengelniederhammer:
> Compression is not really what's happening here, so it seems to give the wrong idea.

Oh I see. I misread the code (just had a quick look over it). I think it would be cleaner if we did not hide metadata processing in "compress" and "decompress" functions. Let's split out that functionality to a dedicated place?

corneliusroemer:
Re how much storage is saved, running this script from ChatGPT

<details>

```
SELECT
    schemaname,
    tablename,
    pg_size_pretty(pg_total_relation_size(schemaname || '.' || tablename)) AS total_size,
    pg_size_pretty(pg_relation_size(schemaname || '.' || tablename)) AS table_size,
    pg_size_pretty(pg_indexes_size(schemaname || '.' || tablename)) AS indexes_size,
    pg_size_pretty(pg_total_relation_size(schemaname || '.' || tablename) - pg_relation_size(schemaname || '.' || tablename) - pg_indexes_size(schemaname || '.' || tablename)) AS toast_size
FROM
    pg_tables
WHERE
    schemaname NOT IN ('pg_catalog', 'information_schema')
ORDER BY
    pg_total_relation_size(schemaname || '.' || tablename) DESC;
```

</details>

on this branch vs main gives:

### This branch (compressed)

```
loculus-#     pg_total_relation_size(schemaname || '.' || tablename) DESC;
      schemaname       |             tablename              | total_size | table_size | indexes_size | toast_size
-----------------------+------------------------------------+------------+------------+--------------+------------
 public                | sequence_entries_preprocessed_data | 74 MB      | 28 MB      | 2512 kB      | 43 MB
 public                | sequence_entries                   | 57 MB      | 37 MB      | 2120 kB      | 17 MB
... (truncating small/irrelevant tables)
```

### main

```
      schemaname       |             tablename              | total_size | table_size | indexes_size | toast_size
-----------------------+------------------------------------+------------+------------+--------------+------------
 public                | sequence_entries_preprocessed_data | 104 MB     | 35 MB      | 3160 kB      | 66 MB
 public                | sequence_entries                   | 56 MB      | 37 MB      | 2144 kB      | 17 MB
... (truncating small/irrelevant tables)
```

So we're shaving off 30% in total size, 20% in table size and 30% in toast size for our largest table which is the preprocessed data. In an actual deployment that table is even more the largest component.

This is a cheap substantial perf win, do you agree now @fengelniederhammer :)

backend/src/test/kotlin/org/loculus/backend/controller/submission/GetReleasedDataEndpointTest.kt
anna-parker: Ah true! Yes then deleting it sounds good!

corneliusroemer:
>  I think it would be cleaner if we did not hide metadata processing in "compress" and "decompress" functions. Let's split out that functionality to a dedicated place?

I disagree, I think this is a great place, note I renamed the function to no longer mention sequences as we now compress sequences _and_ metadata. I agree that it might be confusing to you as when you wrote this you only compressed sequences, but that's just because you wrote the code.

Maybe we could rename the service to something more general than compression. I wouldn't move it somewhere else, that would just complicate logic - the fact the required changes are so small is a good sign that this is the right place to make those changes.

I first had a separate service, but that would then have to be called in a number of places per endpoint (the way I had set it up) - now this is much cleaner.

The current location is good because we want this hydration/dehydration (or compression or whatever) to happen in one place before things go to db and when they come out of db. And it's natural to do it in the same place that we compress sequences.

fengelniederhammer:
> @fengelniederhammer feel free to propose a split into something that makes sense for you, maybe in a PR that targets this PR here - I'm probably thinking of something different (and worse) than what you have in mind.

-> #3241

backend/src/test/resources/backend_config.json
anna-parker: You can change to camelCase in both places if you would like

backend/src/test/kotlin/org/loculus/backend/service/ProcessedMetadataPostprocessorTest.kt
```diff
@@ -0,0 +1,65 @@
...
+        assertThat(compressed.metadata, not(hasKey(unconfiguredNull)))
+        assertThat(compressed.metadata, hasKey(configuredPresent))
+        assertThat(compressed.metadata, hasKey(unconfiguredPresent))
+        assertEquals(compressed.metadata[configuredPresent], testData.metadata[configuredPresent])
+
+        // Check decompression behavior
+        assertEquals(decompressed.metadata[configuredPresent], testData.metadata[configuredPresent])
+        assertEquals(decompressed.metadata[configuredNull], testData.metadata[configuredNull])
+        assertThat(decompressed.metadata, not(hasKey(unconfiguredPresent)))
+        assertThat(decompressed.metadata, not(hasKey(unconfiguredNull)))
```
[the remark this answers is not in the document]
fengelniederhammer: I like the wording

backend/src/main/kotlin/org/loculus/backend/service/submission/ProcessedDataPostprocessor.kt
```diff
@@ -0,0 +1,19 @@
+package org.loculus.backend.service.submission
+
+import org.loculus.backend.api.Organism
+import org.loculus.backend.api.ProcessedData
+import org.springframework.stereotype.Service
+
+@Service
+class ProcessedDataPostprocessor(
```
chaoran-chen, line 8: Would something like `ProcessedDataCompressionService` be a better name? I think that's more specific and explains better that it doesn't really change the data but only compresses them.

chaoran-chen (approved):
Great, thank you!

backend/src/main/kotlin/org/loculus/backend/service/submission/ProcessedDataPostprocessor.kt
fengelniederhammer: I find`Compression` too misleading, because the metadata are not compressed. But in general I feel like this class could have a better name.

chaoran-chen: They are not compressed into a binary (and not compressed very strongly) but at least in a broader sense, they are compressed, aren't they? But I can see that the name is also not perfect as people often tend to associate "compression" with other algorithms.

fengelniederhammer: The term "compression" is taken in software. I'd avoid that. That's why I like @anna-parker's suggestion of using "condensed".

backend/src/test/kotlin/org/loculus/backend/service/ProcessedMetadataPostprocessorTest.kt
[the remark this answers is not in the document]
corneliusroemer: good for me!

PR #3246, opened 2024-11-19
fix(backend): Fix submitter of revoked sequences to be the revoker

### LINKED ISSUE #3244: Revoked sequences always have original submitter of the revoked version as the submitter, rather than the revoker
Just noticed that when I revoke an INSDC ingested sequence as a curator on staging ingest seems to auto approve the revocation. That's probably not what we want.

theosanderson:
Fixes #3244

https://theosanderson-fix-revoked.loculus.org/

Update the submitter of a revoked sequence to the revoker

* Modify `backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt` to set the `submitterColumn` to the revoker's account during the revocation process
* Change the `revoke` function to use the revoker's username for the `submitterColumn` instead of cloning the old submitter information

Tested that autoapprove no longer happens:

image

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/loculus-project/loculus/pull/3246?shareId=b9d63876-b575-4564-b2fb-3c100518daf2).

### DIFF
--- a/backend/src/test/kotlin/org/loculus/backend/controller/submission/RevokeEndpointTest.kt
+++ b/backend/src/test/kotlin/org/loculus/backend/controller/submission/RevokeEndpointTest.kt
@@ -8,7 +8,9 @@ import org.loculus.backend.controller.DEFAULT_ORGANISM
 import org.loculus.backend.controller.DEFAULT_USER_NAME
 import org.loculus.backend.controller.EndpointTest
 import org.loculus.backend.controller.OTHER_ORGANISM
+import org.loculus.backend.controller.SUPER_USER_NAME
 import org.loculus.backend.controller.assertStatusIs
+import org.loculus.backend.controller.assertSubmitterIs
 import org.loculus.backend.controller.expectUnauthorizedResponse
 import org.loculus.backend.controller.generateJwtFor
 import org.loculus.backend.controller.jwtForSuperUser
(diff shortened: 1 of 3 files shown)


corneliusroemer:
Thanks! Oof, how did this live so long without us noticing 😬

anna-parker (approved):
Thanks @theosanderson!

corneliusroemer:
I'm adding a test for this in #3248

backend/src/test/kotlin/org/loculus/backend/controller/submission/RevokeEndpointTest.kt
```diff
@@ -36,7 +38,7 @@ class RevokeEndpointTest(
     }
 
     @Test
-    fun `GIVEN entries with 'APPROVED_FOR_RELEASE' THEN returns revocation version in status AWAITING_APPROVAL`() {
+    fun `GIVEN entries with 'APPROVED_FOR_RELEASE' THEN returns revocation version in status PROCESSED`() {
```
corneliusroemer, line 41: This was something left over from @fhennig's refactor

corneliusroemer:
@fengelniederhammer maybe you have some thoughts on how to better test the created sequence entries - we've now got a bunch of helpers, that's not terrible but maybe there's a neater way you know. Also, we might want to test similar basic things explicitly for other parts of the submission flow.

fengelniederhammer:
> we've now got a bunch of helpers

The helpers already look pretty helpful to me. The tests are nice to read.

corneliusroemer:
@fengelniederhammer I just wasn't sure whether creating a helper for each property was a good way to go. Could come up with a better way (limited Kotlin knowledge).

PR #3247, opened 2024-11-19
chore(backend): fix some more racy backend tests by not asserting on exact order

### LINKED ISSUE #3242: Flaky backend test with wrong assumptions about ordering
Yet another test that expects exact ordering when we don't actually promise this (we could order in SQL but we don't):

```
ApproveProcessedDataEndpointTest > WHEN user approves with submitterNamesFilter THEN only approves entries from submitter in filter() FAILED
    java.lang.AssertionError: JSON path "$[*].accession" expected:<[LOC_000001Y, LOC_000002W, LOC_000003U, LOC_000004S, LOC_000005Q, LOC_000006N, LOC_000007L, LOC_000008J, LOC_000009G, LOC_00000AE]> but was:<["LOC_000006N","LOC_000004S","LOC_000007L","LOC_000003U","LOC_000009G","LOC_000002W","LOC_000008J","LOC_000005Q","LOC_00000AE","LOC_000001Y"]>
        at org.springframework.test.util.AssertionErrors.fail(AssertionErrors.java:59)
        at org.springframework.test.util.AssertionErrors.assertEquals(AssertionErrors.java:122)
        at org.springframework.test.util.JsonPathExpectationsHelper.assertValue(JsonPathExpectationsHelper.java:123)
        at org.springframework.test.web.servlet.result.JsonPathResultMatchers.lambda$value$2(JsonPathResultMatchers.java:111)
        at org.springframework.test.web.servlet.MockMvc$1.andExpect(MockMvc.java:214)
        at org.loculus.backend.controller.submission.ApproveProcessedDataEndpointTest.WHEN user approves with submitterNamesFilter THEN only approves entries from submitter in filter(ApproveProcessedDataEndpointTest.kt:383)
```

corneliusroemer:
resolves #3242 

This should avoid most of the flakes due to similar wrongly strong assumptions about SQL result order being stable without explicit sorting.

I can't find any more instances of `jsonPath("\$[*]`. 

But I should also look for `jsonPath("\$[0]` which I haven't done. Just noticed some tests like this in the revocation tests.

### DIFF
--- a/backend/src/test/kotlin/org/loculus/backend/controller/TestHelpers.kt
+++ b/backend/src/test/kotlin/org/loculus/backend/controller/TestHelpers.kt
@@ -9,6 +9,9 @@ import kotlinx.datetime.plus
 import kotlinx.datetime.toLocalDateTime
 import org.hamcrest.MatcherAssert.assertThat
 import org.hamcrest.Matchers
+import org.hamcrest.Matchers.allOf
+import org.hamcrest.Matchers.containsInAnyOrder
+import org.hamcrest.Matchers.hasEntry
 import org.hamcrest.Matchers.`is`
 import org.hamcrest.Matchers.not
 import org.loculus.backend.api.AccessionVersion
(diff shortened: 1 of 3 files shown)


fengelniederhammer (approved):
LGTM

PR #3252, opened 2024-11-19
fix(backend): delete processed entries as well when deleting unreleased sequence entries

### LINKED ISSUE #3250: When deleting unreleased sequence entries, the entry remains in processed data table blocking pipeline upgrades

corneliusroemer:
Fixes #3250

preview URL: https://delete-processed.loculus.org

Generated partially with help from Copilot, see #3251 

### Summary

When deleting entries from sequence_entries table, we also now delete from sequence_entries_preprocessed_data table.

### Followup

We should still add foreign key constraints - all sorts of race conditions are still possible until then, but this is better than nothing.

### Testing

I've manually tested that if I delete something from sequence_entries it also disappears from sequence_entries_preprocessed_data.

I've looked into whether we can easily test programmatically that rows have disappeared from processed data but I couldn't find good ways of doing this, we don't seem to be doing anything like that so far, or maybe I missed it. @fengelniederhammer any ideas on how to assert that after deletion there are no more corresponding entries in processed data table? I guess foreign key constraints are the best way to ensure, then almost no need to test at all.
### PR Checklist

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/service/submission/SequenceEntriesPreprocessedDataTable.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/service/submission/SequenceEntriesPreprocessedDataTable.kt
@@ -1,6 +1,7 @@
 package org.loculus.backend.service.submission
 
 import org.jetbrains.exposed.sql.SqlExpressionBuilder.eq
+import org.jetbrains.exposed.sql.SqlExpressionBuilder.inList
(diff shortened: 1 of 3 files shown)


chaoran-chen:
Thanks for identifying the bug! Should we maybe just add a foreign key now instead? Should be similar easy and I'm happy to create a PR.

corneliusroemer:
Superseded by #3259

Issue #3249, opened 2024-11-19
More useful log message when "Checking for newer processing pipeline versions"


corneliusroemer:
This might get addressed by @theosanderson's dashboard WIP, but in any case it'd be nice to have better information in the log other than that something was checked. We'd want to know:
- For each version how many are still to be processed
- For each version, how many released now fail processing (this would indicate this processing version can never succeed)
- possibly more

PR #3253, opened 2024-11-19
fix(backend): ignore deleted sequences in processing pipeline updates

### LINKED ISSUE #3250: When deleting unreleased sequence entries, the entry remains in processed data table blocking pipeline upgrades

corneliusroemer:
fixes #3250 independently of #3252, but that PR is worth it anyways as we don't want processed data lying around for deleted sequence entries.

preview URL: https://ignore-deleted.loculus.org

### Summary
Solve the issue of #3250 narrowly, where we noticed it. This is to make sure that even if #3252 doesn't delete 100% of entries and some slip through it has no effect on processing pipeline updates. This also means existing instances don't need to clean up processed data to avoid #3250.

It'd be nice if we could rewrite the single big SQL update statement into chunks so we can do some informative logging of intermediate results to help with debugging updates that are in process, see #3249 - but that's for another time.

Once we have foreign key constraints we can probably remove the inner join again.

### Testing

I tested deleting some sequences then updating prepro version and it worked (PR #3256)

Brave Browser 2024-11-19 23 08 59

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
@@ -1156,13 +1156,19 @@ private fun Transaction.findNewPreprocessingPipelineVersion(): Long? {
                     not exists( -- ...for which no sequence exists...
                         select
                         from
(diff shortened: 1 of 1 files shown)


corneliusroemer:
Superseded by #3259

PR #3255, opened 2024-11-19
feat(config): allow dev standin dbs to be semi-persistent through values.yaml

### LINKED ISSUE #3254: Allow standin database to be made persistent through values.yaml
For semi-permanent previews and tests, it's sufficient to use our standin db that's running inside the deployment and just not make it reload on config/version changes.

Right now this requires fairly deep changes in the helm chart, see: https://github.com/loculus-project/loculus/pull/2807

Would be nice if we simply had a flag that would allow a test/preview deployment to have a persistent db.

corneliusroemer:
resolves #3254

preview URL: https://persistence.loculus.org

Need to revert from true to false before merging.

### Summary

Allows not reloading the main and keycloak dbs for each commit in preview/dev deployments through values.yaml

Note that the database is 20+ minutes old while backend was replaced recently, that's what we can do now:
image

### DIFF
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -1529,9 +1529,8 @@ secrets:
       username: "dummy"
       password: "dummy"
 runDevelopmentKeycloakDatabase: true
-runDevelopmentKeycloakDatabaseWithPersistence: true
 runDevelopmentMainDatabase: true
-runDevelopmentMainDatabaseWithPersistence: true
+developmentDatabasePersistence: true
 enforceHTTPS: true
 registrationTermsMessage: >
   You must agree to the <a href="http://main.loculus.org/terms">terms of use</a>.
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -1530,6 +1530,7 @@ secrets:
       password: "dummy"
 runDevelopmentKeycloakDatabase: true
 runDevelopmentMainDatabase: true
+runDevelopmentMainDatabaseWithPersistence: false
 enforceHTTPS: true
 registrationTermsMessage: >
   You must agree to the <a href="http://main.loculus.org/terms">terms of use</a>.
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -1529,6 +1529,7 @@ secrets:
       username: "dummy"
       password: "dummy"
 runDevelopmentKeycloakDatabase: true
+runDevelopmentKeycloakDatabaseWithPersistence: true
 runDevelopmentMainDatabase: true
 runDevelopmentMainDatabaseWithPersistence: true
 enforceHTTPS: true
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -1530,7 +1530,7 @@ secrets:
       password: "dummy"
 runDevelopmentKeycloakDatabase: true
 runDevelopmentMainDatabase: true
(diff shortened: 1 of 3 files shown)


theosanderson:
(if we do this [which sounds good] we probs want to do it for KC DB too?)

corneliusroemer:
I was wondering how to name it, it seems a bit much to have booleans but then if we are happy with changing the config setting liberally, especially dev settings I'm happy not to worry too much.

Ideally, we'd back these dbs with a persistent volume claim but I don't know enough kubernetes to know how easy or hard that would be.

I thought about doing it for keycloak as well, just wanted to test it first for one.

corneliusroemer:
The need to do this now arose because I was surprised that virus2's db got replaced after an update, so having the persistence option there through values yaml would be really good.

kubernetes/loculus/values.yaml
```diff
@@ -1529,7 +1529,9 @@ secrets:
       username: "dummy"
       password: "dummy"
 runDevelopmentKeycloakDatabase: true
+runDevelopmentKeycloakDatabaseWithPersistence: true
```
corneliusroemer, line 1532:
```suggestion
runDevelopmentKeycloakDatabaseWithPersistence: false
```

kubernetes/loculus/values.yaml
```diff
@@ -1529,7 +1529,9 @@ secrets:
...
 runDevelopmentMainDatabase: true
+runDevelopmentMainDatabaseWithPersistence: true
```
corneliusroemer, line 1534:
```suggestion
runDevelopmentMainDatabaseWithPersistence: false
```

kubernetes/loculus/values.yaml
theosanderson, line 1532: IMO `run` is confusing naming here - we've used it above to mean turn on the database. I would guess that I would only need to set `runDevelopmentMainDatabaseWithPersistence` and not `runDevelopmentMainDatabase` (we I think intend to set `runDevelopmentMainDatabase`, etc. to default to `false` at some point).

theosanderson: For me the ideal would be `persistentDevDatabases` which would apply to both as I don't really see a reason one would want to have one persistent but not the other

theosanderson: or `devDatabasePersistence`

corneliusroemer: For development one will always want it to be cluster-internal and not externally managed?

Yeah for future would be better to make this enum, something like:

```
keycloakDatabase: 'ephemeral' | 'semi-persistent' | 'persistent' | 'external'
```

Feel free to change it to whatever, in my view we shouldn't waste too much thought on such dev-only features for now. Happy to rename anytime.

corneliusroemer: ok it's `developmentDatabasePersistence: true` for all

kubernetes/loculus/values.yaml
```diff
@@ -1530,6 +1530,7 @@ secrets:
       password: "dummy"
 runDevelopmentKeycloakDatabase: true
 runDevelopmentMainDatabase: true
+developmentDatabasePersistence: true
```
theosanderson, line 1533: back to false before merge?

theosanderson (approved):
Thanks

kubernetes/loculus/values.yaml
corneliusroemer, line 1533:
```suggestion
developmentDatabasePersistence: false
```

PR #3258, opened 2024-11-20
Delete-processed-tests

anna-parker:
resolves #

preview URL:

### Summary

Adds tests for https://github.com/loculus-project/loculus/pull/3252 to ensure we don't run into any pipeline update issues as in https://github.com/loculus-project/loculus/issues/3250 - I just verified that this test fails locally on main but now succeeds :-) 

Failure on main:
```
DeleteAllSequenceDataEndpointTest > GIVEN preprocessing pipeline version 1 WHEN some sequences deleted THEN can update pipeline to version 2() FAILED
    java.lang.AssertionError: An update to v2 should be possible
    Expected: is <2L>
         but: was null
```


corneliusroemer:
Nice! A good test of this test would be whether it fails on current main! You can just cherry pick and see that it fails. Maybe you've done that already but always good to test your test (expected fail)

corneliusroemer:
This is great, can be reused for #3259

PR #3259, opened 2024-11-20
fix(backend): add foreign key to delete preprocessed data

### LINKED ISSUE #3250: When deleting unreleased sequence entries, the entry remains in processed data table blocking pipeline upgrades

chaoran-chen:
resolves #3250, alternative to #3252

preview URL: https://delete-processed2.loculus.org

### Summary

This adds a foreign key to `sequence_entries_preprocessed_data` for `accession` and `version`.

### PR Checklist

### DIFF
--- a/backend/src/test/kotlin/org/loculus/backend/controller/EndpointTestExtension.kt
+++ b/backend/src/test/kotlin/org/loculus/backend/controller/EndpointTestExtension.kt
@@ -177,11 +177,11 @@ class EndpointTestExtension :
 private fun clearDatabaseStatement(): String = """
         truncate table $GROUPS_TABLE_NAME cascade;
         update $CURRENT_PROCESSING_PIPELINE_TABLE_NAME set version = 1, started_using_at = now();
-        truncate table $SEQUENCE_ENTRIES_TABLE_NAME;
-        truncate table $SEQUENCE_ENTRIES_PREPROCESSED_DATA_TABLE_NAME;
+        truncate table $SEQUENCE_ENTRIES_TABLE_NAME cascade;
+        truncate table $SEQUENCE_ENTRIES_PREPROCESSED_DATA_TABLE_NAME cascade;
         alter sequence $ACCESSION_SEQUENCE_NAME restart with 1;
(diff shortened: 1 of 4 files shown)


corneliusroemer:
This is definitely the right way to go! We should probably add a few more foreign keys eventually. But great start and should fix the bug once it passes!

theosanderson:
The problem might be that you just need to push an empty commit to this branch because the CI doesn't trigger images to build

theosanderson:
(I just revised which will trigger that, hope that's ok)

corneliusroemer:
I'll cherry pick this test from Anya: https://github.com/loculus-project/loculus/pull/3258

chaoran-chen:
I was referring to many tests failing – a bit surprising but haven't yet thought/look into them deeper

corneliusroemer:
Any progress @chaoran-chen? I have a suspicion that the tests might not properly respect the (new) foreign key constraint. So things are either not created or deleted now without that being expected.

Good reason to add those constraints sooner rather than later, now we might have to fix a lot of tests 😬 

@fengelniederhammer what are your thoughts as you set up most of the tests here?

chaoran-chen:
I fixed it, I just need get my GPG working again so that I can commit...

backend/src/test/kotlin/org/loculus/backend/controller/EndpointTestExtension.kt
corneliusroemer, line 1: Why do we need to add `cascade` to the clearDatabaseStatement? Should the `on update cascade on delete cascade` of the migration be enough?

Does this suggest we're missing something in the fkey definitition?

corneliusroemer (approved):
See #3278 for a way to avoid similar test confusion in the future. Feel free to merge that PR into this PR before merging this PR.

backend/src/test/kotlin/org/loculus/backend/controller/EndpointTestExtension.kt
chaoran-chen: The foreign key is for on delete, not on truncate (and there is no on truncate).

PR #3272, opened 2024-11-21
upgrade to Keycloakify 11

fhennig:
resolves #

preview URL:

### PR Checklist


fhennig:
In [5e0c509](https://github.com/loculus-project/loculus/pull/3272/commits/5e0c50911ac20a8e8db6f225f1c9360c70459e58) I'm applying the fix from @theosanderson from this PR: https://github.com/loculus-project/loculus/pull/3271

Weird that this just happens 'out-of-the-box' with the starter. Thanks again to @theosanderson for finding this!

keycloak/keycloakify/LICENSE
```diff
@@ -1,7 +1,5 @@
 MIT License
 
-This license applies specifically to this directory
```
theosanderson, line 3: (good to revert this :) )

fhennig: oh yes :D 

811b519e21a9dbf37648bbd0b0a51b065587e31d

fhennig:
It's working in the preview! 🥳

corneliusroemer:
After I found the root cause of the CI failure (wrong casing of kcContext that didn't show on macOS due to file system casing insensitivity) I wanted to report the issue to keycloakify but it turns out that the casing isn't in the starter: https://github.com/keycloakify/keycloakify-starter/blob/main/src/login/KcContext.ts

@fhennig how did you create the starter? Did you maybe not copy it over manually somehow rather than a complete removal over folder and replacement with current starter?

As a result, the fix is easier than Theo's which just happened to fix casing as well but the types extension was unnecessary.

fhennig:
What I did before was clone into `keycloakify-starter`, `mv keycloakify keycloakify9` and `mv keycloakify-starter keycloakify`. I have no idea why the file was lowercase :O Well spotted! I thought I actually checked that as well, oh well.

I pushed the change now! It's actually good, because there is `kc.gen.tsx` which gets regenerated if the theme name changes and it contains a reference to `KcContext`, so it's nice that the file will be generated with the correct name.

Wow I feel stupid now 🤦 😅

theosanderson:
The file was named wrongly on main by me in the old version – now that it occurs to me I think I have dim memories of that. Some original eslint thing was giving me warnings and I briefly tried to fix them and then gave up or something

corneliusroemer:
You weren't stupid at all @fhennig - you actually did everything right and macOS (and git) played a trick on us! This is just something we need to remember can happen 🤷

corneliusroemer:
I've now added the keycloak update to 26, might as well test it with that.

corneliusroemer:
It's super confusing that version 26 is more like version 21 than 22-25 here: https://github.com/keycloakify/keycloakify/issues/731

fhennig:
Still WIP: The "Update user profile" page that you get to after signup with ORCID doesn't have the tickbox yet.

fhennig:
I think some docs might be missing still, but I think the new theme is feature-complete now!

fhennig:
I tried registering with KC in the preview and get an email, but I think no email arrived.

theosanderson:
> I tried registering with KC in the preview and get an email, but I think no email arrived.

I just tried and seemed to see the same. I tried an unrelated preview (https://profile-get-seqs.loculus.org/) and registration was fine, so that seems like a problem

fhennig:
Alright so I just removed the e-mail theme and then I got an email when registering.

fhennig:
Maybe I shouldn't have deleted the `theme.properties` 😆  I'll see if that maybe was the issue.

theosanderson:
I changed something (elsewhere) so the ORCiD flows now work - and they lgtm

fhennig:
Found what I did wrong, I'll have a fix in a bit.

theosanderson (approved):
Thanks this looks nice! It's the sort of thing that is pretty hard to review confidently for code, but I've tested the functionality in various ways and things seem good!

fhennig:
great! I have also updated the kc26 branch here: https://github.com/loculus-project/loculus/pull/3299 and it also looked good there! 🥳

corneliusroemer (changes requested):

PR #3275, opened 2024-11-22
Switch keycloakify PR to use Keycloak 26

corneliusroemer:
@fhennig this uses your keycloakify v11 theme with keycloak v26. The few lines of code do very much underestimate the amount of blood and sweat that have gone into getting this to work 😆 

So now you can test the theme with the Keycloak we want to get to asap!

https://kc26-on-kcify.loculus.org


fhennig:
too many force pushes, closing in favor of #3299

PR #3278, opened 2024-11-23
test(backend): throw when database clearing fails in tests

corneliusroemer:
This would have made it immediately clear what the reason for the test failures in https://github.com/loculus-project/loculus/pull/3259 was. So good to have for future.

Also truncate in one go for simplicity and maybe also performance.


chaoran-chen (approved):
Thanks!

Issue #3267, opened 2024-11-21
Backend log message is wrong, it doesn't mention all the extra filters that have been added to endpoint


corneliusroemer:
Going through backend logs I noticed that this line is completely missing all the extra filters that have been added since. I wish we logged a little more explicitly, shouldn't be too hard to ajust add all the filters here and do likewise with other endpoints so we understand better what's going on in logs:

https://github.com/loculus-project/loculus/blob/445f81aae75c5511fe19a6610f195bb27acfa437/backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt#L455

PR #3279, opened 2024-11-23
fix(backend,website): calculate status counts efficiently & request only aggregates for submission portal

### LINKED ISSUE #1612: Review page (and `/get-sequences` endpoint) is slow

### LINKED ISSUE #2880: submission page freezes when too many sequences waiting for approval
As a curator when I log in to submit I am first logged in in the INSDC ingest user group, the page freezes before it loads, probably because there are so many sequences: 
image

### LINKED ISSUE #3095: Review page: total number of sequences is wrong
I uploaded 50,000 sequences and observed a strange behavior of the total number of sequences:

https://github.com/user-attachments/assets/665f23f1-44fc-4c54-ba89-e46055527615

I believe to have seen this before: is there already a ticket?
(I guess it's a website bug but it could also be a backend bug.)

### LINKED ISSUE #3269: Submission portal takes multiple seconds to load if there are 10k sequences to review
We no longer have OOM (at least for 10k sequences) as mentioned here: https://github.com/loculus-project/loculus/issues/2881#issue-2550791971 but load times are still unacceptably long if there are thousands of sequences to review/approve (e.g. curators).

West Nile shows blank screen to superusers for 10+ seconds.

This wasn't fixed in #3031 

The issue isn't that there are many request made. It's that the website server seems to block, waiting for stuff.

Maybe mitigating the stall is as simple as disabling SSR of this page?

It's particularly the `/[organism]/submission/[groupid]` page that's slow. It's blocking on getSequences on the server side. 

There are a number of options here for speed up:
- We could refactor it to get the count of sequences on the client side
- Speed up getSequences
- ...

Relevant code is here:

https://github.com/loculus-project/loculus/blob/531c2bc3055ab344ad11035dfd93c20709692453/website/src/pages/%5Borganism%5D/submission/%5BgroupId%5D/index.astro#L19-L44

corneliusroemer:
resolves #3095, resolves #1612, resolves #3269, resolves #2880

helps with #3267

preview URL: https://profile-get-seqs.loculus.org

### Summary

@fhennig correctly identified that the way we calculated status counts in getSequences was the culprit behind the inconsistent total counts. 

We issued 4 independent db queries one after the other. In between statuses changed. What made it worse was that each of the calls was hugely inefficient, using a join on data use terms, ordering by accession column and much more that is totally unnecessary for the counts calculation.

The fluctuation could be fixed by higher transaction isolation but this is much much faster and efficient so should be enough (I've profiled with simple logs and tested).

Also fixes fact that log message misses mention of some filters.

### Future work

There's still lots to be done but this goes some way _and_ fixes the inconsistency bug.

We should not have this single endpoint that calculates much more than necessary. It's behind the review page slowness #3269: we really just want total numbers of sequences to review but the backend needs to send loads of unnecessary information to the website simply because we have no simpler endpoint.

Instead, we should spin up a few simple endpoints that do what the website needs - let's mark them as "non-public" in the backend API docs so that we don't need stability guarantees, they should be used solely by the website.

### Screenshot

Reload the page and look at both screencasts together to see speedup.

This PR:

Before (main):

### PR Checklist

### DIFF
--- a/website/src/pages/[organism]/submission/[groupId]/index.astro
+++ b/website/src/pages/[organism]/submission/[groupId]/index.astro
@@ -1,5 +1,4 @@
 ---
-import { cleanOrganism } from '../../../../components/Navigation/cleanOrganism';
 import SubmissionPageWrapper from '../../../../components/Submission/SubmissionPageWrapper.astro';
 import { routes } from '../../../../routes/routes';
 import { BackendClient } from '../../../../services/backendClient';
@@ -12,9 +11,9 @@ import IcOutlineUpload from '~icons/ic/outline-upload';
 import MdiViewListOutline from '~icons/mdi/view-list-outline';
 
 const session = Astro.locals.session!;
+const organism = Astro.params.organism!;
 const accessToken = getAccessToken(session)!;
 
-const { organism } = cleanOrganism(Astro.params.organism);
 const groupsResult = await getGroupsAndCurrentGroup(Astro.params, Astro.locals.session);
 
 async function getSequenceCounts(organism: string, groupId: number) {
@@ -25,6 +24,8 @@ async function getSequenceCounts(organism: string, groupId: number) {
             headers: createAuthorizationHeader(accessToken),
             queries: {
                 groupIdsFilter: groupId.toString(),
+                page: 0,
+                size: 0,
             },
         });
 
@@ -48,25 +49,25 @@ async function getSequenceCounts(organism: string, groupId: number) {
     {
         groupsResult.match(
             async ({ currentGroup: group }) => {
-                const { othersTotal, approvedTotal } = await getSequenceCounts(organism!.key, group.groupId);
+                const { othersTotal, approvedTotal } = await getSequenceCounts(organism, group.groupId);
 
                 const options = [
                     {
                         title: 'Submit',
                         description: 'Upload new sequences.',
-                        route: routes.submitPage(organism!.key, group.groupId),
+                        route: routes.submitPage(organism, group.groupId),
                         icon: IcOutlineUpload,
                     },
                     {
                         title: 'Revise',
                         description: 'Upload revisions for existing sequences.',
-                        route: routes.revisePage(organism!.key, group.groupId),
+                        route: routes.revisePage(organism, group.groupId),
                         icon: F7Arrow2Circlepath,
                     },
                     {
                         title: 'Review',
                         description: "Review your group's unreleased submissions.",
-                        route: routes.userSequenceReviewPage(organism!.key, group.groupId),
+                        route: routes.userSequenceReviewPage(organism, group.groupId),
                         icon: GgCheckO,
                         count: othersTotal,
                         countClass: 'text-primary-600 font-semibold',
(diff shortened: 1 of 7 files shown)


chaoran-chen (dismissed):

corneliusroemer:
@chaoran-chen don't approve too early, I'm not done yet. I've figured out how to solve all review page speed issues with a little extra hack 😁

corneliusroemer:
Now it's really ready @chaoran-chen

fhennig:
Hey, I see it's not done yet, but great work already!

corneliusroemer:
@fhennig it has been done for a few hours! Perf and inconsistency fixed completely, better than I could have imagined!

website/src/pages/[organism]/submission/[groupId]/index.astro
```diff
@@ -1,5 +1,4 @@
 ---
-import { cleanOrganism } from '../../../../components/Navigation/cleanOrganism';
```
Comment (fhennig) [inline website/src/pages/[organism]/submission/[groupId]/index.astro:2]: Why is the `cleanOrganism` function not needed anymore?

Comment (corneliusroemer) [reply website/src/pages/[organism]/submission/[groupId]/index.astro:2]: It was never needed here and just complicated accessing the organism name, we can get it simply from the organism path.

Comment (corneliusroemer) [reply website/src/pages/[organism]/submission/[groupId]/index.astro:2]: It doesn't seem to be doing anything useful, we don't use anything but organism!.key later on, so let's cut out the indirection

Comment (fhennig) [reply website/src/pages/[organism]/submission/[groupId]/index.astro:2]: If the user puts a garbage organism in the path, they'll get a 500 error. The fix seems to be to properly guard against this, like most other pages in the codebase do:

```ts
const organism = Astro.params.organism!;

if (!cleanedOrganism) {
    return {
        statusCode: 404,
        body: 'Organism not found',
    };
}
```

Comment (fhennig) [reply website/src/pages/[organism]/submission/[groupId]/index.astro:2]: Maybe Astro middleware could be used to do this validation in a more DRY way.

https://docs.astro.build/en/guides/middleware

Comment (corneliusroemer) [reply website/src/pages/[organism]/submission/[groupId]/index.astro:2]: I don't see a 500, where do you?
image

Comment (fhennig) [reply website/src/pages/[organism]/submission/[groupId]/index.astro:2]: I didn't actually check, sorry. That was just my understanding from ChatGPT and the astro docs 😅 

In that case please don't do an incomplete change like this: remove _all_ usages of the function as well as the function itself!

anna-parker (approved):
LGTM!

PR #3308, opened 2024-11-29
feat(website): Adapt collectionDate search to work with date ranges

### LINKED ISSUE #3285: Implement date search that works intuitively with new date range fields
We will soon no longer have a single date field: there will be two dates for each sample instead: lower range and upper range of what the date string we got means. For complete dates, the 2 are the same. But for incomplete they aren't.

This makes search non-intuitive to users. Most users will initially not care about the concept of date ranges, they just want something that works for filtering dates somehow.

What we get instead after #3263 is two dates to filter by, and this causes cognitive strain (even to me having thought about this feature for weeks).

image

There's a simple way of having intuitive search: show just one date search bar with _two_ modes (selectable through radio button):
- strict (sample date range must be entirely contained inside the query range)
- lax (sample date range must at least overlap the query range)

People probably want strict to be the default, but we can change this once we know what's more intuitive. The idea to have strict and lax search is something that I haven't seen much on other websites so even this way will take some getting used to (the good thing is that people won't need to know about strict vs lax, it'll just intuitively work ok).

Implementation:
1. Create a new custom search field type `date_range` (exact names tbd, just for scoping I choose some)
2. It takes as config two config values: `date_lower` and `upper`, to refer to the metadata fields that should be used
3. It displays a normal date search but with _added_ radio buttons for "strict" and "lax" modes, defaults to strict. For what follows I call the fields `range_min` and `range_max`
4. The LAPIS query is constructed based on mode as follows in pseudocde
```
if strict:
    filter((date_lower >= range_min) and (date_upper <= range_max))
if lax:
    filter((date_lower <= range_min) and (date_upper >= range_max))
```

That's it, I hope that's clear and should be doable. Let me know if you have questions.

fhennig:
resolves #3285 

preview URL: https://date-ranges-ui.loculus.org

### Summary

Add config bit to allow configuring the `rangeOverlapSearch` on a property (or rather two properties belonging together). Add a new `DateRangeField` search field type for these kinds of date ranges.

This is the first time we don't have a 1:1 mapping of search fields and API parameters anymore, so some bits in the code are a bit ... complicated maybe? But so far this way of doing it still seems fine. A "proper" solution would (IMO) involve decoupling the metadata field definition and the search field definitions, but seems kind of heavy handed at the moment.

### Screenshot
I wanted to maybe use a toggle, but ran into [this UI bug](https://github.com/saadeghi/daisyui/discussions/2451) (clash with tailwind forms).

image

#### with tooltip

image

### PR Checklist


website/src/utils/search.ts
```diff
@@ -63,7 +63,13 @@ const getFieldOrColumnVisibilitiesFromQuery = (
...
             return;
         }
-        visibilities.set(field.name, initiallyVisibleAccessor(field) === true);
+
+        let fieldName = field.name;
+
+        if (field.rangeOverlapSearch) {
+            fieldName = field.rangeOverlapSearch.rangeName;
+        }
+        visibilities.set(fieldName, initiallyVisibleAccessor(field) === true);
```
fhennig, line 72: The modifications in this function could be avoided if we force users to uses a `field.name` as the `rangeName`. This can't really be enforced. IMO this can't be avoided because we have to group fields two distinct fields.

Happy for input though if anybody happens to look at the PR already!

fhennig, line 72: The modifications in this function could be avoided if we force users to uses a `field.name` as the `rangeName`. This can't really be enforced. IMO this can't be avoided because we have to group fields two distinct fields.

Happy for input though if anybody happens to look at the PR already!

theosanderson:
The preview doesn't seem to be showing this UI for me atm?

fhennig:
> The preview doesn't seem to be showing this UI for me atm?

Yes I noticed as well, was a bit confused, it was due to the rangeSearch being enabled because of the `date` type and it taking precedence in the code. I think The last commit will fix it.

website/src/components/SearchPage/fields/DateRangeField.tsx
```diff
@@ -0,0 +1,134 @@
...
+                        checked={strictMode}
+                        onChange={(event) => setStrictMode(event.target.checked)}
+                    />
+                </label>
+            </div>
+
+            <DateField
+                field={{
+                    name: 'not used',
+                    label: 'Lower',
```
theosanderson, line 117: IMO `From` and `To` is significantly more understandable

theosanderson: (to be clear, for code the current is fine, and arguably better given we use `From` for the individual LAPIS query - just talking in terms of UI)

fhennig: https://github.com/loculus-project/loculus/pull/3308/commits/1daa04f01bd7f9b8f2db084d6e18a4204a85accf

theosanderson:
(looks really nice!  will try to dive in more soon!)

website/src/components/SearchPage/fields/DateRangeField.tsx
```diff
@@ -0,0 +1,134 @@
...
+            <div className='flex flex-row justify-between items-baseline mb-2'>
+                <h3 className='text-gray-500 text-sm'>{field.displayName}</h3>
+                <CustomTooltip id={'strict-tooltip' + field.name}>
+                    <div className='w-52'>{TOOLTIP_TEXT}</div>
+                </CustomTooltip>
+                <label data-tooltip-id={'strict-tooltip' + field.name}>
+                    <span className='text-gray-400 text-sm mr-2'>strict</span>
+                    <input
+                        type='checkbox'
+                        className='checkbox checkbox-sm [--chkbg:theme(colors.gray.400)]'
```
theosanderson, line 107: > the checkbox looks a bit like it's disabled, but with more solid colors it draws a lot of attention. But making it smaller would also not be great for clicking it I think. 🤔 Haven't come up with a good solution for this yet. (but maybe not critical, I don't think it looks terrible)

I don't think it looks terrible and I don't think we need to block, but I imagine it could be better with white bg and gray tick rather than the reverse

website/src/utils/search.ts
theosanderson: I think we need some more changes in this general area (appreciate the PR is still a draft): atm Collection date shows up twice on the list of fields to change the visibility of. I'd need to think about exactly the best way to handle this.

fhennig: I think it's just there twice because of the second _Collection date_ `string` field which is just above the range field.

I had a vague feeling that we shouldn't look for Metadata field visiblity, but rather MetadataFilter visibility! But also haven't looked into it more yet. But good to know that you think this area could use some more work! I'll spend some more time on it then (probably monday).

website/src/components/SearchPage/fields/DateRangeField.tsx
fhennig: That's a great idea! I implemented it in https://github.com/loculus-project/loculus/pull/3308/commits/5d5ea4f354871414fc4b513fd1f29957e46c6e2c (also updated the screenshots)

website/src/utils/search.ts
theosanderson: > I think it's just there twice because of the second Collection date string field which is just above the range field.

OK! That may well be true - my bad - and in that case it may well not need more work - sorry for my misunderstanding!

fhennig:
Next I think I want to add a test for new new component

corneliusroemer:
Looks great! 

For test, did you think of a component test? For E2E we'd need to modify test data, which might be possible as it's only an addition we need here, but we could also wait for Theo's more realistic new test setup.

corneliusroemer:
@fhennig sorry for force pushing the base branch, this causes merge conflicts, I'll try to resolve them - note that the last two pushes had no CI running due to the conflicts: 

image

corneliusroemer:
The rebase was easy, but there was another conflict due to reordering of values.yaml, I hope I did it correctly, if there are issues beware of this potentially being the source @fhennig

kubernetes/loculus/values.yaml
```diff
@@ -101,7 +91,24 @@ defaultOrganismConfig: &defaultOrganismConfig
             releaseDate: ncbiReleaseDate
           args:
             fieldType: dateRangeUpper
+        rangeOverlapSearch:
+          rangeName: sampleCollectionDateRange
+          rangeDisplayName: Collection date
+          bound: upper
         noInput: true
+      - name: sampleCollectionDate
+        displayName: Collection date (text)
```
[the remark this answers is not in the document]
fhennig: Is the displayName also used as the column title?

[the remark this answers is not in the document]
theosanderson: So we can add `notSearchable: true` for search.

In terms of columns it looks like we have a bug here in that there is a `Collection date` in the column selector which when ticked doesn't make a column appear. (For understandable reasons!)

corneliusroemer:
Did another merge conflict resolution post-merge of the base, this should have been the last resolution we need.

kubernetes/loculus/values.yaml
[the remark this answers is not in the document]
fhennig: > In terms of columns it looks like we have a bug here in that there is a Collection date in the column selector which when ticked doesn't make a column appear

Should we just tolerate this and make an issue for it and try and adress soonish? Besides this I think the solution is good.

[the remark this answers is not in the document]
fhennig: I removed the "(text)" and added the "notSearchable" bit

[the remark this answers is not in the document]
theosanderson: (i.e. we can remove it specifically by name/id or whatever in a totally hardcoded way)

corneliusroemer (changes requested):
I _think_ the current implementation doesn't allow going from URL to search state.

When I uncheck the strict box and have a date in the bar, copy to new browser tab, I get a different result.

It seems that the strictness checkbox deselection state isn't correctly deduced from the presence of `UpperFrom` (which indicates lax search) https://date-ranges-ui.loculus.org/west-nile/search?sampleCollectionDateRangeUpperFrom=2024-07-09

Instead of checkbox deselected and date 2024-07-09 being in "from" part of the date range picker, I get strict being selected and no date being entered:

image

fhennig:
@corneliusroemer Ah thanks for checking that! I'll fix it. I somehow completely forgot that you don't need to always give both sides of the range. If you give both sides of the range, it should correclty infer it from the URL. I know where to make the changes, I'll go ahead and do so!

kubernetes/loculus/values.yaml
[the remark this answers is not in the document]
fhennig: hacky fix: https://github.com/loculus-project/loculus/pull/3308/commits/4cc6f4a59f225b7aaf55e3a406900d43b543805d

issue: https://github.com/loculus-project/loculus/issues/3325

`notSearchable` didn't actually work, I fixed it here: https://github.com/loculus-project/loculus/pull/3308/commits/2cca86efecc3cba09fb0e0dffa64ea00ca1ee00c - not sure if this might have other consequences we didn't yet think about?

theosanderson (approved):
LGTM - really great work, thanks

corneliusroemer:
This still doesn't work: https://date-ranges-ui.loculus.org/west-nile/search?sampleCollectionDateRangeUpperFrom=2024-10-14 - the URL params get wiped and UI is without

fhennig:
Ah yes I see it too for the not-strict case. I'll get to it tomorrow

kubernetes/loculus/templates/_common-metadata.tpl
```diff
@@ -202,6 +202,9 @@ organisms:
   {{- if or (or (eq .type "timestamp") (eq .type "date")) .rangeSearch }}
   rangeSearch: true
   {{- end }}
+  {{- if .rangeOverlapSearch }}
+  rangeOverlapSearch: {{ .rangeOverlapSearch | toJson }}
+  {{- end}}
```
chaoran-chen, line 207: Could you please document the new config options in `/docs`?

website/src/components/SearchPage/fields/DateRangeField.tsx
```diff
@@ -0,0 +1,131 @@
...
+                    <div className='w-52'>
+                        <p>
+                            <span className='font-bold'>strict: </span>
+                            {field.displayName} range must be <span className='italic'>entirely </span>
+                            inside of the search range.
+                        </p>
+                        <p>
+                            <span className='font-bold'>not strict: </span>
+                            {field.displayName} range must have <span className='italic'>some overlap </span>
+                            with the search range.
```
chaoran-chen, line 96: I think that it would be great to have some description about this in the [search sequences documentation](https://loculus.org/for-users/search-sequences/). The tooltips are already really nice but on the documentation page, we can a bit more precise and maybe have an example?

chaoran-chen (approved):
Thanks very much, Felix! I didn't look at the code but only tested it and it looks really nice! I just have two reminders to add some documentation.

fhennig:
Wow getting my image to work in starlight is a pain. It won't take the SVG and it's scaling the PNG to look blurry.

anna-parker:
@fhennig not sure what image you mean but https://pictogrammers.com/library/mdi/ has a ton of great images that can be easily resized and are free for use.

fhennig:
@anna-parker Ah I meant the diagram i added to the docs

website/src/components/SearchPage/fields/DateRangeField.tsx
fhennig: https://docs-date-ranges-ui.loculus.org/for-users/search-sequences/

(Image not loading yet, maybe just not deployed yet?)

chaoran-chen: Nice! Looks amazing!

corneliusroemer:
Thanks for fixing the bug, indeed it works now!

Something I noticed that doesn't need fixing: with lax mode we show results even if the range is negative with starting date _after_ ending date. That's because we use 2 dates here rather than a single range picker - which is good because a range picker would always require start+end when we often just want one of the two

Slack 2024-12-03 17 15 43

website/src/types/config.ts
```diff
@@ -27,6 +27,19 @@ export const customDisplay = z.object({
     displayGroup: z.string().optional(),
 });
 
+/**
+ * RangeOverlapSearch to configure on two fields that together allow to query
+ * For an overlap of a search range and a targer range
+ */
+export const rangeOverlapSearch = z.object({
+    /**
+     * specify rangeName in both (upper and lower) fields to link them later.
```
corneliusroemer, line 36: Still not a fan of using 3 lines for comments when 1 would do

corneliusroemer (approved):
I'm not sure I'm convinced it's good to use "non strict" in docs as opposed to _lax_, don't think this was discussed much but not that important so I'll let go

fhennig:
Sorry about that! I had it in mind in the beginning but then lost track of it.

Thanks for approving 🙏

theosanderson:
> I'm not sure I'm convinced it's good to use "non strict" in docs as opposed to lax, don't think this was discussed much but not that important so I'll let go

I support the current state in the sense that the UI doesn't show the word "lax" so having it in the tooltip would IMO be confusing about what it would refer to

website/src/types/config.ts
fhennig: I can try and remember, would be nice to update the linter/formatter if possible with this.

fhennig:
Yeah, initially I tried using a toggle with _strict_ and _lax_ on either side of it, but the toggle component was buggy so I dropped it. I linked to the bug report somewhere, it was a conflict somehow with the tailwind forms.

I decided it wasn't critical and moved on.

PR #3322, opened 2024-12-02
feat: Change data use terms in bulk

### LINKED ISSUE #992: Change data use term in bulk
As a sequence owner, I want to be able to change the data use terms of my sequences on the website.

We thought about implementing it on the user sequences page:
grafik

Two options came to our minds:
* add a button to each sequence line to change the license of this entry
* add a button on the top right (next to "Revoke") to change the license of all selected entries.

The backend endpoint accepts a list of accessions.

fhennig:
resolves #992 

preview URL: http://datausetermsbulkchange.loculus.org

### Summary

Work in progress!

### PR Checklist


fhennig:
I have my function in the LapisClient ready to fetch the data, just looking into how to call it

fhennig:
I managed to fetch data from the LAPIS API!

Now I'm wondering whether it made sense to refactor the filter bits so I can use them in here too ... I'm not sure what the hiddenValues are for exactly.

Anyways, off to lunch for now.

fhennig:
Ok so here's my plan for how to implement this: 

Fetch info like this from LAPIS:

```
{"accession":"LOC_00002LR","dataUseTerms":"RESTRICTED","dataUseTermsRestrictedUntil":"2025-06-03"},{"accession":"LOC_00001QJ","dataUseTerms":"RESTRICTED","dataUseTermsRestrictedUntil":"2025-06-03"},{"accession":"LOC_000028G","dataUseTerms":"RESTRICTED","dataUseTermsRestrictedUntil":"2025-06-03"},
```

And then in the component find the date range that we can allow the user to set a new date in, and then update to an endpoint in the backend.

I can see a potential performance issue here when querying for a couple thousand accessions, but:
- I think LAPIS doesn't support doing the calculation of finding the range in its query endpoints
- We can't use the backend because it doesn't have the same filtering capabilities as LAPIS

Given that the performance issues are just speculation and only for large numbers of accessions, I'm going to go ahead with my 'naive' approach for now. Just wanted to outline it here in case someone already wants to interject

fhennig:
I couldn't replicate the E2E test failure locally. We should really do something about this flakiness.

fhennig:
**Status Update:** Everything is wired up the way it should be in the end.

Next steps:
- Test whether the Update actually works (doesn't seem so in the preview? But I didn't get an error)
- Make the UI nicer
- Restrict the edit stuff to just the "Submit > View" page.

fhennig:
related: https://github.com/loculus-project/loculus/issues/3354

corneliusroemer:
This is the test that failed:

```
1 failed
    [chromium] › pages/search/index.spec.ts:61:5 › The search page › should search many sequence entries by accession 
  1 flaky

  1) [chromium] › pages/search/index.spec.ts:61:5 › The search page › should search many sequence entries by accession 

    Error: expect(received).toBeTruthy()

    Received: false

      74 |         expect(newAccessions.length).toBe(3);
      75 |         expect(newAccessions.includes(previousAccessions[0])).toBeTruthy();
    > 76 |         expect(newAccessions.includes(previousAccessions[1])).toBeTruthy();
         |                                                               ^
      77 |         expect(newAccessions.includes(previousAccessions[2])).toBeTruthy();
      78 |     });
      79 |
        at /home/runner/work/loculus/loculus/website/tests/pages/search/index.spec.ts:76:63
```

website/src/components/DataUseTerms/EditDataUseTermsModal.tsx
theosanderson, line 1: For now reviewing the functionality (really great!):

I find it confusing to select some restricted sequences and hit the edit button and get:

image

Despite the left hand side (which isn't super emphasised) my intuition is that when you hit "edit" you will be presented with the current state and will then change it to new state. But the radio button here doesn't reflect the current state. I appreciate that it's impossible for it to, because one could have a mixture. But IMO if we use radio buttons, they should start off with neither option selected. And IMO the UI should say above the choices "Choose the new data use terms for these sequences" or similar, to emphasise this.

website/src/components/DataUseTerms/EditDataUseTermsModal.tsx
```diff
@@ -0,0 +1,287 @@
...
+            );
+        case 'mixed':
+            return (
+                <div className='space-y-4'>
+                    <ActiveFilters sequenceFilter={sequenceFilter} />
+                    <p>
+                        {state.openCount} open and {state.restrictedCount} restricted sequences selected.
+                    </p>
+                    <p>
+                        You can release all the {state.restrictedCount} restricted sequences as open. If you want to
```
theosanderson, line 179:
```suggestion
                        You can release all the {state.restrictedCount} restricted sequences, moving them to the Open Data Use Terms. If you want to
```

theosanderson: We should also make these descriptions of terms links that open in a new window to the terms (which are available in the Values.yaml, unsure if currently passed to the website)

website/src/components/DataUseTerms/EditDataUseTermsModal.tsx
fhennig: > when you hit "edit" you will be presented with the current state and will then change it to new state

I agree with this!

I think we should stick with the radio buttons because we use the same UI in the submission, and I think having the concistency is helpful.

> if we use radio buttons, they should start off with neither option selected. And IMO the UI should say above the choices "Choose the new data use terms for these sequences" or similar, to emphasise this.

I like that! 

I think I will ditch the 2 column layout, I thought it would be nice and save some space, but now I think it's acutally not that great. I think I still have an idea how to save space so we don'thave to scroll.

fhennig: https://github.com/loculus-project/loculus/pull/3322/commits/8fa8dd1312c13a0400724f334f3a09aa5792b7f4

screenshot updated too.

website/src/components/DataUseTerms/EditDataUseTermsModal.tsx
fhennig: https://github.com/loculus-project/loculus/pull/3322/commits/4f25f4423a2a2cec8a7b512e7cf2693e874f5974

I used a link to the same place that the radio buttons link to. Neither that link nor the radio button links open in a new tab (I agree that it would probably be useful, but then for all of them, right?)

website/src/pages/[organism]/submission/[groupId]/released.astro
Comment (theosanderson) [inline website/src/pages/[organism]/submission/[groupId]/released.astro:1]: It would be great to make this page display the "data use terms" column by default.

A few ways one could do this:
- hackily patch the schema on this page (feels sketchy)
- give SearchFullUI an additional property `extraColumnsToShowByDefault`, and here to pass in `dataUseTerms`
- use `showEditDataUseTermsButton` or a renamed version of it to enable this too.

Sorry not to have thought of this sooner. I don't mind it being a follow-up issue but it may make sense to do now.

website/src/components/DataUseTerms/EditDataUseTermsModal.tsx
theosanderson: Thanks!

I think the "Choose the new data use terms for 3 restricted sequences" should go immediately above the choice of _Open_ and _Restricted_.

I also don't think the note about the release date is relevant unless they choose restricted? So maybe the note could go above the date picker? But if this is hard we can leave it for the future because this feature won't be used sooo much.

website/src/components/DataUseTerms/DataUseTermsSelector.spec.tsx
theosanderson, line 1: Nice!

website/src/components/DataUseTerms/DataUseTermsSelector.tsx
```diff
@@ -1,25 +1,82 @@
...
+import { DateTime } from 'luxon';
+import { useState, type FC } from 'react';
 
+import { DateChangeModal, datePickerTheme } from './DateChangeModal.tsx';
+import { getClientLogger } from '../../clientLogger.ts';
 import { routes } from '../../routes/routes.ts';
-import { type DataUseTermsType, openDataUseTermsType, restrictedDataUseTermsType } from '../../types/backend.ts';
+import {
+    type DataUseTermsType,
+    openDataUseTermsType,
```
theosanderson, line 10: observation (not requesting code change atm): we should have named this `openDataUseTermsOption` - calling it `type` makes me think its a TypeScript type and confused me

website/src/components/SearchPage/DownloadDialog/SequenceFilters.tsx
theosanderson, line 1: Nicely done

website/src/components/DataUseTerms/EditDataUseTermsModal.tsx
```diff
@@ -0,0 +1,289 @@
...
+}) => {
+    const setDataUseTermsHook = backendClientHooks(clientConfig).useSetDataUseTerms(
+        { headers: createAuthorizationHeader(accessToken) },
+        {
+            onError: (error) =>
+                toast.error('Failed to edit terms of use: ' + stringifyMaybeAxiosError(error), {
+                    position: 'top-center',
+                    autoClose: false,
+                }),
+            onSuccess: () => location.reload(),
```
theosanderson, line 251: Fairly important change: we need to make a toast for the success operation here or things are super unclear. Maybe for now as a kind of hack: 

```
onSuccess: () => {
toast.success("Data use terms updated successfully, page will refresh in 5 seconds");
setTimeout(()=>location.reload(),5000);
}
```

theosanderson (commented):
Really good work, I had a look through all the code and it looks great to my eyes

fhennig:
Some more feedback I got from @corneliusroemer 

- The two column open/restricted selector doesn't look good on mobile (-> maybe make the layout responsive)
- One can apply a "non-update" and it shows in the data use terms history (should just not update)
- There is an error message "Please chose a date before ..." (see image below) and it should say "before or equal to", because once can 'equalize' the dates on multiple sequences.

image

Thanks Theo and Cornelius for the review!!

theosanderson:
To my mind we don't need to worry that much about mobile responsivity for submission-related pages (but ofc. no problem with making it better if that's not too hard)

website/src/components/DataUseTerms/EditDataUseTermsModal.tsx
fhennig: all done!

[`dd7860c` (#3322)](https://github.com/loculus-project/loculus/pull/3322/commits/dd7860cc5b270e683efc6829950f5970ebb0462e)

website/src/components/DataUseTerms/EditDataUseTermsModal.tsx
fhennig: I added a popup with info [`c716111` (#3322)](https://github.com/loculus-project/loculus/pull/3322/commits/c716111f5e7acde1c506350152a289f1a67f8b47)

I don't think we need the timeout for reload, I just removed the reload alltogether (not sure why we would reload the page. I got this from the EditDataUseTermsButton, but since the data won't be there, I think it doesn't make sense to reload)

fhennig: I have refactored both places to use the same toasts: [`9a56b19` (#3322)](https://github.com/loculus-project/loculus/pull/3322/commits/9a56b19443ebfb136f7557f9b9c98b3d99343161)

website/src/components/DataUseTerms/DataUseTermsSelector.tsx
fhennig: Oh, actually I think the same! I think naming things accurately is great. done: [`9b87eed` (#3322)](https://github.com/loculus-project/loculus/pull/3322/commits/9b87eedccbf344b50bf48154719d3127cff1e426)

website/src/components/DataUseTerms/EditDataUseTermsModal.tsx
```diff
@@ -0,0 +1,289 @@
...
+        case 'mixed':
+            return (
+                <div className='space-y-4'>
+                    <ActiveFilters sequenceFilter={sequenceFilter} />
+                    <p>
+                        {state.openCount} open and {state.restrictedCount} restricted sequences selected.
+                    </p>
+                    <p>
+                        You can release all the {state.restrictedCount} restricted sequences, moving them to the{' '}
+                        <a href={routes.datauseTermsPage()} className='text-primary-600'>
```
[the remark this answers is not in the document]
fhennig: I traced where this value gets placed and it's only available in the backend AFAIKT. You said it's not blocking, so I'm resolving this as "won't fix"

website/src/components/DataUseTerms/DataUseTermsSelector.spec.tsx
fhennig: Yay, thanks! 😃

website/src/pages/[organism]/submission/[groupId]/released.astro
Comment (fhennig) [reply website/src/pages/[organism]/submission/[groupId]/released.astro:1]: I went for "use showEditDataUseTermsButton or a renamed version of it to enable this too." which felt reasonably clean, although maybe the "extraColumnsToShowByDefault" is a bit nicer, but I think it's fine. We might touch the column visibilities again anyway because of the range stuff.

[`fdba6ea` (#3322)](https://github.com/loculus-project/loculus/pull/3322/commits/fdba6ea4e60eae05e33d9189928e2597a5eb7647)

fhennig:
regarding "One can apply a "non-update" and it shows in the data use terms history"

I implemented a fix here: [`d2e5503` (#3322)](https://github.com/loculus-project/loculus/pull/3322/commits/d2e550392c8b9ffbe1a5f079d6eab2580fca9201)

It wasn't possible to do the filtering in the backend without breaking up the `batchInsert` into individual inserts, and so I decided to calculate the affected accessions in the frontend. This is actually cool, because now the number on the button shows correctly how many sequences actually receive a changed date!

website/src/components/DataUseTerms/EditDataUseTermsModal.tsx
theosanderson: Yes, they should probably both open in new tabs - sorry for slow reply

website/src/components/DataUseTerms/DataUseTermsSelector.tsx
theosanderson: Thanks!

website/src/components/DataUseTerms/EditDataUseTermsModal.tsx
theosanderson: Yeah, the reload did make sense in the context of an individual sequence details page , as the sequence details page is instaneously up to date. But when the sequence details are open in a modal it loses the modal, so that's bad anyway so I think your new approach sounds good for now.

theosanderson (approved):
I didn't look through all the code again, but functionality seems good now and we've had a pass through the code so I'm approving. Thanks for all the hard work!

Issue #3268, opened 2024-11-21
We need to reduce the frequency at which we poll Github for curated data updates


corneliusroemer:
Github doesn't like how often we're polling it. We should hit github pages for this and not raw user content:

```
02:50:19    ERROR (trigger_submission_to_ena.py:  71) - Failed to retrieve file due to requests exception: 403 Client Error: Forbidden for url: https://raw.githubusercontent.com/pathoplexus/ena-submission/main/test/approved_ena_submission_list.json
```

I saw this in staging. It's also happening in production.

We currently hit Github every 2 minutes from that one deployment alone.

PR #3327, opened 2024-12-02
feat(deposition): Use github pages to not get rate limited, assert sequences with ena specific fields were submitted by us or the insdc_submission_group

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/3268, https://github.com/loculus-project/loculus/issues/3328

preview URL: http://ena-polling-fixes.loculus.org/

### Summary

1. Use github pages to reduce polling of github (we were being rate limited)
2. Fix logger.warning syntax (error due to comma splitting strings) and use logging instance logger instead of logging base class (make ruff happy).
3. Check if sequences with ena specific metadata fields were submitted by us or the insdc_submission_group (previously used submitter name insdc_ingest_user but this then warns us if a sequence has been curated).

### PR Checklist

### DIFF
--- a/ena-submission/config/defaults.yaml
+++ b/ena-submission/config/defaults.yaml
@@ -1,7 +1,7 @@
 username: external_metadata_updater
 password: external_metadata_updater
 keycloak_client_id: backend-client
-ingest_pipeline_submitter: insdc_ingest_user
+ingest_pipeline_submission_group: 1
 db_name: Loculus
 unique_project_suffix: Loculus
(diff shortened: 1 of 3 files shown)


theosanderson:
Just to flag that because GH pages can cache stuff and because this file changes this will potentially cause a bit of confusion while the cache has yet to be updated (I recall that in the past this could be a significant length of time). But I expect this is OK for now.

theosanderson (approved):

chaoran-chen (approved):

Issue #3341, opened 2024-12-02
Formatting issues in "data use terms history" modal 


rneher:
On Firefox (Ubuntu) the data use terms history model has rather unfortunate line wraps. 

![image](https://github.com/user-attachments/assets/5870f3a7-3fc9-4061-8df8-819018f5b18c)

theosanderson:
I'm not actually 100% sure how fixed this will be (the User might not be) as I haven't tested in this environment, so reopening for now

rneher:
this is partially fixed, but `insdc_ingest_user` is still too long. I think we should just increase the width of this column

PR #3342, opened 2024-12-02
fix(website): Adjust break all to break word

theosanderson:
resolves https://github.com/loculus-project/loculus/issues/3341 and various other problems

preview URL: https://breakword.loculus.org/

## Screenshot

image
 becomes
image

We were using `break-all` which allows wrapping newlines very arbitrarily.  Now we only use `break-words`
.


anna-parker (approved):
LGTM - noticed this fixes cases where we would split a line in the middle of a word

PR #3348, opened 2024-12-03
feat(website, backend, prepro)!: Update error schema with input and output metadata field names

anna-parker:
resolves #

preview URL: https://update-error-schema-anya.loculus.org/

### Summary

Follow up of https://github.com/loculus-project/loculus/pull/932
image

This PR:
1. Adds both unprocessed and processed metadata field names to the error annotations, updating the types and tests in preprocessing, website and backend.
2. Use the unprocessed and processed (mapped to displayName) metadata field names for errors and warnings in the edit form.
3. Removed the duplicated uneditable processed metadata section from the edit page and move the edit button to the bottom of the edit page.
4. Add a tooltip with information on the name of the input metadata field (currently we just show the display Name which is confusing for users): 
image

## Info

This change isn't "breaking per-se as we include a flyway migration that updates the schema into the new format - however the results of the error schema won't be fully accurate until the preprocessing pipeline reprocesses all data and adds both unprocesssed and processed metadata field names to the error and warning output. 

### PR Checklist

- [x] The structure of the errrors and warnings column in the DB will now change - this needs to be handled correctly: DB migration, prepro pipeline update etc. -> fully tested in https://github.com/loculus-project/loculus/pull/3393, when migrations occur users are logged out of their accounts and when they log back in again the changes have occurred. The schema changes have been added in a flyway migration that can be rerun after the schema is corrected without overwriting values.

### DIFF
--- a/backend/src/main/resources/db/migration/V1.8__update_error_schema.sql
+++ b/backend/src/main/resources/db/migration/V1.8__update_error_schema.sql
@@ -1,0 +1,67 @@
+update sequence_entries_preprocessed_data
+set warnings = (
+  select jsonb_agg(
+    jsonb_build_object(
+      'unprocessedFields', (
+        select jsonb_agg(
+          jsonb_build_object(
+            'name', source->>'name',
+            'type', source->>'type'
+          )
+        )
+        from jsonb_array_elements(warning->'source') as source
+      ),
+      'processedFields', (
+        select jsonb_agg(
+          jsonb_build_object(
+            'name', source->>'name',
+            'type', source->>'type'
+          )
+        )
+        from jsonb_array_elements(warning->'source') as source
+      ),
+      'message', warning->>'message'
+    )
+  )
+  from jsonb_array_elements(warnings) as warning
+)
+where warnings is not null
+  and exists (
+    select 1
+    from jsonb_array_elements(warnings) as warning
+    where warning->'source' is not null
+  );
+
+update sequence_entries_preprocessed_data
+set errors = (
+  select jsonb_agg(
+    jsonb_build_object(
+      'unprocessedFields', (
+        select jsonb_agg(
+          jsonb_build_object(
+            'name', source->>'name',
+            'type', source->>'type'
+          )
+        )
+        from jsonb_array_elements(error->'source') as source
+      ),
+      'processedFields', (
+        select jsonb_agg(
+          jsonb_build_object(
+            'name', source->>'name',
+            'type', source->>'type'
+          )
+        )
+        from jsonb_array_elements(error->'source') as source
+      ),
+      'message', error->>'message'
+    )
+  )
+  from jsonb_array_elements(errors) as error
+)
+where errors is not null
+and exists (
+    select 1
+    from jsonb_array_elements(errors) as error
+    where error->'source' is not null
+  );
--- a/backend/src/test/kotlin/org/loculus/backend/controller/submission/PreparedProcessedData.kt
+++ b/backend/src/test/kotlin/org/loculus/backend/controller/submission/PreparedProcessedData.kt
@@ -396,7 +396,13 @@ object PreparedProcessedData {
         accession = accession,
         errors = listOf(
             PreprocessingAnnotation(
-                source = listOf(
+                unprocessedFields = listOf(
+                    PreprocessingAnnotationSource(
+                        PreprocessingAnnotationSourceType.Metadata,
+                        "host",
+                    ),
+                ),
+                processedFields = listOf(
                     PreprocessingAnnotationSource(
                         PreprocessingAnnotationSourceType.Metadata,
                         "host",
@@ -405,7 +411,13 @@ object PreparedProcessedData {
                 "Not this kind of host",
             ),
             PreprocessingAnnotation(
-                source = listOf(
+                unprocessedFields = listOf(
+                    PreprocessingAnnotationSource(
+                        PreprocessingAnnotationSourceType.NucleotideSequence,
+                        MAIN_SEGMENT,
+                    ),
+                ),
+                processedFields = listOf(
                     PreprocessingAnnotationSource(
                         PreprocessingAnnotationSourceType.NucleotideSequence,
                         MAIN_SEGMENT,
@@ -420,7 +432,13 @@ object PreparedProcessedData {
         accession = accession,
         warnings = listOf(
             PreprocessingAnnotation(
-                source = listOf(
+                unprocessedFields = listOf(
+                    PreprocessingAnnotationSource(
+                        PreprocessingAnnotationSourceType.Metadata,
+                        "host",
+                    ),
+                ),
+                processedFields = listOf(
                     PreprocessingAnnotationSource(
                         PreprocessingAnnotationSourceType.Metadata,
                         "host",
@@ -429,7 +447,13 @@ object PreparedProcessedData {
                 "Not this kind of host",
             ),
             PreprocessingAnnotation(
-                source = listOf(
+                unprocessedFields = listOf(
+                    PreprocessingAnnotationSource(
+                        PreprocessingAnnotationSourceType.NucleotideSequence,
+                        MAIN_SEGMENT,
+                    ),
+                ),
+                processedFields = listOf(
                     PreprocessingAnnotationSource(
                         PreprocessingAnnotationSourceType.NucleotideSequence,
                         MAIN_SEGMENT,
--- a/preprocessing/nextclade/src/loculus_preprocessing/datatypes.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/datatypes.py
@@ -36,14 +36,16 @@ class AnnotationSource:
 
 @dataclass(frozen=True)
 class ProcessingAnnotation:
-    source: tuple[AnnotationSource, ...]
+    unprocessedFields: tuple[AnnotationSource, ...]  # noqa: N815
+    processedFields: tuple[AnnotationSource, ...]  # noqa: N815
     message: str
 
     def __post_init__(self):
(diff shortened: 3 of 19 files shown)


backend/src/test/kotlin/org/loculus/backend/controller/submission/PreparedProcessedData.kt
theosanderson, line 1: (thanks for looking at this!!)
Just reminding myself that we'll need to add a DB migration. I would hope Postgres may be expressive enough to make the JSON surgery fairly doable

anna-parker: I wonder if just reprocessing the data would be enough? i.e. by increasing the pipeline version? In the DB we just store an errors and warnings column for the processed data and don't define the structure. I can test tomorrow on a persistent instance.

theosanderson: Yes, I agree it would be

theosanderson: Just things would be broken in the interim (which doesn't seem a big deal atm)

anna-parker: Uggh I just realized - either way if we make this schema change in one go there will be a period where the backend/website will not be able to read in the old processed data with the "source" annotation until it is all reprocessed to contain the new "unprocessedFields" and "processedFields". Are we ok with a period of time when users will not be able to submit or should we try to split this schema change up into portions so that we have a smooth roll-out?

anna-parker: @theosanderson maybe I'm missing sth - did you have an idea how to do this with a DB migration without interim breakage? If thats possible it would be great

theosanderson: So I asked Claude and it suggested this:

```
UPDATE table_name
SET warnings = (
  SELECT json_agg(
    json_build_object(
      'unprocessedFields', (
        SELECT json_agg(
          json_build_object(
            'name', source->>'name',
            'type', source->>'type'
          )
        )
        FROM json_array_elements(warning->'source') AS source
      ),
      'processedFields', (
        SELECT json_agg(
          json_build_object(
            'name', source->>'name',
            'type', source->>'type'
          )
        )
        FROM json_array_elements(warning->'source') AS source
      ),
      'message', warning->>'message'
    )
  )
  FROM json_array_elements(warnings) AS warning
)
WHERE warnings IS NOT NULL;
```

I have no idea if that would as is work but I vaguely expect there is some way to do it

theosanderson: I got Claude to generate a script to test the migration (there's a subtle difference in the function names vs above `jsonb` vs `json`)

https://gist.github.com/theosanderson/5fedae89d85b895b05d42213f33440cb

And I pasted it into the terminal in https://www.crunchydata.com/developers/playground/learn-sql and it seemed to work

anna-parker: Ah true - I guess we would still have to re-process everything once more as the schema would be correct like this but the content of the new unprocessedFields wouldn't necessarily be correct as we currently do not add input fields to the source field only the processedFields. 

I also assume we wouldn't add this to the flyway migration as we won't want it to be performed each time we restart the DB. 

So the process would be: 
1. roll out the changes from this PR.
2. Perform manual DB migration by connecting to the DB (there will still be a short period where the state will be out of sync)
3. Reprocess all sequences by increasing the preprocesing pipeline version. 

I can try to test this today on a persistent preview.

theosanderson: I don't think there's a problem with adding this to flyway. It won't really do anything if the table is empty

preprocessing/nextclade/tests/test_config.yaml
```diff
@@ -64,7 +64,7 @@ processing_spec:
       type: date
     function: parse_and_assert_past_date
     inputs:
-      date: required_collection_date
```
anna-parker, line 67: I changed this to ensure the tests work when the input and output field do not have the same name

backend/src/main/resources/db/migration/V1.8__update_error_schema.sql
```diff
@@ -0,0 +1,67 @@
...
+    )
+  )
+  from jsonb_array_elements(warnings) as warning
+)
+where warnings is not null
+  and exists (
+    select 1
+    from jsonb_array_elements(warnings) as warning
+    where warning->'source' is not null
+  );
```
theosanderson, line 33: I have no problem with us having this check, but as a more academic point - I can't see any reason why we would need to have it. This migration would only be running on an empty DB except in the situation where this change is needed as far as I can see.

theosanderson: (not requesting any changes!)

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -389,9 +407,15 @@ def add_input_metadata(
...
-                    source=(
+                    unprocessedFields=(
+                        AnnotationSource(
+                            name=segment,
+                            type=AnnotationSourceType.NUCLEOTIDE_SEQUENCE,
+                        ),
+                    ),
+                    processedFields=(
                         AnnotationSource(
-                            name="segment",
```
theosanderson, line 394: ooh was this a bug on main?

preprocessing/nextclade/src/loculus_preprocessing/datatypes.py
theosanderson, line 1: Please could you update the dummy preprocessing pipeline too

theosanderson (commented):
Thanks so much for taking this on!

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
anna-parker: Yeah... hihi

preprocessing/nextclade/src/loculus_preprocessing/datatypes.py
anna-parker: Good catch - took me a good minute to get this working but shoudl be good now  :-)

website/src/components/ReviewPage/ReviewCard.tsx
```diff
@@ -201,20 +205,27 @@ const MetadataList: FC<MetadataListProps> = ({ data, isLoading, metadataDisplayN
...
+                        .map((field) => metadataDisplayNames.get(field.name) ?? field.name)
+                        .join(', ');
+                    const unprocessedFieldName = error.unprocessedFields.map((field) => field.name).join(', ');
                     return (
                         <div key={uniqueKey} className='flex flex-shrink-0'>
                             <p
                                 className='text-red-600'
                                 data-tooltip-id={'error-tooltip-' + accession + '-' + uniqueKey}
                             >
+                                {processedFieldName} (produced from input fields: {unprocessedFieldName}):{' '}
```
theosanderson, line 228: image

I don't think this is great - I would leave it as is and if we want to rethink it later we can

website/src/components/ReviewPage/ReviewCard.tsx
```diff
@@ -238,14 +249,18 @@ const Warnings: FC<WarningsProps> = ({ warnings, accession }) => {
...
-                ))}
+                {warnings.map((warning) => {
+                    const processedFieldName = warning.processedFields.map((field) => field.name).join(', ');
+                    const unprocessedFieldName = warning.unprocessedFields.map((field) => field.name).join(', ');
+                    return (
+                        <p
+                            key={warning.processedFields.map((field) => field.type + field.name).join('.') + accession}
+                            className='text-yellow-500'
+                        >
+                            {processedFieldName} (produced from input fields: {unprocessedFieldName}): {warning.message}
```
theosanderson, line 260: as above

website/src/components/ReviewPage/ReviewCard.tsx
anna-parker: So just get rid of the (produced from ... section)? I do think it is nice to map to the display Name in the warning - otherwise it is confusing as the error is about the geoLocCountry and you see Collection country in the details

theosanderson: I think this UI is worse than previously atm. E.g. `main (produced from input fields: main)` when main isn't actually field is one example, but I think it applies to all of them here really. I agree we can put some effort into making this better at some point. (TBH I think one might do it by making the submission page allow you to upload stuff with display names as the column headers, as part of https://github.com/loculus-project/loculus/issues/3399) but I don't think we need to block on this for this PR.

theosanderson: (So yes, my suggestion was just getting rid for now)

anna-parker: True - it does look very odd for the sequence errors - I have now removed :-)

theosanderson (approved):
Approved but it would be great for @corneliusroemer to look too

backend/src/main/resources/db/migration/V1.8__update_error_schema.sql
anna-parker: Im just going to leave as is as teh current state has been tested and it doesn't break anything

Issue #3305, opened 2024-11-27
Replace zodios


fhennig:
As Cornelius reported here: https://github.com/loculus-project/loculus/issues/3295, Zodios seems to be unmaintained, and we're now behind a major version on react-query: https://github.com/loculus-project/loculus/pull/3294.

fengelniederhammer:
We could simply build our own wrapper around axios with the zod types that we have? It's a bit of manual effort, but in the end probably easier to maintain than the complicated zodios stuff.

fhennig:
Yes, @corneliusroemer also said so on Slack. I think it sounds like a good way to solve this. From what I could tell, it shouldn't be too much boilerplate to write.

corneliusroemer:
There might be some less heavy wrappers as well, something like: https://github.com/mattpocock/zod-fetch

fengelniederhammer:
Seems like this is using `fetch`. Personally, I like `axios` better than fetch (mostly because the errors are better)

Issue #3295, opened 2024-11-25
Zodios maintenance status: last commit 14 months ago, pinning us to tanstack react query v4


corneliusroemer:
No breakage so far, but we might get into some issues at some point

https://loculus.slack.com/archives/C05G172HL6L/p1732565097248449

anna-parker:
This has blocked the zod update in: https://github.com/loculus-project/loculus/pull/6654

PR #3352, opened 2024-12-03
chore(deps): disable unupgradable deps in dependabot.yaml

corneliusroemer:
see https://github.com/loculus-project/loculus/issues/3305
see https://github.com/loculus-project/loculus/issues/3295


anna-parker (approved):
lgtm

Issue #3343, opened 2024-12-02
Customizable column widths


theosanderson:
We should allow the Values yaml to specify widths for each column when displayed in the table, this should be a property of the metadata item.

This would require changing the values.yaml, the templates that parse it, the config.ts and the Table.tsx files.

PR #3353, opened 2024-12-03
feat(website): specify width for column(s) in search table

theosanderson:
resolves https://github.com/loculus-project/loculus/issues/3343

preview URL: https://colwidths2.loculus.org/ebola-sudan/search?

### Summary
Allows specifying column widths (which are actual minimum widths). If a column has no set width, behaviour is unchanged from pre-existing situation.

I haven't added too many for Loculus as this doesn't transfer across to our real deployments.

### Screenshot
Before:
image

After:
image


docs/src/content/docs/for-administrators/setup-with-kubernetes.md
```diff
@@ -134,7 +134,7 @@ organisms:
...
+Note the metadata section includes various fields for how the metadata of specific sequences should be displayed. Each metadata item must have a `name` which will also be displayed on the page unless `displayName` is also set. The `type` of the data, as well as if the field is `required` and if `autoComplete` is enabled can also be added. Additionally, links from metadata entries to external websites can be added using the `customDisplay` option. We also allow metadata to be grouped in sections, specified by the `header` field. The `noInput` parameter specifies that a parameter is generated internally by loculus (can be specified in the preprocessing pipeline) and should not be expected as input. You can optionally add a `columnWidth` for the column in the search table, in pixels, which will be the minimal width for the column.
```
chaoran-chen, line 137: Why don't we then also call the field `columnMinWidth`? Do you think that the logic of the field will likely change in the future?

kubernetes/loculus/templates/_common-metadata.tpl
```diff
@@ -211,6 +213,9 @@ organisms:
   {{- if .truncateColumnDisplayTo }}
   truncateColumnDisplayTo: {{ .truncateColumnDisplayTo }}
   {{- end }}
+  {{- if .columnWidth }}
+  columnWidth: {{ .columnWidth }}
+  {{- end }}
```
chaoran-chen, line 218: Could you please also document this in the Helm chart config?

website/src/components/SearchPage/Table.tsx
```diff
@@ -145,6 +146,9 @@ export const Table: FC<TableProps> = ({
                                     key={c.field}
                                     onClick={() => handleSort(c.field)}
                                     className='px-2 py-3 text-xs font-medium tracking-wider text-gray-500 uppercase cursor-pointer last:pr-6 text-left'
+                                    style={{
+                                        minWidth: c.columnWidth != null ? `${c.columnWidth}px` : undefined,
+                                    }}
```
chaoran-chen, line 151: I was confused at first for a moment and wondered whether this was catching `c.columnWidth === undefined` until I saw that this was using `!=`. What would you think about storing `c.columnWidth !== null && c.columnWidth !== undefined ? `${c.columnWidth}px` : undefined` in a separate variable and then use it here and below?

chaoran-chen (approved):
Thanks! Looks good but just a few minor comments.

kubernetes/loculus/templates/_common-metadata.tpl
theosanderson: Done (I hadn't seen it in a search because it was `.mdx` not `.md`) - though I agree with Cornelius that HTML isn't a good format to maintain this in

chaoran-chen: Thanks! (Yes, it would definitely be nice to create a proper schema with autogenerated documentation but until then, this is the simplest way to have something.)

website/src/components/SearchPage/Table.tsx
theosanderson: Ah, I hadn't particularly taken in that we were using `!=` as I did this with `aider`. Funny, I actually thought our linter didn't allow that!

Happy to go for your approach.

docs/src/content/docs/for-administrators/setup-with-kubernetes.md
theosanderson: IMO this is the better name, and I don't think the details of the implementation are too important. It is the width provided there are sufficient columns that they take up the other space. I originally had it as affecting CSS "width", but I found that that didn't actually work (the columns became narrower than this width) for reasons I don't understand. I don't feel that strongly so can change it if you do.

PR #3356, opened 2024-12-03
fix(backend): reset etag for data_use_terms_table updates

### LINKED ISSUE #3354: Updating data use terms doesn't trigger an update of SILO
I have released some sequences on main.loculus.org and the update didn't show up even after a few minutes.

I suspect that we require

something like `SELECT create_update_trigger_for_table('user_groups_table');` but for `data_use_terms_table`.

The https://github.com/loculus-project/loculus/blob/8a22aa84649ab2499edcc8c0b58eab7b15ce8391/backend/docs/db/schema.sql seems to confirm that a trigger is indeed not set for this table.

corneliusroemer:
fixes #3354

preview URL: https://fix-etag.loculus.org

### Summary
I had forgotten to include `data_use_terms_table` updates in the list of tables whose update should reset the etagg

### DIFF
--- a/backend/docs/db/schema.sql
+++ b/backend/docs/db/schema.sql
@@ -739,6 +739,13 @@ CREATE TRIGGER update_tracker_trigger AFTER INSERT OR DELETE OR UPDATE OR TRUNCA
 
 
 --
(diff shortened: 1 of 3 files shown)


theosanderson (approved):
lgtm

PR #3359, opened 2024-12-04
feat(website): show API URLs on API documentation page

### LINKED ISSUE #3358: Make API URLs easier to find
As mentioned in #3357, the API URLs can be very different depending on the setup. We should make it easier to find out the host of the different APIs and I propose displaying them directly on the API documentation page.

chaoran-chen:
resolves #3358

preview URL: https://api-urls.loculus.org/api-documentation

### Summary

This displays the API URLs directly on the API documentation page.

### PR Checklist


fhennig:
As Anya noted, the KC URL is the internal one

chaoran-chen:
@fhennig Yes.. Any thoughts on how I can get the Keycloak root API?

Now, I'm also noticing that `host` might actually not be the right term because a link with ".../west-nile" is not really a host. What do you think?

fhennig (approved):
looks correct!

But the KC server seems to have issues? If I just try to log in, it doesn't work. But I think it's not related to your changes

PR #3362, opened 2024-12-04
docs: add API overview and describe where to find the APIs

### LINKED ISSUE #3357: Fix (generalize) API links in docs
In some of our docs pages, we making assumptions of where the APIs can be reached which are not necessarily true. For example, [here](https://github.com/loculus-project/loculus/blob/main/docs/src/content/docs/for-users/submit-sequences.md#api), we write that the backend is at `https://backend.<URL of the Loculus website>` but that's not always the base. In our previews, it is at `https://backend-<URL of the Loculus website>` and in GenSpectrum, it is at `https://api.loculus.genspectrum.org/backend/`. We should phrase it more generally.

chaoran-chen:
resolves #3357

Preview: https://docs-docs-api-links.loculus.org

This depends on #3359 which should be merged first.


fhennig (approved):
LGTM!

PR #3376, opened 2024-12-05
chore(deps): Upgrade to astro v5

fhennig:
preview URL: http://astro-v5.loculus.org

### Summary
Using the guide: https://5-0-0-beta.docs.astro.build/en/guides/upgrade-to/v5/

Breaking changes to still check out:
- https://5-0-0-beta.docs.astro.build/en/guides/upgrade-to/v5/#changed-actions-submitted-by-html-forms-no-longer-use-cookie-redirects

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -21,9 +21,9 @@
         "e2e:headed": "npx playwright test --headed"
     },
     "dependencies": {
-        "@astrojs/mdx": "^3.1.9",
-        "@astrojs/node": "^8.3.4",
         "@emotion/react": "^11.14.0",
+        "@astrojs/mdx": "^4.0.1",
+        "@astrojs/node": "^9.0.0",
         "@headlessui/react": "^2.2.0",
         "@mui/material": "~5.14.20",
         "@svgr/core": "^8.1.0",
@@ -31,7 +31,7 @@
         "@tanstack/react-query": "^4.36.1",
         "@zodios/core": "^10.9.6",
         "@zodios/react": "^10.4.5",
-        "astro": "^4.16.17",
+        "astro": "^5.0.2",
         "axios": "^1.7.9",
         "change-case": "~5.3.0",
         "chart.js": "^4.4.7",
@@ -55,7 +55,7 @@
     },
     "devDependencies": {
         "@astrojs/check": "^0.9.4",
-        "@astrojs/react": "^3.6.3",
+        "@astrojs/react": "^4.0.0",
         "@astrojs/tailwind": "^5.1.3",
         "@emotion/styled": "^11.14.0",
         "@iconify/json": "^2.2.279",
--- a/website/tsconfig.json
+++ b/website/tsconfig.json
@@ -7,5 +7,6 @@
         "types": ["unplugin-icons/types/react"],
         "allowImportingTsExtensions": true
     },
+    "include": [".astro/types.d.ts", "**/*"],
     "exclude": ["playwright-report", "dist", "node_modules"]
 }
--- a/website/package-lock.json
+++ b/website/package-lock.json
@@ -8,8 +8,8 @@
             "name": "loculus-website",
             "version": "0.0.1",
             "dependencies": {
-                "@astrojs/mdx": "^3.1.9",
-                "@astrojs/node": "^8.3.4",
+                "@astrojs/mdx": "^4.0.1",
+                "@astrojs/node": "^9.0.0",
                 "@emotion/react": "^11.14.0",
                 "@headlessui/react": "^2.2.0",
                 "@mui/material": "~5.14.20",
@@ -18,7 +18,7 @@
                 "@tanstack/react-query": "^4.36.1",
                 "@zodios/core": "^10.9.6",
                 "@zodios/react": "^10.4.5",
-                "astro": "^4.16.17",
+                "astro": "^5.0.2",
                 "axios": "^1.7.9",
                 "change-case": "~5.3.0",
                 "chart.js": "^4.4.7",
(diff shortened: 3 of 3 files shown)


fhennig:
From argocd: 

```
[ERROR] Error: CONFIG_DIR environment variable was not set during build time, is 'undefined'
```

The website pods don't want to come up. Strange.

fhennig:
Alright. So I figured out that I can fix it by prefixing CONFIG_DIR with PUBLIC_, but I don't understand why. The CONFIG_DIR should only be relevant on the server side, not on the client side! 

We could just change the config and then I think it'd be fine.

@corneliusroemer , what do you think?

fhennig:
Re: https://5-0-0-beta.docs.astro.build/en/guides/upgrade-to/v5/#changed-actions-submitted-by-html-forms-no-longer-use-cookie-redirects

I _think_ it isn't relevant for us? But can't really tell.

Besides that I think we'd be good to go!

corneliusroemer:
Nice! Now we're ready for Astro 5 before it is even released!

website/src/config.ts
```diff
@@ -11,7 +11,7 @@ let _config: WebsiteConfig | null = null;
 let _runtimeConfig: RuntimeConfig | null = null;
 
 function getConfigDir(): string {
-    const configDir = import.meta.env.CONFIG_DIR;
+    const configDir = import.meta.env.PUBLIC_CONFIG_DIR;
```
corneliusroemer, line 14: I saw some comment in your PR but could you add a summary of the reasoning behind this change in the main PR message (i.e. top post) - I don't know whether this env variable is used by other parts.

Is there something in the astro changelog?

corneliusroemer (commented):
Oh nice, so Astro 5 is actually out already!

website/src/config.ts
corneliusroemer: Reading more on this, and looking at where we specify `CONFIG_DIR` I'm pretty sure we should not do it this way. Can you revert and try to find an alternative solution that doesn't involve renaming the variable?

One definitely shouldn't rename a variable only partially. And in this case an env variable is part of the public API so that's two reasons not to change.

corneliusroemer (changes requested):
The CONFIG_DIR rename doesn't look like a good idea

website/src/config.ts
fengelniederhammer: We could use `process.env`. We don't need the variable at build time and we know that we're running on node (i.e. `process.env` is available).

fhennig:
I'll try what @fengelniederhammer suggested 👌

fhennig:
Interesting, I still get errors. I think there are code paths in the client that are leading to the config getting accessed? But I _do_ wonder why we didn't have issues with this before?

website/src/config.ts
```diff
@@ -11,7 +12,7 @@ let _config: WebsiteConfig | null = null;
 let _runtimeConfig: RuntimeConfig | null = null;
 
 function getConfigDir(): string {
-    const configDir = import.meta.env.CONFIG_DIR;
+    const configDir = process.env.CONFIG_DIR;
     if (typeof configDir !== 'string' || configDir === '') {
         throw new Error(`CONFIG_DIR environment variable was not set during build time, is '${configDir}'`);
```
fengelniederhammer, line 17: If we use `process` now, the error message is not really correct anymore - the env variable will not be baked into the code at build time anymore but instead actually be read at runtime.

fhennig: Yeah I was just trying out stuff

fhennig:
Ok using dotenv like that won't work.

fhennig:
I think I'll just roll back from the `process` change and then I'll have to debug more thoroughly what's going on with the errors.

fhennig:
Tests pass! Is everything looking good? I'm suspicious, but can't see anything that looks wrong. Maybe @corneliusroemer you want to have a look?

website/tsconfig.json
```diff
@@ -7,5 +7,6 @@
         "types": ["unplugin-icons/types/react"],
         "allowImportingTsExtensions": true
     },
+    "include": [".astro/types.d.ts", "**/*"],
```
chaoran-chen, line 10: Oh, interesting! How did you come up with that? And doesn't `**/*` include `.astro/types.d.ts`?

fengelniederhammer: What's the magic trick here? Does this line make the `meta.env` thing work?

fhennig: I just followed the migration guide 😆 

https://5-0-0-beta.docs.astro.build/en/guides/upgrade-to/v5/#what-should-i-do-13

> What's the magic trick here? Does this line make the meta.env thing work?

No, that line was already there before, when it wasn't working. I honestly don't know anymore what it was. I rebased the branch, cleaned up the history a bit by just getting rid of commits, and then it just worked 🤷 

> And doesn't **/* include .astro/types.d.ts?

It seems like it does, actually! But I got this line from the migration guide, maybe it's just like that for clarity?

chaoran-chen: Ah, if it's from the migration guide, then I think it's good! Maybe `**` doesn't include folders starting with a `. `?

chaoran-chen (approved):
Looks good to me!

website/package.json
```diff
@@ -21,17 +21,17 @@
         "e2e:headed": "npx playwright test --headed"
     },
     "dependencies": {
-        "@astrojs/mdx": "^3.1.9",
-        "@astrojs/node": "^8.3.4",
         "@emotion/react": "^11.14.0",
+        "@astrojs/mdx": "^4.0.1",
```
corneliusroemer, line 25: Now the deps are no longer sorted :)

PR #3380, opened 2024-12-05
feat: earliest release date

### LINKED ISSUE #2894: preprocessing: Add "consensus" release date field (with whatever nomenclature)
Currently on Pathoplexus we default to showing an "NCBI release date" field. This is good because if we used the Pathoplexus release date field all sequences ingested from INSDC would have the same release date which wouldn't be useful. But it's bad because it is undefined for sequences submitted directly to pathoplexus. IMO we should create a consensus field which is the NCBI release date if set (which will generally also be the Pathoplexus release date for items submitted to INSDC after launch) and otherwise the Pathoplexus release date. I'm raising this in Loculus as the code implementation would be here.

fhennig:
resolves #2894 

preview URL: https://earliest-release-date.loculus.org

### Screenshot

image

### PR Checklist


backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
```diff
@@ -93,6 +98,14 @@ open class ReleasedDataModel(
             NullNode.getInstance()
         }
 
+        backendConfig.
+
+        // TODO in here we can calculate the earlierstReleaseDate and add it to the metadata
+        // Settings are pulled from backendConfig
+        // Or maybe we can't do it here, because we can't do it on a line-by-line basis?
+        // is it  sufficient to just take the earliest of 'releasedAtTimestamp' and any other
```
theosanderson, line 106: There's this other element that it's across versions – for version 3 it will be version 1s release date, or the earliest of the external fields, whichever is earliest

fhennig:
is it sufficient to just add a metadata entry for the new field, to get it displayed? (Looking into it)

fhennig:
I've added the field (manually for now) - I think it's required for being able to actually display the value?

fhennig:
The Helm Chart now generates a default, because Jackson can't cope with the field missing AFAIKT

fhennig:
**Status update**: It's working in the preview

Current thoughts:
- I'd really like a test to somehow test and make sure that the "stream is ordered" invariant is upheld
- It'd be nice to have test for the "business logic" of actually finding the earliest date, and maybe cover some edge cases
- How do handle parsing errors? Where is the "validation boundary"?

chaoran-chen:
> - I'd really like a test to somehow test and make sure that the "stream is ordered" invariant is upheld
- It'd be nice to have test for the "business logic" of actually finding the earliest date, and maybe cover some edge cases

Sounds sensible to me!

> - How do handle parsing errors? Where is the "validation boundary". If a date isn't parsed correctly (at the moment) the whole endpoint fails ... probably not good.

Would it be easy to validate the config at the beginning when the server starts to ensure that the earlierst release date fields are all of type date? Or do you refer to something else?

fhennig:
I was just thinking out loud, I have now just added a try/catch block so the endpoint still returns data.

I _think_ you said at some point that there is some validation on the `processedData`?

chaoran-chen:
When the pipeline submits data to the backend, the backend checks the types of the metadata: https://github.com/loculus-project/loculus/blob/main/backend/src/main/kotlin/org/loculus/backend/service/submission/ProcessedSequenceEntryValidator.kt#L223 – is that what you mean?

fhennig:
I didn't think of this:

> Would it be easy to validate the config at the beginning when the server starts to ensure that the earlierst release date fields are all of type date?

But it's a good idea and I did it now.

chaoran-chen (commented):
Thanks a lot, Felix! The code looks great to me! I just have a few small suggestions.

backend/src/main/kotlin/org/loculus/backend/utils/EarliestReleaseDateFinder.kt
```diff
@@ -0,0 +1,56 @@
...
+import kotlinx.datetime.LocalDateTime
+import kotlinx.datetime.LocalTime
+import mu.KotlinLogging
+import org.loculus.backend.service.submission.RawProcessedData
+
+private val log = KotlinLogging.logger { }
+
+/**
+ * Calculate the earliest release date for rows of sequence entries given to it one by one.
+ * Assumes that rows are sorted: all accession entries are given in a block, and with ascending versions.
```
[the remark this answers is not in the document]
fhennig: you mean add asserts in the class? I think it should be fine without adding stuff here, because I specifically added the test for that reason. But I'm also not opposed. Maybe others want to add their opinion?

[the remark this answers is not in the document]
chaoran-chen: Yes, I meant asserting within the class (either only during debug mode or always)

backend/src/test/kotlin/org/loculus/backend/controller/submission/GetReleasedDataEndpointTest.kt
```diff
@@ -343,6 +351,48 @@ class GetReleasedDataEndpointTest(
         assertThat(data[0].metadata, `is`(not(emptyMap())))
     }
 
+    @Test
+    fun `GIVEN multiple accessions with multiple versions THEN results are sorted`() {
```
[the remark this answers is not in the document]
fhennig: Yes, I've added a good comment!

[`b1d94b2` (#3380)](https://github.com/loculus-project/loculus/pull/3380/commits/b1d94b24d74d478efd2abc03a10d6238d63ec01d)

backend/src/main/kotlin/org/loculus/backend/utils/EarliestReleaseDateFinder.kt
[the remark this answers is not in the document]
fhennig: I added an assertion that the versions are incrementing: [`f0258ec` (#3380)](https://github.com/loculus-project/loculus/pull/3380/commits/f0258ec899abbd088dfb0218e2ccf1cd861fc84f)

I think with this and the test pointing to this class, it should be sufficient.

The assertion doesn't check whether all accessions are coming in a block together, but to keep track of that we'd need to track all accessions any ways, at that point we could even ditch the requirement for accessions coming in as chunks 😄 

IMO this is sufficient, so I'm marking this as resolved (also because you weren't sure whether to add anyways). If you think more should be done, feel free to un-resolve!

docs/src/content/docs/reference/helm-chart-config.mdx
```diff
@@ -535,6 +535,27 @@ Each organism object has the following fields:
             <td></td>
             <td>Metadata fields associated with the organism.</td>
         </tr>
+        <tr>
+            <td>`earliestReleaseDate`</td>
+            <td>Object</td>
+            <td></td>
+            <td>Configuration object for enabling and configuring the `earliestReleaseDate` metadata field.</td>
```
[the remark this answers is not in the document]
fhennig: Added [`e51ae77` (#3380)](https://github.com/loculus-project/loculus/pull/3380/commits/e51ae77190899d895e865d0f2c84c38335c89f1b) [`14cba2e` (#3380)](https://github.com/loculus-project/loculus/pull/3380/commits/14cba2ed08a365027051bf25c99af8d0a3836cd7)

(In the long term I think it would be good to have this in a more dedicated [explanation](https://diataxis.fr/explanation/) section -- this is supposed to be only [reference](https://diataxis.fr/reference/) information I think)

backend/src/main/kotlin/org/loculus/backend/utils/EarliestReleaseDateFinder.kt
```diff
@@ -0,0 +1,56 @@
...
+
+        fields.forEach { field ->
+            rawProcessedData.processedData.metadata[field]?.textValue()?.let { dateText ->
+                val date = try {
+                    LocalDateTime(LocalDate.parse(dateText), LocalTime.fromSecondOfDay(0))
+                } catch (e: IllegalArgumentException) {
+                    log.warn {
+                        "Incorrectly formatted date on ${rawProcessedData.accession}." +
+                            "${rawProcessedData.version} on field $field: $dateText"
+                    }
```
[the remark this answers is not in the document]
fhennig: Yeah indeed this should never happen because inputs are already validated (as you pointed out earlier).

> If the latter, I think that we should indicate it more clearly and e.g. encourage the reader to report it.

Where would they report it? Could you maybe give me the message you'd like to see here?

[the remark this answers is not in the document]
chaoran-chen: I'd use at least log level `error` – maybe something even higher.

What about that? "Unexpected error: Incorrectly formatted date on ${rawProcessedData.accession}. ${rawProcessedData.version} on field $field: $dateText – Something is wrong with this instance: it might be a configuration error or a bug of the software. Please feel free to reach out to the developers if you need advise (https://github.com/loculus-project/loculus/issues)"

docs/src/content/docs/reference/helm-chart-config.mdx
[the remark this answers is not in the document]
chaoran-chen: Agreed, having an explanation section would be great! (Feel free to introduce one whenever you like.)

backend/src/main/kotlin/org/loculus/backend/utils/EarliestReleaseDateFinder.kt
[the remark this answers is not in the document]
chaoran-chen: sounds good! thanks!

backend/src/test/kotlin/org/loculus/backend/config/BackendSpringConfigTest.kt
```diff
@@ -0,0 +1,74 @@
...
+    fun `GIVEN an empty config THEN the it is valid`() {
+        val conf = backendConfig(emptyList(), EarliestReleaseDate(false, emptyList()))
+
+        val errors = validateEarliestReleaseDateFields(conf)
+
+        assertTrue(errors.isEmpty())
+    }
+
+    @Test
+    fun `GIVEN a config with two external fields that exist and are of type date THEN it is valid`() {
```
chaoran-chen, line 21: Given that the test class is generally for the backend spring config, I think it's not clear from the name that the "external fields" related to the EarliestReleaseDate

backend/src/test/kotlin/org/loculus/backend/controller/submission/GetReleasedDataEndpointTest.kt
```diff
@@ -343,6 +351,51 @@ class GetReleasedDataEndpointTest(
         assertThat(data[0].metadata, `is`(not(emptyMap())))
     }
 
+    /**
+     * This test ist relevant for EarliestReleaseDateFinder -- It relies on this particular ordering to be returned.
```
chaoran-chen, line 355:
```suggestion
     * This test ist relevant for EarliestReleaseDateFinder which relies on this particular ordering to be returned.
```

Otherwise, one could read it as that the test relies on the particular ordering.

chaoran-chen (approved):
Thanks @fhennig! There are a few minor comments left but otherwise, it looks great and I'll already approve. Feel free to ignore the comments if you like.

fhennig:
Thanks! I'll still incorporate the last few bits, but thanks for approving already!

backend/src/test/kotlin/org/loculus/backend/config/BackendSpringConfigTest.kt
fhennig: good point, fixed here: https://github.com/loculus-project/loculus/pull/3380/commits/f1f5eb49900cdd4412c9ffe0fad9fd7a22f3eaf7

kubernetes/loculus/values.yaml
```diff
@@ -131,6 +135,12 @@ defaultOrganismConfig: &defaultOrganismConfig
             timestamp: ncbiReleaseDate
         noInput: true
         columnWidth: 100
+      - name: earliestReleaseDate
```
corneliusroemer, line 138: This should go into the default metadata part of the helm template as it's a magic value and not something that's calculated by preprocessing

corneliusroemer (commented):
We might want to rename it to dateFirstPublic as earliestReleaseDate is an implementation focused name, it's the earliest across versions and including INSDC, but to end users they just want to know when was this sequence first public.

backend/src/main/kotlin/org/loculus/backend/utils/EarliestReleaseDateFinder.kt
```diff
@@ -0,0 +1,74 @@
...
+/**
+ * Calculate the earliest release date for rows of sequence entries given to it one by one.
+ * Assumes that rows are sorted: all accession entries are given in a block, and with ascending versions.
+ *
+ * The earliest release date of a sequence is the earliest date of:
+ * - the internal release date
+ * - any date from a given list of fields
+ * - the earliest release date from the previous version (if it exists)
+ */
+class EarliestReleaseDateFinder(private val fields: List<String>) {
```
corneliusroemer, line 20: Does this need to be so complicated? 

Earliest release date of Loculus is always the release date of version 1. We can get this at db level in the view.

Then calculating the minimum of that and whatever else comes in from processed data is trivial as it can be done at the single sequence level.

backend/src/main/kotlin/org/loculus/backend/utils/EarliestReleaseDateFinder.kt
```diff
@@ -0,0 +1,74 @@
...
+    private val earliestReleaseDateCache = mutableMapOf<String, LocalDateTime>()
```
corneliusroemer, line 21: Instead of `String` this should probably be `AccessionVersion` for clarity?

backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
```diff
@@ -93,6 +109,8 @@ open class ReleasedDataModel(
             NullNode.getInstance()
         }
 
+        val earliestReleaseDate = earliestReleaseDateFinder?.calculateEarliestReleaseDate(rawProcessedData)
```
corneliusroemer, line 112: It would be more in line with how we compute other same-accession across-version fields to make a db query and follow the pattern for `getLatestVersions` and `getLatestRevocationVersions` as part of `computeVersionStatus` and `computeDataUseTerm` above

Relying on accession/versions to be sorted is a bit brittle and probably quite a few more lines of code as well.

backend/src/main/kotlin/org/loculus/backend/utils/EarliestReleaseDateFinder.kt
fhennig: The specifics of calculating the earliest release date can be changed (like you said) quite easily, and it's all in this class, and there are tests for the class. Not sure what you find complicated about it.

backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
fhennig: the examples you gave don't access JSON fields, so it's not quite the same. Personally I'm not sure whether it's possible to do this with SQL -- now I think it might, but it would definitely be a very complicated query.

I think the way it works now doesn't introduce any problems?

backend/src/main/kotlin/org/loculus/backend/utils/EarliestReleaseDateFinder.kt
fhennig: I think it's clear from context. If the PR would still be open I would be open to change it but I don't think it warrants a new PR.

corneliusroemer:
I don't think this comment was ever addressed, this is blocking in my mind as once set the name will likely stick:
- https://github.com/loculus-project/loculus/pull/3380#pullrequestreview-2496119705

chaoran-chen:
>What's breaking
>To enable earliest release date, one needs to change the values.yaml, see the Pathoplexus PR for example edit: [pathoplexus/pathoplexus#341 (files)](https://github.com/pathoplexus/pathoplexus/pull/341/files)

@corneliusroemer, why is this breaking? Isn't it just a new feature but if an instance doesn't enable it, nothing would change?

corneliusroemer:
Right, I missed that since it didn't quote the text. To keep this PR on topic, I'll take the side discussion to Slack

corneliusroemer:
> @corneliusroemer, why is this breaking? Isn't it just a new feature but if an instance doesn't enable it, nothing would change?

Our convention is that things that usually imply config/code changes for known users (PPX/Genspectrum) should be marked as breaking. I'm happy to call it a different name than breaking. It's not the name that matters but the concept: PPX needs to be able to quickly tell when the config should be changed. Does that make sense?

So I agree that this isn't breaking in the strict sense, but it's something to draw attention to, and the way to draw attention is by adding a `!`.

theosanderson:
> Our convention is that things that usually imply config/code changes for known users (PPX/Genspectrum) should be marked as breaking. I'm happy to call it a different name than breaking. It's not the name that matters but the concept: PPX needs to be able to quickly tell when the config should be changed. Does that make sense?

Yeah I think new features are kind of an edge case where I don't have strong feelings either way. E.g. here is another new feature that won't take effect until we enable it in PP: https://github.com/loculus-project/loculus/pull/3532 but we don't necessarily have to enable it at the point when we merge (though, yes, sometimes it may be useful to do so so we keep track). So I wouldn't object if someone added `!` to this one, but also wouldn't object if it was missed. If it's missed it's probably useful to try to create an issue on PP to add the config in due course.

fhennig:
> Our convention is that things that usually imply config/code changes for known users (PPX/Genspectrum) should be marked as breaking. I'm happy to call it a different name than breaking. It's not the name that matters but the concept: PPX needs to be able to quickly tell when the config should be changed. Does that make sense?

I wasn't aware of this but happy to keep it in mind for the future.

PR #3383, opened 2024-12-05
feat(website): Column order for search table

### LINKED ISSUE #3378: Respect table column order from config in browse page table
The config specifies table columns as an array, but we seem to not respect this when creating the actual table: https://github.com/loculus-project/loculus/blob/5f7ca953b29ca07e3dc923249f8141660c155436/website/src/types/config.ts#L98

It would be nice to be able to customize the order in which we want columns to appear, e.g. clade followed by lineage without interspersed country etc.

For avoidance of doubt, I'm talking about the order of these columns (`Authors`, `Clades` etc):
image

I'm surprised we don't have an issue for this, seems like little work and large impact.

See: https://loculus.slack.com/archives/C05G172HL6L/p1733380853207909?thread_ts=1733339244.515559&cid=C05G172HL6L @emmahodcroft

theosanderson:
resolves #3378

preview URL: https://column-order.loculus.org/

### Summary
Adds an optional index to metadata items that determines their order. Currently this applies only to the search page table columns. We could expand this to e.g. the search fields but this would be a bit more complex, so is considered out of scope for now. The order of fields is `accession, [lower order index], [higher order index], [higher still..], [fields without order index specified]`

### Screenshot
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ node_modules/
 temp/
 
 .aider*
+.env
--- a/docs/src/content/docs/reference/helm-chart-config.mdx
+++ b/docs/src/content/docs/reference/helm-chart-config.mdx
@@ -650,6 +650,12 @@ Each organism object has the following fields:
             <td>The minimum column width for this field on the search table.</td>
         </tr>
         <tr>
+            <td>`order`</td>
+            <td>Number</td>
+            <td></td>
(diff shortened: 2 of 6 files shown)


theosanderson:
It makes it easier to add intermediate stuff later :)

corneliusroemer:
The smaller the order number the earlier columns appear, and without number a very high number is imputed, correct?

Issue #3382, opened 2024-12-05
Adjust how table expands on narrow screens so download button is visible


theosanderson:
On a narrow screen the search table looks like this:

![image](https://github.com/user-attachments/assets/69bf2fdb-6c9c-4a45-bb21-6b9778933513)

mostly this is ok, horizontal scrolling is acceptable, but the download button should be kept on the screen with no need to scroll to reach it

PR #3385, opened 2024-12-05
fix(website): don't allow search page to expand beyond window width - instead make the table scroll

theosanderson:
resolves https://github.com/loculus-project/loculus/issues/3382

preview URL: https://search-table-2.loculus.org/ebola-sudan/search?

### Summary
Ensures that the download button always stays on the page even when the table is wider than the screen, which doesn't happen at the moment

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ node_modules/
 temp/
 
 .aider*
+.env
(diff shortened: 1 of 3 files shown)


anna-parker (approved):
Awesome!

PR #3386, opened 2024-12-05
Search table 3

theosanderson:
See https://github.com/loculus-project/loculus/issues/3387

https://search-table-3.loculus.org/ebola-sudan/search?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ node_modules/
 temp/
 
 .aider*
+.env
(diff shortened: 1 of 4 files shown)


anna-parker:
I just tested this locally by minimizing the window and I don't see a scroll bar at the bottom (horizontal direction) but only on the right side (vertical), could we also add one on the bottom? 

https://github.com/user-attachments/assets/e6ab956c-a9d8-4175-9936-6ad8eae64b2a

theosanderson:
Ah sorry – looks like I failed to push commits or something.

theosanderson:
OK this is now pushed properly - I'm aware that it's totally broken but it just provides a sense

PR #3396, opened 2024-12-06
perf(silo-prepro): speed up by using more efficient jq command

corneliusroemer:
It truned out that `jq -c '.'` was rate limiting and causing significant extra processing time.

New command has been tested for correctness and performance, still validates.

see https://loculus.slack.com/archives/C05G172HL6L/p1733500150097089?thread_ts=1733478069.696049&cid=C05G172HL6L

https://speed-up-silo-prepro.loculus.org


anna-parker (approved):
lgtm

PR #3397, opened 2024-12-06
style: Trying out a new active-filter style

fhennig:
https://filterstyle.loculus.org

Idea is to make it less like an input, and maybe also use some color as accents on the modal. Also use _display names_ for the fields instead of the URL params.

I'll see after the weekend if i still like it!

### Screenshot

image

<details>
<summary>New 'selected sequences' filter</summary>

https://github.com/user-attachments/assets/b0d4be11-d960-443d-b1a9-7939301932bb

</details>
<details>
<summary>before</summary>

image

</details>

### PR Checklist

### DIFF
--- a/website/src/components/SearchPage/DownloadDialog/DownloadDialog.spec.tsx
+++ b/website/src/components/SearchPage/DownloadDialog/DownloadDialog.spec.tsx
@@ -49,7 +49,7 @@ describe('DownloadDialog', () => {
     });
 
     test('should generate the right download link from filters', async () => {
-        await renderDialog(new FieldFilter({ accession: ['accession1', 'accession2'], field1: 'value1' }, {}));
+        await renderDialog(new FieldFilter({ accession: ['accession1', 'accession2'], field1: 'value1' }, {}, []));
         await checkAgreement();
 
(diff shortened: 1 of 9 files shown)


theosanderson:
Definite improvement! It maybe now feels more salient than the form inputs, which may not be exactly what we want, but it looks really nice.

fhennig:
Thank you! Yes I agree regarding saliency. I think on here it's better:

image

Maybe more color in the controls is the way to go? I might go ahead and try that for the download page.

fhennig:
I tried it

image

But I can't tell, it might be too colorful 😅 I think it's time to go to bed lol

theosanderson:
Yeah, I think the Download form has always felt a bit of a distinct design from our main one - so personally I like the new approach for consistency!

emmahodcroft:
I really like this! Very cool. And thanks for the video and the before!

theosanderson (approved):
Thanks!

fhennig:
Thanks for the review/approval!

I want to merge this into main on it's own though, to keep the history a bit cleaner (so I'm waiting for the other PR to get merged first)

PR #3402, opened 2024-12-09
fix(prepro): don't crash when backend sends error instead of valid ndjson

corneliusroemer:
preview URL: https://no-crash.loculus.org

### Summary
- Backend sometimes sends error message rather than valid ndjson in `extract-unprocessed-entries` response, in particular when there's more than 1 prepro pipeline replicate
- Prepro shouldn't crash for known issues, just sleep for a bit then continue


theosanderson (approved):

anna-parker (approved):
Thanks!!

PR #3403, opened 2024-12-09
fix(prepro): avoid type error when raising exception

corneliusroemer:
Fix small bug in #3402

### DIFF
--- a/preprocessing/nextclade/src/loculus_preprocessing/backend.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/backend.py
@@ -81,7 +81,6 @@ def parse_ndjson(ndjson_data: str) -> Sequence[UnprocessedEntry]:
             json_object = json.loads(json_str_processed)
         except json.JSONDecodeError as e:
             error_msg = f"Failed to parse JSON: {json_str_processed}"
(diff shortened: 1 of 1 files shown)


preprocessing/nextclade/src/loculus_preprocessing/backend.py
```diff
@@ -81,7 +81,8 @@ def parse_ndjson(ndjson_data: str) -> Sequence[UnprocessedEntry]:
             json_object = json.loads(json_str_processed)
         except json.JSONDecodeError as e:
             error_msg = f"Failed to parse JSON: {json_str_processed}"
-            raise json.JSONDecodeError(error_msg) from e
+            # raise json.JSONDecodeError(errog. ValueError
```
anna-parker, line 84:
```suggestion
            # json.JSONDecodeError(error) leads to ValueError 
```

anna-parker (approved):

preprocessing/nextclade/src/loculus_preprocessing/backend.py
corneliusroemer, line 84:
```suggestion
```

corneliusroemer:
Ah thanks I had a stupid comment left over there!

PR #3409, opened 2024-12-09
feat(website): Limit organism card container width when there are 5 organisms

theosanderson:
Limits organisms to be 3 wide when there are 5 organisms

https://theosanderson-patch-1.loculus.org/


corneliusroemer (approved):
Thanks! Centering would be nice eventually, but this is better than 4-1 in any case :)

PR #3410, opened 2024-12-09
fix(website): word-break: break-word;

theosanderson:
Found by @corneliusroemer. Fixes overflows on seq details page. https://loculus.slack.com/archives/C05G172HL6L/p1733766470652209?thread_ts=1733766069.701169&cid=C05G172HL6L

----
Co-authored-by: Cornelius Roemer <corneliusroemer@users.noreply.github.com>

### DIFF
--- a/website/src/styles/base.scss
+++ b/website/src/styles/base.scss
@@ -109,6 +109,7 @@ td {
 
 * {
     @apply break-words;
+    word-break: break-word;
 }
 
 .rs-input {


corneliusroemer (approved):
We might actually want to use `@apply break-words` instead of `word-break: break-word`

Brave Browser 2024-12-09 19 04 33

https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap

theosanderson:
We already have that - see the diff

theosanderson:
image

PR #3419, opened 2024-12-09
fix(website): Set min width to 130px for columns without defined widths

theosanderson:
Now that we have lots of columns with defined min widths on PPX, when you add a new column it is given essentially zero width, and looks absurd.

image

Preview: https://theosanderson-patch-1.loculus.org/

Instead we should give a sensible minimum width, for now I'm going for 130px.

image


anna-parker (approved):
Thanks @theosanderson Much better than the current state! would be nice to make the togglebar a bit more visible when we add more columns as its easy to miss that there are more columns added on the right - but not blocking :-)

PR #3421, opened 2024-12-10
feat(ingest): Improve memory usage

anna-parker:
resolves #

preview URL: https://ingest-mem-improvements.loculus.org/

### Summary

Currently format_ncbi_metadata.py formats each line in a jsonl but then saves each line to a dictionary before writing to a tsv file. Instead we could immedietly write each line to a tsv using csv. 

This is a drastic memory reduction for large datasets, e.g. the peak memory usage for Influenza A is currently 2600M (using htop) with this change we decrease to a constant 14M memory usage. Additionally, we even get a small speed-up (shaving off 0.3sec - seen by adding benchmarking to snakemake rule) as I remove the intermediate pandas df conversion step. 

### PR Checklist

### DIFF
--- a/ingest/scripts/filter_out_depositions.py
+++ b/ingest/scripts/filter_out_depositions.py
@@ -1,3 +1,4 @@
+import csv
 import logging
 from dataclasses import dataclass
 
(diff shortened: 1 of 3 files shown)


chaoran-chen (approved):

anna-parker:
Ohh god this is so stupid - I just remembered that I used pandas for this because of escape characters... when I use pandas it nicely removes the escape characters from \"VECTOR\" so that they are not shown on the website: 
image
But I cannot do this with the csv library for some reason - it is stricter and has to write ",),( with the escape character - which isn't removed:
image 

I think it would be good to just replace all " in ingest with ' as the metadata tsv now looks like `"State Research Center of Virology and Biotechnology ""VECTOR"", Laboratory of Virology"` for these cases which is ugly.

PR #3425, opened 2024-12-11
feat(website): Make submission template configurable

### LINKED ISSUE #3331: Submission template: have a subset of fields in template
We have an overwhelming number of fields. We should make the default template only have a subset of fields, and in a specific order that is useful. People can always add more fields based on the docs.

### Acceptance criteria

The values.yaml `schema` object has a new optional property, that looks like this:

```
metadataTemplate: [sampleCollectionDate, geoLocCountry]
```

It has a list of field names, which are used to construct the metadata template in place of the full list if set. (submissionId is always added at the front to whatever is in the list).

fhennig:
resolves #3331

preview URL: http://submission-template.loculus.org

### PR Checklist

### DIFF
--- a/website/src/config.ts
+++ b/website/src/config.ts
@@ -68,6 +68,11 @@ export function getSchema(organism: string): Schema {
     return getConfig(organism).schema;
 }
 
+export function getMetadataTemplateFields(organism: string): string[] {
+    return getConfig(organism).schema.inputFields.map((field) => field.name);
+    // TODO
+}
+
 export function getRuntimeConfig(): RuntimeConfig {
     if (_runtimeConfig === null) {
         _runtimeConfig = readTypedConfigFile('runtime_config.json', runtimeConfig);
--- a/website/src/config.ts
+++ b/website/src/config.ts
@@ -69,8 +69,11 @@ export function getSchema(organism: string): Schema {
 }
 
 export function getMetadataTemplateFields(organism: string): string[] {
+    const schema = getConfig(organism).schema;
+    if (schema.metadataTemplate !== undefined) {
+        return schema.metadataTemplate;
+    }
     return getConfig(organism).schema.inputFields.map((field) => field.name);
-    // TODO
 }
 
 export function getRuntimeConfig(): RuntimeConfig {
--- a/website/src/config.ts
+++ b/website/src/config.ts
@@ -18,9 +18,33 @@ function getConfigDir(): string {
     return configDir;
 }
 
+function validateWebsiteConfig(config: WebsiteConfig): Error[] {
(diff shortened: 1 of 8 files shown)


chaoran-chen:
> maybe validate that the fields exist?

The fields don't need to be the same as the fields that LAPIS receives and the backend cares about, so at the moment, I don't think that there is much to validate.

fhennig:
I'm validating that the fields are part of the `inputFields` [`03f4267` (#3425)](https://github.com/loculus-project/loculus/pull/3425/commits/03f42672d57929c60b66849cfcfc5633d4dfbea5) (Theo told me about this)

website/src/config.ts
```diff
@@ -18,9 +18,33 @@ function getConfigDir(): string {
     return configDir;
 }
 
+function validateWebsiteConfig(config: WebsiteConfig): Error[] {
```
theosanderson, line 21: I suspect think this is probably OK, but we should be wary that we need to get the config for pretty much every single thing we ever do on the website, so if we start to add significant milliseconds of validation that's probably a problem, so we should be wary of expanding this pattern too much

theosanderson (approved):
lgtm, thanks!

website/src/config.ts
fhennig: The validated config is cached in _config, validation only happens on initial read

fhennig:
```typescript
export function getWebsiteConfig(): WebsiteConfig {
    if (_config === null) {
        _config = readTypedConfigFile('website_config.json', websiteConfig);
        const config = readTypedConfigFile('website_config.json', websiteConfig);
        const validationErrors = validateWebsiteConfig(config);
        if (validationErrors.length > 0) {
            throw new AggregateError(validationErrors, 'There were validation errors in the website_config.json');
        }
        _config = config;
    }
    return _config;
}
```

But now that I read it, if there are errors, they will keep getting parsed 😅

theosanderson: Are you saying you think this code won't be executed every time a page is loaded? I'm 90% sure that it will - you can check by adding a print statement.

theosanderson: (but no that's probably not what you are saying - I was unclear, I meant "every route" essentially)

fhennig: I got curious! I just checked with a print. The part inside of the `if` is only executed once! My understanding is that `_config` is a global variable that exists once in the server, so it is also there across page loads.

theosanderson: oh what! ok sorry, my bad!!

theosanderson: Thanks for educating me :)

Issue #3334, opened 2024-12-02
Accept Excel files for metadata


emmahodcroft:
A _lot_ of people use Excel to handle and store metadata. Whether rightly or wrongly, it is what it is. We will lower the bar for upload significantly if we can support Excel metadata uploads. 

We would probably need to force people to set the dates into ISO still, unless Excel encodes whether dates are US format or not (which it well might) to avoid confusion of whether `04/07/2022` is 7 of April or 4 of July.

anna-parker:
I fully agree, I think every beta tester we have talked to so far has raised this exact point, they use excel and would like more support for it

fhennig:
Good point about the dates! I did the PoC now https://github.com/loculus-project/loculus/pull/3426 and looked into the docs of the library I used. It seems like excel internally stores date as an offset, the stuff seen in the UI is not actually stored in the file verbatim (source: [SheetJS docs](https://docs.sheetjs.com/docs/csf/features/dates/#how-sheetjs-handles-dates-and-times)). The library seems to come with some date parsing support but I'm not sure which information you need to give it (i.e. which columns are dates?) -- would need to look into it more.

PR #3426, opened 2024-12-11
POC: Excel upload

fhennig:
resolves #

preview URL: http://xslx-poc.loculus.org

### Summary

WIP

### Screenshot

image
image

### PR Checklist

### DIFF
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -1,0 +1,115 @@
+{
+  "name": "backend",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "dependencies": {
+        "xlsx": "^0.18.5"
+      }
+    },
+    "node_modules/adler-32": {
+      "version": "1.3.1",
+      "resolved": "https://registry.npmjs.org/adler-32/-/adler-32-1.3.1.tgz",
+      "integrity": "sha512-ynZ4w/nUUv5rrsR8UUGoe1VC9hZj6V5hU9Qw1HlMDJGEJw5S7TfTErWTjMys6M7vr0YWcPqs3qAr4ss0nDfP+A==",
+      "license": "Apache-2.0",
+      "engines": {
+        "node": ">=0.8"
+      }
+    },
+    "node_modules/cfb": {
+      "version": "1.2.2",
+      "resolved": "https://registry.npmjs.org/cfb/-/cfb-1.2.2.tgz",
+      "integrity": "sha512-KfdUZsSOw19/ObEWasvBP/Ac4reZvAGauZhs6S/gqNhXhI7cKwvlH7ulj+dOEYnca4bm4SGo8C1bTAQvnTjgQA==",
+      "license": "Apache-2.0",
+      "dependencies": {
+        "adler-32": "~1.3.0",
+        "crc-32": "~1.2.0"
+      },
+      "engines": {
+        "node": ">=0.8"
+      }
+    },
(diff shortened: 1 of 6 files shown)


fhennig:
While doing this I realized that by default (I think) the file wouldn't be read entirely, but just streamed to the API by the browser. If the files are large, it would be cool if this functionality could be preserved. We could just handle the conversion on the server. _But_ if we want to allow the user to select a column mapping, then we have to read the file (at least in part)

the `xlsx` library seems to make it really easy though!

fhennig:
Ok, I think I'm done exploring!

I wanted to find out: _How easy is it to allow users to upload excel files?_ and _can we allow users to map table columns in the UI? (instead of having to do it in the file themselves.

My result:
- Excel seems very easy (at least with simple files)
- table column mapping too

On the branch so far:
- you should be able to upload excel files and they will be converted. You will see that it says `converted.tsv` as the filename.
- There is a button "Edit columns" and a modal will open where you can define a mapping from metadata columns to your table columns. The dropdowns have the file table columns in them already. But you can't _actually_ do the edit yet (I just wanted to see whether it's feasible).

Only caveat: excel files will be read into memory entirely - but this is only a problem for very large files, which we won't encounter I believe. (If you upload a TSV file and don't edit the columns, it will work as before)

rneher:
Very nice! 

I think the standard metadata columns should be available in the client (@theosanderson, @corneliusroemer ?) and it would be nice to highlight the required ones visually. 

I am not sure whether mapping input column to database columns, or vice versa is more intuitive. Since we have an enormous number of potential columns, showing the columns detected in the uploaded file and allowing to pick the corresponding Loculus columns might be easier (required columns could be a top section in a drop-down). Unmapped columns could be ignored.

theosanderson:
Great stuff!

> I think the standard metadata columns should be available in the client (@theosanderson, @corneliusroemer ?) and it would be nice to highlight the required ones visually.

The only bit missing for this is that `inputFields` which are the inputs that people supply, don't currently have a `required` attribute to specify if they are required. That is something we should have. (It's confusing because fields as a whole do have this, but not `inputFields` which are technically separate). This is hard to describe and it would be easy to derail this whole thread discussing it so we should probably discuss in a new issue. I created #3430.

> showing the columns detected in the uploaded file and allowing to pick the corresponding Loculus columns might be easier (required columns could be a top section in a drop-down)

This seems to be the actual behaviour of Felix's mock-up atm (which I agree is good) but the column headings are mislabelled (I know it's just a test!).

image

Something that would be nice would be to automatically map on the basis of display names as well as technical names. Then I think we might change the templates to use display names `Country` not `geoLocCountry`.

fhennig:
Thanks both of you! 

To track and prioritize I've created issues for both excel file upload (https://github.com/loculus-project/loculus/issues/3431) and column remapping (https://github.com/loculus-project/loculus/issues/3432) -- I captured your remarks about auto mapping and required fields in there. I also added some questions I have still to both (more of a technical nature).

fhennig:
I'm closing this for now but reference the PR in both issues.

fhennig:
Emma pointed out here https://github.com/loculus-project/loculus/issues/3334 that dates might need special handling, and indeed the SheetJS docs show that dates aren't stored in some ISO format, but as an integer offset to some base date. Should I maybe investigate a bit more? 

SheetJS docs: https://docs.sheetjs.com/docs/csf/features/dates/#how-sheetjs-handles-dates-and-times

PR #3428, opened 2024-12-11
chore(helm): move earliestReleaseDate to _common-metadata.tpl

corneliusroemer:
Followup to #3380

Hard coded metadata fields are configured in _common-metadata.tpl as they are not something to be configured per instance via values.yaml

Disabling per organism isn't easy for common-metadata fields, so let's remove the ability to disable, earliest release is something that can always be calculated so no need to disable it. Let's only make configurable what truly needs to be configurable.

resolves #

preview URL:

### PR Checklist

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/config/Config.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/config/Config.kt
@@ -78,4 +78,4 @@ data class ExternalMetadata(
     override val required: Boolean = false,
 ) : BaseMetadata()
 
-data class EarliestReleaseDate(val enabled: Boolean = false, val externalFields: List<String>)
+data class EarliestReleaseDate(val externalFields: List<String>)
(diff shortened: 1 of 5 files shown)


chaoran-chen (changes requested):
Could you please explain more why this should be not disableable? I don't see how the current code is problematic and I think that the feature is useful. I can well imagine using it very soon for one of the wastewater instances.

corneliusroemer:
Well, I never approved the original PR.

`earliestReleaseDate` is a magic field name, hence it should live in the `_common-metadata.tpl` like other computed fields.

The argument must be the other way round, why _should_ it be disableable, this wasn't part of the original discussion so it's an extra feature that was added. What harm is done?

Disabling at per-organism level is not straightforward with current setup of common-metadata. What would work is disable entirely for the entire instance, which should be enough.

theosanderson:
My 2 cents:

I agree with Cornelius that the current version doesn't fit into our logic so well.

Nevertheless, it doesn't seem a big problem for me, and if it's going to be useful for Chaoran the current behaviour seems ok for me for now.

I think we should aim to make common metadata items configurable in terms of questions like whether they appear in tables and stuff relatively soon, and would support a change somewhat like this after that.

chaoran-chen:
> this wasn't part of the original discussion so it's an extra feature that was added

@corneliusroemer, that's not true, we discussed it a lot in #2894. You didn't respond at the end to [the question](https://github.com/loculus-project/loculus/issues/2894#issuecomment-2506390303) whether adding the disabling is acceptable for you but didn't object either but said that the issue was ready-to-go.

> Disabling at per-organism level is not straightforward with current setup of common-metadata. What would work is disable entirely for the entire instance, which should be enough.

An instance-wide disabling would be fine with me as well.

PR #3429, opened 2024-12-11
feat(ingest): Add option to use nextclade sort to identify segments for multi-segmented organisms

anna-parker:
resolves #

preview URL: https://add-fast-segment-identifi.loculus.org/

### Summary

1. Add option in Snakefile to either determine segments via [nextclade sort](https://docs.nextstrain.org/projects/nextclade/en/stable/user/nextclade-cli/reference.html#nextclade-sort) or nextclade align. Nextclade sort is extremely fast and when used on influenza data produces highly accurate results. Similar to the nextclade dataset requirement for nextclade align nextclade sort requires a minimzer.json. 
2. Add script to parse nextclade sort output, set dataset with highest score as segment -> allow option to parse `dataset` (identified sorted group's name) into additional metadata fields and add them to the metadata tsv during prepare_metadata. 
3. Fix small bug in prepro.

See this isn action here: https://virus4.loculus.org/ code here: https://github.com/loculus-project/private_deployments/blob/main/deploy/virus4/values.yaml

### PR Checklist

- [ ] Update config docs?

### Future work
- [ ] Use for CCHF to make ingest faster/ less CPU intensive.

### DIFF
--- a/ingest/scripts/call_loculus.py
+++ b/ingest/scripts/call_loculus.py
@@ -205,8 +205,8 @@ def post_fasta_batches(
     logger.info(df.columns)
     sequences_dict = {record.id: record for record in SeqIO.parse(fasta_file, "fasta")}
     submission_ids = df["submissionId"].tolist()
-    for batch_num, submission_id_chunk in (
-        (i, submission_ids[i : i + chunk_size]) for i in range(0, len(submission_ids), chunk_size)
+    for batch_num, submission_id_chunk in enumerate(
+        submission_ids[i : i + chunk_size] for i in range(0, len(submission_ids), chunk_size)
     ):
         matching_fasta_records = filter_fasta_by_submission_ids(
             sequences_dict, submission_id_chunk, config
@@ -219,7 +219,10 @@ def post_fasta_batches(
         metadata = df[df["submissionId"].isin(submission_id_chunk)]
         metadata_output_file = "results/batch_metadata.tsv"
         metadata.to_csv(metadata_output_file, sep="\t", index=False, float_format="%.0f")
-        with open(metadata_output_file, "rb") as metadata_ , open(sequences_output_file, "rb") as fasta_:
+        with (
+            open(metadata_output_file, "rb") as metadata_,
+            open(sequences_output_file, "rb") as fasta_,
+        ):
             files = {
                 "metadataFile": metadata_,
                 "sequenceFile": fasta_,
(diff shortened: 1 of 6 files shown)


preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -591,7 +591,9 @@ def process_single(  # noqa: C901
         unaligned_nucleotide_sequences = unprocessed.unalignedNucleotideSequences
 
     for segment in config.nucleotideSequences:
-        sequence = unaligned_nucleotide_sequences[segment]
+        sequence = unaligned_nucleotide_sequences.get(segment, None)
```
anna-parker, line 594: This was only a bug for the not-aligned organism pipeline - I guess I never tested it on a multi-segmented organism.

ingest/scripts/call_loculus.py
```diff
@@ -171,6 +173,65 @@ def get_or_create_group_and_return_group_id(config: Config, allow_creation: bool
...
+        ]
+    return [
+        fasta_dict[submission_id] for submission_id in submission_ids if submission_id in fasta_dict
+    ]
+
+
+def post_fasta_batches(
+    url,
+    fasta_file: str,
+    metadata_file: str,
```
fhennig, line 198: I _think_ the types here might be off? In line 273 you pass in a file pointer not a file name, so it's not `str` at least for the `fasta_file`. If `SeqIO.parse` can read a filename, you might be able to just ditch the `open(..)` in line 273 and instead pass in file names?

ingest/scripts/call_loculus.py
```diff
@@ -171,6 +173,65 @@ def get_or_create_group_and_return_group_id(config: Config, allow_creation: bool
...
+    config: Config,
+    params: dict[str, str],
+    chunk_size=60000,
+) -> requests.Response:
+    """Chunks metadata files, joins with sequences and submits each chunk via POST."""
+    df = pd.read_csv(metadata_file, sep="\t")
+    logger.info(df.columns)
+    sequences_dict = {record.id: record for record in SeqIO.parse(fasta_file, "fasta")}
+    submission_ids = df["submissionId"].tolist()
+    for batch_num, submission_id_chunk in (
```
fhennig, line 208: `batch_num` sounds like it'd be a the batch number (0, 1, 2, 3, ...) but it's actually (0, 60000, ...)

Another variable name would be good, but I also can't think of one so feel free to ignore.

fhennig (approved):
LGTM! Just a few minor things, but feel free to ignore

ingest/scripts/prepare_metadata.py
```diff
@@ -71,15 +71,14 @@ def main(
...
-        # Segments are a tsv file with the first column being the fasta id
-        # and the second being the segment
-        segments_dict: dict[str, str] = {}
-        with open(segments, encoding="utf-8") as file:
-            for line in file:
-                if line.startswith("seqName"):
-                    continue
-                fasta_id, segment = line.strip().split("\t")
-                segments_dict[fasta_id] = segment
+        segments_dict: dict[str, dict[str, str]] = {}
```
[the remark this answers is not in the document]
anna-parker: Ah yes I can do that (you don't even need to import TypeAlias in python
```
AccessionID = str
```
is enough)

fhennig:
Thanks!!

PR #3433, opened 2024-12-12
feat(backend): Return `pipelineVersion` in `get-released-data`

### LINKED ISSUE #3209: Include pipeline version in 'get-released-sequences' as a metadata field
@fengelniederhammer raised a good point that we don't currently emit the pipeline version used to process sequences at all. This isn't great, the only way to work around this right now is to query the db.

We should somehow include the pipeline version in the response of `get-released-sequences`.

As it's NDJSON we can't put it in an info top-level-key unfortunately.

So we have options to:
- include in HTTP header
- emit as metadata field

We should probably just put it as a metadata field, that would also allow us in the future to use different pipeline versions for different organisms, something we've wanted to enable in the future to not have to reprocess everything if a single organism's Nextclade dataset is updated.

Some background discussion: https://loculus.slack.com/archives/C05G172HL6L/p1731500439841509

_______

Acceptance criteria:

- The `get-released-sequences` endpoint contains a `pipelineVersion` field for each item with the pipeline version (which will be the same for all sequences)
- The kubernetes config is adjusted to add this field to `kubernetes/loculus/templates/_common-metadata.tpl`

fhennig:
resolves #3209 

preview URL: https://backend-pipeline-v-in-metadata.loculus.org/swagger-ui/index.html#/submission-controller/getReleasedData
### Summary

WIP

### Screenshot

n/A

### PR Checklist

### DIFF
--- a/backend/docs/db/schema.sql
+++ b/backend/docs/db/schema.sql
@@ -464,7 +464,11 @@ CREATE VIEW public.sequence_entries_view AS
     sepd.finished_processing_at,
     sepd.processed_data,
     (sepd.processed_data || em.joint_metadata) AS joint_metadata,
-    sepd.pipeline_version,
+        CASE
+            WHEN se.is_revocation THEN ( SELECT current_processing_pipeline.version
+               FROM public.current_processing_pipeline)
+            ELSE sepd.pipeline_version
+        END AS pipeline_version,
     sepd.errors,
     sepd.warnings,
         CASE
--- a/backend/docs/db/schema.sql
+++ b/backend/docs/db/schema.sql
@@ -464,6 +464,7 @@ CREATE VIEW public.sequence_entries_view AS
     sepd.finished_processing_at,
     sepd.processed_data,
     (sepd.processed_data || em.joint_metadata) AS joint_metadata,
+    sepd.pipeline_version,
     sepd.errors,
     sepd.warnings,
         CASE
--- a/backend/src/main/resources/db/migration/V1.9__update_sequence_entries_view_add_pipeline_version.sql
+++ b/backend/src/main/resources/db/migration/V1.9__update_sequence_entries_view_add_pipeline_version.sql
@@ -7,7 +7,10 @@ select
     sepd.finished_processing_at,
     sepd.processed_data as processed_data,
     sepd.processed_data || em.joint_metadata as joint_metadata,
-    sepd.pipeline_version,
+    case
+        when se.is_revocation then (select version from current_processing_pipeline)
+        else sepd.pipeline_version
+    end as pipeline_version,
     sepd.errors,
     sepd.warnings,
     case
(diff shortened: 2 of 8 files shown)


backend/src/main/resources/db/migration/V1.9__update_sequence_entries_view_add_pipeline_version.sql
```diff
@@ -0,0 +1,41 @@
...
+select
+    se.*,
+    sepd.started_processing_at,
+    sepd.finished_processing_at,
+    sepd.processed_data as processed_data,
+    sepd.processed_data || em.joint_metadata as joint_metadata,
+    case
+        when se.is_revocation then (select version from current_processing_pipeline)
+        else sepd.pipeline_version
+    end as pipeline_version,
```
fhennig, line 13: These lines are the ones I actually changes, just FYI

fhennig:
Not sure about if/where to document this.

fhennig:
Oops sorry Fabian I missclicked, didn't mean to ping you about this 😅

theosanderson (approved):
lgtm!

backend/src/main/resources/db/migration/V1.9__update_sequence_entries_view_add_pipeline_version.sql
chaoran-chen: But why not just this?

```suggestion
    (select version from current_processing_pipeline) as pipeline_version,
```

fhennig:
Thanks for the quick review @theosanderson  🥳

backend/src/main/resources/db/migration/V1.9__update_sequence_entries_view_add_pipeline_version.sql
chaoran-chen: We know that this is always the same as `sepd.pipeline_version` as we explicitly filter for it below

```sql
        and sepd.pipeline_version = (select version from current_processing_pipeline)
```

chaoran-chen: Thanks Felix for explaining over a call: the results are different for unprocessed entries and I agree that the current solution is better.

backend/docs/db/schema.sql
```diff
@@ -464,6 +464,11 @@ CREATE VIEW public.sequence_entries_view AS
     sepd.finished_processing_at,
     sepd.processed_data,
     (sepd.processed_data || em.joint_metadata) AS joint_metadata,
+        CASE
+            WHEN se.is_revocation THEN ( SELECT current_processing_pipeline.version
```
corneliusroemer, line 468: Why don't we just send the pipeline version here for everything? Maybe I'm missing something, might be good to explain why the complexity is necessary?

theosanderson: discussed in https://github.com/loculus-project/loculus/pull/3433#discussion_r1882006866 (I haven't really followed the discussion)

Issue #3074, opened 2024-10-25
Lineage search field with sub-lineage queries


chaoran-chen:
In [LAPIS v0.3.7](https://github.com/GenSpectrum/LAPIS/releases/tag/v0.3.7), the sub-lineage filter was generalized, making it possible to include sub-lineages of a lineage by appending the query with a `*`. In Loculus, we should incorporate this feature.

Sub-tasks include (probably not complete):

- Website: add UI to send corresponding queries and allow users to decide whether to include sublineages
- Deployment and config: pass the lineage tree definition file to SILO (and LAPIS?) and the backend
- Backend: If a field has the new lineage type, SILO expects that only values appear that are also present in the lineage tree definition file. The backend, therefore, should validate the lineages that the preprocessing pipeline generate.

Taepper:
We might need to reconsider responsibilities regarding lineage validations.

For SILO 0.3.0 it was decided, that we validate all input values for lineage columns. This means that they need to be either `NULL` or a _valid_ lineage. Validity is defined through presence in the lineage definition file ([covid example](https://github.com/GenSpectrum/LAPIS-SILO/blob/main/testBaseData/exampleDataset/lineage_definitions.yaml)) as defined [here](https://github.com/GenSpectrum/LAPIS-SILO/issues/458).

This validation has advantages as any _invalid_ values could lead to misleading behavior to the user. Incorrect specification of lineage definition file might make the user assume that lineages are case-insensitive as there is no error when he inputs lower-case lineages. Also, absence of errors might make users not detect that a new lineage is in the data, without it being present in the tree definition.

However, SILO preprocessing that produces errors can be a little cumbersome to deal with from the loculus side of things. Therefore some proper validation needs to be present in a solution to this issue. It would also be possible (and not too much effort) to make SILO "errors on unknown values" configurable and thus optional. If this were chosen, the warnings emitted by SILO should be propagated _somewhere_ to prevent the above mentioned misleading behavior from appearing.

Otherwise, proper validation should most likely appear before the SILO import script is called, as the errors would be a lot easier to deal with in the backend

chaoran-chen:
[...]

PR #3434, opened 2024-12-12
feat: Add lineage definition file to SILO import config

fhennig:
~~resolves~~ part of #3074 

preview URL: https://lineage-validation.loculus.org

### Summary
- make the lineage definition configurable in the helm chart
- provide the lineage definition file URLs to the silo import job
- configure silo import to use the lineage definition file

### Screenshot

n/A

### PR Checklist

### DIFF
--- a/docs/src/content/docs/reference/helm-chart-config.mdx
+++ b/docs/src/content/docs/reference/helm-chart-config.mdx
@@ -871,7 +871,7 @@ Each organism object has the following fields:
                 Use this on string fields that contain lineages.
                 The value needs to be a lineage system that is defined under the `lineageSystemDefinitions` key.
             </td>
-        <tr>
+        </tr>
     </tbody>
 </table>
 
--- a/docs/src/content/docs/reference/helm-chart-config.mdx
+++ b/docs/src/content/docs/reference/helm-chart-config.mdx
@@ -451,10 +451,10 @@ Here's an example of a `lineageDefinitions` section:
 ```yaml
 lineageSystemDefinitions:
   pangoLineage:  # Lineage name to use in metadata fields
-    "1": https://example.org/lineage_defintions_v1.yaml  # Definition per pipeline version
-    "2": https://example.org/lineage_defintions_v2.yaml
+    1: https://example.org/lineage_defintions_v1.yaml  # Definition per pipeline version
+    2: https://example.org/lineage_defintions_v2.yaml
   myLineage:
-    "1": ...
+    1: ...
 ```
 
 <table>
--- a/docs/src/content/docs/reference/helm-chart-config.mdx
+++ b/docs/src/content/docs/reference/helm-chart-config.mdx
@@ -868,7 +868,7 @@ Each organism object has the following fields:
             <td>String</td>
             <td></td>
             <td>
-                Use this on string fields that contain lineages.
+                Use this on string fields that contain lineages, if you want to enable searches that can include sublineages.
                 The value needs to be a lineage system that is defined under the `lineageSystemDefinitions` key.
             </td>
         </tr>
--- a/docs/src/content/docs/reference/helm-chart-config.mdx
+++ b/docs/src/content/docs/reference/helm-chart-config.mdx
@@ -434,6 +434,51 @@ For production environments, these should always be set to false. Instead, exter
             <td></td>
             <td>An object where the keys are the organism IDs and values are an [Organism (type)](#organism-type)</td>
         </tr>
+        <tr>
+            <td>`lineageSystemDefinitions`</td>
+            <td>Object</td>
+            <td></td>
+            <td>An object where the keys are the lineage system names and values are links to lineage system definition files per pipeline version (See [Lineage system defintions](#lineage-type))</td>
+
+        </tr>
+    </tbody>
+</table>
+
+### Lineage system definitions
+
+Here's an example of a `lineageDefinitions` section:
+
+```yaml
+lineageSystemDefinitions:
+  pangoLineage:  # Lineage name to use in metadata fields
+    "1": https://example.org/lineage_defintions_v1.yaml  # Definition per pipeline version
+    "2": https://example.org/lineage_defintions_v2.yaml
+  myLineage:
+    "1": ...
+```
+
+<table>
+    <thead>
+        <tr>
+            <th className='w-48'>Field</th>
+            <th>Type</th>
+            <th>Default</th>
+            <th>Description</th>
+        </tr>
+    </thead>
+    <tbody>
+        <tr>
+            <td>`lineageSystemDefinitions.<name>`</td>
+            <td>Object</td>
+            <td></td>
+            <td>A map from pipeline versions to file URLs.</td>
+        </tr>
+        <tr>
+            <td>`lineageSystemDefinitions.<name>.<pipelineVersion>`</td>
+            <td>`String`</td>
+            <td></td>
+            <td>The URL to the lineage defintion file for that lineage system and that pipeline version.</td>
+        </tr>
     </tbody>
 </table>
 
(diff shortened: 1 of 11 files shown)


kubernetes/loculus/templates/lapis-silo-deployment.yaml
```diff
@@ -2,6 +2,9 @@
 {{- $keycloakTokenUrl := "http://loculus-keycloak-service:8083/realms/loculus/protocol/openid-connect/token" }}
 
 {{- range $key, $_ := (.Values.organisms | default .Values.defaultOrganisms) }}
+# TODO go through the schema for the organism and gather all the metadata entries that are of lineage type
+# TODO There can be multple fields, but they all need to have the same lineageSystem ... not sure how to handle yet
```
theosanderson, line 6: As a starting point we could potentially just state that there can be only one `lineage` field

theosanderson: your solution looks nice!

fhennig: I just added a commit, where I was thinking about having a new lineageSystems section in the Helm chart. It's then kind of easy to ensure only one system is sued per organism. Good, or a needless level of indirection?

Also happy to go with the "Only one lineage field" constraint for now (will make templating a bit simpler)

theosanderson: Happy with all options. Until LAPIS/SILO supports multiple lineage systems, it's all a bit academic. We always can change it when LAPIS/SILO does implement something, so if it was me, I'd probably just have the simplest option for now of per-organism:

```
  name: organismName
  lineageDefs:
    1: http://....
```

but I don't mind

fhennig:
I think I got something that _should work_ in principle, will try and test it tomorrow/monday. It's not super obvious how to test this. Could we define an actual lineage field with a lineage etc. for one of the test organisms we have? Would be cool.

fhennig:
The file is downloaded fresh every time. _Maybe_ the files for all versions could be downloaded once and then just copied around but IMO I wouldn't introduce all this complexity in the shell script just to save us a download every now and then. But happy to implement if there's disagreement.

fengelniederhammer:
> The file is downloaded fresh every time. _Maybe_ the files for all versions could be downloaded once and then just copied around but IMO I wouldn't introduce all this complexity in the shell script just to save us a download every now and then. But happy to implement if there's disagreement.

It's past time we migrate the silo import script to a proper language (Python?) - it grows and grows and becomes more unmaintainable with every bit that we add.

kubernetes/loculus/silo_import_job.sh
```diff
@@ -148,6 +151,46 @@ download_data() {
...
+      echo "Error: Failed to download file from $lineage_url."
+      exit 1
+    fi  
+  else
+    echo "Multiple pipeline versions in data to import: $pipelineVersion"
+    exit 1
+  fi
+
+  cp $preprocessing_config_file $preprocessing_config_file_merged
+  echo -e "lineageDefinitionsFilename: \"$lineage_definition_file\"\n" >> $preprocessing_config_file_merged
```
fengelniederhammer, line 191: Does it make sense to wait for the new SILO version that supports supplying this via command line args? I don't know how urgent this change is. https://github.com/GenSpectrum/LAPIS-SILO/pull/633 shouldn't take too long anymore.

fhennig: I don't think this PR is urgent in principle, I think we can wait a bit more, but I would probably want to get this PR closed this week, so I can close everything before the holidays. In that case I'd rather make an issue and link the PR so we can change it later. It's not really that bad the way the config is assembled now I think.

kubernetes/loculus/values.yaml
```diff
@@ -28,6 +28,9 @@ logo:
   url: "/favicon.svg"
   width: 100
   height: 100
+lineageSystemDefinitions:
+  pangoLineage:
+    "1": https://raw.githubusercontent.com/loculus-project/loculus/refs/heads/lineage-validation/preprocessing/dummy/lineage.yaml
```
fhennig, line 33: This link points to this branch, would need to be changed before merging

docs/src/content/docs/reference/helm-chart-config.mdx
```diff
@@ -818,6 +863,15 @@ Each organism object has the following fields:
                 [here](https://github.com/loculus-project/loculus/blob/main/preprocessing/nextclade/README.md#custom-preprocessing-functions).
             </td>
         </tr>
+        <tr>
+            <td>`lineageSystem`</td>
+            <td>String</td>
+            <td></td>
+            <td>
+                Use this on string fields that contain lineages.
```
theosanderson, line 871:
```suggestion
                Use this on string fields that contain lineages, if you want to enable searches that can include sublineages.
```

kubernetes/loculus/silo_import_job.sh
```diff
@@ -148,6 +151,48 @@ download_data() {
   echo
 }
 
+# Generate the preprocessing config file with the lineage file for the current pipeline version.
+# the lineage definition file needs to be downloaded first.
+prepare_preprocessing_config() {
+  rm -f $lineage_definition_file $preprocessing_config_file_merged
+
+  if [[ -z "$LINEAGE_DEFINITIONS" ]]; then
+    echo "No LINEAGE_DEFINITIONS given, nothing to configure;"
```
theosanderson, line 160: I guess this probably means that if someone configures the metadata type to lineage but doesn't define a lineageSystem, that will crash SILO. But to me that feels acceptable

theosanderson: Ah no, maybe that's not even possible

docs/src/content/docs/reference/helm-chart-config.mdx
```diff
@@ -434,6 +434,51 @@ For production environments, these should always be set to false. Instead, exter
...
+</table>
+
+### Lineage system definitions
+
+Here's an example of a `lineageDefinitions` section:
+
+```yaml
+lineageSystemDefinitions:
+  pangoLineage:  # Lineage name to use in metadata fields
+    "1": https://example.org/lineage_defintions_v1.yaml  # Definition per pipeline version
```
theosanderson, line 454: Does this need to be quoted btw? I presume so?

theosanderson (approved):
All looks very reasonable to me! Thanks!!

docs/src/content/docs/reference/helm-chart-config.mdx
fhennig: good point, they actually don't need to be. I was just overly cautious 😅 

I'll remove it.

fhennig: [`25ca1cf` (#3434)](https://github.com/loculus-project/loculus/pull/3434/commits/25ca1cfd34f650a7c315064d447caaaf6d81c063)

fhennig:
@theosanderson What about the link to the lineage defintion file, that currently points to a branch? Should we merge it as is, and then I'd make a second PR to change just that line (once the file is actually in main)?

theosanderson:
> @theosanderson What about the link to the lineage defintion file, that currently points to a branch? Should we merge it as is, and then I'd make a second PR to change just that line (once the file is actually in main)?

If it was me I'd just change it to `main` now. BTW we will hit problems as GitHub will likely start blocking access to this, but I guess we can cross that bridge when we come to it

theosanderson:
(also fine with 2 PRs btw)

fhennig:
> If it was me I'd just change it to main now.

I _think_ then the tests might break after the commit get's pushed.

> (also fine with 2 PRs btw)

Let's just do that then!

theosanderson:
> I think then the tests might break after the commit get's pushed - and then merging is prohibited.
I don't think we enforce (programmatically) tests passing - but your approach is all good!

PR #3450, opened 2024-12-17
chore(website): migrate to eslint 9, fix linter issues where reasonably possible

### LINKED ISSUE #3069: chore: Migrate ESLint 8 to 9
I have disabled the dependabot upgrades for ESLint 9 for now: https://github.com/loculus-project/loculus/pull/3057

A move from 8 to 9 will have to include migrating to the new config format.

I have decided to track this in it's own issue, so we can prioritize doing this. Not sure if it's really that important.

We need to migrate the config file and the automatic [migration tool](https://eslint.org/docs/latest/use/configure/migration-guide#migrate-your-config-file) doesn't work well for `.js` eslintrc files, so it'll be a manual effort.

links:
- https://eslint.org/docs/latest/use/migrate-to-9.0.0
- https://eslint.org/docs/latest/use/configure/migration-guide

### Tasks

- [ ] migrate to the new config format first
- [ ] then bump to the new major version

fengelniederhammer:
resolves #3069

preview URL: https://eslint9.loculus.org/

### Summary

This migrates to Eslint 9 and simplifies the config (more plain plugins, less custom rules). I mostly copied the config from the GenSpectrum dashboards: https://github.com/GenSpectrum/dashboards/pull/373

The main issue in the code was that `@typescript-eslint/no-explicit-any` was disabled.
https://github.com/loculus-project/loculus/blame/d3505b76aa45ef278868c63c55f5d040a29d42c1/website/.eslintrc.cjs#L104C51-L104C51

IMO this is one of the most important rules and must be enabled. `any` should be avoided, because it essentially disabled Typescript. If you really don't know the type, you should use `unknown` and check in the code whether you can access fields/methods of a variable. 

I considered improving the current code outside the scope of this PR. I applied the small and easy fixes where possible and tried to change as little behavior as possible. Fixing the `any` issues should be done in a separate PR. There are parts of the code that basically lack all type information and should be refactored to ensure maintainability of the codebase. If you think that this approach is good, I'll create an issue.

### PR Checklist


chaoran-chen:
Thank you very much, Fabian! Creating an issue to fix the missing types / `any` would be great!

fengelniederhammer:
I created #3451 to fix the `any`s.

corneliusroemer:
Impressive PR, thanks for getting us to eslint 9!

website/package.json
```diff
@@ -11,8 +11,8 @@
         "build": "astro build",
         "preview": "astro preview",
         "astro": "astro",
-        "check-format": "eslint --cache \"./**/*.{ts,tsx,astro}\" && prettier --ignore-path \"../.gitignore\" --check \"./**/*.{ts,tsx,json,astro,cjs,mjs,css,scss}\"",
-        "format": "eslint --cache \"./**/*.{ts,tsx,astro}\" --fix && npm run format-fast",
+        "check-format": "eslint --cache . && prettier --ignore-path \"../.gitignore\" --check \"./**/*.{ts,tsx,json,astro,cjs,mjs,css,scss}\"",
+        "format": "eslint --cache . --fix && npm run format-fast",
```
corneliusroemer, line 15: I think this caused #3461, it's a shame one can't comment inside the package.json - given one can't comment to explain a certain construct it's sometimes better not to change even if one doesn't understand why something is done a certain way, there might be a reason 🙃

Issue #3461, opened 2024-12-17
Eslint errors in `./dist` folder since eslint 9 PR


corneliusroemer:
In a directory where I've run `npm run build` I now get eslint errors since the eslint 9 PR got merged: #3450 

```
/Users/corneliusromer/code/loculus/website/dist/server/chunks/ErrorFeedback_ByRDamcW.mjs
   732:3  warning  Unused eslint-disable directive (no problems were reported)
   922:7  error    Definition for rule 'react/prop-types' was not found            react/prop-types
  1208:5  error    Definition for rule 'react-hooks/rules-of-hooks' was not found  react-hooks/rules-of-hooks

/Users/corneliusromer/code/loculus/website/dist/server/chunks/Pagination_DlSmtfe7.mjs
  75:3  warning  Unused eslint-disable directive (no problems were reported from 'no-nested-ternary')
  80:3  warning  Unused eslint-disable directive (no problems were reported from 'no-nested-ternary')

/Users/corneliusromer/code/loculus/website/dist/server/chunks/useTheme_xLVzobZB.mjs
   615:7  warning  Unused eslint-disable directive (no problems were reported from 'no-restricted-globals')
   692:7  warning  Unused eslint-disable directive (no problems were reported from 'guard-for-in' or 'no-restricted-syntax')
  1412:5  error    Definition for rule 'react-hooks/rules-of-hooks' was not found                                             react-hooks/rules-of-hooks
  1622:5  error    Definition for rule 'react-hooks/rules-of-hooks' was not found                                             react-hooks/rules-of-hooks

/Users/corneliusromer/code/loculus/website/dist/server/pages/seqsets.astro.mjs
   174:5  error    Definition for rule 'react-hooks/exhaustive-deps' was not found  react-hooks/exhaustive-deps
   364:3  warning  Unused eslint-disable directive (no problems were reported)
   449:3  warning  Unused eslint-disable directive (no problems were reported)
  1466:7  error    Definition for rule 'react/prop-types' was not found             react/prop-types
  1895:7  error    Definition for rule 'react/prop-types' was not found             react/prop-types

✖ 14 problems (7 errors, 7 warnings)
  0 errors and 7 warnings potentially fixable with the `--fix` option.
```

fhennig:
I looked at the PR but don't quite understand what the issue was and why the fix works - can you explain? So I don't make the same mistake in the future 😁

PR #3463, opened 2024-12-17
chore(ci): run website build before check scripts to avoid regression of #3461

corneliusroemer:
- Test for regression of #3461
- Also invalidate cache on changes to package.json just in case


website/package.json
```diff
@@ -11,9 +11,9 @@
         "build": "astro build",
         "preview": "astro preview",
         "astro": "astro",
-        "check-format": "eslint --cache \"./**/*.{ts,tsx,astro}\" && prettier --ignore-path \"../.gitignore\" --check \"./**/*.{ts,tsx,json,astro,cjs,mjs,css,scss}\"",
+        "check-format": "eslint --cache \"./**/*.{ts,tsx,astro}\" && prettier --ignore-path \"../.gitignore\" \"./.gitignore\" --check \"./**/*.{ts,tsx,json,astro,cjs,mjs,css,scss}\"",
```
corneliusroemer, line 14:
```suggestion
        "check-format": "eslint --cache \"./**/*.{ts,tsx,astro}\" && prettier --ignore-path \"../.gitignore\" --ignore-path \"./.gitignore\" --check \"./**/*.{ts,tsx,json,astro,cjs,mjs,css,scss}\"",
```

theosanderson (commented):
Sorry I didn't review this at the time: my opinion would be that the time cost to get the formatting results doesn't justify the benefit - we could add a special alternative special workflow e.g. on main to test this but I'm generally not sure it's so crucial

PR #3464, opened 2024-12-17
fix a couple any types

fhennig:
resolves #

preview URL:

### PR Checklist

### DIFF
--- a/website/src/components/DataUseTerms/EditDataUseTermsModal.tsx
+++ b/website/src/components/DataUseTerms/EditDataUseTermsModal.tsx
@@ -47,8 +47,7 @@ type LoadedState = {
     earliestRestrictedUntil: DateTime | null;
 };
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any -- TODO(#3451) use proper types
-function getLoadedState(rows: Record<string, any>[]): LoadedState {
+function getLoadedState(rows: Record<string, string | number | boolean | null>[]): LoadedState {
     const openAccessions: string[] = [];
     const restrictedAccessions = new Map<string, string>();
     let earliestRestrictedUntil: DateTime | null = null;
--- a/website/src/components/DataUseTerms/EditDataUseTermsModal.tsx
+++ b/website/src/components/DataUseTerms/EditDataUseTermsModal.tsx
@@ -12,6 +12,7 @@ import {
     type DataUseTerms,
     type DataUseTermsOption,
 } from '../../types/backend';
+import type { Details } from '../../types/lapis';
 import type { ClientConfig } from '../../types/runtimeConfig';
 import { createAuthorizationHeader } from '../../utils/createAuthorizationHeader';
 import type { SequenceFilter } from '../SearchPage/DownloadDialog/SequenceFilters';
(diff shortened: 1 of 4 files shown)


website/src/components/DataUseTerms/EditDataUseTermsModal.tsx
```diff
@@ -47,8 +47,7 @@ type LoadedState = {
     earliestRestrictedUntil: DateTime | null;
 };
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any -- TODO(#3451) use proper types
-function getLoadedState(rows: Record<string, any>[]): LoadedState {
+function getLoadedState(rows: Record<string, string | number | boolean | null>[]): LoadedState {
```
fengelniederhammer, line 50: This data is coming from the LAPIS details endpoint, isn't it? We already have a type for that:
```suggestion
function getLoadedState(rows: Details[]): LoadedState {
```
fromk `import type {Details} from "../../types/lapis.ts";`

website/src/utils/search.ts
```diff
@@ -229,6 +228,7 @@ const makeCaseInsensitiveLiteralSubstringRegex = (s: string): string => {
     return `(?i)${s.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')}`;
 };
 
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
fengelniederhammer, line 231:
```suggestion
/* eslint-disable @typescript-eslint/no-explicit-any -- TODO(#3451) use proper types */
```

fengelniederhammer (changes requested):
Thanks, nice improvement! I think we can improve a little more ;)

website/src/components/DataUseTerms/EditDataUseTermsModal.tsx
fhennig: https://github.com/loculus-project/loculus/pull/3464/commits/b817828476b2acfab01801f968fe73cae9dca8ad

website/src/components/Navigation/DocsMenu.tsx
```diff
@@ -5,7 +5,7 @@ import { type FC } from 'react';
 import XIcon from '~icons/material-symbols/close';
 import MenuIcon from '~icons/material-symbols/menu';
 
-type Page = MDXInstance<Record<string, any>>; // eslint-disable-line @typescript-eslint/no-explicit-any -- TODO(#3451) use a proper type
+type Page = MDXInstance<Record<string, unknown>>;
```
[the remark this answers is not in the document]
fhennig: I reverted the change.

Looking "upstream" where the pages are coming from, they just come in as 'any' and I thought it wouldn't be very helpful to cast there, without actually validating. And I didn't want to validate now (I just wanted to do a few quick changes)

I thought 'unknown' is fine in the sense that we do not know what the frontmatter looks like. We'd have to properly validate (and come up with what to do if i.e. 'title' is missing). Maybe doing this one is for another PR 😄

fengelniederhammer (approved):

website/src/components/Navigation/DocsMenu.tsx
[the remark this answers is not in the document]
fengelniederhammer: Sounds good to me. At least its obvious that this part needs improvemenets :+1:

PR #3473, opened 2024-12-19
fix: false positive test 'notSearchable' test

### LINKED ISSUE #3326: false positive test on `notSearchable`
We have a test for the `notSearchable` parameter: https://github.com/loculus-project/loculus/blob/0fdac51a1db9357bf24267e824307b0b026b715d/website/src/components/SearchPage/SearchFullUI.spec.tsx#L170

but @fhennig found in https://github.com/loculus-project/loculus/pull/3308/files that the notSearchable feature didn't work (and indeed it's not wired up to anything, though Felix has done work to do so now).

When we have time we should remove or fix the broken test.

fhennig:
resolves #3326

### Summary
- The test failed because it searched for the search field by `name`, but the if the `label` is not given, the sentence case of the name is used. That's why the query returned no results, but the search field was actually there.
- The test now has the label set explicitly.

### DIFF
--- a/website/src/components/SearchPage/SearchFullUI.spec.tsx
+++ b/website/src/components/SearchPage/SearchFullUI.spec.tsx
@@ -168,9 +168,9 @@ describe('SearchFullUI', () => {
     it('should not render the form with fields with flag notSearchable', () => {
         renderSearchFullUI({
             searchFormFilters: [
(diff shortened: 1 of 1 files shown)


theosanderson (approved):
Thanks!

PR #3477, opened 2024-12-20
feat(website): generate metadata docs from config

anna-parker:
resolves #

preview URL:

### PR Checklist

### DIFF
--- a/website/src/components/MetadataTable.astro
+++ b/website/src/components/MetadataTable.astro
@@ -1,0 +1,18 @@
+---
+import { getConfiguredOrganisms, getSchema } from '../config';
+import OrganismTableSelector from './OrganismTableSelector';
+import type { OrganismMetadata } from './OrganismTableSelector';
+
+
+const configuredOrganisms = getConfiguredOrganisms();
+const organisms: OrganismMetadata[] = configuredOrganisms.map((organism) => {
+  return {
+    key: organism.key,
+    displayName: organism.displayName,
+    metadata: getSchema(organism.key).metadata,
+    inputFields: getSchema(organism.key).inputFields
+  };
+});
+---
+
+<OrganismTableSelector organisms={organisms} client:load/>
--- a/website/src/components/OrganismTableSelector.tsx
+++ b/website/src/components/OrganismTableSelector.tsx
@@ -1,0 +1,138 @@
+import React, { useState, useEffect } from 'react';
+import type { FC } from 'react';
+import type { Metadata, InputField } from '../types/config';
+
+import { groupFieldsByHeader } from '../utils/groupFieldsByHeader';
+import { routes } from '../routes/routes.ts';
+import IwwaArrowDown from '~icons/iwwa/arrow-down';
+
+export type OrganismMetadata = {
+    key: string;
+    displayName: string;
+    metadata: Metadata[];
+    inputFields: InputField[];
+};
+
+type Props = {
+    organisms: OrganismMetadata[];
+};
+
+const OrganismTableSelector: FC<Props> = ({ organisms }) => {
+    const [selectedOrganism, setSelectedOrganism] = useState<OrganismMetadata | null>(null);
+    const [groupedFields, setGroupedFields] = useState<Map<string, InputField[]>>(new Map());
+    const [expandedHeaders, setExpandedHeaders] = useState<Set<string>>(new Set(['Required fields']));
+
+    const handleOrganismSelect = (event: { target: { value: string } }) => {
+        const organismKey = event.target.value;
+        const organism = organisms.find((o) => o.key === organismKey);
+        console.log(organism);
+        setSelectedOrganism(organism || null);
+    };
+
+    const toggleHeader = (header: string) => {
+        const updatedExpandedHeaders = new Set(expandedHeaders);
+        if (updatedExpandedHeaders.has(header)) {
+            updatedExpandedHeaders.delete(header); // Close the table if already expanded
+        } else {
+            updatedExpandedHeaders.add(header);
+        }
+        setExpandedHeaders(updatedExpandedHeaders);
+    };
+
+    useEffect(() => {
+        if (selectedOrganism) {
+            setGroupedFields(groupFieldsByHeader(selectedOrganism.inputFields, selectedOrganism.metadata));
+            console.log(groupedFields);
+        }
+    }, [selectedOrganism]);
+
+    return (
+        <div>
+            <div>
+                <select id='organism-select' onChange={handleOrganismSelect} className='border border-gray-300 p-2'>
+                    <option value=''>-- Select an Organism --</option>
+                    {organisms.map((organism) => (
+                        <option key={organism.key} value={organism.key}>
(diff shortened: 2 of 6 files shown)


website/src/utils/groupFieldsByHeader.ts
```diff
@@ -0,0 +1,37 @@
...
+    displayName: 'Submission ID',
+    definition: 'FASTA ID',
+    guidance:
+        'Your sequence identifier; should match the FASTA file header - this is used to link the metadata to the FASTA sequence',
+    example: 'GJP123',
+    noEdit: true,
+    required: true,
+};
+
+export const groupFieldsByHeader = (inputFields: InputField[], metadata: Metadata[]): Map<string, InputField[]> => {
```
fhennig, line 14: This grouping by Header will come in very handy also for the dropdowns in the Column remapping!

I'm wondering, is there a meaningful way we can sort the `InputField[]`s here? The way it comes out of this function now is it's just ordered the same way as in the config; is that good? For larger lists of fields, maybe alphabetical sorting is easier?

Not sure if you made any considerations regarding sorting yet (or maybe this is also not the right place to sort things), feel free to leave it as is!

website/src/types/config.ts
```diff
@@ -62,6 +62,11 @@ export const inputField = z.object({
     name: z.string(),
     displayName: z.string().optional(),
     noEdit: z.boolean().optional(),
+    required: z.boolean().optional(),
+    definition: z.string().optional(),
+    example: z.union([z.string(), z.number()]).optional(),
+    guidance: z.string().optional(),
```
fhennig, line 68: I know so far none of the other properties are documented, but I'd really like some documentation on "guidance" and also "definition" (at first I thought it's some technical definition).

but not a blocking thing.

website/src/pages/[organism]/metadata-overview/index.ts
```diff
@@ -0,0 +1,36 @@
...
+    const headers: Record<string, string> = {
+        'Content-Type': 'text/tsv', // eslint-disable-line @typescript-eslint/naming-convention
+    };
+
+    const filename = `${organism.displayName.replaceAll(' ', '_')}_metadata_overview.tsv`;
+    headers['Content-Disposition'] = `attachment; filename="${filename}"`;
+
+    const fieldNames = inputFields.map(
+        (field) =>
+            `${field.name}\t${field.required ?? ''}\t${field.definition ?? ''} ${field.guidance ?? ''}\t${field.example ?? ''}`,
```
Comment (fhennig) [inline website/src/pages/[organism]/metadata-overview/index.ts:29]: I'm assuming the space (instead of tab) is between definition and guidance is deliberate, but just wanted to point it out in case it wasn't.

website/src/pages/[organism]/metadata-overview/index.ts
```diff
@@ -0,0 +1,36 @@
...
+    );
+    const tsvTemplate = [...extraFields, ...fieldNames].join('\n');
```
Comment (fhennig) [inline website/src/pages/[organism]/metadata-overview/index.ts:31]: I think adding a header row would be nice, in particular for the 'required' column since it'll otherwise just be a column of boolean values and it's not clear what they mean.

website/src/components/OrganismTableSelector.tsx
```diff
@@ -0,0 +1,139 @@
...
+                            <h3
+                                className='text-lg font-semibold mb-4 cursor-pointer'
+                                onClick={() => toggleHeader(header)}
+                            >
+                                {header}
+                                <IwwaArrowDown className='inline-block -mt-1 ml-1 h-4 w-4' />
+                            </h3>
+                            <div
+                                className={`transition-all duration-300 ${
+                                    expandedHeaders.has(header) ? 'block' : 'sr-only'
```
fhennig, line 83: Cool, I didn't know about 'sr-only' before!

website/src/components/OrganismTableSelector.tsx
```diff
@@ -0,0 +1,139 @@
...
+    fields: InputField[];
+    metadata: Metadata[];
+};
+
+const MetadataTable: FC<TableProps> = ({ fields, metadata }) => {
+    return (
+        <table className='table-auto border-collapse border border-gray-200 w-full'>
+            <thead>
+                <tr>
+                    <th className='border border-gray-300 px-4 py-2' style={{ width: '20%' }}>
```
fhennig, line 109:
```suggestion
                    <th className='border border-gray-300 px-4 py-2 w-[20%]'>
```

(In this case you could actually use `w-1/5`, but for the other special cases you can't)

website/src/utils/groupFieldsByHeader.ts
anna-parker: I think for the moment using the input order might be easiest as I can see us wanting to have more important fields first and they wouldn't necessarily be first alphabetically. I don't see a way to customize the order without adding a new metadata flag though (maybe you have a suggestion) - I'm not sure the order is important enough to complicate the values.yaml with yet another flag - but we can come back to it later if it becomes an issue :-)

website/src/pages/[organism]/metadata-overview/index.ts
Comment (anna-parker) [reply website/src/pages/[organism]/metadata-overview/index.ts:31]: Great point! I thought this was added but it doesn't look like that - I will try to do that now!

kubernetes/loculus/templates/_inputFieldsFromValues.tpl
```diff
@@ -2,7 +2,7 @@
 {{- $data := . }}
 {{- $metadata := $data.metadata }}
 {{- $extraFields := $data.extraInputFields }}
-{{- $TO_KEEP := list "name" "displayName" "definition" "guidance" "example" "required" "noEdit"}}
+{{- $TO_KEEP := list "name" "displayName" "definition" "guidance" "example" "required" "noEdit" "desired"}}
```
chaoran-chen, line 5: It would be great if you could add desired also to https://loculus.org/reference/helm-chart-config/#metadata-type, and could you maybe add some description for the other, related fields like guidance and definition?

website/src/components/OrganismTableSelector.tsx
```diff
@@ -0,0 +1,131 @@
...
+    displayName: string;
+    metadata: Metadata[];
+    inputFields: InputField[];
+};
+
+type Props = {
+    organisms: OrganismMetadata[];
+};
+
+const OrganismTableSelector: FC<Props> = ({ organisms }) => {
```
chaoran-chen, line 20: The name seems a bit too generic, would something like `OrganismMetadataTableSelector` be better?

website/src/components/MetadataTable.astro
chaoran-chen, line 1: I think this feature is amazing and would be useful generally! Could we make it accessible also for other instances? Either on an own page or in a modal?

fhennig (approved):
Thanks!

website/src/components/MetadataTable.astro
anna-parker: I can't quite envision this - can we move this to a new issue and PR and get this merged :-) ?

chaoran-chen: Sure, I created a new issue: #3525

PR #3478, opened 2024-12-23
feat(website): Column remapping when submitting metadata files

### LINKED ISSUE #3432: feat(website): support column remapping for uploaded metadata files
As a user I want to be able to upload metadata files without having to edit them, even if the columns do not match the metadata template given by Loculus. Instead I want to be able to select which column in my file to map to which column in the metadata template. I expect the UI to be smart about the column mapping, and automatically map columns with identical names to each other.

Relevant PoC: https://github.com/loculus-project/loculus/pull/3426

Image

I want to know which columns I _have_ to supply, and which are optional. This requires https://github.com/loculus-project/loculus/issues/3430 (not blocking though I think)

Some open questions (not blocking):
- Which way should the mapping go? given file columns -> expected columns or the other way round? (This can probably be changed easily, so possibly no need to discuss a lot; just pick one way for now)
- If the user discards the file, should we try to keep the mapping around? (Might be annoying to edit data, the file disappears, I have to select the mapping all over again)
  - Idea: Maybe keep the file handle around, detect if the file changed, make the user re-submit it, but cache the mapping as good as possible (just on the site - gone if page is refreshed)

fhennig:
resolves #3432 

preview URL: https://column-remapping.loculus.org/

### Summary

- When a metadata file is uploaded, it is now possible to add a column mapping
- The mapping is only applied when the file is uploaded
- If a new file is uploaded, the mapping is preserved as best as possible.
- If no mapping is configured, the file isn't read
- The mapping cannot map the same input field twice
- The mapping needs to contain all required input fields

### Screenshot

image

### PR Checklist


theosanderson:
I think the logic needs to be reversed, with a row for each "Column in your file", for which you can choose the inputField to map to. This is because often the uploaded file may only have like 4 columns from the selection.

fhennig:
ok!

chaoran-chen:
What happens if you have a column that is not mentioned in the `inputFields`? The API/backend would accept it and leave it to the pipeline to decide what to do with it. With this PR, would the website now reject files that would still be possible to submit directly through the API?

fhennig:
Files are generally not rejected.

The column remapping however would force you to either map a column to a column that is specified in the input fields, or to just not include the column.

You can also just ignore column remapping altogether, and still submit files with whatever structure you like to the backend.

chaoran-chen:
Ah, so it's possible to skip the remapping? (Sorry, I haven't tried out the feature yet!)

fhennig:
yes. Remapping is only done if you click the button to add a column mapping. I also took care to not decompress TSV files if no column mapping is to be applied! (I still need to test if this actually works the way I implemented it)

fhennig:
TODO replace `getMetadataTemplateFields` usage with using all possible input fields.

rneher:
I just tried this on the preview. 

Overall, the flow seems sensible to me. But there were a few things I stumbled on. 

 - we effectively do not have a required metadata format anymore. So the text on the left needs adjusting
 - it would be good if the button "Column mapping" had help tool-tip or similar that explains what this does (e.g. "If you are not using our metadata template, this allows you to map columns in your file to the fields expected by the database")
 - It was quite hard to find the entries in the long list. the list seems vaguely alphabetical, but important fields like "collection date" were all the way down in a seemingly random place
 



image

rneher:
Ideally, I think we should have required columns at the top of the list. And if easy to do, we could highly columns that are already selected with a different background color or similar?

rneher:
beyond that, I think it would be good to merge this in asap after some testing.

fhennig:
I checked about "And if easy to do, we could highly columns that are already selected with a different background color or similar?" -> Unfortunately the native select element options cannot be styled. I experimented with putting UTF-8 symbols with the text but wasn't happy. I think it's quite helpful though to see which ones are already mapped. A proper solution would, however, involve getting rid of the native select element. Probably not for this PR.

fhennig:
Ok, I decided to use ListBox by headlessUI since we already use that for the Autocomplete Field as well, and now I have this:

image

website/src/utils/groupFieldsByHeader.ts
```diff
@@ -11,15 +12,36 @@ const SUBMISSION_ID_FIELD: InputField = {
     required: true,
 };
 
-export const groupFieldsByHeader = (inputFields: InputField[], metadata: Metadata[]): Map<string, InputField[]> => {
+const ACCESSION_INPUT_FIELD: InputField = {
+    name: ACCESSION_FIELD,
+    displayName: 'Accession',
+    definition: 'TODO',
```
anna-parker, line 18: Hmm... so we should actually make this customizable for each `name` and `accessionPrefix` (see values.yaml) - sorry for complicating this!

 But users would really need to see an example accession and we have a different `accessionPrefix` in Loculus and Pathoplexus. 
```suggestion
    definition: 'The {name, e.g. Loculus} accession (without version) of the sequence you would like to revise',
    guidance: ''
    example: '{accessionPrefix, e.g. LOC_}000P97Y`
```
You can probably leave guidance empty I think with an example and a definition this should be quite clear :-)

fhennig: I moved getGroupedInputFields into the `config.ts`, so it's easier to get these properties. Let me know what you think! https://github.com/loculus-project/loculus/pull/3478/commits/84c843ca3b9aebb6502ff94649345bf6792af67b

website/package.json
```diff
@@ -52,6 +52,7 @@
         "react-toastify": "^11.0.2",
         "react-tooltip": "^5.28.0",
         "rsuite": "^5.76.2",
+        "string-similarity-js": "^2.1.4",
```
chaoran-chen, line 55: Given that the library doesn't seem to be actively maintained (last commit 4 years ago, and not many stars) and just consists of [one file](https://github.com/stephenjjbrown/string-similarity-js/blob/master/src/string-similarity.ts), I wonder whether it would be better to simply copy over the code (of course with a reference back to the repository) instead of including the package to reduce the surface for supply chain attacks.

website/src/components/Submission/DataUploadForm.tsx
```diff
@@ -122,9 +126,12 @@ const InnerDataUploadForm = ({
     onError,
     group,
     referenceGenomeSequenceNames,
+    metadataTemplateFields,
 }: DataUploadFormProps) => {
-    const [metadataFile, setMetadataFile] = useState<File | null>(null);
-    const [sequenceFile, setSequenceFile] = useState<File | null>(null);
+    const [metadataFile, setMetadataFile] = useState<ProcessedFile | null>(null);
+    // can be null; if null -> don't apply mapping.
```
chaoran-chen, line 132:
```suggestion
    // The columnMapping can be null; if null -> don't apply mapping.
```

To avoid potential misunderstanding to which line this comment belongs to

anna-parker:
@fhennig the preview is broken, maybe you need to rebase?

fhennig:
@anna-parker preview working again!

anna-parker:
I was just testing and I cannot upload the test data anymore (even if I do not do any mapping), I tried uploading the example ebola data here: https://github.com/pathoplexus/example_data but it fails, I checked and this still works on main: 
image

fhennig:
oh no :( Thanks for finding that Anna!

website/src/components/Submission/FileUpload/ColumnMapping.ts
```diff
@@ -0,0 +1,99 @@
...
+            const bestMatch = this.getBestMatchingTargetColumn(sourceColumn, availableFields);
+            mapping.set(sourceColumn, bestMatch);
+            availableFields = availableFields.filter((field) => field.name !== bestMatch);
+        });
+        return new ColumnMapping(mapping);
+    }
+
+    /* Update the mapping with new source and target columns, trying to keep as much of the 
+       mapping intact as possible. */
+    public update(newSourceColumns: string[], inputFields: InputField[]): ColumnMapping {
```
chaoran-chen, line 39: Would it be better/more consistent to call the second parameter `newInputFields`?

website/src/components/Submission/FileUpload/ColumnMapping.ts
```diff
@@ -0,0 +1,99 @@
...
+        const newMapping = new Map(this.map);
+        newMapping.set(sourceColumn, targetColumn);
+        this.map.forEach((targetCol, srcCol) => targetCol === targetColumn && newMapping.set(srcCol, null));
+        return new ColumnMapping(newMapping);
+    }
+
+    /* Apply this mapping to a TSV file, returning a new file with remapped columns. */
+    public async applyTo(tsvFile: ProcessedFile): Promise<File> {
+        const text = await tsvFile.text();
+        const inputRows = text.trim().split('\n');
```
chaoran-chen, line 75: Oh, you're not using a full TSV parser? This would cause problems if there is a multiline cell which, I think, is a reasonable use case for fields like version comment. The following file doesn't work:

```
submissionId	versionComment	geoLocCountry	sampleCollectionDate
test1	"Paragraph 1...

...and paragraph 2"	USA	2011-07-18
test2	"Paragraph 1...

...and paragraph 2"	USA	2011-07-18
```

chaoran-chen (dismissed):

website/src/components/Submission/FileUpload/ColumnMapping.ts
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    /* Create a new mapping with the given columns, doing a best-effort to pre-match columns. */
+    public static fromColumns(sourceColumns: string[], inputFields: InputField[]) {
+        const mapping = new Map();
+        let availableFields = inputFields;
+        sourceColumns.forEach((sourceColumn) => {
+            const bestMatch = this.getBestMatchingTargetColumn(sourceColumn, availableFields);
+            mapping.set(sourceColumn, bestMatch);
+            availableFields = availableFields.filter((field) => field.name !== bestMatch);
```
anna-parker, line 32: Hey! I found another issue here, so first you need to do an exact match and then only fuzzy matching, for example if I have my list of columns [geoLocAdmin3, geoLocAdmin1] then geoLocAdmin3 will get mapped to geoLocAdmin1 and geoLocAdmin1 will get mapped to geoLocAdmin2, see example below where I did this on the preview: 

image

It might also make sense to mark which fields have not been mapped exactly (e.g. score under 1) with a different color so that the user will quickly see that this mapping happened - otherwise they might miss sth being mapped incorrectly

anna-parker (changes requested):
Hey Felix - this is looking really great I found one bug if you could fix it - I was also wondering if you could add a couple more desired fields to the values.yaml to check this works correctly - this is also a place where there could be duplications - a field can be desired and also sample-specific - so I wanted to test duplications are handled correctly there :-)

website/src/components/Submission/FileUpload/ColumnMappingModal.tsx
```diff
@@ -0,0 +1,276 @@
...
+                    e.preventDefault();
+                    openDialog();
+                }}
+            >
+                {openModalButtonText}
+            </button>
+            <Tooltip id='columnMapping' place='bottom'>
+                If you are not using our metadata template, this allows you to map
+                <br />
+                columns in your file to the fields expected by the database.
```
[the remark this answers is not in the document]
fhennig: https://github.com/loculus-project/loculus/pull/3478/commits/f21819012be66a90789dacc8ad1603ab893a94a2

website/src/components/Submission/FileUpload/ColumnMapping.ts
```diff
@@ -0,0 +1,99 @@
+import { type ProcessedFile } from './fileProcessing';
+import type { InputField } from '../../../types/config';
+import stringSimilarity from '../../../utils/stringSimilarity';
+
+export class ColumnMapping {
+    private readonly map: ReadonlyMap<string, string | null>;
+
+    private constructor(map: ReadonlyMap<string, string | null>) {
+        this.map = map;
+    }
```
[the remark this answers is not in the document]
fhennig: https://github.com/loculus-project/loculus/pull/3478/commits/be7b42dfeeb693ddb120b95cc9a40b2c1091b88b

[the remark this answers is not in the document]
fhennig: That is very good to know!

website/src/components/Submission/FileUpload/ColumnMapping.ts
fhennig: https://github.com/loculus-project/loculus/pull/3478/commits/d8d1616fa42ab8bc048a70c0ed016472a9ab8b5c

website/src/components/Submission/FileUpload/ColumnMappingModal.tsx
```diff
@@ -0,0 +1,276 @@
...
+                    className='z-20 max-w-80 space-y-2'
+                    delayShow={200}
+                >
+                    <p>
+                        <span className='font-mono font-semibold text-gray-300'>{field.name}</span>
+                    </p>
+                    {field.definition && <p>{field.definition}</p>}
+                    {field.guidance && <p>{field.guidance}</p>}
+                </Tooltip>
+            )}
```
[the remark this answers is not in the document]
fhennig: true!

https://github.com/loculus-project/loculus/pull/3478/commits/e8f4f31983cc82e207dd16d548fb6c19ee6af88c

website/src/components/Submission/FileUpload/ColumnMapping.ts
fhennig: Ha! I had a murky feeling about the mapping, good think you spotted it. I've fixed it now. Also the non-exact matches are not in italic, does that work?

fhennig: I also added all the 'desired' flags!

website/src/components/Submission/FileUpload/ColumnMapping.ts
fhennig: update: used a different lib.

rneher:
I like the new UI. I would probably prefer if the required and desired fields would not be there twice. But not blocking.

website/src/components/Submission/FileUpload/ColumnMappingModal.tsx
```diff
@@ -0,0 +1,276 @@
...
+            </Tooltip>
+            <BaseDialog title='Remap Columns' isOpen={isOpen} onClose={closeDialog} fullWidth={false}>
+                {currentMapping === null || inputColumns === null ? (
+                    'Loading ...'
+                ) : (
+                    <div className='space-y-4'>
+                        <table>
+                            <thead>
+                                <tr>
+                                    <th className='pr-16 py-2'>Column in your file</th>
```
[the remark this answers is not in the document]
chaoran-chen: The mobile view still isn't really good but agree that it's not a priority, so no need to work on it for now.

website/src/components/Submission/FileUpload/ColumnMapping.ts
anna-parker: Thanks it looks great!

website/src/components/Submission/FileUpload/UploadComponent.tsx
```diff
@@ -19,6 +19,8 @@ export const UploadComponent = ({
     const [isDragOver, setIsDragOver] = useState(false);
     const isClient = useClientFlag();
 
+    // TODO It feels like it would be useful to return a ProcessedFile here, so we know if it's compressed or not etc.
```
anna-parker, line 22: can this TODO be removed now?

anna-parker: There seems to be some issues with uploading .gz files with these changes (.gz files are still working on main): 
image

fhennig:
@corneliusroemer In the monday meeting you said you wanted to have a look, and this is arguably a non-trivial PR; Your input is appreciated!

Chaoran already said he's fine with how things are, I've adress all of Richards points and I think all the compression related stuff that Anya mentioned is also sorted out.

website/src/components/Submission/FileUpload/ColumnMappingModal.tsx
```diff
@@ -0,0 +1,291 @@
...
+                )}
+            </BaseDialog>
+        </>
+    );
+};
+
+async function extractColumns(tsvFile: ProcessedFile): Promise<Result<string[], Error>> {
+    let text;
+    try {
+        // there is potential to optimize: don't read the whole file, just the header (read in chunks)
```
anna-parker, line 169: is this an open TODO or can it be removed?

anna-parker (approved):
LGTM! 

I tested with metadata in all zip combinations except for `.xz` and this works, I also tried out some column renaming and read through the code. I'm not a typescript expert but everything looks very good and the feature is working really nicely!

website/src/components/Submission/FileUpload/ColumnMappingModal.tsx
fhennig: I didn't write TODO because I didn't intend _to do_ it (😄)  but thought it'd still be a good note to have there. But maybe I just remove it, I guess from the code it's obvious that this improvement could be made; and the whole file will still be read later in the process anyways, so we wouldn't even gain that much. i'll remove it.

fhennig: removed: https://github.com/loculus-project/loculus/pull/3478/commits/49108a0f9f438b9b8790b0b6735c4e7b2a574a88

PR #3483, opened 2024-12-29
refactor(backend): reduce redundancy in DataUseTerms.fromParameters()

chaoran-chen:
A tiny refactoring to avoid having the check `restrictedUntilString == null` at two places.


anna-parker (approved):
lgtm

PR #3505, opened 2025-01-08
fix(ingest): decrease ingest memory usage for large datasets

anna-parker:
resolves #

preview URL: https://ingest-memory-fixes.loculus.org/

### Summary

When dealing with organisms with a large number of sequences (e.g. influenza A has 1.3M sequences) even just reading it the entire metadata is highly memory intensive - this PR switches to streaming metadata where ever possible and is similar to the change @corneliusroemer made in https://github.com/loculus-project/loculus/issues/2277 to stream sequences. 

Note that in the case where segments are grouped the entire file still must be read into a dictionary. 

### Testing
How best to test this? Use https://github.com/pathoplexus/pathoplexus/tree/main/data-integrity-tests/regression-testing and switch the URLs to this branch and main :-) 

### PR Checklist

- [x] Improve batching: this still reads in both fasta and metadata and is thus the current memory bottleneck -> to remove this the input files need to be sorted => I will do this in a different PR to make it easier to review this PR
- [x] extend tests to ndjson files


fhennig (approved):
looks sensible! 

still @corneliusroemer because I think it'd be good if he had a look.

anna-parker:
As discussed with @corneliusroemer offline I will merge

PR #3508, opened 2025-01-10
feat(website): Improve message when gapped sequences uploaded.

### LINKED ISSUE #3507: Make 'non-IUPAC symbols' error message a bit clearer
Currently we only allow `ATCG` and ambig nucleotides (including `N`) in uploaded sequences (as I understand) - and in particular do not allow gaps (`-`). However if you have gaps (and only gaps) the error message shows as thus:

```
Found non-IUPAC symbols in the main sequence: -
```
screenshot:
image

This can look a bit confusing as it's perhaps not clear that the last `-` is a reference to the symbol that's not allowed, rather than a stray character or 'list separator' or even an indication that the list is empty. 

It may be good to just expand this error message a little to explicitly say that only `ATCGN` and ambiguous characters are not allowed?

theosanderson:
resolves #3507

preview URL: https://gap-message.loculus.org/ebola-sudan/submission/2/review

image

### DIFF
--- a/preprocessing/nextclade/src/loculus_preprocessing/sequence_checks.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/sequence_checks.py
@@ -47,7 +47,7 @@ def errors_if_non_iupac(
                         ],
                         message=(
                             f"Found non-IUPAC symbols in the {segment} sequence: "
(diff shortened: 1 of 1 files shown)


anna-parker (approved):
Thanks!

PR #3509, opened 2025-01-10
feat(ingest): Allow option to group segments with ground truth

anna-parker:
resolves #

preview URL: https://group-using-assemblies.loculus.org/

### Summary

Currently we only group segments heuristically but many organisms also have assembly information where we have ground truth grouping information - sadly this is not an element of the ncbi dataset that we download but it can be downloaded separately to create a source of truth JSON that can be used for grouping segments. 

This is great as we cannot group segments when there is any ambiguity, and it is not uncommon for there to be metadata differences between segments within an assembly. 

### PR Checklist

- [x] remove grouping ground truth from CCHF (as this changes the current grouping) -> should be added later
- [x] extend tests to ndjson files


kubernetes/loculus/values.yaml
```diff
@@ -1485,6 +1485,7 @@ defaultOrganisms:
         taxon_id: 3052518
         nextclade_dataset_server: https://raw.githubusercontent.com/nextstrain/nextclade_data/cornelius-cchfv/data_output
         nextclade_dataset_name: nextstrain/cchfv/linked
+        grouping_ground_truth: https://anna-parker.github.io/influenza-a-groupings/results/cchf_groups.json
```
anna-parker, line 1488: I added this so that the preview can test the grouping - currently we actually don't group the reference sequences so this should be removed before merging at the moment

ingest/Snakefile
```diff
@@ -255,7 +257,7 @@ if ALIGN:
             """
 
 if not ALIGN:
-    rule download:
+    rule download_minimizer:
         output:
             results="results/minimzer.json",
```
corneliusroemer, line 262:
```suggestion
            results="results/minimizer.json",
```

corneliusroemer (approved):

ingest/Snakefile
```diff
@@ -324,18 +326,60 @@ rule prepare_metadata:
...
+        output:
+            results="results/groups.json",
+        params:
+            grouping=config.get("grouping_ground_truth")
+        shell:
+            """
+            curl -L -o {output.results} {params.grouping}
+            """
+
+    rule group_segments:
```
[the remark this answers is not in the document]
anna-parker: what about group_predefined_segments and algorithmic_group_segments? as override would kind of imply that the override grouping happens after the algorithmic grouping but it actually happens before the algorithmic grouping.

PR #3529, opened 2025-01-15
chore(deps): SILO v0.5.0

Taepper:
preview URL: http://tae-silo-0-5-0.loculus.org

### Summary

This updates the dependency SILO to version `0.5.0`. 
To note: This version has some nice changes to the CLI interface and config handling of SILO. A refactor of `silo_import_job.sh` should be easier now, as some things that were previously specified by config files can now be directly supplied via CLI. This can reduce the "magic" some commands seemed to exhibit before.

The help-text SILO provides has also been improved a lot, which can be consulted for the usability improvements mentioned above.

Moreover, the built binary is now called `silo` instead of `siloApi`. `api` and `preprocessing` are supplied instead of `--api` and `--preprocessing`.

### PR Checklist

~~- [ ] All necessary documentation has been adapted.~~
~~- [ ] The implemented feature is covered by an appropriate test.~~

PR #3530, opened 2025-01-15
argocd: replace all '.' in branch names with '-' for preview URLs

Taepper:
resolves preview not showing up in https://github.com/loculus-project/loculus/pull/3529

preview URL:

### Summary

I opened the PR to update SILO with a branch containing the version number `0.5.0`. This prevented argocd to spin up a preview instance with the following error:

image

...while dots in branch names are arguably a bad idea in general, we already have special logic in place to replace many other characters, therefore, I think it is worthwhile to also add this case. In case someone else does the same mistake in the future it will not clutter our PRs as it is not possible to change the branch of a PR. (I would need to close the PR and open a new one, adding noise to our closed PR list)

### PR Checklist

### DIFF
--- a/kubernetes/appset.yaml
+++ b/kubernetes/appset.yaml
@@ -24,23 +24,23 @@ spec:
   goTemplate: true
   template:
     metadata:
-      name: pp-{{ (printf "%.25s" .branch) | replace "_" "-" | replace "/" "-" | trimSuffix "-" | lower }}-{{.number}}
+      name: pp-{{ (printf "%.25s" .branch) | replace "_" "-" | replace "/" "-" | replace "." "-" | trimSuffix "-" | lower }}-{{.number}}
     spec:
       destination:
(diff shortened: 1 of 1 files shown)


chaoran-chen (approved):
Makes sense to me, thanks!

theosanderson (approved):
This is good, but it would have needed to be deployed manually - I've moved the change to the top level git ops file so these changes can be made without my intervention in future

Taepper:
Hmm, I would like to squash merge as it would add a reference to this PR in the commit message (there is no issue). But this would obfuscate @theosanderson's commit. What is the suggested procedure in this case. Should I add `(#3530)` to the commit message of both commits manually?

theosanderson:
(I don't mind at all how we merge - I don't think it will have major practical effects. I updated the PR description so I think a squash merge would be ok)

PR #3532, opened 2025-01-15
Make it possible to hide fields from the table

### LINKED ISSUE #3325: feat: Support making a field hidden from the column selection
Initial comment: https://github.com/loculus-project/loculus/pull/3308#discussion_r1865803286

As a maintainer I want to be able to configure metadata fields as 'hidden' in the table column selection, because certain metadata fields do not make sense to be displayed there and would only confuse the user.

This would already be useful to hide the 'Collection date' text field, which we currently hide in a hardcoded way.

fhennig:
resolves #3325 

preview URL: http://column-hiding.loculus.org

### Summary

- Adds a new metadata config setting: `hideInSearchResultsTable` which you can set to hide a column in the Column visibility dialog.

### Screenshot

Nothing to see really.

### PR Checklist

### DIFF
--- a/docs/src/content/docs/reference/helm-chart-config.mdx
+++ b/docs/src/content/docs/reference/helm-chart-config.mdx
@@ -815,6 +815,12 @@ Each organism object has the following fields:
             <td>If true, hides the field on the sequence details page.</td>
         </tr>
         <tr>
+            <td>`hideInSearchResultsTable`</td>
+            <td>Boolean</td>
+            <td></td>
+            <td>If true, hides the field in the search results table (and makes it impossible to show).</td>
+        </tr>
+        <tr>
             <td>`perSegment`</td>
             <td>Boolean</td>
             <td></td>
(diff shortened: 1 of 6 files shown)


theosanderson (approved):
Looks very reasonable to me! If one wanted to test this one way would be:
- add a field with this to e.g. `test-organism` in `values.yaml`
- in the integration tests, assert that it is not displayed when you go and list the fields for this organism (and assert that some other field is, as a positive control)

But not demanding that

fhennig:
Thanks! I added a test, maybe you can have a look again?

theosanderson (approved):
Nice - formatting issue but otherwise good.

My inclination is that we should aim in say a year's time to have most of what is currently the "E2E" tests instead moved to the "integration" tests, so if I were adding this I'd have probably added it to the integration tests, but I appreciate we don't have e.g. a `searchPage` set up there so it's more work / less nice. We can probably move whole sets of tests over in due course.

theosanderson:
huh or maybe the format issue is unrelated?

PR #3533, opened 2025-01-16
Remove silo preprocessing_config_merged

Taepper:
resolves `TODO` in comment of `silo_import_job.sh`

preview URL: https://tae-remove-silo-preproces.loculus.org/

### Summary

~SILO 0.5.0 has much better CLI support. We now use this to simplify some of the logic which was in place for the silo import job. This also removed a file template and accompanying docker file mount.~

It is cumbersome to decide whether to supply the lineageDefinitionsFilename or not, if we do it via CLI. Therefore, moving it to the template where the `preprocessing_config` is generated is a much simpler change. Also this lends itself for extensibility once more lineage definitions are supplied for the same organism

### PR Checklist

### DIFF
--- a/kubernetes/loculus/silo_import_job.sh
+++ b/kubernetes/loculus/silo_import_job.sh
@@ -5,8 +5,6 @@ set -e
 root_dir=""
 last_etag=""
 lineage_definition_file=/preprocessing/input/lineage_definitions.yaml
-preprocessing_config_file=preprocessing_config.yaml
-preprocessing_config_file_merged=preprocessing_config_merged.yaml
 
 # Parse command-line arguments
 usage() {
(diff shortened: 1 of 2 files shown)


fhennig:
Oh! If I read this right, I could've done it like this from the start? Damn, I didn't think about that the config is also generated.

The preview doesn't work, can you maybe fix it? I see you already rebased, maybe push an empty commit to make Argo re-deploy the preview?

Code looks goo though!

fengelniederhammer:
> If I read this right, I could've done it like this from the start?

Yes ;)

Taepper:
Are the integration tests known to be flaky? They failed both on main and on the latest push to this branch for me now. One workflow restart was sufficient to make it green again. Now I am wondering whether this only started occuring after I merged silo 0.5.0 PR or was already happening before?

fengelniederhammer (approved):

theosanderson:
> Are the integration tests known to be flaky? 

They are very new so probably are flaky but it's not known - but I wouldn't worry that you have done anything. Apologies, will try to look into it soon.

Issue #1517, opened 2024-03-29
Use `text` type for strings in PostgreSQL and associated code


chaoran-chen:
Currently, most string columns in the PostgreSQL database ([init.sql](https://github.com/loculus-project/loculus/blob/256d17781c63985a49e2185d652f345072def1e3/backend/src/main/resources/db/migration/V1__init.sql)) are typed as `text` with a few as `varchar(255)`. In the Kotlin code, however, we often use `varchar` (e.g., [here](https://github.com/loculus-project/loculus/blob/256d17781c63985a49e2185d652f345072def1e3/backend/src/main/kotlin/org/loculus/backend/service/submission/SequenceEntriesTable.kt#L19)).

I think that we should only use `text`. There isn't really a technical/efficiency advantage of using `varchar` and we should leave it to the backend to enforce any length restrictions.

PR #3534, opened 2025-01-16
Make pipeline version per Organism

### LINKED ISSUE #1728: Make pipeline version scoped at organism level rather than global
As a maintainer, I want to be able to update the processing pipeline of a single organism at a time without bumping versions of the pipelines of other organisms (which then also leads to the reprocessing of data of the other organisms).

At the moment, the backend maintains a single processing pipeline version and only accepts data that have the current version or a higher version. This is maintained across all organisms. I.e., if we want to upgrade the pipeline for one organism, we have to upgrade the pipelines for all other organisms as well.

This should be changed. We should be able to have pipelines with different version numbers for different organisms. I think that we would add an `organism` column to the `current_processing_pipeline`.

fhennig:
resolves #1728

preview URL: https://pipeline-v-per-organism.loculus.org

### Summary
The `current_processing_pipeline` table now has a new column: `organism`. We now store the current pipeline for each organism. 

TODO: How does this affect users? Cornelius said there was previously a constraint to have to use the same version in the values.yaml for all organisms, does that still need to be removed?

#### Implementation notes
Before this change, we stored "V1" in this table (`current_processing_pipeline`)  - that was simple and could be done without knowing the user configuration. Some of the additional complexity in this PR arises from having to configure _a version_ for the CurrentPipelineVersion for each organism (and also i.e. having to reset this in the tests).

Future work could either include making entries in the table optional, and only filling it as soon as the first sequence for an organism is submitted, or maybe introducing a mechanism to register pipeline versions via API. I opted against making any of these changes to keep the PR small and not change too much of how things are currently functioning.

#### Testing
- Exisiting tests were updated
- A backend test was added to check that if one pipeline version updates for one organism, another organisms pipeline version is not changed.
- A backend test was added to check that if the initialization runs again, we don't insert v1 rows for organisms that are already in the table.

#### Docs
Document pipeline version concept better. Relevant link: https://github.com/loculus-project/loculus/blob/main/preprocessing/specification.md#reprocessing

### Screenshot
backend change - n/a

### PR Checklist

### DIFF
--- a/backend/docs/db/schema.sql
+++ b/backend/docs/db/schema.sql
@@ -609,7 +609,7 @@ ALTER TABLE ONLY public.audit_log
 --
 
 ALTER TABLE ONLY public.current_processing_pipeline
-    ADD CONSTRAINT current_processing_pipeline_pkey PRIMARY KEY (organism, version);
+    ADD CONSTRAINT current_processing_pipeline_pkey PRIMARY KEY (organism);
 
 
 --
--- a/backend/docs/db/schema.sql
+++ b/backend/docs/db/schema.sql
@@ -468,16 +468,27 @@ CREATE VIEW public.sequence_entries_view AS
     sepd.finished_processing_at,
     sepd.processed_data,
     (sepd.processed_data || em.joint_metadata) AS joint_metadata,
+        CASE
+            WHEN se.is_revocation THEN ( SELECT current_processing_pipeline.version
+               FROM public.current_processing_pipeline
+              WHERE (current_processing_pipeline.organism = se.organism))
+            ELSE sepd.pipeline_version
+        END AS pipeline_version,
     sepd.errors,
     sepd.warnings,
         CASE
             WHEN (se.released_at IS NOT NULL) THEN 'APPROVED_FOR_RELEASE'::text
-            WHEN se.is_revocation THEN 'AWAITING_APPROVAL'::text
+            WHEN se.is_revocation THEN 'PROCESSED'::text
             WHEN (sepd.processing_status = 'IN_PROCESSING'::text) THEN 'IN_PROCESSING'::text
-            WHEN (sepd.processing_status = 'HAS_ERRORS'::text) THEN 'HAS_ERRORS'::text
-            WHEN (sepd.processing_status = 'FINISHED'::text) THEN 'AWAITING_APPROVAL'::text
+            WHEN (sepd.processing_status = 'PROCESSED'::text) THEN 'PROCESSED'::text
             ELSE 'RECEIVED'::text
-        END AS status
+        END AS status,
+        CASE
+            WHEN (sepd.processing_status = 'IN_PROCESSING'::text) THEN NULL::text
+            WHEN ((sepd.errors IS NOT NULL) AND (jsonb_array_length(sepd.errors) > 0)) THEN 'HAS_ERRORS'::text
+            WHEN ((sepd.warnings IS NOT NULL) AND (jsonb_array_length(sepd.warnings) > 0)) THEN 'HAS_WARNINGS'::text
+            ELSE 'NO_ISSUES'::text
+        END AS processing_result
    FROM ((public.sequence_entries se
      LEFT JOIN public.sequence_entries_preprocessed_data sepd ON (((se.accession = sepd.accession) AND (se.version = sepd.version) AND (sepd.pipeline_version = ( SELECT current_processing_pipeline.version
            FROM public.current_processing_pipeline
--- a/backend/docs/db/schema.sql
+++ b/backend/docs/db/schema.sql
@@ -602,7 +602,7 @@ ALTER TABLE ONLY public.audit_log
 --
 
 ALTER TABLE ONLY public.current_processing_pipeline
-    ADD CONSTRAINT current_processing_pipeline_pkey PRIMARY KEY (version);
+    ADD CONSTRAINT current_processing_pipeline_pkey PRIMARY KEY (organism, version);
 
 
 --
--- a/backend/docs/db/schema.sql
+++ b/backend/docs/db/schema.sql
@@ -194,6 +194,28 @@ CREATE TABLE public.data_use_terms_table (
 ALTER TABLE public.data_use_terms_table OWNER TO postgres;
 
 --
+-- Name: sequence_entries; Type: TABLE; Schema: public; Owner: postgres
+--
+
+CREATE TABLE public.sequence_entries (
+    accession text NOT NULL,
+    version bigint NOT NULL,
+    organism text NOT NULL,
+    submission_id text NOT NULL,
+    submitter text NOT NULL,
+    approver text,
+    group_id integer NOT NULL,
+    submitted_at timestamp without time zone NOT NULL,
+    released_at timestamp without time zone,
+    is_revocation boolean DEFAULT false NOT NULL,
+    original_data jsonb,
+    version_comment text
+);
+
+
+ALTER TABLE public.sequence_entries OWNER TO postgres;
+
+--
 -- Name: sequence_entries_preprocessed_data; Type: TABLE; Schema: public; Owner: postgres
 --
 
@@ -235,7 +257,10 @@ CREATE VIEW public.external_metadata_view AS
             sequence_entries_preprocessed_data.finished_processing_at
            FROM public.sequence_entries_preprocessed_data
           WHERE (sequence_entries_preprocessed_data.pipeline_version = ( SELECT current_processing_pipeline.version
-                   FROM public.current_processing_pipeline))) cpd
+                   FROM public.current_processing_pipeline
+                  WHERE (current_processing_pipeline.organism = ( SELECT se.organism
+                           FROM public.sequence_entries se
+                          WHERE ((se.accession = sequence_entries_preprocessed_data.accession) AND (se.version = sequence_entries_preprocessed_data.version))))))) cpd
      LEFT JOIN public.all_external_metadata ON (((all_external_metadata.accession = cpd.accession) AND (all_external_metadata.version = cpd.version))));
 
 
@@ -423,28 +448,6 @@ CREATE TABLE public.seqsets (
 ALTER TABLE public.seqsets OWNER TO postgres;
 
 --
--- Name: sequence_entries; Type: TABLE; Schema: public; Owner: postgres
---
-
-CREATE TABLE public.sequence_entries (
-    accession text NOT NULL,
-    version bigint NOT NULL,
-    organism text NOT NULL,
-    submission_id text NOT NULL,
-    submitter text NOT NULL,
-    approver text,
-    group_id integer NOT NULL,
-    submitted_at timestamp without time zone NOT NULL,
-    released_at timestamp without time zone,
-    is_revocation boolean DEFAULT false NOT NULL,
-    original_data jsonb,
-    version_comment text
-);
-
-
-ALTER TABLE public.sequence_entries OWNER TO postgres;
-
---
 -- Name: sequence_entries_view; Type: VIEW; Schema: public; Owner: postgres
 --
 
@@ -465,29 +468,22 @@ CREATE VIEW public.sequence_entries_view AS
     sepd.finished_processing_at,
     sepd.processed_data,
     (sepd.processed_data || em.joint_metadata) AS joint_metadata,
-        CASE
-            WHEN se.is_revocation THEN ( SELECT current_processing_pipeline.version
-               FROM public.current_processing_pipeline)
-            ELSE sepd.pipeline_version
-        END AS pipeline_version,
     sepd.errors,
     sepd.warnings,
         CASE
             WHEN (se.released_at IS NOT NULL) THEN 'APPROVED_FOR_RELEASE'::text
-            WHEN se.is_revocation THEN 'PROCESSED'::text
+            WHEN se.is_revocation THEN 'AWAITING_APPROVAL'::text
             WHEN (sepd.processing_status = 'IN_PROCESSING'::text) THEN 'IN_PROCESSING'::text
-            WHEN (sepd.processing_status = 'PROCESSED'::text) THEN 'PROCESSED'::text
+            WHEN (sepd.processing_status = 'HAS_ERRORS'::text) THEN 'HAS_ERRORS'::text
+            WHEN (sepd.processing_status = 'FINISHED'::text) THEN 'AWAITING_APPROVAL'::text
             ELSE 'RECEIVED'::text
-        END AS status,
-        CASE
-            WHEN (sepd.processing_status = 'IN_PROCESSING'::text) THEN NULL::text
-            WHEN ((sepd.errors IS NOT NULL) AND (jsonb_array_length(sepd.errors) > 0)) THEN 'HAS_ERRORS'::text
-            WHEN ((sepd.warnings IS NOT NULL) AND (jsonb_array_length(sepd.warnings) > 0)) THEN 'HAS_WARNINGS'::text
-            ELSE 'NO_ISSUES'::text
-        END AS processing_result
+        END AS status
    FROM ((public.sequence_entries se
      LEFT JOIN public.sequence_entries_preprocessed_data sepd ON (((se.accession = sepd.accession) AND (se.version = sepd.version) AND (sepd.pipeline_version = ( SELECT current_processing_pipeline.version
-           FROM public.current_processing_pipeline)))))
+           FROM public.current_processing_pipeline
+          WHERE (current_processing_pipeline.organism = ( SELECT se_1.organism
+                   FROM public.sequence_entries se_1
+                  WHERE ((se_1.accession = sepd.accession) AND (se_1.version = sepd.version)))))))))
      LEFT JOIN public.external_metadata_view em ON (((se.accession = em.accession) AND (se.version = em.version))));
 
 
--- a/backend/docs/db/schema.sql
+++ b/backend/docs/db/schema.sql
@@ -488,7 +488,7 @@ CREATE VIEW public.sequence_entries_view AS
         END AS processing_result
    FROM (((public.sequence_entries se
      LEFT JOIN public.sequence_entries_preprocessed_data sepd ON (((se.accession = sepd.accession) AND (se.version = sepd.version))))
-     JOIN public.current_processing_pipeline ccp ON (((se.organism = ccp.organism) AND (sepd.pipeline_version = ccp.version))))
+     LEFT JOIN public.current_processing_pipeline ccp ON (((se.organism = ccp.organism) AND (sepd.pipeline_version = ccp.version))))
      LEFT JOIN public.external_metadata_view em ON (((se.accession = em.accession) AND (se.version = em.version))));
 
 
--- a/backend/docs/db/schema.sql
+++ b/backend/docs/db/schema.sql
@@ -171,7 +171,8 @@ ALTER SEQUENCE public.audit_log_id_seq OWNED BY public.audit_log.id;
 
 CREATE TABLE public.current_processing_pipeline (
     version bigint NOT NULL,
-    started_using_at timestamp without time zone NOT NULL
+    started_using_at timestamp without time zone NOT NULL,
+    organism text NOT NULL
 );
 
 
--- a/backend/docs/db/schema.sql
+++ b/backend/docs/db/schema.sql
@@ -246,21 +246,18 @@ CREATE VIEW public.external_metadata_view AS
             WHEN (all_external_metadata.external_metadata IS NULL) THEN jsonb_build_object('metadata', (cpd.processed_data -> 'metadata'::text))
             ELSE jsonb_build_object('metadata', ((cpd.processed_data -> 'metadata'::text) || all_external_metadata.external_metadata))
         END AS joint_metadata
-   FROM (( SELECT sequence_entries_preprocessed_data.accession,
-            sequence_entries_preprocessed_data.version,
-            sequence_entries_preprocessed_data.pipeline_version,
-            sequence_entries_preprocessed_data.processed_data,
-            sequence_entries_preprocessed_data.errors,
-            sequence_entries_preprocessed_data.warnings,
-            sequence_entries_preprocessed_data.processing_status,
-            sequence_entries_preprocessed_data.started_processing_at,
-            sequence_entries_preprocessed_data.finished_processing_at
-           FROM public.sequence_entries_preprocessed_data
-          WHERE (sequence_entries_preprocessed_data.pipeline_version = ( SELECT current_processing_pipeline.version
-                   FROM public.current_processing_pipeline
-                  WHERE (current_processing_pipeline.organism = ( SELECT se.organism
-                           FROM public.sequence_entries se
-                          WHERE ((se.accession = sequence_entries_preprocessed_data.accession) AND (se.version = sequence_entries_preprocessed_data.version))))))) cpd
+   FROM (( SELECT sepd.accession,
+            sepd.version,
+            sepd.pipeline_version,
+            sepd.processed_data,
+            sepd.errors,
+            sepd.warnings,
+            sepd.processing_status,
+            sepd.started_processing_at,
+            sepd.finished_processing_at
+           FROM ((public.sequence_entries_preprocessed_data sepd
+             JOIN public.sequence_entries se ON (((sepd.accession = se.accession) AND (sepd.version = se.version))))
+             JOIN public.current_processing_pipeline cpp ON (((se.organism = cpp.organism) AND (sepd.pipeline_version = cpp.version))))) cpd
      LEFT JOIN public.all_external_metadata ON (((all_external_metadata.accession = cpd.accession) AND (all_external_metadata.version = cpd.version))));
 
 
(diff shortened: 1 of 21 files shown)


fengelniederhammer:
Wow, the pipeline version in the backend is not per organism? But in the `values.yml` it's per organism, isn't it?

fhennig:
What do you mean? The pipeline version is internal, so it doesn't appear in the values.yaml if I'm not mistaken.

chaoran-chen:
The pipeline version also occurs in the Values.yaml because it's also used to deploy the pipelines.

fengelniederhammer:
https://github.com/loculus-project/loculus/blob/9aa28aba623ac8dcb5c0008b5d0ae583e10532e8/kubernetes/loculus/values.yaml#L1148-L1150

This is the pipeline version, isn't it?

theosanderson:
Currently for the backend there is a single latest pipeline version. So if you make a new pipeline version for one organism in the values yaml you have to do so for all organisms. This issue is (intended to be) about addressing that.

fhennig:
Note to self: The table now starts out empty, somehow it needs to be filled. (I don't think there is a way around the table starting empty now? Because we don't know which organisms there are)

fhennig:
Update: I found a way to initialize the table on backend-startup.

Now I marked some places in the schema that still need to be adjusted, that currently lead to: "ERROR: more than one row returned by a subquery used as an expression"

fhennig:
Ok at least some of the test failures are because the `flyway` property I added cannot be injected. maybe I can create a dummy bean for it in the test, or I can somehow make the dependency optional.

Ideally I'd solve the problem of "run x after flyway is finished running" in a different way maybe? Maybe the PostConstruct shouldn't be in the SubmissionDatabaseService, but somewhere else.

Also for anyone who is following along, I wanted to do it this way as a quick way to get it to run, and then iterate on it to make it prettier/less invasive, just so y'all know.

backend/src/main/kotlin/org/loculus/backend/service/submission/CurrentProcessingPipelineTable.kt
```diff
@@ -1,11 +1,42 @@
...
+            this[startedUsingAtColumn] = now
+        }
+
+    fun pipelineNeedsUpdate(foundVersion: Long, organism: String) = CurrentProcessingPipelineTable
+        .selectAll()
+        .where { versionColumn neq foundVersion }
+        .andWhere { organismColumn eq organism }
+        .limit(1)
+        .empty()
+        .not()
```
chaoran-chen, line 31: I think that we need to check whether the found version is larger than all existing versions?

fhennig: Yes! Fixed: https://github.com/loculus-project/loculus/pull/3534/commits/fdc37b286142a7b22df39f78968dbf44e4af780c

(It worked before due to constraints elsewhere in the codebase, but this is more robust and correct 👍 )

fhennig:
I've moved the PR out of Draft, because I'm done with the code+tests. I'll look at the docs today.

fhennig:
I have added some docs stuff:

image

Not sure if the new subsection is a good idea or not. But I added a page to explain a bit more what a preprocessing pipeline even is, taking some of the text from the specification. 

I imagine that in the near future it'd be good to have a couple of sections like "setup", "preprocessing pipeline", "user management", "<something else>" sort of like chapters in a manual.

backend/src/main/resources/db/migration/V1.10__pipeline_version_per_organism.sql
```diff
@@ -0,0 +1,116 @@
...
+from
+    (
+        -- Get only the preprocessed data for the current pipeline version
+        select * from sequence_entries_preprocessed_data
+        where pipeline_version = (select version from current_processing_pipeline
+                                  where organism = (
+                                      select organism from sequence_entries se
+                                      where se.accession = sequence_entries_preprocessed_data.accession
+                                      and se.version = sequence_entries_preprocessed_data.version
+                                  ))
```
chaoran-chen, line 64: Could we keep this simpler and use a less nested query? Would this be logically equivalent?

```suggestion
        select sepd.*
        from
            sequence_entries_preprocessed_data sepd
            join sequence_entries se on sepd.accession = se.accession and sepd.version = se.version
            join current_processing_pipeline cpp on se.organism = cpp.organism
                and sepd.pipeline_versioin = cpp.version
```

backend/src/main/resources/db/migration/V1.10__pipeline_version_per_organism.sql
```diff
@@ -0,0 +1,116 @@
...
+    sequence_entries se
+    left join sequence_entries_preprocessed_data sepd on
+        se.accession = sepd.accession
+        and se.version = sepd.version
+        and sepd.pipeline_version = (select version from current_processing_pipeline
+                                     where organism = (
+                                         select organism from sequence_entries se
+                                         where se.accession = sepd.accession
+                                         and se.version = sepd.version
+                                     ))
```
chaoran-chen, line 109: Same comment as above: I think that it would be more readable to just join the tables instead of using nested queries. (But this is also correct, so not blocking)

backend/src/main/kotlin/org/loculus/backend/service/submission/CurrentProcessingPipelineTable.kt
```diff
@@ -1,11 +1,53 @@
...
+
+    /**
+     * Given a version that was found, is the currently stored 'current' pipeline version for this organism
+     * less than the one that was found? If so, the pipeline needs to 'update' i.e. reprocess older entries.
+     */
+    fun pipelineNeedsUpdate(foundVersion: Long, organism: String) = CurrentProcessingPipelineTable
+        .selectAll()
+        .where { versionColumn less foundVersion }
+        .andWhere { organismColumn eq organism }
+        .limit(1)
```
chaoran-chen, line 37: The limit is not really needed, right? Since it should be at most one?

backend/src/main/kotlin/org/loculus/backend/service/submission/SequenceEntriesTable.kt
```diff
@@ -44,6 +44,14 @@ object SequenceEntriesTable : Table(SEQUENCE_ENTRIES_TABLE_NAME) {
         )
     }
 
+    fun distinctOrganisms() = SequenceEntriesTable
+        .select(SequenceEntriesTable.organismColumn)
+        .withDistinct()
```
chaoran-chen, line 49: Just a side note related to what we chatted about yesterday: this is a good sign that it would be nice to have a dedicated organism table.

backend/src/test/kotlin/org/loculus/backend/service/submission/UseNewerProcessingPipelineVersionTaskTest.kt
```diff
@@ -32,19 +39,69 @@ class UseNewerProcessingPipelineVersionTaskTest(
...
         useNewerProcessingPipelineVersionTask.task()
-        assertThat(submissionDatabaseService.getCurrentProcessingPipelineVersion(), `is`(3L))
+        assertThat(submissionDatabaseService.getCurrentProcessingPipelineVersion(Organism(DEFAULT_ORGANISM)), `is`(3L))
 
         submissionControllerClient.extractUnprocessedData(numberOfSequenceEntries = 10, pipelineVersion = 2)
             .andExpect(status().isUnprocessableEntity)
     }
+
+    @Test
+    fun `GIVEN the pipeline v for one organism updates THEN the pipeline v for another organism is not updated`() {
```
chaoran-chen, line 59:
```suggestion
    fun `GIVEN the pipeline version for one organism updates THEN the pipeline version for another organism is not updated`() {
```

chaoran-chen (approved):
Thank you very much, Felix! The code looks good to me and the tests are reassuring!

As it is a rather complicated change and it is not easy to test manually, I think that it would be great to have another review by someone. @corneliusroemer, @theosanderson, would you like to take a look?

corneliusroemer (changes requested):
I'll review this by Thursday EOD

corneliusroemer:
A simple/cheap way to test is to use different values for different pipelines in our default values.yaml

There are lines like this per organism (right now version is shared, we can now undo this constraint)

```
  preprocessing:
    - &preprocessing
      version: 1
      image: ghcr.io/loculus-project/preprocessing-nextclade
```

backend/src/main/kotlin/org/loculus/backend/service/submission/CurrentProcessingPipelineTable.kt
fhennig: I based this on previous code where the same applied, I thought maybe it was there to make the query faster or something. But I can remove it

backend/src/test/kotlin/org/loculus/backend/service/submission/UseNewerProcessingPipelineVersionTaskTest.kt
fhennig: That'll be more than 120 characters and the linter complains and I think it's not possible to split the function name over multiple lines 😅

fhennig: Maybe I can also just disble the linting rule for that line! Didn't think about that before.

backend/src/main/kotlin/org/loculus/backend/service/submission/SequenceEntriesTable.kt
corneliusroemer: I agree this is a non-ideal function: go through all entries, possibly millions to just find out which organisms there are? We could do the same by looking at values.yaml - or as you suggest add an organisms table.

If we do that, we could then make all organism columns across all tables foreign keys into that organism table.

backend/src/test/kotlin/org/loculus/backend/service/submission/UseNewerProcessingPipelineVersionTaskTest.kt
corneliusroemer: Yep, disabling ktlint is perfectly fine here, also: you can cut out the "the"

corneliusroemer:
```suggestion
    fun `GIVEN pipeline v for 1 organism updates THEN pipeline v for other organism is not updated`() {
```

something like that works as well for shortening

backend/docs/db/schema.sql
```diff
@@ -234,7 +257,10 @@ CREATE VIEW public.external_metadata_view AS
             sequence_entries_preprocessed_data.finished_processing_at
            FROM public.sequence_entries_preprocessed_data
           WHERE (sequence_entries_preprocessed_data.pipeline_version = ( SELECT current_processing_pipeline.version
-                   FROM public.current_processing_pipeline))) cpd
+                   FROM public.current_processing_pipeline
```
corneliusroemer, line 260: This is quite hard to read and probably clearer to read (and possibly also more performant) by joining the organism and current_processing_pipeline_version for a column.

backend/docs/db/schema.sql
```diff
@@ -486,7 +491,10 @@ CREATE VIEW public.sequence_entries_view AS
         END AS processing_result
    FROM ((public.sequence_entries se
      LEFT JOIN public.sequence_entries_preprocessed_data sepd ON (((se.accession = sepd.accession) AND (se.version = sepd.version) AND (sepd.pipeline_version = ( SELECT current_processing_pipeline.version
-           FROM public.current_processing_pipeline)))))
+           FROM public.current_processing_pipeline
```
corneliusroemer, line 494: Here again, this becomes rather unreadable - best to join the  public.current_processing_pipeline table

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -1152,21 +1169,32 @@ private fun Transaction.findNewPreprocessingPipelineVersion(): Long? {
...
                 where
                     not exists( -- ...for which no sequence exists...
                         select
                         from
                             ( -- ...that was processed successfully with the current version...
-                                select accession, version
-                                from sequence_entries_preprocessed_data
+                                select sep.accession, sep.version
+                                from sequence_entries_preprocessed_data sep
+                                join sequence_entries se
```
corneliusroemer, line 1189: To avoid this join, we might as well add the organism to the processed table, it's just one column - a little bit of denormalization might not harm, especially as organism isn't supposed to change

corneliusroemer (commented):
Thanks for this PR!

A few thoughts:

> This is an internal optimization and has no effect on any user facing API.

That's not quite true, it changes what's allowed in the values.yaml: processing version no longer has to be identical across all organisms.

> Before this change, we stored "V1" in this table

Improve clarity: What is "this table". IIUC, we now add an extra column to the processing pipeline version table that keeps track of organisms - previously there was just 1 row for all organisms.

---

Right now we need to join the sequence entry table to find out the organism in the processed data table. That's kind of annoying as we need the organism now a lot in the processed data table.

We could just duplicate that information by adding an organism column to the processed data table - then we don't need to join every time we want to find the current processing version.

backend/src/main/kotlin/org/loculus/backend/service/submission/CurrentProcessingPipelineTable.kt
```diff
@@ -1,11 +1,53 @@
...
     val startedUsingAtColumn = datetime("started_using_at")
+
+    /**
+     * Every organism needs to have a current pipeline version in the CurrentProcessingPipelineTable.
+     * This function sets V1 for all given organisms, if no version is defined yet.
+     */
+    fun setV1ForOrganismsIfNotExist(organisms: Collection<String>, now: LocalDateTime) =
+        CurrentProcessingPipelineTable.batchInsert(organisms, ignore = true) { organism ->
+            this[organismColumn] = organism
+            this[versionColumn] = 1
```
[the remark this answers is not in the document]
chaoran-chen: (This was also previously the case, nothing changed is this PR.)

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -124,10 +125,11 @@ class SubmissionDatabaseService(
         )
     }
 
-    fun getCurrentProcessingPipelineVersion(): Long {
+    fun getCurrentProcessingPipelineVersion(organism: Organism): Long {
         val table = CurrentProcessingPipelineTable
         return table
             .select(table.versionColumn)
```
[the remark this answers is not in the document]
chaoran-chen: In SQL, it is entirely fine to write `select version from current_preprocessing_table where organism = XXX`, you don't need to have `organism` in SELECT.

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -1152,21 +1169,32 @@ private fun Transaction.findNewPreprocessingPipelineVersion(): Long? {
...
-                        where pipeline_version > (select version from current_processing_pipeline)
+                        select distinct sep.pipeline_version
+                        from sequence_entries_preprocessed_data sep
+                        join sequence_entries se
+                            on se.accession = sep.accession
+                            and se.version = sep.version
+                        where 
+                            se.organism = ?
+                            and sep.pipeline_version > (select version from current_processing_pipeline
+                                                        where organism = ?)
```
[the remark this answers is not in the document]
chaoran-chen: I hope/expect that the database will only perform the query once (which doesn't really matter in this case given that the table is tiny). Felix sugggested in #3571 to split up the query to improve the maintainability and readability but I think that's beyond the scope of the PR.

backend/src/main/kotlin/org/loculus/backend/service/submission/SequenceEntriesTable.kt
```diff
@@ -44,6 +44,14 @@ object SequenceEntriesTable : Table(SEQUENCE_ENTRIES_TABLE_NAME) {
...
+        .asSequence()
+        .map {
+            it[SequenceEntriesTable.organismColumn]
```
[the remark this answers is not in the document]
fhennig: This is _Exposed_, and my understanding in that this step is the deserialization of the Row, and thus cannot be skipped - AFAIK. Maybe @fengelniederhammer can chime in?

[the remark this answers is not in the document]
fengelniederhammer: I didn't look at the code, but most probably you have a sequence of exposed results before the map (those results only contain one column, because you selected only one). But you still need to map it to something more useful.

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
fhennig: True, it would be good to make this query simpler, but now that this already works I'm not sure I would make this change to this table (and I won't unless you insist). In other places, other potential changes to the handling of organisms have been mentioned, and so I would rather not denormalize so we can avoid renormalizing in the future, which would just be wasted effort on both ends.

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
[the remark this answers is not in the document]
fhennig: Yes IMO let's leave it for now and rather do a comprehensive change when we get to #3571. Pulling this out will make the place where we inject vars even more full and then that also becomes errorprone/hard to follow, so let's just do it properly in one go if we do it.

backend/src/main/kotlin/org/loculus/backend/service/submission/SequenceEntriesTable.kt
fhennig: Yep, always room for improvement!

backend/src/main/kotlin/org/loculus/backend/service/submission/CurrentProcessingPipelineTable.kt
```diff
@@ -1,11 +1,53 @@
...
+import org.jetbrains.exposed.sql.andWhere
+import org.jetbrains.exposed.sql.batchInsert
 import org.jetbrains.exposed.sql.kotlin.datetime.datetime
+import org.jetbrains.exposed.sql.selectAll
+import org.jetbrains.exposed.sql.update
 
 const val CURRENT_PROCESSING_PIPELINE_TABLE_NAME = "current_processing_pipeline"
 
 object CurrentProcessingPipelineTable : Table(CURRENT_PROCESSING_PIPELINE_TABLE_NAME) {
+    val organismColumn = varchar("organism", 255)
```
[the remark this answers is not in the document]
fhennig: This was taken from other places in the code, where it is done similarly. There is a refactoring issue for it: https://github.com/loculus-project/loculus/issues/1517 .

If this would change to `text` - it would be inconsistent with the rest of the Kotlin code. You can see the organism column definitions for i.e. the SequenceEntriesView or MetadataUploadAuxTable.

backend/src/main/kotlin/org/loculus/backend/service/submission/CurrentProcessingPipelineTable.kt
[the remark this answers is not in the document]
fhennig: Sure! I added it to the other PR so we won't forget it. (But just to be clear: This is not a novel constraint, like Chaoran said).

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -1115,35 +1117,50 @@ class SubmissionDatabaseService(
...
 
-    fun useNewerProcessingPipelineIfPossible(): Long? {
-        log.info("Checking for newer processing pipeline versions")
+    fun useNewerProcessingPipelineIfPossible(): Map<String, Long?> =
+        SequenceEntriesTable.distinctOrganisms().map { organismName ->
+            Pair(organismName, useNewerProcessingPipelineIfPossible(organismName))
+        }.toMap()
+
+    /**
+     * Returns the new Version that was set for the organism, or null if nothing was changed.
```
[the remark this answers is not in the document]
fhennig: Well, just, nothing in general haha. No state was modified - at all. But one can always be clearer!

https://github.com/loculus-project/loculus/pull/3534/commits/bf4e0e9c1f07eb90d13c323e7d8057429d128079

backend/src/main/resources/db/migration/V1.10__pipeline_version_per_organism.sql
fhennig: https://github.com/loculus-project/loculus/pull/3534/commits/9c8e7cac8bc748c5a2c544339d0d522262f66d8c

backend/src/main/resources/db/migration/V1.10__pipeline_version_per_organism.sql
fhennig: https://github.com/loculus-project/loculus/pull/3534/commits/9c8e7cac8bc748c5a2c544339d0d522262f66d8c

much better

backend/src/main/kotlin/org/loculus/backend/service/submission/CurrentProcessingPipelineTable.kt
fhennig: https://github.com/loculus-project/loculus/pull/3534/commits/9c8e7cac8bc748c5a2c544339d0d522262f66d8c

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
theosanderson: (I like normalization personally [even here])

backend/src/test/kotlin/org/loculus/backend/service/submission/UseNewerProcessingPipelineVersionTaskTest.kt
fhennig: disabled the linter: https://github.com/loculus-project/loculus/pull/3534/commits/4c6a05c2e39424f8b26fbd43865d7c1425525889

backend/docs/db/schema.sql
```diff
@@ -466,7 +470,8 @@ CREATE VIEW public.sequence_entries_view AS
     (sepd.processed_data || em.joint_metadata) AS joint_metadata,
         CASE
             WHEN se.is_revocation THEN ( SELECT current_processing_pipeline.version
-               FROM public.current_processing_pipeline)
+               FROM public.current_processing_pipeline
```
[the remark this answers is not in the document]
fhennig: Selecting the version in this way could be viewed as dummy value I guess! 

I think changing this is not in scope of the current PR. The change doesn't modify how the table works - merely updates it so it works with the new ccp table.

fhennig:
> That's not quite true, it changes what's allowed in the values.yaml: processing version no longer has to be identical across all organisms.

Ah yes, good point! I meant that it isn't breaking (but I know that we use 'breaking' a little differently). From what I gathered, this constraint isn't enforced anywhere though, or is it? AFAIKT the Helm chart just passes the version through.

kubernetes/loculus/values.yaml
```diff
@@ -1212,10 +1213,8 @@ defaultOrganisms:
     <<: *defaultOrganismConfig
     preprocessing:
       - <<: *preprocessing
-        configFile:
-          <<: *preprocessingConfigFile
-          nextclade_dataset_server: https://raw.githubusercontent.com/nextstrain/nextclade_data/ebola/data_output
-          nextclade_dataset_name: nextstrain/ebola/sudan
+      - <<: *preprocessing # test having 2 simultaneous versions
+        version: "2"
```
chaoran-chen: @corneliusroemer, this is breaking the test dummy organism because the lineage system definitions is only provided for version 1. This is causing the SILO import job to throw "No URL defined for pipeline version 2."

chaoran-chen (changes requested):

kubernetes/loculus/values.yaml
```diff
@@ -1212,10 +1211,13 @@ defaultOrganisms:
...
-          <<: *preprocessingConfigFile
-          nextclade_dataset_server: https://raw.githubusercontent.com/nextstrain/nextclade_data/ebola/data_output
-          nextclade_dataset_name: nextstrain/ebola/sudan
+        version: 1
+        nextclade_dataset_name: nextstrain/ebola/sudan
+        nextclade_dataset_server: https://raw.githubusercontent.com/nextstrain/nextclade_data/ebola/data_output
+      - <<: *preprocessing # test having 2 simultaneous versions
+        version: 2
+        nextclade_dataset_name: nextstrain/ebola/sudan
+        nextclade_dataset_server: https://raw.githubusercontent.com/nextstrain/nextclade_data/ebola/data_output
```
chaoran-chen, line 1220: @corneliusroemer, @fhennig, this is breaking the dummy test organism because no lineage system definition is specified for pipelien version 2. This causes the SILO import job to throw "No URL defined for pipeline version 2"

chaoran-chen (changes requested):

backend/docs/db/schema.sql
chaoran-chen, line 1: The ingest on the preview is not working. I haven't checked whether this is caused by this PR.

kubernetes/loculus/values.yaml
fhennig: attempt at a fix: https://github.com/loculus-project/loculus/pull/3534/commits/49225145e159d12586bc4085da8f00504b443659

(But maybe we want to add better warnings/tests around this?)

fhennig: maybe the way the lineages are defined could also be changed to be put into the preprocessing config

backend/docs/db/schema.sql
fhennig: the fix worked, but it's not good that the missing URL broke all organisms maybe?

fhennig: (worked as in, things are back now)

chaoran-chen (approved):
Thanks for fixing!

fhennig:
Okay I tested it with west nile, and it works! My now data gets processed by the bumped pipeline, and now the old data is getting reprocessed. Great! Now I'M happy.

fhennig:
waiting for tests to turn green, and will merge then.

PR #3535, opened 2025-01-16
feat(ingest): decrease memory usage of submission batching

anna-parker:
resolves #

preview URL: https://stream-batching.loculus.org/

### Summary

The backend cannot handle submission of over 60k sequences at once, therefore they must be batched. I introduced batching in https://github.com/loculus-project/loculus/pull/3429 but my script required ingest to read both the sequences and metadata files into memory to merge the batches. This was very memory inefficient. 

I now switch to sorting the files using gnu-sort (which should even work for files that are larger than working memory) and then refactor my batching script to read in both metadata and sequence files line by line, creating batched files to be then submitted to the backend. 

### Testing
Use https://github.com/pathoplexus/pathoplexus/tree/main/data-integrity-tests/regression-testing and switch the URLs to this branch and main :-)

### DIFF
--- a/ingest/Snakefile
+++ b/ingest/Snakefile
@@ -36,14 +36,26 @@ if SEGMENTED:
         ALIGN = False
     if ALIGN:
         for segment in config["nucleotide_sequences"]:
-            if config.get("nextclade_dataset_server_map") and segment in config["nextclade_dataset_server_map"]:
-                dataset_server_map[segment] = config["nextclade_dataset_server_map"][segment]
+            if (
+                config.get("nextclade_dataset_server_map")
+                and segment in config["nextclade_dataset_server_map"]
+            ):
+                dataset_server_map[segment] = config["nextclade_dataset_server_map"][
+                    segment
+                ]
             else:
                 dataset_server_map[segment] = config.get("nextclade_dataset_server")
-            if config.get("nextclade_dataset_name_map") and segment in config["nextclade_dataset_name_map"]:
-                dataset_name_map[segment] = config["nextclade_dataset_name_map"][segment]
+            if (
+                config.get("nextclade_dataset_name_map")
+                and segment in config["nextclade_dataset_name_map"]
+            ):
+                dataset_name_map[segment] = config["nextclade_dataset_name_map"][
+                    segment
+                ]
             else:
-                dataset_name_map[segment] = config.get("nextclade_dataset_name") + "/" + segment
+                dataset_name_map[segment] = (
+                    config.get("nextclade_dataset_name") + "/" + segment
+                )
 
 if os.uname().sysname == "Darwin":
     # Don't use conda-forge unzip on macOS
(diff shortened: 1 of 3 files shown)


ingest/scripts/call_loculus.py
```diff
@@ -173,52 +173,20 @@ def get_or_create_group_and_return_group_id(config: Config, allow_creation: bool
...
-
-
 def post_fasta_batches(
     url,
     fasta_file: str,
     metadata_file: str,
     config: Config,
     params: dict[str, str],
-    chunk_size=60000,
+    chunk_size=1000,
```
anna-parker, line 182: I decreased this to see it in action on the preview - it can be increased before submission
```suggestion
    chunk_size=10000,
```

ingest/scripts/call_loculus.py
```diff
@@ -263,18 +300,15 @@ def submit_or_revise(
...
         "groupId": group_id,
     }
     if mode == "submit":
         params["dataUseTermsType"] = "OPEN"
 
-    response = post_fasta_batches(
-        url, sequences, metadata, config, params=params, chunk_size=60000
-    )
-    logger.debug(f"{logging_strings["noun"]} response: {response.json()}")
+    response = post_fasta_batches(url, sequences, metadata, config, params=params, chunk_size=1000)
```
anna-parker, line 311: I decreased this so that the previews test batching but we should increase it again before merging
```suggestion
    response = post_fasta_batches(url, sequences, metadata, config, params=params, chunk_size=10000)
```

ingest/Snakefile
```diff
@@ -468,11 +483,41 @@ rule prepare_files:
...
 
 
+rule sort_fasta:
+    input:
+        sequences="results/{basename}.fasta",
+    output:
+        sorted="results/{basename}_sorted.fasta",
+    shell:
+        """
+        awk '/^>/ {{if (seq) print seq; seq=""; print; next}} {{seq = seq $0}} END {{if (seq) print seq}}' {input.sequences} | \
```
corneliusroemer, line 493: What does this do? Some comment might be nice.
Anya just said: This turns fasta into tab separated, so each sequence and id on one line.
Can use seqkit instead here for readability.

Even better:
- Extract just the ids
- Sort the ids
- Then stream grep/join the full sequences onto the sorted ids

ingest/Snakefile
```diff
@@ -468,11 +483,41 @@ rule prepare_files:
...
+
+
+rule sort_metadata:
+    input:
+        metadata="results/{basename}.tsv",
+    output:
+        sorted="results/{basename}_sorted.tsv",
+    shell:
+        """
+        columnNumber=$(awk -F'\\t' '{{for(i=1;i<=NF;i++) if($i=="submissionId") print i}}' {input.metadata});
```
corneliusroemer, line 507: `tsv-sort` script from tsv-utils

corneliusroemer (approved):

ingest/scripts/call_loculus.py
```diff
@@ -173,73 +172,110 @@ def get_or_create_group_and_return_group_id(config: Config, allow_creation: bool
...
+    number_of_submissions = -1
+    fasta_submission_id = None
+    fasta_record_header = None
+    metadata_header = None
+
+    sequences_batch_output = []
+    metadata_batch_output = []
+
+    with (
+        open(fasta_file, encoding="utf-8") as fasta_file_stream,
```
anna-parker, line 212: potentially replace with seqIOs SeqIO.parse() which should read in the file record by record and handle multi-line sequences which will not be correctly handled in this case

PR #3536, opened 2025-01-16
chore(deployment): Change python3 to python in deploy.py

theosanderson:
I always find myself having to make this change locally as I use conda environments and they don't change `python3`, just `python` to point to the right place. Seeing if (A) it passes CI [it does] (B) anyone objects to it.


anna-parker (approved):
lgtm!

Issue #3460, opened 2024-12-17
Preprocessing errors user experience is degraded (for collectionDate, maybe elsewhere)


theosanderson:
One of our user stories involves submitting a misformatted collection date, observing the error from the preprocessing pipeline and correcting it. This now yields:

<img width="1317" alt="image" src="https://github.com/user-attachments/assets/178d66c2-fe28-4e30-92a3-86d99714825f" />

Problems:
* displayName is highlighted in red and shown below. While the `displayName` is technically derived from the date, this isn't something we need to highlight as it will just cause confusion about what the actual problem is. We should not propagate this error to the displayName
* collectionDate is not shown in the card, it should be (and in red)
* Displaying messages about the lower and the upper range is also confusing. Here we should have a single message referring to the main field
* Displaying "is required" is confusing - the problem is the parsing error

theosanderson:
IMO the change to this should be in preprocessing, and the website should just faithfully display the preprocessing messages. (That's not a statement about who should do the work to address it! And for now we can put it on the backlog)

PR #3567, opened 2025-01-22
fix(website): Don't propagate date parsing errors to concatenated fields

### LINKED ISSUE #3566: Review page: Confusing error message regarding "display name" if underlying field is wrong
I uploaded a slightly adapted version of the CCHF data from the example data (to Pathoplexus staging) and then changed the collection date on the review page to "2020-01-00" and got the below view and error message:

Image

It is unexpected that the "Display Name" field is marked as an error and there is an error message "displayName: Metadata field displayName: Date format is not recognized.".

I assume that this is caused because the collection date is part of the display name and, therefore, the display name could not be computed but the error message is quite confusing.

theosanderson:
resolves #3566 and partly https://github.com/loculus-project/loculus/issues/3460

We may want to make this configurable in some way in the future, but for now it addresses the issue that displayName gets referenced when there are date problems which is confusing to users.

When would this change potentially cause a problem: if an admin had a field that was only part of a concatenation, and not used as a standalone field, so the date parsing would never cause an error.

preview URL: https://prop.loculus.org/

image

### DIFF
--- a/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
@@ -619,15 +619,11 @@ class ProcessingFunctions:
                         {"date": input_data[order[i]]}, output_field, input_fields
                     )
                     formatted_input_data.append("" if processed.datum is None else processed.datum)
-                    errors += processed.errors
-                    warnings += processed.warnings
                 elif type[i] == "timestamp":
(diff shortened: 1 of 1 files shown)


anna-parker (approved):
I think this is an ok fix for the moment - @corneliusroemer can you have a quick look to make sure I haven't forgotten anything?

theosanderson:
I plan to merge this on Monday if I don't hear otherwise

PR #3570, opened 2025-01-22
feat(website): Use a row separator in the search table

theosanderson:
resolves #

preview URL: https://tablerowsep.loculus.org/ebola-sudan/search?

### Summary

This PR adds separators between table rows on the search page. I think it looks quite a bit cleaner:

image

In the process of implementing it I realised that we, I guess accidentally, basically disabled borders on all tables with hardcoded CSS. Therefore I've removed that.


fhennig:
I think a separator between the headers and the first line would also look nice. Especially with some cells having multiple lines. But yes, looks cleaner I agree!

fhennig:
thanks! Just format and then it LGTM

fhennig (approved):

Issue #3574, opened 2025-01-26
Document customDisplay config


chaoran-chen:
We should document how `customDisplay` works. This is currently not documented in https://loculus.org/reference/helm-chart-config/#metadata-type.

PR #3576, opened 2025-01-26
feat(website): support HTML custom display of table entries

chaoran-chen:
preview URL: https://html-custom-display.loculus.org

### Summary

This adds a new type of custom display "html" that allows the maintainer to provide an HTML template to render a table entry.

Background: I'd like to use that to improve the display of the frequency JSONs [here](https://wise-loculus.genspectrum.org/seq/LOC_0000BEV.1).

To demonstrate the feature, the feature was used to make the country bold in the below screenshot.

### Screenshot

image

### PR Checklist

- ~~[ ] All necessary documentation has been adapted.~~ Documentation is missing for the custom display feature and I propose adding it separately (#3574).
- ~~[ ] The implemented feature is covered by an appropriate test.~~ We don't have any tests for the other custom displays and I hope it's ok that we don't have one in this PR.


kubernetes/loculus/values.yaml
```diff
@@ -168,6 +168,9 @@ defaultOrganismConfig: &defaultOrganismConfig
         guidance: Geo-coordinate longitude in decimal degree (WGS84) format, i.e. values in range -180 to 180, where positive values are east of the Prime Meridian.
       - name: geoLocCountry
         displayName: Collection country
+        customDisplay:
+          type: html
+          html: "<b>__value__</b>"
```
chaoran-chen, line 173:
```suggestion
```

Revert before merge

website/src/components/SequenceDetailsPage/DataTableEntryValue.tsx
```diff
@@ -70,4 +78,13 @@ const PlainValueDisplay: React.FC<{ value: TableDataEntry['value'] }> = ({ value
     return <span className='italic'>None</span>;
 };
 
+const generateCleanHtml = (trustedHtml: string, userValue: string): string => {
+    const cleanedValue = sanitizeHtml(userValue, {
```
theosanderson, line 82: Could you explain the behaviour here and how it helps with your goal? When I read this it sounds like it doesn't allow any tags? I don't see how that will let you display a table (except one with only one variable cell)?

theosanderson: Or will you have some complicated javascript in the HTML?

chaoran-chen: My plan was to use a custom element. The `html`-config would be set to something like `<frequency-custom-display>__value__</frequency-custom-display>` and through `additionalHeadHTML`, I'd include a JS file that registers the element. Here's a prototype:

[example.zip](https://github.com/user-attachments/files/18551143/example.zip)

theosanderson: Cool - my only feedback is I feel like`html` is a confusing name. To me it implies that one can supply HTML. Something like `wrappedInHtml` or `htmlTemplate` would be a better name IMO for the `type`.

theosanderson (approved):
happy with this. I think if we decide to actually support tables we should actually implement a table feature directly - and I'd be uncertain as to whether Loculus should document the idea of using a custom element for this, but in terms of having the raw feature - sounds fair!

website/src/components/SequenceDetailsPage/DataTableEntryValue.tsx
chaoran-chen: I like `htmlTemplate`, just renamed it.

chaoran-chen:
Thanks for reviewing, @theosanderson! (As it's not urgent, I'll keep the PR open for another day to see whether someone else has comments.)

website/src/components/SequenceDetailsPage/DataTableEntryValue.tsx
theosanderson: I would also guess that your solution might yield weird behaviour with hydration? Idk if it's worth testing it before merging in case we're adding something that's not of any use?

theosanderson: (maybe you have!)

chaoran-chen: Hmm, good question, haven't thought about that – yes, might be good to give it a try first.

theosanderson: (I edited my comment as I felt less confident but yeah, I am bouncing back and forth so maybe worth checking)

fhennig:
In the files it seems like the config option is still called `html` (not `htmlTemplate`).

I see how this is useful for this JSON thing! Can you maybe come up with more examples? I think having more could be nice to document this feature better.

website/src/components/SequenceDetailsPage/DataTableEntryValue.tsx
chaoran-chen: image

It works!

theosanderson: Did you check the console says nothing about hydration?

theosanderson: Just checked - can't see a message - great!

chaoran-chen:
@fhennig, I think it's fine to keep the value just `html` and call the type `htmlTemplate`? But no strong preference, let me know if you think it's better to change.

This can be used for all kinds of things: to embed an image, a video, map, plot, tree – whatever you can imagine!

kubernetes/loculus/values_preview_server.yaml
```diff
@@ -32,4 +32,4 @@ reduceResourceRequest: true
 previewDocs: true
 robotsNoindexHeader: true
 disableEnaSubmission: false
-additionalHeadHTML: '<script defer data-domain="loculus.org" src="https://plausible.io/js/script.js"></script>'
+additionalHeadHTML: '<script defer data-domain="loculus.org" src="https://plausible.io/js/script.js"></script><script src="https://chaoran-chen.github.io/test-custom-display/frequency-custom-display.js"></script>'
```
chaoran-chen, line 35: Revert before merge

kubernetes/loculus/values.yaml
```diff
@@ -168,6 +168,9 @@ defaultOrganismConfig: &defaultOrganismConfig
         guidance: Geo-coordinate longitude in decimal degree (WGS84) format, i.e. values in range -180 to 180, where positive values are east of the Prime Meridian.
       - name: geoLocCountry
         displayName: Collection country
+        customDisplay:
+          type: htmlTemplate
+          html: "<frequency-custom-display>__value__</frequency-custom-display>"
```
chaoran-chen, line 173: Revert before merge

PR #3580, opened 2025-01-27
chore(deps): switch from no-longer-npm-published xlsx to mirror @lokalise/xlsx

corneliusroemer:
resolves #

preview URL:

### PR Checklist


fhennig:
for the record, we could also intall the SheetJS version that is not on NPM; non-NPM dependencies are do work: https://docs.sheetjs.com/docs/getting-started/installation/nodejs/

Whether this is better than this other mirror maintained by another random person, I don't know. looks fine in principle though!

corneliusroemer:
@fhennig I know we could install the tarball, but I doubt it'll work with dependabot, for example.

I did my due diligence on the maintainer, it's not a random person - or do you disagree?

fhennig (approved):

anna-parker (approved):
oh cool - this actually means far fewer dependencies required overall

Thanks for doing this @corneliusroemer!

PR #3581, opened 2025-01-27
feat!: Allow raw reads input

fhennig:
Requires values.yaml changes.


anna-parker (approved):
could you update the commit message to feat(config)! when you squash&merge?

PR #3588, opened 2025-01-29
feat(Helm): Add JSON schema for `values.yaml` validation

### LINKED ISSUE #1207: Organism technical names must be lower case
Otherwise it leads to a deployment error.

Document this or remove the restriction.

### LINKED ISSUE #2796: Config file schema
We should fix a format for the Loculus configuration and create a schema.

### LINKED ISSUE #2798: Config file validation
After #2796, we should add validation of the configuration files that gives useful feedback to an admin setting up an instance.

fhennig:
resolves #2796, resolves #2798, resolves #1207

preview URL for the new docs: https://docs-values-yaml-schema.loculus.org/reference/helm-chart-config/

### Summary
The main change in this PR is the addition of the `values.schema.json` file to the Helm Chart, which Helm will by default use to validate `values.yaml` (You can do this with `helm lint .`).

The Helm Chart reference is now completely generated from the schema (it was written by hand before).

The new schema has extra keys in the properties (`groups`, `placeholder`, `docsIncludePrefix`) to facilitate generating the documentation in individual chunks.

A new github action was added to lint the existing values.yaml and make sure that we add all the new things we add to the `values.yaml` also to the schema.

The `values.schema.json` now contains really all things we set anywhere in CI or whereever, some of the settings still need to be documented.

### PR Checklist


corneliusroemer:
The schema already works, showing you the schema doesn't allow current values:
image

fhennig:
Ah damn, looks like this sort of composing is actually not supported: https://github.com/helm/helm/issues/10481

That's a shame.

fhennig:
I like how currently the docs are split into sections: https://loculus.org/reference/helm-chart-config/ and I thought I could replicate that with using multiple files as sources - I guess I can't.

fhennig:
A quit search showed that it should be possible! That's a great idea!

fhennig:
The schema should include that the pipeline version starts at v1: https://github.com/loculus-project/loculus/pull/3534#discussion_r1936497899 (thanks Cornelius!)

fhennig:
Maybe it would also be nice to have a few example yamls of stuff that users might want to configure, and then we can also lint/validate those?

fhennig:
In an effort to maybe extract examples and look further into which properties are required, I'm going to look into untangling our huge values.yaml into a minimal set of sensible defaults, and a more "example"-specific part that can be applied on top of the actual defaults.

theosanderson:
> In an effort to maybe extract examples and look further into which properties are required, I'm going to look into untangling our huge values.yaml into a minimal set of sensible defaults

I think one can make a case for making the `defaultOrganisms` much more stripped back, which would make it easier for maintainers to read the master values file, while complicating our dev set up a bit.
But with that exception, I think we broadly do things according to Helm standard practices, which are to have tons of defaults: https://github.com/bitnami/charts/blob/main/bitnami/postgresql/values.yaml

fhennig:
True! I think there should be as many defaults as possible, and all switches etc should have a sensible default state. But exactly the actual "content" (i.e. organisms) and other instance specific defaults I think should be in there. Example:

```
auth:
  smtp:
    host: "in-v3.mailjet.com"
    port: 587
    user: "fafd505de339dd2e9c3e85ad9981af8a"
    replyTo: "noreply@loculus.org"
    from: "noreply@loculus.org"
    envelopeFrom: "noreply@loculus.org"
  identityProviders:
    orcid:
      clientId: "APP-P1P7N7T9YVBHQ4EH"
```

```
seqSets:
  crossRef:
    DOIPrefix: "placeholder"
    endpoint: "https://test.crossref.org"
```

```
lineageSystemDefinitions:
  pangoLineage:
    1: https://raw.githubusercontent.com/loculus-project/loculus/c400348ea0ba0b8178aa43475d5c7539fc097997/preprocessing/dummy/lineage.yaml
```

I also think this isn't great (from a `tpl` file):

```
...
{{- range $key, $instance := (.Values.organisms | default .Values.defaultOrganisms) }}
...
```

By default I would expect the Chart to not contain data (going back to your PSQL example, the defaults don't create any schema or data by default). 

IMO we're using these defaults because it's convenient to us, but from an outside perspective it isn't super obvious what I have to un-set to get a "clean" install.

Of course one could argue that looking into this isn't in the scope of the issue being adressed here, and I can also drop it, but since I've spend so much time looking at the Chart now, I felt like it'd be sensible to tackle this.

theosanderson:
> smtp config

totally agree

>seqSets

I think these are broadly reasonable defaults (placeholders) - one could make it more sophisticated but it seems ok given where we are atm

> lineage systems

This is probably a reason to put these within the organism as per https://github.com/loculus-project/loculus/issues/3603

> By default I would expect the Chart to not contain data (going back to your PSQL example, the defaults don't create any schema or data by default).

So we have considered this, and decided that we wanted to allow people to have a working copy of loculus where they could actually see working organisms with a single command without needing to provide any additional configuration, just by deploying a bare helm chart. We can of course rethink that decision, but it is one we considered: it was considered around external users not just around us, and our tutorials are structured around it.

fhennig:
Ok! If that was intentional (it looked to me like it was just out of convenience for us) then I'll just not do anything to change the values.yaml for now!

theosanderson:
I guess the thought process is that there are lots of times when you would want to deploy an empty postgres DB instance and then populate it (without touching helm again), whereas there are no circumstances where you'd want to want to run loculus without any organisms.

fhennig:
I'll add something like this to the schema: 

```json
{
    "$schema": "http://json-schema.org/draft-07/schema#",
    "definitions": {
        "organism": {
            "type": "object",
            "properties": {
                "property1": { "type": "string" },
                "property2": { "type": "number" }
            },
            "required": ["property1", "property2"]
        }
    },
    "type": "object",
    "properties": {
        "organisms": { "$ref": "#/definitions/organism" },
        "defaultOrganisms": { "$ref": "#/definitions/organism" }
    }
}

```

for the default organisms, so they are also covered by the schema, even though we won't surface them in the docs

kubernetes/loculus/values.schema.json
```diff
@@ -0,0 +1,998 @@
...
+      "required": ["name"]
+    },
+    "organism": {
+      "type": "object",
+      "properties": {
+        "schema": {
+          "groups": ["organism"],
+          "docsIncludePrefix": false,
+          "type": "object",
+          "description": "Object of type [Schema](#schema-type)",
```
theosanderson, line 176:
```suggestion
          "description": "Object of type [Schema](#schema-type)",
```

IMO this would be significantly clearer if called an `Organism Schema`

theosanderson: (or similar)

kubernetes/loculus/values.schema.json
```diff
@@ -0,0 +1,998 @@
+{
+  "title": "values.yaml Schema for the Loculus Helm Chart",
+  "description": "Also has 'groups' and 'placeholder' keys in some places, which are used during docs generation. You can set 'docsIncludePrefix' to false to cut of a path when generating docs. Have a look at the SchemaDocs.astro component in the Loculus docs.",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "definitions": {
+    "metadata": {
```
theosanderson, line 6: image
would be clearer if it was about a `Metadata item` or similar (metadata is plural)

theosanderson:
Great work!

Have you checked it against the Pathoplexus schema?

docs/src/components/SchemaDocs.astro
```diff
@@ -0,0 +1,142 @@
...
+ * Properties in the schema can be assigned to groups by adding a new "groups": ["group1", "group2"] key to them.
+ * This way, you can split schema definition into semantic groups.
+ */
+import { type JSONSchema7, type JSONSchema7Definition, type JSONSchema7TypeName, type JSONSchema7Type } from 'json-schema';
+import MarkdownRenderer from './MarkdownRenderer.astro';
+import rawSchema from '../values.schema.json';
+
+// the 'as any' isn't pretty but needed because our schema doesn't actually conform to the type.
+// it's still nice to use the type to get at least some type hinting when developing.
+let schema: JSONSchema7 = rawSchema as any;
```
chaoran-chen, line 13: What are the differences? Would it be easy to extend/adapt the type to fit?

kubernetes/loculus/values.schema.json
```diff
@@ -0,0 +1,998 @@
+{
+  "title": "values.yaml Schema for the Loculus Helm Chart",
+  "description": "Also has 'groups' and 'placeholder' keys in some places, which are used during docs generation. You can set 'docsIncludePrefix' to false to cut of a path when generating docs. Have a look at the SchemaDocs.astro component in the Loculus docs.",
```
chaoran-chen, line 3:
```suggestion
  "description": "Also has 'groups' and 'placeholder' keys in some places, which are used during docs generation. You can set 'docsIncludePrefix' to false to cut off a path when generating docs. Have a look at the SchemaDocs.astro component in the Loculus docs.",
```

kubernetes/loculus/values.schema.json
chaoran-chen: Maybe "Metadata field"?

docs/src/content/docs/reference/helm-chart-config.mdx
```diff
@@ -22,144 +28,6 @@ The configuration for the Helm chart is provided as a YAML file. It has the foll
...
-            <td>The host URL for CrossRef callbacks</td>
-        </tr>
-        <tr>
-            <td>`gitHubMainUrl`</td>
-            <td>String</td>
-            <td>"https://github.com/loculus-project/loculus"</td>
-            <td>The link that the GitHub icon in the footer points to</td>
-        </tr>
         <tr>
             <td>`dataUseTerms.enabled`</td>
```
chaoran-chen, line 32: Why is dataUseTerms treated differently and not rendered through the schema?

kubernetes/loculus/values.schema.json
```diff
@@ -0,0 +1,998 @@
...
+          "description": "Key used across app to refer to this field."
+        },
+        "displayName": {
+          "groups": ["metadata"],
+          "type": "string",
+          "description": "Name displayed to users."
+        },
+        "type": {
+          "groups": ["metadata"],
+          "type": "string",
```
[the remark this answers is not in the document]
fhennig: Can't the user define arbitrary metadata fields? I could make field `myFooField` right?

docs/src/components/SchemaDocs.astro
fhennig: Hmm, since it's all just in this class and just a few lines I didn't think it's worth it - even with the type fixed the code below would still have to check for presence/absence of the fields. The types are recursively defined and it's a bit of a complex structure, so I didn't bother. I think it's definitely not _easy_ (maybe also not _hard_ but I think it's tedious and possibly a lot of lines of code)

docs/src/content/docs/reference/helm-chart-config.mdx
fhennig: That was a copy-paste leftover, whoops! it was also in the schema already (but without the doc strings) - I've fixed it here now, thanks!: https://github.com/loculus-project/loculus/pull/3588/commits/b4a6ce4881531763f429f8e6b13b0afaba6d3203

kubernetes/loculus/values.schema.json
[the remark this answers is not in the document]
theosanderson: To my knowledge, there are a limited set of `type`s for fields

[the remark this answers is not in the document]
chaoran-chen: There is only a fixed set of possible types: https://github.com/loculus-project/loculus/blob/main/backend/src/main/kotlin/org/loculus/backend/config/Config.kt#L36

docs/src/components/SchemaDocs.astro
chaoran-chen: ok, makes sense!

kubernetes/loculus/values.schema.json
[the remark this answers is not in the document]
fhennig: Ah yes I thought somehow this was about `name` - I misread. Yes should definitely be an enum.

kubernetes/loculus/values.schema.json
fhennig: https://github.com/loculus-project/loculus/pull/3588/commits/8905db7f6628fcf9650c371719cd8bdbfc1f46c3

kubernetes/loculus/values.schema.json
fhennig: https://github.com/loculus-project/loculus/pull/3588/commits/3f8ca53e7ace5617a1bbf92e03c253b1d8ae47d0

fhennig:
I just checked against the PPX values (good point Theo!) and found only one issue, I have made a PPX PR for it: https://github.com/pathoplexus/pathoplexus/pull/379

theosanderson:
Thanks is it at all possible to allow numbers for the example fields? This actually feels a kind of annoying restriction (not important of course)

kubernetes/loculus/values.schema.json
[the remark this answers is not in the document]
fhennig: https://github.com/loculus-project/loculus/pull/3588/commits/dce4259a7ec46ec5a553591694945e8f36d001fd

fhennig:
I think it should be possible! Not sure where the example fields are used, I'll dig a little

theosanderson:
Just used for docs I think

chaoran-chen:
It's here defined as "string | number"

docs/src/content/docs/reference/helm-chart-config.mdx
chaoran-chen: @fhennig, the datauseterms are not mentioned in the schema docs at all anymore somehow?

chaoran-chen:
Strange.. Try rerunning the pipelines again?

fhennig:
green now, maybe a hiccup because we did a prod deployment.

kubernetes/loculus/values.schema.json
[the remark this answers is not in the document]
fhennig: I also checked this with the ppx values.yaml etc.

docs/src/content/docs/reference/helm-chart-config.mdx
fhennig: oh yes damn, https://github.com/loculus-project/loculus/pull/3588/commits/13fd107ed5b2bb063a09266ef029541ae3f49644

chaoran-chen (approved):
Nice, it looks all good to me now!

kubernetes/loculus/values.schema.json
```diff
@@ -0,0 +1,1005 @@
...
+            "cpu": { "type": "string" }
+          },
+          "required": ["memory", "cpu"],
+          "additionalProperties": false
+        }
+      },
+      "required": ["requests", "limits"],
+      "additionalProperties": false
+    },
+    "resources": {
```
theosanderson, line 950: sorry but this will likely need tweaking as a result of https://github.com/loculus-project/loculus/pull/3680 which was a hotfix

fhennig: Could you kindly provide a values.yaml snippet that I can use as a basis? You probably have one somewhere right?

theosanderson: That PR does contain one :)  as does https://github.com/pathoplexus/pathoplexus/pull/381

fhennig: Alright! Thanks sorry I didn't see.

https://github.com/loculus-project/loculus/pull/3588/commits/030bcd25c434048345bc96f6e50dccbfee3919e2

chaoran-chen (approved):

PR #3593, opened 2025-01-31
chore(deps): Bump backend JVM version to 23

fhennig:
21 (current) is LTS and I understand if we want to stay on LTS, but 23 would be latest.

### PR Checklist


fengelniederhammer:
I'd prefer to stay on LTS. We don't depend on JVM features that much since we use Kotlin. Thus I would rather go with less maintenance.

fhennig:
ok makes sense.

PR #3594, opened 2025-01-31
fix(CI): extend/adjust timeout for pods to be ready before integration tests

theosanderson:
We are seeing flakiness due to the wait for pods to be ready not being met in time on `main`. Now I have copied the more complex configuration over from the E2E tests.


anna-parker (approved):
lgtm

PR #3604, opened 2025-02-03
fix(website): ....

fhennig:
resolves #

preview URL:

### PR Checklist


theosanderson (approved):
Thanks!

theosanderson:
(If you go to the bulk page, then turn off the data use terms display, then refresh then it reappears - but that's incredibly minor and unlikely to happen)

fhennig:
Ah yes true. I now briefly looked into how to instead pass that in the URLSearchParams but yeah not worth the effort ATM.

PR #3605, opened 2025-02-03
fix-www-redirect

theosanderson:
I will try to test this


corneliusroemer (approved):
Thanks! I tested the preview and except for certificate warning due to wildcard it worked.

PR #3628, opened 2025-02-05
fix(website): avoid duplicated items in download url and respect choice to include old versions

### LINKED ISSUE #3626: Repetition of fields in Download URLs and we were not respecting the choice to include all old versions
Here is a download URL:

`https://lapis.pathoplexus.org/mpox/sample/details?downloadAsFile=true&downloadFileBasename=mpox_metadata_2025-02-05T0105&versionStatus=LATEST_VERSION&isRevocation=false&dataFormat=tsv&nucleotideMutations=&aminoAcidMutations=&nucleotideInsertions=&aminoAcidInsertions=&versionStatus=LATEST_VERSION&isRevocation=false`

the `versionStatus=LATEST_VERSION&isRevocation=false` appears twice! 

Also when you tick to include old versions it doesn't work

theosanderson:
resolves #3626

We ended up with duplicate parameters in our download URL - because they were being set both from the download form and from the original search query. Now we overwrite the original search query with whatever is in the form. It's possible that previously we were not actually respecting the form about open data - depending how LAPIS processed duplicate entries. Also we were not respecting the choice about including old versions.

preview URL: http://dupe.loculus.org (no sequences due to ingest issues currently, but you can check the download URLs)

### DIFF
--- a/website/src/components/SearchPage/DownloadDialog/DownloadUrlGenerator.ts
+++ b/website/src/components/SearchPage/DownloadDialog/DownloadUrlGenerator.ts
@@ -40,7 +40,6 @@ export class DownloadUrlGenerator {
         params.set('downloadAsFile', 'true');
         params.set('downloadFileBasename', this.generateFilename(option.dataType));
 
-        console.log('option', option);
         excludedParams.add(VERSION_STATUS_FIELD);
         excludedParams.add(IS_REVOCATION_FIELD);
         if (!option.includeOldData) {
--- a/website/src/components/SearchPage/DownloadDialog/DownloadUrlGenerator.ts
+++ b/website/src/components/SearchPage/DownloadDialog/DownloadUrlGenerator.ts
@@ -46,9 +46,7 @@ export class DownloadUrlGenerator {
         if (!option.includeOldData) {
             params.set(VERSION_STATUS_FIELD, versionStatuses.latestVersion);
             params.set(IS_REVOCATION_FIELD, 'false');
-
         }
-
 
         if (!option.includeRestricted) {
             params.set('dataUseTerms', 'OPEN');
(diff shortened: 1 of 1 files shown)


website/src/components/SearchPage/DownloadDialog/DownloadUrlGenerator.ts
```diff
@@ -53,7 +53,7 @@ export class DownloadUrlGenerator {
         }
 
         downloadParameters.toUrlSearchParams().forEach(([name, value]) => {
-            if (value.length > 0) {
+            if (value.length && !params.has(name)) {
```
theosanderson, line 56: Nope, this rightly makes tests fail as it would mess up the entry of multiple accessionVersions

chaoran-chen (approved):
Thanks! Looks good as a hotfix.

corneliusroemer (commented):
We should add at least a unit test for this

corneliusroemer:
> It's possible that previously we were not actually respecting the form about open data - depending how LAPIS processed duplicate entries.

Could we get an answer to this question? It determines how bad the issue is here

website/src/components/SearchPage/DownloadDialog/DownloadUrlGenerator.ts
```diff
@@ -35,28 +35,38 @@ export class DownloadUrlGenerator {
     public generateDownloadUrl(downloadParameters: SequenceFilter, option: DownloadOption) {
         const baseUrl = `${this.lapisUrl}${getEndpoint(option.dataType)}`;
         const params = new URLSearchParams();
+        const excludedParams = new Set<string>();
 
         params.set('downloadAsFile', 'true');
         params.set('downloadFileBasename', this.generateFilename(option.dataType));
+
+        excludedParams.add(VERSION_STATUS_FIELD);
```
corneliusroemer, line 43: I'd probably just initialize the set immediately with these 2 params? Bit cleaner :)

PR #3634, opened 2025-02-06
feat!(website): make metadata docs generally available

### LINKED ISSUE #3525: Make metadata docs generally available
#3477 added a `MetadataTable.astro` component to show a table with metadata description information but it is not used in the Loculus code but only in the overlay version for Pathoplexus. This feature would be also useful for other Loculus instances and we should generalize it.

## Acceptance criteria

- Clicking on the link to "metadata format" on the submit page should show the metadata table.
- The metadata table could be in a modal or in a separate page.
- Optional: The maintainer can decide in the config on a per-organism level whether a metadata docs. should be available.

chaoran-chen:
resolves #3525

preview URL: https://metadata-docs.loculus.org

### Summary

Previously, when clicking on the "metadata format" link on a submit page, the user is redirected to a page that just said "Metadata Format Fill this in with tips for your loculus instance."

With this PR, the user gets redirected to `/<organism>/submission/metadataformat` which the organism-specific metadata docs. I am putting it under `/submission` as it is specific to the submission but might not reflect the format of the metadata that the data user receives.

Pathoplexus can overwrite the link, I prepared a PR: https://github.com/pathoplexus/pathoplexus/pull/367

### Screenshot

image

### PR Checklist


chaoran-chen:
Thanks for reviewing! Yes, `MetadataTable` is used in Pathoplexus [here](https://github.com/pathoplexus/pathoplexus/blob/main/monorepo/website/src/pages/docs/concepts/metadataformat.mdx?plain=1#L27). In fact, quite a lot of component are only used in Pathoplexus but dead code here and it would be great to find a way to address it better.

PR #3635, opened 2025-02-06
feat(website): Make all metadata sections in format page expanded by default

theosanderson:
Otherwise searching is really hard
https://expanded.loculus.org/ebola-sudan/submission/metadataformat

https://loculus.slack.com/archives/C061ZQQM3N1/p1738851331300799

Stop gap while we consider other options

Also expands the width of the name column to avoid too much wrapping


emmahodcroft (approved):
looks good to me, cannot comment on actual code

PR #3640, opened 2025-02-07
feat(website): surface reference sequence more clearly on mutations

### LINKED ISSUE #3599: Surface reference sequence ID for mutations, etc. more prominently
We do surface the reference sequence (#2238) but under a relatively subtle icon. It has been suggested that we do this more prominently. (I take some responsibility for us deciding to make it subtle - my bad).

https://microbial-bioinfo.slack.com/archives/C2BE1SW91/p1738433750298669

theosanderson:
resolves #3599

preview URL: https://mutationinfo.loculus.org/ebola-sudan/search?

image
image

### Summary
Adds a more explicit description of the reference sequence. We keep the subtler version with an icon and a popup for the QC info as otherwise it would be overwhelming to see this text repeated so many times.

### PR Checklist


chaoran-chen (approved):
Looks good to me, thanks!

PR #3643, opened 2025-02-07
add support for NCBI gateway URL

### LINKED ISSUE #3629: Ingest breakage
Ingest is broken on main.loculus.org

Logs show (for example):

```
Error: 3048448 appears to be an invalid tax name.  This could be due to unexpected special characters.
```

Presumably relates to NCBI virus or our interaction with it

theosanderson:
https://support-gateay-url.loculus.org/

resolves #3629 for now

We are currently unable to connect from NCBI datasets from our Hetzner server

We add the capacity for NCBI datasets to be configured to receive an alternative gateway URL. Atm for our preview servers this is a proxy I am hosting on one of my servers. Hopefully if this causes our traffic to decrease we will get unblacklisted and we can move to a proxy hosted here at Loculus.

### DIFF
--- a/ingest/Snakefile
+++ b/ingest/Snakefile
@@ -96,7 +96,7 @@ rule fetch_ncbi_dataset_package:
             --no-progressbar \
             --filename {output.dataset_package} \
             --api-key {params.api_key} \
-            {f'--gateway-url {params.gateway_url}' if params.gateway_url else ''}\
+            {('--gateway-url ' + params.gateway_url) if params.gateway_url else ''}
         """
 
 
--- a/ingest/Snakefile
+++ b/ingest/Snakefile
@@ -23,6 +23,7 @@ TAXON_ID = config["taxon_id"]
 SEGMENTED = config["segmented"]
 LOG_LEVEL = config.get("log_level", "INFO")
 NCBI_API_KEY = os.getenv("NCBI_API_KEY")
+NCBI_GATEWAY_URL = config.get("ncbi_gateway_url")
 FILTER_FASTA_HEADERS = config.get("filter_fasta_headers", None)
 APPROVE_TIMEOUT_MIN = config.get("approve_timeout_min")  # time in minutes
 CHECK_ENA_DEPOSITION = config.get("check_ena_deposition", False)
@@ -88,12 +89,14 @@ rule fetch_ncbi_dataset_package:
     params:
         taxon_id=TAXON_ID,
         api_key=NCBI_API_KEY,
+        gateway_url=NCBI_GATEWAY_URL,
     shell:
         """
         datasets download virus genome taxon {params.taxon_id} \
             --no-progressbar \
             --filename {output.dataset_package} \
-            --api-key {params.api_key}\
+            --api-key {params.api_key} \
+            {f'--gateway-url {params.gateway_url}' if params.gateway_url else ''}\
         """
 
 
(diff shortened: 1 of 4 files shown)


corneliusroemer (dismissed):
Nice! doesn't actually seem to allow datasets to work though 😬

kubernetes/loculus/values.yaml
```diff
@@ -1,5 +1,6 @@
 environment: server
 robotsNoindexHeader: false
+ncbiGatewayUrl: "https://ncbi-temp-proxy.api.taxonium.org/datasets/v2"
```
chaoran-chen, line 3: Shouldn't this rather go to the preview config and not be set here as a default?

theosanderson:
> doesn't actually seem to allow datasets to work though 😬
I've tested it locally and it (the proxy) does. There was a snakemake formatting issue.

ingest/Snakefile
```diff
@@ -88,12 +89,14 @@ rule fetch_ncbi_dataset_package:
     params:
         taxon_id=TAXON_ID,
         api_key=NCBI_API_KEY,
+        gateway_url=NCBI_GATEWAY_URL,
```
corneliusroemer, line 92:
```suggestion
        gateway_url=f'--gateway-url {NCBI_GATEWAY_URL}' if NCBI_GATEWAY_URL else '',
```

ingest/Snakefile
```diff
@@ -88,12 +89,14 @@ rule fetch_ncbi_dataset_package:
...
     shell:
         """
         datasets download virus genome taxon {params.taxon_id} \
             --no-progressbar \
             --filename {output.dataset_package} \
-            --api-key {params.api_key}\
+            --api-key {params.api_key} \
+            {('--gateway-url ' + params.gateway_url) if params.gateway_url else ''}
```
corneliusroemer, line 99:
```suggestion
            {params.gateway}
```

theosanderson:
Please can you ask before pushing to other people's branches in general - I think I had potentially already fixed it?

corneliusroemer (approved):
It actually works now! :) 
image

Good point @chaoran-chen re adding to preview overrides - otherwise good to go I think.

kubernetes/loculus/values.yaml
chaoran-chen: (Just for later, when this works and should be merged: this config is probably best to go under `ingest.configFile`?)

theosanderson: I don't think we have an ingest configuration that sits outside of organism configuration? IMO this should

corneliusroemer:
> Please can you ask before pushing to other people's branches in general - I think I had potentially already fixed it?
I'll make a PR next time, I just wanted to help fix it quickly - your fix wouldn't have worked. I don't think snakemake supports arbitrary Python in the shell string param.

Brave Browser 2025-02-07 16 55 54

kubernetes/loculus/values.yaml
chaoran-chen: We could introduce one, maybe something as simple as this for now?

```
ingest:
  ncbiGatewayUrl: ...
```

Otherwise, it seems very unclear to me what `ncbiGatewayUrl` applies to/means.

theosanderson:
OK - thanks for the info! Suggestions work great too (as you did)

kubernetes/loculus/values.yaml
theosanderson: (implemented, thanks!)

Issue #3457, opened 2024-12-17
UI: Make it more obvious that the table is x-scrollable


fhennig:
This is a screenshot of the search table:

<img width="1076" alt="image" src="https://github.com/user-attachments/assets/289e4559-85c4-425b-9c0f-e5287918ed1e" />

There are a few more columns to the right, but it isn't clear from the UI.

theosanderson:
I also created an issue for this (https://github.com/loculus-project/loculus/issues/3387) but good to have yours to make it clearer what the issue is. I have a clear idea of how we can fix this - using the browser window's scrollbar - and made a partial demo in https://github.com/loculus-project/loculus/pull/3386 but it might be a fair bit of work to get it fully working

theosanderson:
(I guess we could also use some other approach where we make a fake scrollbar)

chaoran-chen:
What would you feel about this solution: we could show an arrow with a message "scroll for more fields" if it is possible to scroll, like this:

<details><summary>Screenshot</summary>

<img width="1128" alt="image" src="https://github.com/user-attachments/assets/ff6eb3b1-e7b6-4234-a5cd-05327efe7899" />

</details>

chaoran-chen:
Disney+ solves this problem by always showing a bit of the next column:

<img width="1276" alt="Image" src="https://github.com/user-attachments/assets/222501c8-4537-4046-8c7f-1d83a2adc778" />

We could try the same by using some smart calculations of the column width.

PR #3644, opened 2025-02-07
feat(website): Improve scroll behaviour in search table

theosanderson:
resolves https://github.com/loculus-project/loculus/issues/3457

preview URL: https://improve-scroll.loculus.org/

image

the scrollbar at the bottom of the page is the result of this PR

Makes it obvious you can scroll the table by implementing a custom horizontal scroll bar that always appears. (Written with o3-mini)
Also truncates tooltip expansions on the field which was always a todo and became more important with this since over-long tooltips make native scrollbars appear.
Negatives: will not match the users expected scroll appearance e.g. on windows probably

### DIFF
--- a/website/src/components/SearchPage/ScrollContainer.tsx
+++ b/website/src/components/SearchPage/ScrollContainer.tsx
@@ -28,7 +28,7 @@ const ScrollContainer: React.FC<ScrollContainerProps> = ({ children }) => {
 
             setMaxScroll(scrollWidth - clientWidth);
 
-            const computedTrackWidth = clientWidth * 0.95;
+            const computedTrackWidth = clientWidth - 40;
             setHandleWidth((clientWidth / scrollWidth) * computedTrackWidth);
 
             const computedLeft = rect.left + (clientWidth - computedTrackWidth) / 2;
--- a/website/src/components/SearchPage/ScrollContainer.tsx
+++ b/website/src/components/SearchPage/ScrollContainer.tsx
@@ -50,7 +50,7 @@ const ScrollContainer: React.FC<ScrollContainerProps> = ({ children }) => {
     setScrollLeft(e.currentTarget.scrollLeft);
   };
 
-  const onMouseDownHandle = (e: ReactMouseEvent<HTMLDivElement, MouseEvent>) => {
+  const onMouseDownHandle = (e: ReactMouseEvent) => {
     setDragging(true);
     setStartX(e.clientX);
     setStartScrollLeft(scrollLeft);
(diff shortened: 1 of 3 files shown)


theosanderson:
Kind of working now: https://improve-scroll.loculus.org/west-nile/search? but needs improvements  and styling

website/src/components/SearchPage/ScrollContainer.tsx
chaoran-chen, line 1: Thanks Theo! It's an interesting approach. On a big screen and mac, it looks not bad but on mobile, it looks rather awkward in my opinion. In particular, it doesn't work well with overlays at the moment – is that something that can be fixed?

<details><summary>Screenshots</summary>


____


</details>

---

What would you feel about this alternative solution: we could show an allow with a message "scroll for more fields" if it is possible scroll, like this:

<details><summary>Screenshot</summary>

image

</details>

theosanderson: Thanks for flagging the overlays point – that is hopefully fairly fixable with z-index.

I just tried it on mobile and it actually felt fine to me – maybe it depends on the device – but I also don't think the old behaviour was a problem on mobile so I'd be open to hiding it.

Re the options you proposed I think I do have a preference for one involving scrollbars (i.e. either this one or using the native scrollbar). Two reasons: A. we show a vertical scrollbar so it seems weird not to show a horizontal one. B. the scrollbar is actually functional, it enables people to scroll with just an old fashioned mouse.

theosanderson: Oh I may have stupidly hidden the vertical scroll bar – if so that's unintentional and will fix!

theosanderson: I fixed the overlay issue, generally improved the styling in minor ways, and added this point o our agenda fro the Monday meeting

website/src/components/SearchPage/Table.tsx
```diff
@@ -134,106 +136,114 @@ export const Table: FC<TableProps> = ({
...
-                                )}
-                            </th>
-                            <th
-                                onClick={() => handleSort(primaryKey)}
-                                className='px-2 py-2 md:pl-6 text-xs font-medium tracking-wider text-gray-500 uppercase cursor-pointer text-left'
-                            >
-                                {capitalCase(primaryKey)} {orderBy.field === primaryKey && orderIcon}
-                            </th>
-                            {columns.map((c) => (
+        <div className='text-sm'>
```
theosanderson, line 139: note to reviewers - the only actual change here is wrapping stuff in ScrollContainer - the diffs are just from the change to the indentation that results

rneher:
Click vs scroll ambiguity on firefox/linux:
image

anna-parker (approved):
lgtm!

website/src/components/SearchPage/ScrollContainer.tsx
theosanderson: Resolving this based on Chaoran's verbal comments in the meeting - though I think it would be good to keep improving the scrollbar post-merge

PR #3645, opened 2025-02-07
fix!(deposition): allow explicit specification of ENA deposition blocking fields

corneliusroemer:
Breaking in that Pathoplexus needs to add:

```yaml
depositionBlockingFields:
  - ncbiReleaseDate
  - ncbiUpdateDate
  - ncbiSubmitterCountry
  - insdcAccessionBase
  - insdcVersion
  - insdcAccessionFull
  - bioprojectAccession
  - gcaAccession
  - biosampleAccession
  - ncbiSourceDb
  - ncbiVirusName
  - ncbiVirusTaxId
```
to values.yaml

We now accept SRA accession from users, so we shouldn't block if that is present anymore.

See https://loculus.slack.com/archives/C0757PTR607/p1738953208387609

Tested on staging and it work:

image

### DIFF
--- a/ena-submission/scripts/get_ena_submission_list.py
+++ b/ena-submission/scripts/get_ena_submission_list.py
@@ -1,12 +1,12 @@
 import json
 import logging
+from collections.abc import Iterator
 from pathlib import Path
 from typing import Any
-from collections.abc import Iterator
 
 import click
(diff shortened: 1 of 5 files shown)


anna-parker:
@corneliusroemer ok if I close this?

Issue #3466, opened 2024-12-18
Allow groups to specify which BioProject we upload their sequences into


anna-parker:
By default we should still create a bioproject for each (group, organism), but we should also let users give a bioproject accession that we can upload sequences to.

At the moment we can hack this with one transaction in the ena DB. In future we should add this as a field when users upload sequences for a specific organism and they should be able to see it on their group page. We will also need to be able to update this in the deposition pod.

anna-parker:
Steps: 

1. Allow input of bioproject and biosample (remove `noInput` in values.yaml), potentially add description to tell users when to add this
2. Modify the DB table PROJECT_table to not use (group_id, organism) as unique ID -> small DB migration
3. Modify the SUBMISSION_table code to store PROJECT_table's new id for new created bioproject
4. Modify code: if bioproject exists create a new entry in PROJECT_table with the details - do not create a new bioproject, else create new bioproject for (group_id, organism) as before
5. Modify code: if biosample exists create a new entry in SAMPLE_table with details - do not create a new biosample, else create a new biosample as before

emmahodcroft:
Ok, so the idea is that users can use these fields, but that we don't really surface/publicise it, and leave it for advanced users/those who ask. 
We should maybe clarify in the docs that if you provide a biosample we do not send on any metadata so it relies on them having the metadata they want in the biosample!

emmahodcroft:
We should also check if the same thing will happen as did with the sraAccession - we had a check in place that we don't upload to ENA if this field is present to prevent double-submission - that check may also be present on these fields, so we will also have to allow upload even if present?

theosanderson:
This concept of accepting metadata that we don't send feels like a big change for me: it breaks the concept of INSDC synchronisation it seems like? IMO that's a real hack that we should only deploy when strictly necessary because of a request from a user and not something that should be a typical practice or advertised. IMO it's very rare that the order of upload is raw reads then consensus (though clearly it does happen)

anna-parker:
> We should also check if the same thing will happen as did with the sraAccession - we had a check in place that we don't upload to ENA if this field is present to prevent double-submission - that check may also be present on these fields, so we will also have to allow upload even if present?

I modified this in https://github.com/loculus-project/loculus/pull/3647 - we should be notified of the sequences with external metadata fields now - I will change the message once these changes have merged

anna-parker:
> This concept of accepting metadata that we don't send feels like a big change for me

[...]

PR #3650, opened 2025-02-10
feat(deposition)!: Allow groups to specify which BioProject (and BioSample) we upload their sequences into 

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/3466

preview URL:

Breaking changes: Please merge the corresponding PP PR with these changes - https://github.com/pathoplexus/pathoplexus/pull/375

### Summary

1. Allow input of bioproject and biosample (remove noInput in values.yaml), add description to tell users when to add this
2. Modify the DB table PROJECT_table to not use (group_id, organism) as unique ID -> small DB migration
3. Modify the SUBMISSION_table code to store PROJECT_table's new id for new created bioproject
4. Modify code: if bioproject exists create a new entry in PROJECT_table with the details - do not create a new bioproject, else create new bioproject for (group_id, organism) as before
5. Modify code: if biosample exists create a new entry in SAMPLE_table with details - do not create a new biosample, else create a new biosample as before
6. Add `RUN-REF` to manifest if `insdcRawReadsAccession` exists in metadata: 
image

### PR Checklist

- [x] Testing: ran locally and submitted to the ENA dev server, checked that I: 
- can submit sequences without bioproject and biosample 
- can submit sequences with preexisting bioproject
- can submit sequences with preexisting bioproject and biosample
- can add RUN_REF if `insdcRawReadsAccession` is given

### DIFF
--- a/ena-submission/src/ena_deposition/create_project.py
+++ b/ena-submission/src/ena_deposition/create_project.py
@@ -178,6 +178,7 @@ def submission_table_update(db_config: SimpleConnectionPool):
             update_values = {
                 "status_all": StatusAll.SUBMITTED_PROJECT,
                 "center_name": corresponding_project[0]["center_name"],
+                "project_id": corresponding_project[0]["project_id"]
             }
             update_db_where_conditions(
                 db_config,
--- a/ena-submission/src/ena_deposition/create_project.py
+++ b/ena-submission/src/ena_deposition/create_project.py
@@ -113,6 +113,9 @@ def submission_table_start(db_config: SimpleConnectionPool):
         seq_key = {"accession": row["accession"], "version": row["version"]}
 
         if "bioprojectAccession" in row["metadata"]:
+            logger.debug(
+                f"Accession {row['accession']} already has bioprojectAccession in metadata"
+            )
             bioproject = row["metadata"]["bioprojectAccession"]
             corresponding_group = find_conditions_in_db(
                 db_config, table_name="project_table", conditions=group_key
@@ -120,9 +123,12 @@ def submission_table_start(db_config: SimpleConnectionPool):
             corresponding_project = [
                 project
                 for project in corresponding_group
-                if project["results"].get("bioproject_accession") == bioproject
+                if project["result"].get("bioproject_accession") == bioproject
             ]
             if len(corresponding_project) == 1:
+                logger.debug(
+                    "bioprojectAccession is already in project_table - adding id to submission_table"
+                )
                 update_values = {
                     "status_all": StatusAll.SUBMITTED_PROJECT,
                     "center_name": corresponding_project[0]["center_name"],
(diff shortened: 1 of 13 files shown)


theosanderson:
Thanks for this! I don't have the chops to review the code, but my overarching thoughts:

I'm fine with this as a hack to support a current user at the moment. In terms of an architected plan for where we're headed, my ideas would be fairly different: I would allow users to optionally select an existing BioProject to associate with their new `group` rather than having a new one created for the `group`, and I think long-term Loculus should always have a means of updating metadata in INSDC whenever there is an inconsistency between the latest data in Loculus and the latest data submitted to INSDC (so I would plan to drop the existing-BioSample support, but possibly after Loculus support raw reads).

anna-parker:
> so I would plan to drop the existing-BioSample support, but possibly after Loculus support raw reads

This sounds good! Think when we support that this will no longer be required

>  In terms of an architected plan for where we're headed, my ideas would be fairly different: I would allow users to optionally select an existing BioProject to associate with their new group rather than having a new one created for the group

I'm still not fully sold on only allowing one `bioproject` per group - but we can discuss this later maybe

> I think long-term Loculus should always have a means of updating metadata in INSDC whenever there is an inconsistency between the latest data in Loculus and the latest data submitted to INSDC

I agree that would be great - I think as brokers we should be able to revise biosamples so maybe we can get to that state

corneliusroemer (commented):
Maybe we can check that the project and sample we got exist before we put them into the project/sample tables.

Should be as simple as making an HTTP call to https://www.ebi.ac.uk/ena/browser/view/PRJEB402 etc.

anna-parker:
Good tests: https://www.ncbi.nlm.nih.gov/bioproject/PRJNA485481 does not exist, https://www.ncbi.nlm.nih.gov/bioproject/PRJNA231221 does

ena-submission/src/ena_deposition/create_project.py
```diff
@@ -88,13 +88,81 @@ def construct_project_set_object(
     return ProjectSet(project=[project_type])
 
 
-def submission_table_start(db_config: SimpleConnectionPool):
+def set_project_table_entry(db_config, config, row):
```
corneliusroemer, line 91: This function would be a logical place to test for existence of the bioproject and error if it doesn't exist

ena-submission/src/ena_deposition/create_sample.py
```diff
@@ -143,13 +143,42 @@ def construct_sample_set_object(
     return SampleSetType(sample=[sample_type])
 
 
+def set_sample_table_entry(db_config, row, seq_key):
```
corneliusroemer, line 146: Same, here, good to check for sample's existence against ENA API like this: https://www.ebi.ac.uk/ena/browser/view/PRJEB402

PR #3671, opened 2025-02-11
(merging into Anya's PR) Suggestions for data use terms selector

theosanderson:
- very slightly rephrases text
- links to the specific pages on the two terms of use
- corrects a reference to "Loculus" to Pathoplexus

At some point we should disentangle all references to Pathoplexus and Pathoplexus paths from this, but:
- we can now disable data use terms entirely
- there were already references to Pathoplexus so this doesn't really make it worse

amends docs (to clarify something that was already true):
image

### DIFF
--- a/website/src/components/DataUseTerms/DataUseTermsSelector.tsx
+++ b/website/src/components/DataUseTerms/DataUseTermsSelector.tsx
@@ -89,7 +89,7 @@ const DataUseTermsSelector: FC<DataUseTermsSelectorProps> = ({
                 <div className='text-xs pl-8 text-gray-500 pb-4'>
                     Anyone can use and share the data (though we believe researchers should exercise scientific
                     etiquette, including the importance of citation).{' '}
-                    <a href={routes.datauseTermsPage()} className='text-primary-600'>
+                    <a href="/about/terms-of-use/open-data" className='text-primary-600'>
                         Find out more
                     </a>
                     .
@@ -114,7 +114,7 @@ const DataUseTermsSelector: FC<DataUseTermsSelectorProps> = ({
                 <div className='text-xs pl-8 text-gray-500 mb-4'>
                     Data use will be restricted for a period of time. The sequences will be available but there will be
                     limitations on how they can be used by others.{' '}
-                    <a href={routes.datauseTermsPage()} className='text-primary-600'>
+                    <a href="/about/terms-of-use/restricted-data" className='text-primary-600'>
                         Find out more
                     </a>
                     .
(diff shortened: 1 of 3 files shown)


website/src/components/DataUseTerms/DataUseTermsSelector.tsx
chaoran-chen, line 1: I don't think that we should hardcode "Pathoplexus" here. [Here](https://loculus.org/for-administrators/data-use-terms/) on this page, we describe the data use terms as a feature that can, in principle, also be used by others. Can't we simply use the site name from the config here?

chaoran-chen: > there were already references to Pathoplexus so this doesn't really make it worse

I'd be in favor of trying to reduce such references instead of increasing them and in this case, I believe that it is simple enough to use the site name?

theosanderson: Yes - just we already used the word "Pathoplexus" twice so I don't think it really makes a difference to whether people can use it.

The tasks to make this PR avoid Pathoplexus-specific stuff would be:

- use site name
- incorporate specific open and closed URLs into routes
- make a PR on Pathoplexus to set those open and closed routes

Given we already hardcode "Pathoplexus" (https://github.com/loculus-project/loculus/blob/a47d51a5ee534884213ee9f50fc8cec7a92a458a/website/src/components/Submission/DataUploadForm.tsx#L350) and so this isn't a regression IMO for maintainers, that's not something I was planning to take on atm

theosanderson: (my reply crossed with yours)

theosanderson: I guess just generally this whole thing feels very Pathoplexus-specific. E.g. will an instance maintainer choose to actually submit to INSDC? The message says they will

chaoran-chen: ok, I see. It's a good summary of what needs to be done to fix this issue: would you maybe like to copy it to an issue?

theosanderson: 👍  I made https://github.com/loculus-project/loculus/issues/3674

chaoran-chen: That's great, thank you!

theosanderson: thanks!

theosanderson:
@anna-parker  - let me know if you'd be ok with us merging this into your PR?

anna-parker (approved):

Issue #3177, opened 2024-11-05
Automate revising sequence metadata in samples submitted to ENA


anna-parker:
1. Ensure metadata field to revise is not authors -> must be revised via email.
2. Ensure metadata field is in submitted sample -> otherwise no update needed, just use previous accessions
3. Send update to ENA

First automate this in dry-run and check it works

PR #3673, opened 2025-02-11
feat(deposition): Automate metadata revision of ENA submissions

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/3177

preview URL:

### Summary

1. Refactors code, adding the mapping of loculus fields to the manifest fields into a dictionary in the config for clarity, additionally correcting the mapping of the consensusSequenceSoftware to PROGRAM and sequencingInstrument to PLATFORM.
2. This enables automatic revision of sequences already submitted to the ENA, it checks for certain edge cases: 
- an older version is added later -> will not revise data
- a version is uploaded where fields have been revised that end up in the assembly manifest -> biosample will be manually updated and db will be put into state has errors as a manual revision is required for the assembly update.
- a revision tries to update a biosample or bioproject accession -> this is not allowed and also puts the db into an error state

### PR Checklist

- [x] Ensure bioproject and biosample cannot be revised
- [x] Test on ENA dev
   - test all current cases still work (with known bioproject, biosample and sra as well)
   - test that if old version is uploaded we error
   - check that sample can be revised
   - check that assembly can be revised
   - check error if changes are in manifest and cannot be automatically revised
- [x] Create change list and check if changed field is in manifest (must be updated via email).


ena-submission/ENA_submission.md
```diff
@@ -310,6 +310,15 @@ When processing is finished the response should look like:
 ]
 ```
 
+# Revising Submissions to ENA
+
+## 1. [Revising Studies (Projects) and Samples](https://ena-docs.readthedocs.io/en/latest/update/metadata/programmatic-study.html)
+Revisions to a study or sample should be submitted the same way as original sequences were submitted, with the `ADD` action in the submission request should be changed to a `MODIFY`. However, the alias must BE THE SAME as the previous version or the assigned accession number must be added for the correct sample/study to be updated. 
+
+## 2. [Revising Assemblies](https://ena-docs.readthedocs.io/en/latest/update/assembly.html)
+It appears that all fields that were explicitly set via the manifest must be updated via an email. This includes changes to any field that is in the `manifest_fields_mapping` field of the default.yaml. Additionally, study and sample reference must stay the same and chromosome names cannot be changed (but new ones can be added).
```
corneliusroemer, line 319: Do you create the email here in his PR automatically, or at least draft the text or is this completely manual right now? Please state what the current state is, what needs to be done manually.

ena-submission/src/ena_deposition/create_assembly.py
```diff
@@ -123,95 +101,137 @@ def create_manifest_object(
...
         raise ValueError(msg) from err
+    return authors
+
+
+def get_flatfile(config: Config, metadata, organism_metadata, unaligned_nucleotide_sequences, dir):
     collection_date = metadata.get("sampleCollectionDate", "Unknown")
     country = metadata.get("geoLocCountry", "Unknown")
     admin1 = metadata.get("geoLocAdmin1", "")
     admin2 = metadata.get("geoLocAdmin2", "")
     country = f"{country}:{admin1}, {admin2}"
```
corneliusroemer, line 140: We should add all divisions we have here and skip those we don't gave

corneliusroemer (approved):
This is great, I have a few questions, see comments.

Would be great if you could also describe briefly what needs to be done by operators manually to submit revisions in the following cases:
- just the metadata changes ok Loculus, e.g collection date
- just sequences change
- both metadata and sequences change

Feel free to merge when you think it's done, as we can improve on this at any time and current is better than what we have right now in any case.

PR #3680, opened 2025-02-12
feat(deployment): organism scoped resource specification

theosanderson:
Preview: https://organism-scoped-resources.loculus.org/

Adds `organismSpecific` resource requirements, keyed by organism. These are optional - without them everything works as before

Example:

```
resources:
  organismSpecific:
    dummy-organism:
      silo:
        requests:
          memory: "10Mi"
        limits:
          memory: "10Mi"
```

Tested working: 
image

### DIFF
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -1692,10 +1692,11 @@ gitHubMainUrl: https://github.com/loculus-project/loculus
 resources:
   organismSpecific:
     dummy-organism:
-      requests:
-        memory: "10Mi"
-      limits:
-        memory: "10Mi"
+      silo:
+        requests:
+          memory: "10Mi"
+        limits:
+          memory: "10Mi"
   website:
     requests:
       memory: "200Mi"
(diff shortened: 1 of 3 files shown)


kubernetes/loculus/values.yaml
```diff
@@ -1690,6 +1690,12 @@ replicas:
 gitHubEditLink: "https://github.com/loculus-project/loculus/edit/main/monorepo/website/"
 gitHubMainUrl: https://github.com/loculus-project/loculus
 resources:
+  organismSpecific:
```
theosanderson, line 1693: testing

theosanderson: Revert all changes to values.yaml before merging

corneliusroemer:
Would it be better to put the organism specific resources under the existing organism key to reduce "foreign keys"?

That way if you add an organism you only need to add things in one place, rather than in 2 disjoint ones matching by organism name.

corneliusroemer:
Sounds ok to me, we can change the values things later, agreed

corneliusroemer (approved):

PR #3681, opened 2025-02-12
feat(website): Add button to flag a sequence

### LINKED ISSUE #3514: Add/Prototype button to report that a sequence might have issues
Need an easy way for authenticated users to report a sequence might have issue -> should be reviewed by curator (how? github issue, email), then discussed in github issue.

Add template to button: which field and why do you think there is an issue

fhennig:
resolves #3514 

preview URL: https://seq-flagging-mvp.loculus.org/seq/LOC_00003QG.1

updated reference docs: https://docs-seq-flagging-mvp.loculus.org/reference/helm-chart-config/#general-settings

### Summary
- Adds a new button on the sequence details page bottom
- The button opens a URL in a new tab.
- The button isn't there if a sequence was revoked.

Config settings:

```yaml
sequenceFlagging:
  github:
    organization: pathoplexus
    repository: curation_reports
    issueTemplate: sequence-metadata-issue.md
```

My reasoning was that this way users can't accidentally mess up the URL, and we could later add other reporting options that are not GitHub.

### Screenshot

image

### PR Checklist

- ~~The implemented feature is covered by an appropriate test.~~ (It's just a link)


fhennig:
TODO:
- Add tests 
- Docs: for admins: how to configure sequence flagging; for users: How to report a sequence

kubernetes/loculus/values.yaml
```diff
@@ -25,6 +25,11 @@ dataUseTerms:
   urls:
     open: https://#TODO-MVP/open
     restricted: https://#TODO-MVP/restricted
+sequenceFlagging:
```
chaoran-chen, line 28: Do we want to have an `enabled: boolean` again?

kubernetes/loculus/values.yaml
```diff
@@ -25,6 +25,11 @@ dataUseTerms:
...
+  github:
+    organization: pathoplexus
+    repository: curation_reports
+    issueTemplate: sequence-metadata-issue.md
```
chaoran-chen, line 32: By default, this should be disabled (we can turn it on for the previews).

chaoran-chen (commented):
Nice :)

kubernetes/loculus/values.yaml
fhennig: I don't think that in this case that would be helpful. If the setting is there, it's enabled. if it isn't there it is disabled, I think that makes sense

kubernetes/loculus/values.yaml
fhennig: makes sense.

fhennig:
@anna-parker You said elsewhere that you think we should just add the button to the full-screen view, but I think this might be confusing and also difficult to explain in the documentation - I think maybe it's better to also add the report button to the pop-up sequence details view?

website/src/components/SequenceDetailsPage/SequencesDataTableTitle.astro
theosanderson, line 1: Thanks for this - sorry to only be looking now. I think this will be a feature used relatively rarely (though super important). IMO it would be a lot better to have a design very similar to the sequence management features at the bottom. (These only appear to sequence owners, but this button would appear to all).

image

I think a new section, below or above "Report an issue" with a button "Create Github issue" would be great. I appreciate you had already refined the issue and I'm raising this late. I can do it in a follow up if preferred. Also other people can disagree.

theosanderson: (I hadn't seen the proposal in the Issue just because so much happening)

chaoran-chen: (Either solution sounds good to me. Should we maybe have a quick poll over Slack to see what others think?)

fhennig: Makes sense to me! I also felt like up there is was a bit too prominent.

I would make the changes as I think it makes sense, but I'll wait a bit for others to chime in.

(Tagging for attention: @chaoran-chen @anna-parker )

anna-parker: This looks great to me!! I like the location - I wouldn't mind having this at the bottom either.

In either case I think we need to remember to only show this button when a user doesn't own the sequence (in which case they should revise and not curate)

theosanderson: I don't necessarily think it's a problem to have it on owned sequences. Use cases:
- "we've noticed that this sequence seems to have a spurious frameshift and will be resequencing it to confirm"
- "Pathoplexus wouldn't accept Country X as valid data so we put in Country Y instead, what should we do?"

anna-parker: Ah ok I guess this is again kind of a mix between curation and flagging. I do agree you should be able to flag your own sequences - but I dont think you should be curating your own sequence. Of the two examples above: 

- "Pathoplexus wouldn't accept Country X as valid data so we put in Country Y instead, what should we do?"
I think should be in the general pathoplexus github (and might get lost in the curation github as an average curator cannot fix it)
- "we've noticed that this sequence seems to have a spurious frameshift and will be resequencing it to confirm"
This does sound valid for flagging but not curating.. so I guess we could just link to the curation github for now but I also wonder if we should in future just add a flag if a user flags their own sequence without it having to go via the curation repo. 

So in summary I guess Im ok with keeping the button for now - but maybe we should rethink the data flow for these cases in future.

theosanderson: Well _curating_ is I think in our lingo what the _curators_ do, not typical users. The reports repo provides information for them to act on in their curation. This would be providing info by saying e.g. you don't need to contact us to propose a change - we're on it, or whatever

fhennig: Updated it!

website/src/components/SequenceDetailsPage/SequenceDataUI.tsx
```diff
@@ -81,41 +86,53 @@ export const SequenceDataUI: FC<Props> = ({
...
-                    <div className='text-sm text-gray-400 mt-4 block'>&nbsp;</div>
-                </div>
+                    </div>
+                </>
+            )}
+            {reportUrl !== undefined && (
+                <>
+                    <hr className='my-4' />
+                    <div className='my-8'>
+                        <h2 className='text-xl font-bold mb-3'>Report an issue</h2>
```
theosanderson, line 130:
```suggestion
                        <h2 className='text-xl font-bold mb-3'>Report an issue with this sequence or metadata</h2>
```

theosanderson (approved):
Thanks! LGTM!

kubernetes/loculus/values.yaml
fhennig: disabled it here: https://github.com/loculus-project/loculus/pull/3681/commits/422d568970e65243593483d4e9011ea6d0f30b8a

corneliusroemer:
I've added an `!` because to use this requires downstream config changes

kubernetes/loculus/values.yaml
corneliusroemer: In the helm charts I've seen, there usually is a simple boolean flag even if there's more complex config - probably to allow disabling at the flip of a boolean, as opposed to the addition/deletion of a lot of yaml.

kubernetes/loculus/values.yaml
corneliusroemer: We should enable for previews

fhennig:
Yes like chaoran said, it's not enabled on main. I haven't enabled something for just previews yet, is it just adding it to `values_preview_server.yaml`? Or how does it work?

Adding the snippet from the PR description should be fine.

fhennig:
https://github.com/loculus-project/loculus/pull/3721

PR #3685, opened 2025-02-13
feat(deployment): skip schema validation for dev

### LINKED ISSUE #3684: Skip schema validation in helm where possible in various dev environments
It's helpful to separate concerns in CI. So for example, we have a CI workflow that checks whether the source code is formatted nicely, but if it's not we can still deploy previews and we can still see if the E2E tests pass. We should do the same for the validation of the helm values. This is achievable and I will make a PR now.

theosanderson:
resolves #3684, apart from for previews which are in a different repo

This will also fix a current broken test on main: https://github.com/loculus-project/loculus/actions/runs/13305856122/job/37156558700 though potentially that should be either added to the schema or refactored to work in a different way

### DIFF
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -1762,4 +1762,3 @@ defaultResources:
     cpu: "20m"
 ingest:
   ncbiGatewayUrl: null
-nonExistentTest: test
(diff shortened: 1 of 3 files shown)


kubernetes/loculus/values.yaml
```diff
@@ -1762,3 +1762,4 @@ defaultResources:
     cpu: "20m"
 ingest:
   ncbiGatewayUrl: null
+nonExistentTest: test
```
theosanderson, line 1765: revert

fhennig (approved):
LGTM!

PR #3686, opened 2025-02-13
fix(website): Fix alignment of seq details page

theosanderson:
https://alignment.loculus.org/

Fixes alignment on sequence details page in full screen mode:

before:
image

after:
image


fhennig (approved):
wow so much better! thanks

PR #3687, opened 2025-02-13
fix(website): improve alignment of download button on search page

theosanderson:
(imo no preview needed)

makes download button line up with table:

image

vs before:

image

(ignore the button styling difference - one is moused over)


fhennig (approved):
thx!

PR #3692, opened 2025-02-14
fix(k8s): Also accept integers as shas/branch names in values schema

### LINKED ISSUE #3691: Previews fail in 3.7% of cases when first 7 chars of sha are all numeric
Apparently our previews don't work when the first 7 chars of a sha are all numeric:

> Failed to load target state: failed to generate manifest for source 1 of 1: rpc error: code = Unknown desc = `helm template . --name-template pp-try-leak-fix-3690 --namespace prev-try-leak-fix --kube-version 1.30 --set shortbranch=try-leak-fix --set sha=5486252 --set branch=try-leak-fix --set host=try-leak-fix.loculus.org --values <path to cached source>/kubernetes/loculus/values.yaml --values <path to cached source>/kubernetes/loculus/values_preview_server.yaml <api versions removed> --include-crds` failed exit status 1: Error: values don't meet the specifications of the schema(s) in the following chart(s): loculus: - sha: Invalid type. Expected: string, given: integer

Image

The probability of hitting such a sha is (10/16)^7 ~ 3.7%

corneliusroemer:
Resolves #3691

Schema was too strict - as strings that happen to also be integers end up interpreted as integers (at least by Helm).

Same issue also manifests in Pathoplexus, so just excluding Loculus previews from schema check isn't enough.

### DIFF
--- a/kubernetes/loculus/values.schema.json
+++ b/kubernetes/loculus/values.schema.json
@@ -1011,15 +1011,15 @@
       "additionalProperties": false
     },
     "branch": {
-      "type": "string",
+      "type": ["string", "integer"],
       "description": "TODO"
     },
     "sha": {
-      "type": "string",
+      "type": ["string", "integer"],
       "description": "TODO"
     },
     "shortbranch": {
-      "type": "string",
+      "type": ["string", "integer"],
       "description": "TODO - this is required/set by ArgoCD"
     },
     "podPriorityClassName": {


theosanderson (approved):
I guess there are two possible fixes - one would be to wrap the commit hash in quotes where it's used and one would be this. I guess this is easier because it's less fussy

Issue #3689, opened 2025-02-14
"Alignment states and QC metrics" is strange phrasing


theosanderson:
@rneher pointed out that this is maybe an odd heading.

Maybe we could just call it "QC metrics" ?

anna-parker:
Maybe "Alignment and QC metrics"?

PR #3696, opened 2025-02-16
Update values.yaml

rneher:
Fix non-sense header promising "alignment states"

making a PR is faster than making an issue.


rneher:
just found the issue https://github.com/loculus-project/loculus/issues/3689
:joy:

theosanderson:
Thanks for pushing this - I think we need to make this change in a few more places in Loculus and also to make changes in Pathoplexus - will draft something

theosanderson:
I made https://github.com/pathoplexus/pathoplexus/pull/388 - I used slightly different phrasing based on Anya's suggestion - feel free to object!

Issue #3682, opened 2025-02-12
Don't submit to ENA with a `segment: main` title


theosanderson:
Segments in the name are useful for multi-segmented viruses but confusing for single segmented viruses

Per Danny:

> is the "segment: main" in the definition line something that ENA does or something Pathoplexus does? (a direct NCBI Genbank submission would probably say something more like "partial genome" or "complete CDS" -- this is added by them, not the submitter)

corneliusroemer:
We need to call it something - what would you prefer instead of main? `genome`?

corneliusroemer:
@anna-parker tried leaving the field empty and found it caused validation errors: https://github.com/loculus-project/loculus/issues/2983#issuecomment-2405104359

theosanderson:
Genome sounds good to me!

anna-parker:
Ah I wonder if the actual problem is that I was using chromosome type `segmented` for all viruses - when it should probably be `monopartite` for all our viruses except CCHF.

PR #3698, opened 2025-02-17
feat(deposition): improve chromosome list

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/3682

preview URL:

### Summary
Read about chromosome list template here: https://ena-docs.readthedocs.io/en/latest/submit/fileprep/assembly.html#chromosome-list-file

1. Rename `main` to `genome` for single-segmented organisms
2. Rename chromosome type of single-segmented organisms to MONOPARTITE instead of SEGMENTED
3. Allow topology to be defined via values.yaml

### PR Checklist

### DIFF
--- a/ena-submission/src/ena_deposition/ena_submission_helper.py
+++ b/ena-submission/src/ena_deposition/ena_submission_helper.py
@@ -322,9 +322,7 @@ def create_flatfile(
 
     embl_content = []
 
-    multi_segment = True
-    if set(unaligned_sequences.keys()) == {"main"}:
-        multi_segment = False
+    multi_segment = set(unaligned_sequences.keys()) != {"main"}
 
     for seq_name, sequence_str in unaligned_sequences.items():
         if not sequence_str:
@@ -387,9 +385,7 @@ def create_fasta(
         with tempfile.NamedTemporaryFile(delete=False, suffix=".fasta.gz") as temp:
             filename = temp.name
 
-    multi_segment = True
-    if set(unaligned_sequences.keys()) == {"main"}:
-        multi_segment = False
+    multi_segment = set(unaligned_sequences.keys()) != {"main"}
 
     with gzip.GzipFile(filename, "wb") as gz:
         if not multi_segment:
--- a/ena-submission/src/ena_deposition/ena_submission_helper.py
+++ b/ena-submission/src/ena_deposition/ena_submission_helper.py
@@ -387,8 +387,12 @@ def create_fasta(
         with tempfile.NamedTemporaryFile(delete=False, suffix=".fasta.gz") as temp:
             filename = temp.name
 
+    multi_segment = True
+    if set(unaligned_sequences.keys()) == {"main"}:
+        multi_segment = False
+
     with gzip.GzipFile(filename, "wb") as gz:
-        if len(unaligned_sequences.keys()) == 1:
+        if not multi_segment:
             entry = chromosome_list.chromosomes[0]
             gz.write(f">{entry.object_name}\n".encode())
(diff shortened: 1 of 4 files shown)


kubernetes/loculus/values.yaml
```diff
@@ -1583,6 +1583,7 @@ defaultOrganisms:
         taxon_id: 3052518
         scientific_name: "Orthonairovirus haemorrhagiae"
         molecule_type: "genomic RNA"
+        topology: "circular"
```
theosanderson, line 1586: (I know this is just a draft ATM: is this just testing or would we actually think CCHF has circular topology?)

anna-parker: It does I believe? https://en.wikipedia.org/wiki/Crimean%E2%80%93Congo_hemorrhagic_fever#Molecular_biology "The genome is circular, negative-sense [RNA](https://en.wikipedia.org/wiki/RNA) in three parts"

theosanderson: Thanks! Agreed

theosanderson: I'm actually not sure

> Alignment of the genomic end sequences with the DUG virus L segment and the M and S segments of both DUG and CCHF viruses demonstrated that the first 9–11 nucleotides are highly conserved. Similar to other bunyaviruses, these conserved ends seem to mediate the formation of terminal base-paired structures (Flick et al., 2002). These formations are involved in viral polymerase binding and initiation of transcription and replication, and give each segment a circular appearance when visualized through electron microscopy (Schmaljohn and Hooper, 2001).

theosanderson: LLMs insist it's linear too, let's do some reading :)

anna-parker: If you look up bunyaviruses they are all circular - this also explains why the terminals are conserved

anna-parker: Or ok "pseudo-circular": https://en.wikipedia.org/wiki/Bunyavirales

anna-parker: this paper also mentions that they are "loosely circular" - and uses this as the reason why the terminals are conserved: https://pmc.ncbi.nlm.nih.gov/articles/PMC2730268/

anna-parker: Found by @corneliusroemer: https://ictv.global/report_9th/RNAneg/Bunyaviridae " The three helical nucleocapsids are circular and comprise one each of the unique ssRNA segments (L, large; M, medium; S, small) encapsidated by N protein and associated with the L protein"

anna-parker: But ok - the reference genome is marked as linear - so I am fine with keeping as linear.

ena-submission/src/ena_deposition/ena_submission_helper.py
```diff
@@ -387,8 +387,12 @@ def create_fasta(
         with tempfile.NamedTemporaryFile(delete=False, suffix=".fasta.gz") as temp:
             filename = temp.name
 
+    multi_segment = True
+    if set(unaligned_sequences.keys()) == {"main"}:
+        multi_segment = False
+
```
theosanderson, line 393:
```suggestion
    multi_segment = set(unaligned_sequences.keys()) != {"main"}
```
(not important, feel free to reject)

theosanderson (approved):
Looks sensible to me, but may be useful for Cornelius to look ideally!

anna-parker:
If its ok I will merge now as I am away this week - @corneliusroemer please feel free to rollback if there are any issues - but I tested this on dev and it looks good!

PR #3704, opened 2025-02-17
feat(deployment): improve observability of silo preprocessing errors by force replacing pod in development and exiting the process if failed

corneliusroemer:
Previously, when silo preprocessing started failing after an update to a preview but there was still successful processed data in the shared volume we would not easily observe the processing failure as "-1/0 sequences" in the UI. One would only see the failure in processing logs which would often not be checked (e.g. in testing https://github.com/loculus-project/loculus/pull/3534).

This PR makes silo preprocessing errors more easily observable _in dev environments only_ by _force replacing_ the LAPIS/SILO pod including the shared volume even if there are no code changes between updates (e.g. when only the backend code changes).

Also, we make the silo preprocessing crash rather than happily error without crashing when there are preprocessing errors.

Force replacing is switched on iff `Values.developmentDatabasePersistence == False` AND `Values.runDevelopmentMainDatabase == True`: this means no force replace in production environments (there we don't run a development database) and also not in development if one sets development database persistence to True.

### DIFF
--- a/kubernetes/loculus/silo_import_wrapper.sh
+++ b/kubernetes/loculus/silo_import_wrapper.sh
@@ -14,6 +14,8 @@ get_value_from_file() {
     fi
 }
 
+exit 1
+
 while true; do
     echo "Checking for new data in SILO"
     last_etag=$(get_value_from_file "$current_etag_path")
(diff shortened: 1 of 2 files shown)


theosanderson (approved):
Thanks - seems good to me!

I'm confident the `Force=true` change is good.

I'm a bit uncertain if there could be something unexpected that makes the exiting from SILO problematic, but I can't think what it is if so!

PR #3710, opened 2025-02-18
feat(website): Add new form to submit individual sequences

### LINKED ISSUE #3226: Submit individual sequences via a form
We should allow users to submit sequences via a form which would just be the same as the edit page, but not prepopulated

### LINKED ISSUE #3722: 'Revise' page incorrectly states open data use terms.
The review page says "open data use terms" but the data use terms are already set at the initial submission.

Image

The revision form should make no statements about the data use terms.

fhennig:
resolves #3226 
resolves #3722

preview URL: https://submission-via-form.loculus.org/dummy-organism/submission/2/submit?inputMode=form

updated docs page: https://docs-submission-via-form.loculus.org/for-users/submit-sequences/

### Summary

- Adds a new `inputMode` URL param to the `submit` endpoint with values `form` and `bulk`. If it is omitted, we default back to the 'bulk' variant.
- There is a tab-style selector for bulk vs form submission (see screenshot below)
- Fixed a mini hydration bug where a `td` was nested in a `div`
- Drop mention of specific data use terms from Acknowledgement section (#3722)
- Small improvement also for the edit form: internally generated TSV file now uses a library with escaping etc instead of pure string interpolation.

#### Tests
- Add new tests for SubmissionRouteUtils
- Added new tests for the input form (`FormOrUploadWrapper.spec.tsx`). Test that errors are raised when files or form input is missing. Test that all of the form that gets included in the TSV and also that escaping works.
- Expanded SubmissionForm tests - submit not only via files, but also via form for some cases.
- New integration test that submits a sequence via the form

#### Implementation notes
- The Edit form as well as the file upload section of the submission form have been extracted into dedicated components
- A new `FormOrUploadWrapper` wraps both and makes it easy to switch between them, returning just files which can then be used for uploading
- Submission form allows manually entering a Submission ID - this is implemented by allowing all `noEdit` fields to be in fact edited on initial submission - AFAICT this was only used for submission ID and accession (only the submission ID is shown in the 'submit' form).

### Screenshot
image

### PR Checklist


website/src/components/Submission/FileUpload/FileUploadComponent.tsx
theosanderson, line 1: (I know it's still a draft). This PR is the sort of core new functionality which IMO should be covered with an integration test. This could be super simple - a single workflow submitting a sequence with some form fields filled in and then checking that sequence indeed appears.

fhennig: I have now added such a test! For checking errors if the user doesn't enter data etc, I added unit tests for that to the website code.

rneher:
just quickly tested this. I think that it is a bit surprising to land on the submission form for a single sequence by default. I would keep the default as the file upload and suggest that prominent button something with something like "Single sequence submission via form" takes you to the form.

rneher:
also. looks like the fields don't accept long enough entries
image

Probably better as a wrapping text field.

website/src/components/Submission/FileUpload/SequenceEntryUploadComponent.tsx
```diff
@@ -0,0 +1,233 @@
...
+                        <DevExampleData
+                            setExampleEntries={setExampleEntries}
+                            exampleEntries={exampleEntries}
+                            handleLoadExampleData={handleLoadExampleData}
+                            dataIsLoaded={!!metadataFile && (!enableConsensusSequences || !!sequenceFile)}
+                        />
+                    )}
+            </div>
+            <form className='sm:col-span-2'>
+                <div className='flex flex-col lg:flex-row gap-6 justify-end'>
```
theosanderson, line 131:
```suggestion
                <div className='flex flex-col lg:flex-row gap-6'>
```

IMO this should be lined up with the UI below as on main

theosanderson:
> also. looks like the fields don't accept long enough entries

I couldn't reproduce this - my theory is it must have been about some field rather than the sequences

> Probably better as a wrapping text field.

These text fields for sequences are definitely really problematic. I mostly of think they should in the more final version be file upload components (for the one segment). Copying and pasting sequences feels generally bad practice?

theosanderson:
I made some UX suggestions at https://subform2.loculus.org/cchf/submission/1/submit (username: superuser, pw: superuser) for some future version of this (some after this PR), which I've also mentioned on Slack.

rneher:
> > also. looks like the fields don't accept long enough entries
> 
> I couldn't reproduce this - my theory is it must have been about some field rather than the sequences

Definitely reproducible for me in CCHF

rneher:
Might have been because my sequences included a fasta header. Seems to work without fasta header.

fhennig:
Thanks for your input! I quite like your suggestions Theo, I think I'd maybe just merge them in?

For the sequence fields, maybe we can look into doing already client-side validation, so the users can't submit incorrect characters.

theosanderson:
> Might have been because my sequences included a fasta header. Seems to work without fasta header.

OK yeah so I think what's happening here is that the whole sequence is being interpreted as the FASTA header

My own preferred way forward on this would be to create a new high priority issue to change all these sequence input text-boxes to `upload-a-fasta-file-with-a-single-sequence` components.

fhennig:
Since this PR is already quite large, I have made an issue for follow up changes to the form itself: https://github.com/loculus-project/loculus/issues/3740

I want to keep this PR to just the initial creation of the single-sequence submission feature, and the refactoring of the internal components.

IMO it'll be easier to review that way.

website/src/components/Edit/EditPage.tsx
```diff
@@ -101,31 +66,33 @@ const InnerEditPage: FC<EditPageProps> = ({
     );
 
     const submitEditedDataForAccessionVersion = () => {
+        const metadataFile = editableMetadata.getMetadataTsv(dataToEdit.submissionId, dataToEdit.accession);
+        if (metadataFile === undefined) {
+            toast.error('Please enter metadata.', { position: 'top-center', autoClose: false });
```
theosanderson, line 71: Not necessarily blocking, but this seems a bit weird. Theoretically one could have have no metadata fields I guess - ideally it feels like the logic for this should be in the backend (which yes, I guess would already require a dummy metadata file with `submissionIds`, but potentially only that?)

theosanderson: I could imagine that this might only appear if they enter no submission id?

website/src/components/Edit/InputForm.tsx
```diff
@@ -0,0 +1,271 @@
...
+     * Return the Metadata information as a TSV. If no information is present, 'undefined' is returned.
+     * @param submissionId optional (might already be in the rows if add to the form initially).
+     *      The submission ID to put into the TSV.
+     * @param accession optional. If an accession is already assigned to this sequence, it should be given.
+     */
+    getMetadataTsv(submissionId?: string, accession?: string): File | undefined {
+        // if no values are set at all, return undefined
+        if (!this.rows.some((row) => row.value !== '')) return undefined;
+
+        // If only the submission ID is set, that't not enough, also return undefined.
```
theosanderson, line 80: ah this relates to my comment above - not sure why we need this?

website/src/components/Edit/InputForm.tsx
```diff
@@ -0,0 +1,271 @@
...
+        return this.rows.reduce((prev, row) => ({ ...prev, [row.key]: row.value }), {});
+    }
+}
+
+type MetadataFormProps = {
+    editableMetadata: EditableMetadata;
+    setEditableMetadata: Dispatch<SetStateAction<EditableMetadata>>;
+    groupedInputFields: Map<string, InputField[]>;
+    /**
+     * If it is a submit form, input fields for noEdit fields will be generated.
```
theosanderson, line 115: I've kind of lost track of the concept of `noEdit` but searching the codebase seems to find it applies to `accession` ? but this wouldn't

website/src/components/Edit/InputForm.tsx
fhennig: I submitted only the submission ID before, and then an error came back from the backend - something like "metadata file does not contain metadata for submissionID X" and I wanted to avoid this error popping up because it mentions a file that the user doesn't know about. So it seemed like the easiest way to prevent this was to just do a check beforehand.

website/src/components/Edit/InputForm.tsx
fhennig: I already thought that this might be confusing, so I added a note about it to the PR description (not suggesting that you should've noticed - just that it's there as well for others that might look):

> Submission form allows manually entering a Submission ID - this is implemented by allowing all noEdit fields to be in fact edited on initial submission - AFAICT this was only used for submission ID and accession (only the submission ID is shown in the 'submit' form).

So yes it applies to the accession field, but the accession field isn't there during submission - and when accession is present during edit, 'isSubmitForm' will be false -> accession not editable.

Does it now make sense to you? Anything that could be better explained in the code? Or is there maybe a better, more obvious way to achieve the editable submission ID?

website/src/components/Edit/InputForm.tsx
theosanderson: Makes sense! We can make a new issue to allow sequences without metadata

fhennig: I'm not entirely sure what to put into such an issue - I'll leave it up to you to maybe create one!

website/src/components/Edit/EditPage.tsx
fhennig: this is also resolved I think?

docs/src/content/docs/for-users/submit-sequences.md
```diff
@@ -4,6 +4,20 @@ title: Submit sequences
 
 You can only upload sequences if you already have an account and [are part of a submitting group](../create-manage-groups/).
 
+You can submit sequences either through uploading files or entering sequence metadata on the website directly.
```
theosanderson, line 7:
```suggestion
You can submit sequences either by uploading files or entering sequence metadata on the website directly.
```

docs/src/content/docs/for-users/submit-sequences.md
```diff
@@ -4,6 +4,20 @@ title: Submit sequences
...
+Submission for the website form is only recommended for uploading a single sequence, for multiple sequences you should use bulk submission with file upload.
```
theosanderson, line 8:
```suggestion
The form submits a single individual sequence entry, to submit multiple sequence entries with one upload you can use the bulk submission tool, uploading a metadata table and a multi-sequence FASTA.
```

website/src/components/Edit/InputForm.tsx
theosanderson: Thanks - OK, doesn't seem a big deal!

website/src/components/Submission/DataUploadForm.tsx
```diff
@@ -448,6 +186,170 @@ const InnerDataUploadForm = ({
...
+    return (
+        <div className='grid sm:grid-cols-3 gap-x-16 gap-y-4'>
+            <div className=''>
+                <h2 className='font-medium text-lg'>Acknowledgement</h2>
+                <p className='text-gray-500 text-sm'>Acknowledge submission terms</p>
+            </div>
+            <div className='gap-x-6 gap-y-8 col-span-2'>
+                <div>
+                    <p className='block text-sm'>
+                        Your data will be available on Pathoplexus, under the selected data use terms. Data with open
```
theosanderson, line 301: Hmm.. so this feels like it has been made out of technical convenience rather than user experience

image

I had thought we would only remove the more specific reference to the data use terms for the revision form. IMO this is a bit of a regression and it would be better to keep the behaviour the same except for the specific case of the revision form where we don't know the terms.

fhennig: Agreed. so you're saying, keep the old behaviour for initial submit, and only have the less specific formulation on the review page? That makes sense to me.

website/src/components/Submission/SubmissionPageWrapper.astro
```diff
@@ -6,10 +6,11 @@ import NeedToLogin from '../common/NeedToLogin.astro';
 
 interface Props {
     title: string;
+    disablePageTitle?: boolean;
```
theosanderson, line 9: nbd but I guess this may not be needed any more? But it's fine if you want to keep it anyway

website/src/components/Submission/FormOrUploadWrapper.tsx
```diff
@@ -0,0 +1,149 @@
...
+    metadataTemplateFields: Map<string, InputField[]>;
+    enableConsensusSequences: boolean;
+};
+
+/**
+ * A component that allows users to upload sequence data. Two modes are supported, see {@link InputMode}.
+ * In 'form' mode, a form is displayed, and the user can directly enter stuff into the form to upload their
+ * metadata. In 'bulk' mode, the user needs to upload files containing the sequences and metadata.
+ * Either way, the component turns the uploaded data into files, so they can be submitted to the API.
+ * Set the 'fileCreatorSetter' to get the files - have a look at existing usage on how this works.
```
theosanderson, line 41: Should this be `setFileFactory`

website/src/components/Submission/FormOrUploadWrapper.tsx
```diff
@@ -0,0 +1,149 @@
...
+        EditableSequences.fromSequenceNames(referenceGenomeSequenceNames.nucleotideSequences),
+    );
+
+    const [metadataFile, setMetadataFile] = useState<ProcessedFile | undefined>(undefined);
+    const [sequenceFile, setSequenceFile] = useState<ProcessedFile | undefined>(undefined);
+    // The columnMapping can be null; if null -> don't apply mapping.
+    const [columnMapping, setColumnMapping] = useState<ColumnMapping | null>(null);
+
+    useEffect(() => {
+        setFileFactory(() => {
```
theosanderson, line 64: I'm not sure I really understand this (which isn't a huge problem). Could you explain a bit about why this works like this (in this thread, not in code necessarily)?

website/src/components/Submission/DataUploadForm.tsx
theosanderson, line 1: It would probably be nice to split some of these components into separate files

theosanderson (commented):
Thanks for putting this together! Looks great

As we've discussed, the sequence input is very suboptimal (as it is on the edit page) but we want to tackle this as the next PR in this theme. 

I'd like to understand why we have this slightly concept logic with a "factory" that's the main remaining thing that I'd like to get my head around

website/src/components/Submission/SubmissionPageWrapper.astro
fhennig: I know it's a bit of a hack - IMO the header shouldn't be in the wrapper, every page should have it's own wrapper.

It's still needed at least if we want the layout to stay the way it is. That is - have the header line up with the form. The header in the wrapper is left-justified, and I think that doesn't look very nice, so I'd keep it for now.

fhennig:
Yes I hope I didn't make a mistake in my reasoning there. Why pass this factory/generator to the parent, instead of just the files directly? 

Usually in react, we'd just pass changed data to the parent directly. In this case, this would however mean, passing directly the files to be submitted out of the component. This would mean (at least in my understanding) that whenever the user changes a field, the files would be regenerated. However, we only really need these files at the very end when we want to submit. I was worried that this would be too much overhead (however I didn't measure anything).

So instead, I decided I'd like make the file generation lazy, so I just pass a function to the parent and the parent can generate the files on demand (i.e. when the button is pressed).

website/src/components/Submission/DataUploadForm.tsx
fhennig: I agree, in fact, I think a lot of files contain too much stuff. So I have left it all in there, because it seemed to me that that's just how we do things in this codebase. In a way, it's also nice to not have so many files.

website/src/components/Submission/FormOrUploadWrapper.tsx
fhennig: explained as a top level comment in the PR

website/src/components/Submission/FormOrUploadWrapper.tsx
fhennig: yes, fixed: https://github.com/loculus-project/loculus/pull/3710/commits/1788f4aa7b1863a8413d0c1b5c8bdd7b6f6ba475

website/src/components/Submission/SubmissionPageWrapper.astro
theosanderson: Thanks for explaining!

docs/src/content/docs/for-users/submit-sequences.md
```diff
@@ -4,6 +4,20 @@ title: Submit sequences
...
+
+## Submit a single sequence through the form
+
+1. Log into your account, and then click 'Submit' in the top-right corner of the website.
+2. Select the organism that you'd like to submit sequences for.
+3. Click on 'Submit single sequence'.
```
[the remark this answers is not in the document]
theosanderson: Sorry for being unclear. Isn't there a missing step of hitting "Submit" after selecting the organism?

website/src/components/Submission/DataUploadForm.tsx
theosanderson: Thanks – I hadn't appreciated nothing had been added

docs/src/content/docs/for-users/submit-sequences.md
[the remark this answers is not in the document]
fhennig: Ah yes true! Updated: https://github.com/loculus-project/loculus/pull/3710/commits/aca18951e019981b7ac5ddb677b9c75fbbab52db

theosanderson:
Right, yeah, I understand now. I think my view of the ideal architecture here would probably be that we to go for a bit less modularity and to allow the top level component to store both/either of files and arbitrary key value pairs from fields (which would then be made into a file with logic at that higher level). IMO the simplicity slightly outweigh the modularity and would be more flexible in terms of e.g. allowing the key value pairs to be processed in a special way depending on the sequence data or something. But it's probably a matter of taste and I don't think it makes sense to ask you to redo everything! We can always refactor later if we think it's an issue!

website/src/components/Submission/FormOrUploadWrapper.tsx
theosanderson, line 1: When swapping from `bulk` to `form` there is some brief kind of nasty CLS that I sometimes get where for a split second `Submit sequences` appears much more centrally, I guess because there's nothing to force the UI to be wide. If it was easy to fix could be nice, but not blocking. (Also you may not be able to reproduce)

theosanderson (approved):
LGTM! There's still stuff we can improve but we can do it in follow-ups

website/src/components/Submission/FormOrUploadWrapper.tsx
fhennig: Ah I see it too. But I'm not sure why it's there. I think maybe if we redesign the form it might get solved along the way, so I'll not touch the form for now.

docs/src/content/docs/for-users/submit-sequences.md
theosanderson, line 1: blocking comment as I realise I didn't test some important stuff

theosanderson, line 1: image

Just stick some extra spacing below to avoid this? I know we'll change this whole thing very soon anyway

theosanderson: (this is on edit page)

theosanderson: image

fhennig:
Thanks for sharing your thoughts on this! I can definitely see your point, passing this function up doesn't feel like the most idiomatic way and I wouldn't be opposed to merging the `FormOrUploadWrapper` into the `DataUploadForm` (again). But yes I won't do it in this PR, possibly in a follow up.

docs/src/content/docs/for-users/submit-sequences.md
fhennig: Ah this is because the section about processed data is gone. yes I will add some space!

theosanderson: TBH we should lose these sequences too, so that's the other option

fhennig: https://github.com/loculus-project/loculus/pull/3710/commits/5f5532976afb8de9c427427d059ec999ce68b939

fhennig:
I'm not gonna merge right away, I'll let it sit at least until the end of tomorrow. Thanks a lot for the review Theo!

chaoran-chen (commented):
Thanks for implementing this, @fhennig and @theosanderson! I briefly tested the feature and it looks great to me! I just noticed a small thing that I commented above. Otherwise, I don't plan to review unless you think that it would be really useful.

docs/src/content/docs/for-users/submit-sequences.md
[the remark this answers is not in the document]
fhennig: As this was already part of the form, and we want to do some other follow-up work on the form, are you fine with moving this into a new PR adressing this ticket? https://github.com/loculus-project/loculus/issues/3740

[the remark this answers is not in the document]
chaoran-chen: Sounds good!

PR #3711, opened 2025-02-18
chore(website): rename `submissionId` to `id` for input

### LINKED ISSUE #2404: Change `submissionId` to `id` in terms of what we expect at submission
If we have time pre-MVP

theosanderson:
resolves #2404

Asking users to supply a metadata file with a "submissionId" for each sequence is confusing. The ID here is actually likely to be the user's own sample ID, as used in their FASTA file. "SubmissionID" implies an association specifically with the Loculus submission - maybe an ID for submission that you get from Loculus.

Here we rename this input field to a bare `id`, and also adjust the templates accordingly. We have back-compatibility that also supports `submissionId` (we can remove that after a while: https://github.com/loculus-project/loculus/issues/4296 it's just in case someone is preparing data right as we roll this out).

This does not rename the field downstream of submission or in our internal variables. After submission this name does make more sense, it distinguishes the name that Loculus assigned (the accession) from the one the user used at submission (the `submissionId`).

This PR adjusts ingest to submit using `id`, but when ingest is calculating hashes it renames this `id` back to `submissionId` for compatibility with the old hashes.

# Testing

This is well covered by a large set of tests, and ingest by the preview instance, additionally the following manual tests have been done:
- [x]  manually submitting example mpox data

I ~propose~ have also done the following 

- [x]  putting this branch onto pathoplexus staging and observing that we don't get version bumps, either on CCHF (segmented) or on non-segmented organisms
    - I checked that ingest appeared to be running with no errors
    - I checked that the number of sequences was the same as prod for all organisms using the admin dashboard (meaning no new versions -[I also checked for errors in preprocessing])

🚀 Preview: https://submissionid2id.loculus.org

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
@@ -10,10 +10,9 @@ import mu.KotlinLogging
 import org.loculus.backend.api.DataUseTerms
 import org.loculus.backend.api.FileCategoryFilesMap
 import org.loculus.backend.api.FileIdAndNameAndReadUrl
-import org.loculus.backend.api.GeneticSequence
 import org.loculus.backend.api.MetadataMap
 import org.loculus.backend.api.Organism
-import org.loculus.backend.api.ProcessedData
+import org.loculus.backend.api.ReleasedData
 import org.loculus.backend.api.VersionStatus
 import org.loculus.backend.api.addUrls
 import org.loculus.backend.config.BackendConfig
@@ -64,7 +63,7 @@ open class ReleasedDataModel(
     private val objectMapper: ObjectMapper,
 ) {
     @Transactional(readOnly = true)
-    open fun getReleasedData(organism: Organism): Sequence<ProcessedData<GeneticSequence>> {
+    open fun getReleasedData(organism: Organism): Sequence<ReleasedData> {
         log.info { "Fetching released submissions from database for organism $organism" }
 
         val latestVersions = submissionDatabaseService.getLatestVersions(organism)
@@ -113,7 +112,7 @@ open class ReleasedDataModel(
         latestVersions: Map<Accession, Version>,
         latestRevocationVersions: Map<Accession, Version>,
         earliestReleaseDateFinder: EarliestReleaseDateFinder?,
-    ): ProcessedData<GeneticSequence> {
+    ): ReleasedData {
         val versionStatus = computeVersionStatus(rawProcessedData, latestVersions, latestRevocationVersions)
 
         val currentDataUseTerms = computeDataUseTerm(rawProcessedData)
@@ -195,14 +194,13 @@ open class ReleasedDataModel(
                 },
             )
 
-        return ProcessedData(
+        return ReleasedData(
             metadata = metadata,
             unalignedNucleotideSequences = rawProcessedData.processedData.unalignedNucleotideSequences,
             alignedNucleotideSequences = rawProcessedData.processedData.alignedNucleotideSequences,
             nucleotideInsertions = rawProcessedData.processedData.nucleotideInsertions,
             aminoAcidInsertions = rawProcessedData.processedData.aminoAcidInsertions,
             alignedAminoAcidSequences = rawProcessedData.processedData.alignedAminoAcidSequences,
-            files = rawProcessedData.processedData.files,
         )
     }
 
--- a/backend/build.gradle
+++ b/backend/build.gradle
@@ -46,6 +46,7 @@ dependencies {
     implementation "org.hibernate.validator:hibernate-validator:9.0.0.Final"
     implementation "org.keycloak:keycloak-admin-client:26.0.5"
     implementation("io.minio:minio:8.5.17")
+    implementation("software.amazon.awssdk:s3:2.31.59")
 
     implementation "org.springframework.boot:spring-boot-starter-oauth2-resource-server"
     implementation "org.springframework.boot:spring-boot-starter-security"
--- a/backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
@@ -173,6 +173,39 @@ data class ProcessedData<SequenceType>(
     val files: FileCategoryFilesMap?,
 )
 
+data class ReleasedData(
+    @Schema(
+        example = """{"date": "2020-01-01", "country": "Germany", "age": 42, "qc": 0.95}""",
+        description = "Key value pairs of metadata, correctly typed",
+    )
+    val metadata: MetadataMap,
+    @Schema(
+        example = """{"segment1": "ACTG", "segment2": "GTCA"}""",
+        description = "The key is the segment name, the value is the nucleotide sequence",
+    )
+    val unalignedNucleotideSequences: Map<SegmentName, GeneticSequence?>,
+    @Schema(
+        example = """{"segment1": "ACTG", "segment2": "GTCA"}""",
+        description = "The key is the segment name, the value is the aligned nucleotide sequence",
+    )
+    val alignedNucleotideSequences: Map<SegmentName, GeneticSequence?>,
+    @Schema(
+        example = """{"segment1": ["123:GTCA", "345:AAAA"], "segment2": ["123:GTCA", "345:AAAA"]}""",
+        description = "The key is the segment name, the value is a list of nucleotide insertions",
+    )
+    val nucleotideInsertions: Map<SegmentName, List<Insertion>>,
+    @Schema(
+        example = """{"gene1": "NRNR", "gene2": "NRNR"}""",
+        description = "The key is the gene name, the value is the amino acid sequence",
+    )
+    val alignedAminoAcidSequences: Map<GeneName, GeneticSequence?>,
+    @Schema(
+        example = """{"gene1": ["123:RRN", "345:NNN"], "gene2": ["123:NNR", "345:RN"]}""",
+        description = "The key is the gene name, the value is a list of amino acid insertions",
+    )
+    val aminoAcidInsertions: Map<GeneName, List<Insertion>>,
+)
+
 data class ExternalSubmittedData(
     override val accession: Accession,
     override val version: Version,
--- a/backend/src/main/kotlin/org/loculus/backend/config/Config.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/config/Config.kt
@@ -51,6 +51,7 @@ data class Schema(
     val externalMetadata: List<ExternalMetadata> = emptyList(),
     val earliestReleaseDate: EarliestReleaseDate = EarliestReleaseDate(false, emptyList()),
     val submissionDataTypes: SubmissionDataTypes = SubmissionDataTypes(),
+    val files: List<FileCategory> = emptyList(),
 )
 
 data class SubmissionDataTypes(
--- a/backend/src/main/kotlin/org/loculus/backend/model/SubmitModel.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/model/SubmitModel.kt
@@ -133,12 +133,6 @@ class SubmitModel(
             val sequenceSubmissionIds = uploadDatabaseService.getSequenceUploadSubmissionIds(uploadId).toSet()
             validateSubmissionIdSetsForConsensusSequences(metadataSubmissionIds, sequenceSubmissionIds)
         }
-        submissionParams.files?.let { submittedFiles ->
-            val fileSubmissionIds = submittedFiles.keys
-            validateSubmissionIdSetsForFiles(metadataSubmissionIds, fileSubmissionIds)
-            // the check below reads the DB, so needs to be after 'insertDataIntoAux'
-            validateFileExistenceAndGroupOwnership(submittedFiles, submissionParams, uploadId)
-        }
 
         if (submissionParams is SubmissionParams.RevisionSubmissionParams) {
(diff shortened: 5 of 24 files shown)


website/src/config.ts
```diff
@@ -150,8 +151,8 @@ function getAccessionInputField(): InputField {
 
 function getSubmissionIdInputField(): InputField {
     return {
-        name: SUBMISSION_ID_FIELD,
-        displayName: 'Submission ID',
+        name: SUBMISSION_ID_INPUT_FIELD,
```
fhennig, line 154: I wouldn't add the `_INPUT` part to this field name - why did you think of doing that? Other fields are also inputs possibly, and the don't have this.

fhennig:
Looks all very sensible, nice!

I would like to have two more backend tests:
- A test that shows that the old submissionId submitting still works
- A test that shows that it's not possible to submit a file with both id and submissionId

If that's there and tests are green I think I would approve.

fhennig:
I've removed the preview label here for now, feel free to add it back! Just a bit many small PRs going on at the moment (I created a lot of them 😬 )

theosanderson:
Thanks Felix – I've now added these tests

anna-parker (dismissed):
I was just looking into the question that prompted this and I think the problem is not the name submissionId but the fact that we allow users to add a submissionId on the single sequence submission form and do not properly explain in the tool tip what it means. 

I think it might be better to not rename the field but improve the tooltip to be clearer or to entirely remove this field for the single-sequence submission form:

image

Update: As a side-note `id` implies a field is unique and this is not the case - even `submissionId` is a bit confusing in this regard

backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
chaoran-chen, line 1: On the website, the old `submissionId` column is not mapped to `id`, I think that such a mapping would be very useful for backward compatibility.

image

chaoran-chen (changes requested):
Thanks very much, looks great! Just a few (hopefully quite minor) comments.

AGENTS.md
chaoran-chen, line 1: Strange.. – the file upload is not working for me (on main, it works). Any idea what's causing this?

https://github.com/user-attachments/assets/b984bdc6-4c9d-440c-9b7e-0c67db8c0014

theosanderson: [haven't tried yet, but sounds very likely I broke something with mapping] Oh wow, sorry and thanks for spotting!! - I was overrelying on automated tests here and probably they mock some crucial component I guess, or don't cover it.

theosanderson: oh weird, when I posted my comment above I was watching the first bit of the video where you upload the metadata file (and assuming something was broken there) - files seems very odd: maybe it could be preview thing with insufficient resources for S3

corneliusroemer (changes requested):
I'll review this by Friday EOB

backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
```diff
@@ -136,7 +136,7 @@ open class ReleasedDataModel(
             mapOf(
                 ("accession" to TextNode(rawProcessedData.accession)),
                 ("version" to LongNode(rawProcessedData.version)),
-                (HEADER_TO_CONNECT_METADATA_AND_SEQUENCES to TextNode(rawProcessedData.submissionId)),
+                ("submissionId" to TextNode(rawProcessedData.submissionId)),
```
corneliusroemer, line 139: Why don't we use `HEADER_TO_CONNECT_METADATA_AND_SEQUENCES_ALTERNATE_FOR_BACKCOMPAT` here?

anna-parker, line 139: maybe we should still call this by its variable name? `HEADER_TO_CONNECT_METADATA_AND_SEQUENCES_ALTERNATE_FOR_BACKCOMPAT`

backend/src/main/kotlin/org/loculus/backend/utils/MetadataEntry.kt
```diff
@@ -17,31 +18,37 @@ fun metadataEntryStreamAsSequence(metadataInputStream: InputStream): Sequence<Me
...
+
+    when {
+        submissionIdHeaders.isEmpty() -> throw UnprocessableEntityException(
+            "The metadata file does not contain either header '$HEADER_TO_CONNECT_METADATA_AND_SEQUENCES' or '$HEADER_TO_CONNECT_METADATA_AND_SEQUENCES_ALTERNATE_FOR_BACKCOMPAT'",
+        )
+        submissionIdHeaders.size > 1 -> throw UnprocessableEntityException(
+            "The metadata file contains both '$HEADER_TO_CONNECT_METADATA_AND_SEQUENCES' and '$HEADER_TO_CONNECT_METADATA_AND_SEQUENCES_ALTERNATE_FOR_BACKCOMPAT'. Only one is allowed.",
         )
     }
+    val submissionIdHeader = submissionIdHeaders.first()
```
corneliusroemer, line 35: Why use first here rather than explicit `HEADER_TO_CONNECT_METADATA_AND_SEQUENCES`?

I see, because this should only contain one item after filtering out the other one.

backend/src/main/kotlin/org/loculus/backend/utils/MetadataEntry.kt
```diff
@@ -60,40 +67,44 @@ fun revisionEntryStreamAsSequence(metadataInputStream: InputStream): Sequence<Re
...
     )
 
-    if (!csvParser.headerNames.contains(HEADER_TO_CONNECT_METADATA_AND_SEQUENCES)) {
-        throw UnprocessableEntityException(
-            "The revised metadata file does not contain the header '$HEADER_TO_CONNECT_METADATA_AND_SEQUENCES'",
+    val headerNames = csvParser.headerNames
+    val submissionIdHeaders = listOf(
+        HEADER_TO_CONNECT_METADATA_AND_SEQUENCES,
+        HEADER_TO_CONNECT_METADATA_AND_SEQUENCES_ALTERNATE_FOR_BACKCOMPAT,
+    ).filter { headerNames.contains(it) }
```
corneliusroemer, line 74: This is really hard to read

backend/src/main/kotlin/org/loculus/backend/utils/MetadataEntry.kt
```diff
@@ -60,40 +67,44 @@ fun revisionEntryStreamAsSequence(metadataInputStream: InputStream): Sequence<Re
...
+
+    when {
+        submissionIdHeaders.isEmpty() -> throw UnprocessableEntityException(
+            "The revised metadata file does not contain either header '$HEADER_TO_CONNECT_METADATA_AND_SEQUENCES' or '$HEADER_TO_CONNECT_METADATA_AND_SEQUENCES_ALTERNATE_FOR_BACKCOMPAT'",
+        )
+        submissionIdHeaders.size > 1 -> throw UnprocessableEntityException(
+            "The revised metadata file contains both '$HEADER_TO_CONNECT_METADATA_AND_SEQUENCES' and '$HEADER_TO_CONNECT_METADATA_AND_SEQUENCES_ALTERNATE_FOR_BACKCOMPAT'. Only one is allowed.",
```
anna-parker, line 81: I don't quite understand this code, isn't the following clearer?

```suggestion
val containsHeader = csvParser.headerNames.contains(HEADER_TO_CONNECT_METADATA_AND_SEQUENCES)
val containsHeaderBackcompat = csvParser.headerNames.contains(HEADER_TO_CONNECT_METADATA_AND_SEQUENCES_ALTERNATE_FOR_BACKCOMPAT)

if (containsHeader1 && containsHeader2) {
    throw UnprocessableEntityException(
        "The revised metadata file contains both '$HEADER_TO_CONNECT_METADATA_AND_SEQUENCES' and '$HEADER_TO_CONNECT_METADATA_AND_SEQUENCES_ALTERNATE_FOR_BACKCOMPAT'. Only one is allowed.",
    )
}
if (!containsHeader && !containsHeader2) {
    throw UnprocessableEntityException(
        "The revised metadata file does not contain either header '$HEADER_TO_CONNECT_METADATA_AND_SEQUENCES' or '$HEADER_TO_CONNECT_METADATA_AND_SEQUENCES_ALTERNATE_FOR_BACKCOMPAT'"
    )
}
```

backend/src/main/kotlin/org/loculus/backend/utils/MetadataEntry.kt
```diff
@@ -17,31 +18,37 @@ fun metadataEntryStreamAsSequence(metadataInputStream: InputStream): Sequence<Me
         CSVFormat.TDF.builder().setHeader().setSkipHeaderRecord(true).build(),
     )
 
-    if (!csvParser.headerNames.contains(HEADER_TO_CONNECT_METADATA_AND_SEQUENCES)) {
-        throw UnprocessableEntityException(
-            "The metadata file headers do not contain the header '$HEADER_TO_CONNECT_METADATA_AND_SEQUENCES': ${csvParser.headerNames}",
+    val headerNames = csvParser.headerNames
```
anna-parker, line 21: see suggestion below

docs/src/content/docs/introduction/glossary.md
```diff
@@ -95,9 +95,9 @@ SILO is an [open-source query engine](https://github.com/GenSpectrum/LAPIS-SILO)
 
 A submission adds new [sequence entries](#sequence-entry). See also [revision](#revision) and [revocation](#revocation).
 
-### Submission ID
+### ID
 
-When users upload [sequence entries](#sequence-entry), they have to provide a submission ID to link the entries in the [metadata](#metadata) file and the FASTA file. Each submission ID must be unique within the submission, but re-use across submissions is acceptable.
+When users upload [sequence entries](#sequence-entry), they have to provide an ID to link the entries in the [metadata](#metadata) file and the FASTA file. Each ID must be unique within the submission, but re-use across submissions is acceptable.
```
corneliusroemer, line 100:
```suggestion
When users upload [sequence entries](#sequence-entry), they have to provide an ID to link the entries in the [metadata](#metadata) file and the FASTA file. Each ID must be unique within the submission, but re-use across submissions is acceptable (though advised against).
```

website/src/components/Submission/FormOrUploadWrapper.spec.tsx
corneliusroemer, line 1: There's a lot of use of literal `ID` - might want to put this into a constant (predates this PR)

website/tests/pages/submission/template.spec.ts
corneliusroemer, line 1: shouldn't we use the constant here that we defined?

```
export const SUBMISSION_ID_INPUT_FIELD = 'id';
```

theosanderson: I think often in tests one doesn't do that - one specifies the raw results one actually expects? - otherwise one can end up saying 

`assert myComplexFn(x) === myComplexFn(x)`

theosanderson:
>The display names could be changed without changing the csv header field. Most of the backend code deals with 1), the website code with 2).

Sure but our default submission method is CSV so that's a pretty important component of this (and that inconsistency wouldn't be great)

> The backend code is quite confusing to me - might there be a simpler way?

I'm happy with any approaches that achieve the same outcome!

backend/src/main/kotlin/org/loculus/backend/utils/MetadataEntry.kt
theosanderson: We have separate for RevisionEntrys and MetadataEntrys - I'm happy to accept any pushes to refactor this but I only think we need to keep the back-compatibility for a few weeks anyway

AGENTS.md
theosanderson: After restarting the preview, this is fine - I made an issue: https://github.com/loculus-project/loculus/issues/4355

chaoran-chen: Thanks very much!

backend/src/main/kotlin/org/loculus/backend/utils/MetadataEntry.kt
theosanderson: I extracted a shared function

backend/src/main/kotlin/org/loculus/backend/utils/MetadataEntry.kt
theosanderson: I refactored to pull out a shared function which should make it more readable in that the intention is clearer

theosanderson:
@corneliusroemer I have now tested on staging

ingest/scripts/override_group_segments.py
```diff
@@ -180,7 +180,7 @@ def write_grouped_metadata(
     groups: Groups,
 ) -> tuple[dict, set]:
     # Map from original accession to the new concatenated accession
-    fasta_id_map: dict[Accession, SubmissionId] = {}
+    fasta_id_map: dict[Accession, id] = {}
```
corneliusroemer, line 183:
```suggestion
    fasta_id_map: dict[Accession, Id] = {}
```

Type should be capitalized?

corneliusroemer (commented):
One small nit (type casing), otherwise looks ok - I'm not a fan of the rename but I won't block

ingest/scripts/override_group_segments.py
theosanderson: Thanks - yeah! Feels like CI should have caught this (not necessarily the capitalisation thing, but the fact that `id` didn't exist as I had fixed the capitalisation above).

corneliusroemer: we're not running type checking in CI yet - though some copilot AI should have! 😁

PR #3715, opened 2025-02-18
chore(deployment): reduce cpu requests to realistic requirements allowing more previews

corneliusroemer:
resolves #

preview URL:

### PR Checklist

### DIFF
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -1719,7 +1719,7 @@ resources:
   ingest:
     requests:
       memory: "1Gi"
-      cpu: "200m"
+      cpu: "20m"
     limits:
       cpu: "200m"
       memory: "10Gi"
@@ -1738,7 +1738,7 @@ resources:
   lapis:
     requests:
       memory: "220Mi"
-      cpu: "30m"
+      cpu: "20m"
     limits:
       memory: "5Gi"
   silo-preprocessing:
@@ -1749,7 +1749,7 @@ resources:
   backend:
     requests:
       memory: "640Mi"
-      cpu: "250m"
+      cpu: "150m"
     limits:
       memory: "3Gi" # Backend requires at least 635741K of memory
   preprocessing:
@@ -1763,6 +1763,6 @@ defaultResources:
     cpu: "20m"
   limits:
     memory: "1Gi"
-    cpu: "20m"
+    cpu: "1000m"
 ingest:
   ncbiGatewayUrl: null
--- a/kubernetes/loculus/templates/keycloak-database-standin.yaml
+++ b/kubernetes/loculus/templates/keycloak-database-standin.yaml
@@ -29,7 +29,7 @@ spec:
               memory: "30Mi"
               cpu: 10m
             limits:
-              memory: "100Mi"
+              memory: "500Mi"
           ports:
             - containerPort: 5432
           env:
--- a/kubernetes/loculus/templates/loculus-database-standin.yaml
+++ b/kubernetes/loculus/templates/loculus-database-standin.yaml
@@ -26,7 +26,7 @@ spec:
         resources:
           requests:
             memory: "200Mi"
-            cpu: "250m"
+            cpu: "50m"
           limits:
             memory: "2Gi"
         ports:


kubernetes/loculus/values.yaml
theosanderson, line 1: > We're currently limited in Loculus preview deployments because of having too much CPU requests. The requests are way higher than actual utilization, which means we could actually have more previews without any issues - if we simply reduced overly high CPU requests.

This wasn't the case earlier today when I had to remove previews - my preview was running but various components dying due to OOM issues. I'm not saying that's always true - I presume you are seeing an issue that has prompted this - but presumably we're relatively balanced

kubernetes/loculus/values.yaml
```diff
@@ -1719,7 +1719,7 @@ resources:
   ingest:
     requests:
       memory: "1Gi"
-      cpu: "200m"
+      cpu: "20m"
```
theosanderson, line 1722: This only makes sense to me as a change specifically for previews?

kubernetes/loculus/values.yaml
```diff
@@ -1763,6 +1763,6 @@ defaultResources:
     cpu: "20m"
   limits:
     memory: "1Gi"
-    cpu: "20m"
+    cpu: "1000m"
```
theosanderson, line 1766: huh yeah this seems an odd decision! I guess we accidentally added this! 🤔  we could just remove the limit entirely?

kubernetes/loculus/values.yaml
theosanderson: (same with the others below)

corneliusroemer: All these values here are for previews, for production and PPX we use different ones and there's no real problem in setting these requests on the lower end as in the worst case it'll cause throttling which isn't a problem for ingest

kubernetes/loculus/values.yaml
corneliusroemer:
```suggestion
```

kubernetes/loculus/values.yaml
theosanderson: Yes but it will affect e.g. CovSpectrum and anyone else who tries out Loculus. In this values file we should aim (mostly) for sensible defaults for a generic production environment. I would think this is a bit low for that

theosanderson: but ok, I see LAPIS below isn't that on main

theosanderson (approved):

kubernetes/loculus/values.yaml
```diff
@@ -1763,6 +1763,5 @@ defaultResources:
     cpu: "20m"
   limits:
     memory: "1Gi"
-    cpu: "20m"
```
theosanderson, line 1766: huh - this isn't allowed - seems like a bug in the schema - but your original `1000m` seems good for now

kubernetes/loculus/values.yaml
theosanderson: or actually maybe LAPIS doesn't need much- I guess SILO does the heavy lifting

kubernetes/loculus/values.yaml
corneliusroemer: Hmm, there were a bunch of events saying "failed scheduling due to insufficient CPU" throughout the day and we were above 93% for CPU utilization earlier

Google Chrome Beta 2025-02-19 00 06 38

kubernetes/loculus/values.yaml
corneliusroemer: Schema complains: 
image

Is that a correct complaint or a false positive complaint?

kubernetes/loculus/values.yaml
corneliusroemer: https://grafana.loculus.org/explore?schemaVersion=1&panes=%7B%22tnw%22:%7B%22datasource%22:%22P8E80F9AEF21F6940%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bkubernetes_events%3D%5C%221%5C%22%7D%20%7C%3D%20%60Insufficient%20cpu%60%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22P8E80F9AEF21F6940%22%7D,%22editorMode%22:%22builder%22%7D%5D,%22range%22:%7B%22from%22:%22now-24h%22,%22to%22:%22now%22%7D%7D%7D&orgId=1

theosanderson: Sure (but in my specific case it was memory that was crashing it - I could see it in argoCD, so both were likely limiting)

kubernetes/loculus/values.yaml
theosanderson: > seems like a bug in the schema

PR #3724, opened 2025-02-20
fix(k8s): treat LAPIS as healthy, even if SILO not up

corneliusroemer:
Switch to using the intrinsic `/actuator/health` instead of the `/sample/info` endpoint for liveness/startupProbe.

`/sample/info` returns 503 when SILO is down, causing LAPIS container to be restarted whenever SILO is down which makes no sense.

We could potentially add a readinessProbe to fine tune restart behaviors.

If we don't want LAPIS to receive traffic if SILO is down, then we could set a _readinessProbe_. But it might be ok for LAPIS to accept traffic if it caches things and doesn't always rely on SILO to be up. Some tradeoffs there.

This PR should remove the restarts that we always get for LAPIS - even under normal operating conditions. Keeps everything cleaner, less noise in event logs etc

image

### DIFF
--- a/kubernetes/loculus/templates/lapis-silo-deployment.yaml
+++ b/kubernetes/loculus/templates/lapis-silo-deployment.yaml
@@ -57,14 +57,14 @@ spec:
               subPath: reference_genomes.json
           startupProbe:
             httpGet:
-              path: /sample/info
+              path: /actuator/health
               port: 8080
             periodSeconds: 10
             failureThreshold: 5
             timeoutSeconds: 5
           livenessProbe:
             httpGet:
-              path: /sample/info
+              path: /actuator/health
               port: 8080
             initialDelaySeconds: 30
             periodSeconds: 10


kubernetes/loculus/templates/lapis-silo-deployment.yaml
theosanderson, line 1: Do we know that this won't lead to LAPIS receiving requests that it cannot serve during the rollout of a new deployment? I.e. under the current system the old pod will be kept around until the new one is fully ready, and requests won't be sent to the new one until then. Will that definitely still be true after this? (You've probably checked / thought it through, but just checking)

theosanderson: I think this probably would be OK atm, though we'd need to change it when we move LAPIS, SILO and SILO-import into their own separate deployments as we plan to one day, so I'm not sure how much value there is

theosanderson: (My comment crossed with yours)

theosanderson: There's a general kubernetes pattern of marking something as unhealthy if it can't function for any reason (including that its dependencies aren't yet met), which applies to lots of aspects of our deployment (e.g. keycloak), and which I don't think is a big problem - it has downsides as you flag of slightly increasing resource usage, but has the upside that you are very confident of something really being healthy.

theosanderson: > If we don't want LAPIS to receive traffic if SILO is down, then we could set a readinessProbe. But it might be ok for LAPIS to accept traffic if it caches things and doesn't always rely on SILO to be up. Some tradeoffs there.

I would be very against LAPIS accepting traffic in this state - I'm pretty sure LAPIS can't respond to arbitrary queries in this state, otherwise one wouldn't need SILO

theosanderson: probably the right way to reduce noise here is indeed to move the current livenessProbe to a readingessProbe

corneliusroemer:
Todo: turn old liveness into readiness

kubernetes/loculus/templates/lapis-silo-deployment.yaml
fengelniederhammer: -> Using `/actuator/health` is probably good for startupProbe and livenessProbe. Would it make sense to add a readinessProbe?

theosanderson: > For sure, I don't think I claimed it would be able to serve arbitrary requests.

Yes but it seems to me not to make sense to degrade users from being able to issue searches for a whole SILO startup cycle if there is an easy way to avoid that..

theosanderson:
We got caught out by this recently (due to an issue we were not aware of that SILO can easily become saturated) so we implemented something very like this (but we also split up LAPIS and SILO) - we now need to make some further changes around the issues I flagged around what happens during deployments because SILO can receive traffic before it has processed the new data, but this PR is implemented

PR #3730, opened 2025-02-21
fix(website)!: use `box-content` for consistent column widths regardless of position

theosanderson:
preview URL: https://box-content.loculus.org/ebola-sudan/search? vs https://main.loculus.org/ebola-sudan/search?

### Summary
We specifically add padding to the last column in the table for aesthetic reasons (`last:pr-6`), but this is problematic because we specify widths for each column but those widths include the padding and we don't know which column the user will choose to be last. Here we use [`box-content`](https://tailwindcss.com/docs/box-sizing) so that the specified width is just for the content, excluding any padding. This will likely require us to update all widths to take account of this.

All columns get a bit bigger on our current previews but I think that's ok as they are not tuned much on Loculus (they are more so on PPX)

### Details of breaking changes

We may want to make all specified column widths on PPX 16 pixels narrower to account for the fact that padding is not longer being taken out

### DIFF
--- a/website/src/components/SearchPage/Table.tsx
+++ b/website/src/components/SearchPage/Table.tsx
@@ -161,7 +161,7 @@ export const Table: FC<TableProps> = ({
                                     <th
                                         key={c.field}
                                         onClick={() => handleSort(c.field)}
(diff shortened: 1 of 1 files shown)


anna-parker (approved):
Im happy to approve this :-)

corneliusroemer:
Ah I think I see it, the last column breaks date into two lines in main but not on preview 😀

Issue #3344, opened 2024-12-02
File/Raw data sharing MVP


chaoran-chen:
I've been talking a lot to @DrYak, @gordonkoehn and @Taepper about how to add raw data sharing capabilities to Loculus. We developed a plan to do this step-by-step and would love to hear everyone's feedback.

Overall, the idea is to stick very closely to the current architecture where the backend only passes data around and is responsible for general things like persistence and versioning while a configurable, external pipeline is responsible for actually processing the data. In following, I will first outline the final picture and then propose three different implementation stages.

## Final picture

**User view:** Depending on the instance, it is possible to upload metadata, consensus sequences and/or raw data. The consensus sequences are, as now, stored in a FASTA. The raw data can be a FASTQ, a BAM file or something else – what exactly one has/can upload depends also on the instance. The pipeline will take the user data and return processed data: this could be only consensus sequences and metadata (i.e., some tabular data), or include some other unaligned raw data (e.g., the raw reads after dehumanizing) or also include aligned raw reads. The user will be able to download all processed data. Aligned reads may be passed into SILO for fast querying.

**Technical view:** If the instance only supports metadata and consensus sequences, everything remains the same. If a user wants to upload raw data as well, their client software (could be the Loculus website or CLI) first submits the metadata (and maybe consensus sequences) with a request to upload raw data. The backend returns ~S3 URLs and some access token~ S3 pre-signed URLs and the client then uploads the raw data to the specified locations.

When the pipeline asks for unprocessed data, the backend will send the S3 URLs of the submitted data in addition to the metadata (and consensus sequences). The backend will also provide S3 locations where the processed raw data should be stored. The pipeline will then download the unprocessed raw data from S3 and then upload the processed raw data again to S3. Then, the pipeline calls the backend again, submitting the processed metadata (and consensus sequences).

The backend will include in `/get-released-data` all S3 URLs to the processed raw data. If there are aligned raw data, the SILO input job can download the processed aligned raw data and give them to SILO preprocessing.

**Use cases:** This will enable various use cases; some examples:

1. Users upload consensus sequences and raw data. The pipeline performs some checks like human reads detection but doesn't otherwise do much. If the data are ok, they pass the pipeline and can be downloaded and potentially subsequently uploaded to INSDC.
2. Users only upload raw data, the pipeline generates consensus sequences.
3. Users upload environmental/wastewater raw data, the pipelne performs some analyses (e.g., lineage deconvolution) and the results are returned as processed data.
[...]

PR #3732, opened 2025-02-24
Add ADR for file sharing

chaoran-chen:
Following the discussion in last week's retro, this is a draft ADR (architectural decision record) for the raw data/file sharing feature (#3344).

https://github.com/loculus-project/loculus/blob/adr-file-sharing/decisions/0001-file-sharing.md

- I used the template from here: https://adr.github.io/adr-templates/
- For now, I just put the file into the `/decisions` folder, happy to get alternative suggestions!
- As I understand, this is designed to be a living file. Please feel free to add options and pro/contra arguments, etc. directly to the file, and comment if you disagree with a point or have questions.

### DIFF
--- a/architecture_docs/decisions/0001-file-sharing.md
+++ b/architecture_docs/decisions/0001-file-sharing.md
@@ -24,32 +24,41 @@ From the user perspective, depending on the instance, it should be possible to u
 * Scalability of traffic: The system should be able to handle the upload and download of large files.
 * Same data privacy, integrity, and persistence model as before: This means that the data uploaded by a group should remain private until it has been approved. Once approved, it should be guaranteed that the originally uploaded data remain available and versioned (unless explicit interventions by a maintainer).
 
+## Implemented solution
+
+We decided to go for option (a), below are some more descriptions from after the implementation.
+
+- S3 storage as the backend to store the actual files. All files are stored there (both released and unreleased) and released files are tagged as `public=true` - a policy is added to the bucket to make these files publicly accessible.
+- upload URLs need to be requested from the backend, under `/files/request-upload`. The endpoint returns presigned URLs where the client uploads the files. For every file and URL, there is also a file ID which is then subsequently used to refer to the file.
+- When submitting (to the `/submit` endpoint) there is a new `fileMapping` parameter which is supplied in the body of the request. It contains a mapping of submission IDs to file IDs that belong to the submission.
+- When the preprocessing pipeline requests unprocessed data, the backend generates presigned GET URLs to the pipeline can access the files. The pipeline can also upload new files, the same way the client can.
+- When a submission is released, the files associated with the submission are tagged as public, and are then also accessible for anyone.
 
 ## Considered Options
 
-### S3 and pre-signed URLs
+### (a) S3 and pre-signed URLs
 
 - To upload files, the authenticated client first calls a new endpoint `/files/request-uploads`. The backend returns S3 pre-signed URLs to which the client uploads the files.
 - When the client calls `/submit` (or `/revise`), they can provide in the metadata file in the column `files` the file IDs of the associated files.
 - The backend will provide to the preprocessing pipeline pre-signed URLs to read the original files. If the pipeline would like to upload processed files, it calls the same `/files/request-uploads` endpoint.
 - During the review, the user receives pre-signed URLs to download the original and processed files.
 
-### S3 and (semi-)public write-only link
+### (b) S3 and (semi-)public write-only link
 
 - Authenticated users can see on the website information to upload to files to a shared S3 bucket.
 - They should choose a random object names to avoid conflicts with uploads of other users.
 - They state the object name in the metadata file when they submit (or revise).
 
-### User-controlled S3
+### (c) User-controlled S3
 
 - Users upload files to their own S3 bucket and give Loculus read access.
 
-### Direct upload to backend
+### (d) Direct upload to backend
 
 - Users directly uploads files to the backend as part of `/submit` or `/revise`.
 - The backend then stores the files on S3 (or potentially at a different location).
 
-### Combination of "S3 and pre-signed URLs" and "Direct upload to backend"
+### (e) Combination of "S3 and pre-signed URLs" and "Direct upload to backend"
 
 - Loculus can support the workflow as described in the option "S3 and pre-signed URLs", targeted at large files and use cases where maintainer does not want to direct large amount of traffic through the backend.
 - In addition, if enabled in the configuration, the user may upload files directly with their call to `/submit` or `/revise`. The maintainer may limit the max. size of files uploaded through this approach to only accept rather small files.
@@ -127,9 +136,3 @@ Chosen option: "{title of option 1}", because {justification. e.g., only option,
 
 * Good, because it saves storage if the same file is uploaded.
 * (We need to clarify whether it is possible to get the hash directly from S3 or need to calculate it ourselves. The latter is bad, because Loculus backend needs to download the files from S3 to hash them which takes time and network traffic.)
-
-
-<!-- This is an optional element. Feel free to remove. -->
-## More Information
-
-{You might want to provide additional evidence/confidence for the decision outcome here and/or document the team agreement on the decision and/or define when/how this decision the decision should be realized and if/when it should be re-visited. Links to other decisions and resources might appear here as well.}


emmahodcroft:
For me for this particular raw-data thing it would be useful to maybe delineate the difference between the aim of what I understand is the current step and how this differs from a more 'full' support of raw-reads (ex: checking for human reads, which is pretty important for making this data more public, and uploading to ENA). Just to be clear how this fits into the large picture. Just in the 'intro' or equivalent.

Not sure if that change would happen in this PR or if this is merged first and then one changes... (I am unlikely to be able to work on this much after today until next week)

decisions/0001-file-sharing.md
```diff
@@ -0,0 +1,103 @@
...
+
+The raw read data of a sample is significantly larger than a consensus sequence. For example, a consensus sequence of a SARS-CoV-2 sample is about 30 KB, while the raw data is around hundred MB to a few GB. We therefore need an approach to store and share large files.
+
+From the user perspective, depending on the instance, it should be possible to upload metadata, consensus sequences and/or raw data. The consensus sequences are, as now, stored in a FASTA. The raw data can be a FASTQ, a BAM file or something else – what exactly one has/can upload depends also on the instance. The maintainer-configurable pipeline will take the user data and return processed data: this could be only consensus sequences and metadata (i.e., some tabular data), or include some other unaligned raw data (e.g., the raw reads after dehumanizing), aligned raw reads, or other type of files. The user will be able to download all processed data. Aligned reads may be passed into SILO for fast querying.
```
anna-parker, line 18: It would be great to add a "Critical User Journey" Section (was popular at Google) where we describe who the user of this is (only WISE or should this also be used by anyone uploading raw reads) and what their user "journey" should be. 

e.g. I am a bioinformatician/ biologist I have raw reads (how many usually? One file, multiple files?) that I store for each sample, I generate consensus sequences and have an excel table with the metadata. I already link the consensus sequences to metadata using a submissionId and I prepare the data in advance. I would like to upload raw reads, consensus sequences and metadata. It would be nice to prepare data, upload, check preprocessing results and approve.

e.g. I am a bioinformatician at WISE I have raw reads stored in an S3, I want to link them to the metadata and get preprocessing results.

fhennig:
Maybe the `decisions` dir can be put into the `architecture_docs` and also individual decisions could be linked in there: https://github.com/loculus-project/loculus/blob/main/architecture_docs/09_architecture_decisions.md

fhennig:
I just had a read through, thanks for picking up my idea and taking the time to write this! At least I now feel like I have a better understanding of the whole situation, thanks!

decisions/0001-file-sharing.md
anna-parker, line 18: > a FASTQ, a BAM file or something else

will there always be one file per raw read sample or will there ever be multiple files?

decisions/0001-file-sharing.md
```diff
@@ -0,0 +1,103 @@
...
+
+
+## Decision Drivers
+
+* Scalability of storage size: The system should be able to scale to TBs or even PBs of data files.
+* Scalability of traffic: The system should be able to handle the upload and download of large files.
+* Same data privacy, integrity, and persistence model as before: This means that the data uploaded by a group should remain private until it has been approved. Once approved, it should be guaranteed that the originally uploaded data remain available and versioned (unless explicit interventions by a maintainer).
+
```
anna-parker, line 26: Could you add a section why we decided against also allowing users to drag and drop fies of raw reads? This would be the nicest user experience

decisions/0001-file-sharing.md
```diff
@@ -0,0 +1,103 @@
...
+## Pros and Cons of the Options
+
+### S3 and pre-signed URLs
+
+* Good, because this gives the backend fine-grained control over the uploaded files:
+  * It controls where files are uploaded to, avoiding users overwriting each other with certainty.
+  * It can set a max file size for each file.
+  * It can limit number of uploads (in total/within a time period/for a user/group/etc.) to limit the use of resources and potential for abuse
+* Bad, because (compared to option "S3 and (semi-)public write-only link"), users need to call an additional endpoint to receive the pre-signed URLs.
+
```
anna-parker, line 80: Additionally users need to update the metadata with the links of each raw read that they have uploaded (additionally your proposal mentioned an additional call to say submission has been completed - so that is one more step)

decisions/0001-file-sharing.md
```diff
@@ -0,0 +1,103 @@
...
+### S3 and (semi-)public write-only link
+
+- Authenticated users can see on the website information to upload to files to a shared S3 bucket.
+- They should choose a random object names to avoid conflicts with uploads of other users.
+- They state the object name in the metadata file when they submit (or revise).
+
+### User-controlled S3
+
+- Users upload files to their own S3 bucket and give Loculus read access.
```
anna-parker, line 45: If WISE is the only user (and they already have an S3 this might actually be enough for the moment)

decisions/0001-file-sharing.md
theosanderson: I don't think we have decided against this?

theosanderson: There will sometimes be multiple files

decisions/0001-file-sharing.md
theosanderson: I don't think we have decided against this?

decisions/0001-file-sharing.md
theosanderson: (sorry I posted a comment here meant for another thread)

decisions/0001-file-sharing.md
```diff
@@ -0,0 +1,103 @@
...
+  * It can set a max file size for each file.
+  * It can limit number of uploads (in total/within a time period/for a user/group/etc.) to limit the use of resources and potential for abuse
+* Bad, because (compared to option "S3 and (semi-)public write-only link"), users need to call an additional endpoint to receive the pre-signed URLs.
+
+### S3 and (semi-)public write-only link
+
+* Good, because users don't need to call an additional endpoint to retrieve pre-signed URLs before each submission.
+* Bad, because Loculus cannot impose limits on the amount of uploaded data.
+* Bad, because once a user has the access information to write to the bucket, they keep their access even after their account has been removed/disabled. The only way to prevent a user from writing would be to change the access information which would affect all users.
+
```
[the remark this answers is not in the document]
theosanderson: The downside of this would be that the backend would have to handle potentially very large files

decisions/0001-file-sharing.md
fhennig: Yeah - _client_ can also be the web frontend, it can also handle the whole back-and-forth of calling endpoints and do it all for you, and you as a _user_ just provide the file.

decisions/0001-file-sharing.md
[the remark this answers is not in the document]
fhennig: I've also looked into writing an endpoint to simply accept files, but I have (from some very quick searches) found that it's actually not that easy to do. The multipart-file-upload we use so far shouldn't really be used for gigabytes of files, and instead it seemed to me like chunked uploads are used instead. That requires you to keep chunks around, and assemble it again, and it looked like Spring doesn't out-of-the-box have good support for this, and at that point it felt to me like it isn't a good idea. 

But all of this is to be taken with a grain of salt, as I didn't spend a lot of time looking into it.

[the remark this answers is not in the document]
anna-parker: I also asked chatGPT about this a couple weeks ago and it seemed relatively straight forward but I didn't implement it (I tried to recreate now): https://chatgpt.com/share/67bc635d-4eb8-8004-ac0e-1120aa50b36e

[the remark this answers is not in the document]
fhennig: Cool! I still have a looming feeling of "It might be more complicated" but that's not really founded on any concrete stuff 😅 

If it really is that easy, I think it'd be very nice if we could basically let the user submit files to our API as normal. The upside would be that we don't have the S3 stuff as a definite extra API surface IMO.

Downside is traffic I guess.

It would be great if this could somehow be worked into the document I think.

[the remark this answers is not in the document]
anna-parker: > The upside would be that we don't have the S3 stuff as a definite extra API surface IMO.

sorry I dont fully understand what you mean by this, could you explain more?

decisions/0001-file-sharing.md
anna-parker: Also good to already mention that ENA actually does quite some validation (or so it appears from the docs) on the metadata fields relating to raw reads - we will probably have to do this as well

chaoran-chen: Yes, it can be multiple files. For the Loculus backend and the architecture that we discuss here, I don't think that we need to limit/define the set of files that can be accepted.

decisions/0001-file-sharing.md
chaoran-chen: Agreed, we should (at a later stage) certainly also implement the upload of raw data in our website.

decisions/0001-file-sharing.md
chaoran-chen: At WISE, there is no maintained S3 instance at the moment (data are just stored on disk in the network, if I'm not mistaken). The goal of this implementation is to enable using Loculus to manage the data.

decisions/0001-file-sharing.md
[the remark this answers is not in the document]
fhennig: Sure! If we let users directly upload to S3, we make this part of our interface to the user. Either the bucket URL, or the pre-signed URL - doesn't really matter. I _guess_ we would probably stick with S3 as a storage mechanism forever, but if we would instead just expose our own API endpoint, S3 becomes a detail that the user doesn't need to know about and we could swap out another storage backend more easily.

IMO it's nice to keep these things flexible if it isn't that much work, that's why I think having our own endpoint would be much nicer.

chaoran-chen:
@emmahodcroft, I think that it makes sense to limit this ADR to the general architecture and not dive deeper into any specific pipeline implementation. As far as I can see, all options presented in this ADR maintain the general separation of concerns that we already have between the backend and the maintainer-configurable pipeline. They are all fully compatible with creating pipelines that do dehumanizing, etc. I also don't see any issues with extending the ENA deposition pipeline in the future, on top of options in this ADR, to upload raw data to ENA.

decisions/0001-file-sharing.md
theosanderson: It was one presigned URL per file, but that doesn't mean there can't be multiple

chaoran-chen: The pre-signed URL approach also works with multiple files. Both the user and the pipeline can upload multiple files by requesting multiple pre-signed URLs. Hereby, I envision that you can specify a number of files that you want to upload when calling `/files/request-uploads` to get that number of URLs. Afterward, when calling `/submit`, `/revise` or `/submit-processed-data`, it is possible to provide the IDs of multiple files.

> I guess when we allow multiple files we must specify that users must upload them in one folder and they give us a link to the folder.

There is no concept of a folder in S3 and it is not possible to generate one pre-signed URL for a folder but only for a specific object.

anna-parker: Ah ok that was were the requesting multiple URLs came in... I thought it was just if the user has multiple raw reads.

I got to admit I do think this becomes quite complex and it gets even harder for users to prepare their data in advance. It is possible to submit folders in aws - so I guess this is a limitation of signed URLs: 

```
aws s3 cp /local/folder/path s3://your-bucket-name/ --recursive
```

decisions/0001-file-sharing.md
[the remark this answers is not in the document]
chaoran-chen: I added the direct upload to the backend as a separate option: daabf50bc651a54bc13c65028c1a9453b4a9665a. Please feel free to suggest additions/edits directly to the document if the description of the idea is not accurate or arguments are missing.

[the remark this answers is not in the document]
theosanderson: Yet another option which would be a new single backend endpoint to upload a single file (which is how I was interpreting @anna-parker's idea, but wrongly)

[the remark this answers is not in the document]
chaoran-chen: Ah, i.e., having a backend endpoint to which you can directly upload a file and then get a file ID?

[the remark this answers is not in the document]
anna-parker: I would have liked to upload a directory to an endpoint and it would either return success or failure. Users would name the directory the same as the submissionId so they would not need to update the metadata - we would know from the folder name which submissionId a raw read was mapped to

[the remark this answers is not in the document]
chaoran-chen: Could you maybe specify an example HTTP request? Method, location, body, etc.? I'm not sure how uploading a directory works.

decisions/0001-file-sharing.md
chaoran-chen: But `aws s3 cp /local/folder/path s3://your-bucket-name/ --recursive` is not done in a single HTTP request, is it? I assume that it is split up into multiple requests.

I envision that most users won't need to care about the API but will just use a CLI or web client (that we will certainly implement in the future).

anna-parker: true! The endpoint would have to take a zipped file (it is possible to unzip while streaming without loading the file into memory so we could still do some checks on the data even)

decisions/0001-file-sharing.md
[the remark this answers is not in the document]
anna-parker: My suggestion would be to have users prepare the raw reads, metadata and consensus sequences. Raw reads should always be in a folder or a single file ideally with the same name as the submissionId and an arbitrary extension. The raw reads file path should be added into the `metadata.tsv`. 

Users would upload raw reads as follows: 
```
TOKEN=$(curl...)
curl -X 'POST' \
  'https://backend-main.loculus.org/{ORGANISM}/submit-raw-reads?groupId={GROUP_ID}' \
  -H 'Authorization: Bearer {TOKEN}' \
  -H 'Content-Type: multipart/form-data' \
  -F 'rawreads=@isolate9.zip' # they could also upload multiple inside a folder with their group_id
# -F 'rawreads=@{GROUP_ID}.zip'
curl -X 'POST' \
  'https://backend-main.loculus.org/{ORGANISM}/submit?groupId={GROUP_ID}' \
  -H 'Authorization: Bearer {TOKEN}' \
  -H 'accept: application/json' \
  -H 'Content-Type: multipart/form-data' \
  -F 'metadataFile=@metadata.tsv;type=text/tab-separated-values' \
  -F 'sequenceFile=@sequences.fasta'
```

Internally `submit-raw-reads` will proxy the POST request to the submission S3 bucket, it will upload all data into a directory called {GROUP_ID} and will rate limit the request. As we can unzip while streaming we can even do some checks on the data here.

After `submit` is called the processing pipeline will:
1. check if the raw reads file linked in the metadata.tsv exists in the submission S3 bucket and will do basic checks (i.e. looks like the correct virus genome, dehumanizing etc.) - this should be fast
2. compress and hash the file
3. move the files to a new loculus-internal S3 bucket - the files will be moved to a folder called LOCULUS_ACCESSION.VERSION. The data in the submission S3 bucket will be deleted. 

If the raw reads errors the user can re-upload the raw reads and resubmit the metadata - until LOCULUS_ACCESSION.VERSION is approved it can be overwritten, afterwards it cannot be overwritten and can only be revised.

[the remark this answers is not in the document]
anna-parker: chatGPT: https://chatgpt.com/share/67bccee8-1644-8004-880d-0406d2507569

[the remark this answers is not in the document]
chaoran-chen: Thanks for the explanations! Would you like to add it as an option to the ADR file, including pro/contra arguments?

> After submit is called the processing pipeline will: …

I'm surprised that you're suggesting the pipeline for doing all of this. I agree that the pipeline should do "the basic checks (i.e. looks like the correct virus genome, dehumanizing etc.)" but checking that the files are linked and moving them into a Loculus-internal bucket sounds like backend tasks to me. They are not domain-specific but tasks that would need to be done for every instance. In our current setting, it is also the backend that checks that the metadata and FASTA entries match and that then ensures that the data are persisted.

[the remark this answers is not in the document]
anna-parker: I actually think having a prepro pipeline do this might not be bad as the files will anyways be stored in an S3 and not the postgres DB. It also means we could already start doing some processing of the raw data before the metadata is uploaded to save wait time for users.
This will also mean that we wont require many changes in the backend. But the details can be refined later :-)

corneliusroemer (changes requested):
Requesting changes preemptively as discussed, as I know I'll have opinions on this one

decisions/0001-file-sharing.md
corneliusroemer, line 1: We already have `loculus/architecture_docs/01_introduction_and_goals.md` etc.

Why are we adding this under `loculus/decisions`?

decisions/0001-file-sharing.md
fhennig: Drag and drop of folders isn't supported AFAIKT, but we totally would if we could. Maybe in the future in the website we can somehow make single file upload easier with drag an drop or something.

decisions/0001-file-sharing.md
fhennig: I have moved it.

PR #3734, opened 2025-02-24
chore(docs): check formatting of docs in ci and do initial formatting

### LINKED ISSUE #2419: Check code formatting of the Loculus docs in CI

corneliusroemer:
TODO: still fails due to some ignore file thing

Closes #2419

- It works, it fails when it's not formatted: 

image

- The biggest diff here is that we need to format to pass things


docs/src/content/docs/for-administrators/sequence-reporting.md
```diff
@@ -9,10 +9,10 @@ You can enable it in the `values.yaml` by configuring the `sequenceFlagging` sec
...
 ```yaml
 sequenceFlagging:
-  github:
-    organization: pathoplexus
-    repository: curation_reports
-    issueTemplate: sequence-metadata-issue.md  # (optional)
+    github:
+        organization: pathoplexus
+        repository: curation_reports
+        issueTemplate: sequence-metadata-issue.md # (optional)
```
fhennig, line 15: Not a fan of the 4 spaces indent for yaml - it's not very common to do this I think.

fhennig: I made an override for this.

fhennig:
@corneliusroemer  you mentioned you have a lot of stuff to do, so I took the liberty to pick up your work here.

I have added the .gitignore file to the prettier check-format command to fix the issue you had initially.

I have also changed the formatting for md and mdx files to 2 space indent, as it's what we always use for YAML, so the YAML examples look right.

I hope that was fine!

docs/src/content/docs/for-administrators/sequence-reporting.md
corneliusroemer: 4 is default sometimes due to Python I think, but agree 2 is better

docs/package.json
```diff
@@ -9,6 +9,7 @@
         "build": "astro check && astro build",
         "preview": "astro preview",
         "astro": "astro",
+        "check-format": "prettier --ignore-path \"../.gitignore\" --ignore-path .gitignore  --check \"./**/*.{ts,tsx,json,astro,md,mdx,mjs,cjs}\"",
         "format": "prettier --ignore-path \"../.gitignore\" --write \"./**/*.{ts,tsx,json,astro,md,mdx,mjs,cjs}\""
```
corneliusroemer, line 13:
```suggestion
        "format": "prettier --ignore-path 
        "../.gitignore\" --ignore-path .gitignore  --write \"./**/*.{ts,tsx,json,astro,md,mdx,mjs,cjs}\""
```

corneliusroemer: I think we should also add the `.gitignore` here? Or do you want to format things that are ignored by git with `format`?

corneliusroemer:
Thanks @fhennig thanks for picking it up! I just have one comment re making format and check-format consistent. Feel free to approve (and merge if you like).

fhennig:
yes good point about the formatting!

Since Chaoran created the issue initially, I'll let him also have a look at do the final approval.

@chaoran-chen

chaoran-chen (approved):
This looks great to me, thank you!

PR #3735, opened 2025-02-24
fix(deployment): add old keycloak flags

### LINKED ISSUE #3618: Keycloak previews no longer allow login with admin:admin
Figure out why and fix. See https://loculus.slack.com/archives/C06JCAZLG14/p1738618231288519

Maybe relevant, recent answer on SO: https://stackoverflow.com/a/79295118/7483211

Weird, nothing has changed in our code:

https://github.com/loculus-project/loculus/blob/d1d145272e60de383f7802f31a59b07e0c0ce5a4/kubernetes/loculus/templates/keycloak-deployment.yaml#L74-L81

And we're still doing this:
https://www.keycloak.org/server/configuration#_creating_the_initial_admin_user

theosanderson:
resolves #3618 for now

preview URL: https://authentication-kcflags.loculus.org/admin/master/console (admin/admin)

### Summary
We upgraded the keycloak environment variable flags without upgrading keycloak itself, meaning an admin account with a known admin password is not created. This is an unfortunate breakage on main for anyone wanting to try out Loculus for their own use, and prevents us logging in ourselves (except on pathoplexus).

Here I add the old flags back as duplicates to mitigate.

### DIFF
--- a/kubernetes/loculus/templates/keycloak-deployment.yaml
+++ b/kubernetes/loculus/templates/keycloak-deployment.yaml
@@ -71,9 +71,16 @@ spec:
                 secretKeyRef:
                   name: keycloak-database
                   key: password
-            - name: KC_BOOTSTRAP_ADMIN_USERNAME
+            - name: KC_BOOTSTRAP_ADMIN_USERNAME # TODO: delete after upgrading keycloak
               value: "admin"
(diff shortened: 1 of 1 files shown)


corneliusroemer:
Made issue #3736 to upgrade keycloak, might be good to reference this issue in the code as well :)

Issue #1981, opened 2024-05-21
Set up simple mirror for `datasets` cli


corneliusroemer:
Multiple reasons:
- Reduce the load on their server (I feel like we might be “hammering” them more than ever in the past with our CI), while our load is not big, it might be enough to break them if they’re inefficient with their queries or have few resources (we're using way less than the official API limit and only using very little bandwidth but bottleneck might be elsewhere)
- Make things faster for us
- Make things less flaky for us

This might involve a single microservice across all of loculus/pathoplexus just doing the following:
1. Every say 30min download all the datasets we use
2. Put them on disk, and update if there’s a change
3. If the mirror gets a request, serve from disk

See https://loculus.slack.com/archives/C05G172HL6L/p1716312340340489

chaoran-chen:
Related issue (about a caching proxy): #3633

Issue #3633, opened 2025-02-06
Set up caching proxy for NCBI Datasets


theosanderson:
We probably hit the NCBI datasets quite hard with our preview instances. We could reduce this by using a proxy.

A command like this:

```
datasets download virus genome taxon 3048448  --gateway-url https://api.ncbi.nlm.nih.gov/datasets/v2           --no-progressbar             --filename results/ncbi_dataset.zip             --api-key None 
```

uses the gateway indicated - meaning we could host our own proxy and run

```
datasets download virus genome taxon 3048448     --gateway-url https://proxy.loculus.org        --no-progressbar             --filename results/ncbi_dataset.zip             --api-key None 
```

theosanderson:
(We have a temporary proxy from https://github.com/loculus-project/loculus/pull/3643 but it's not currently on loculus servers because our IP doesn't work - also it seems like caching would be potentially hard because there are lots of POST requests involved, Cornelius is thinking about mirroring)

chaoran-chen:
Related issue (about mirroring): #1981

theosanderson:
I will close this since we are using mirroring

PR #3754, opened 2025-02-25
feat!(ingest): allows the option to use the ncbi datasets mirror

anna-parker:
fully resolves https://github.com/loculus-project/loculus/issues/1981 and https://github.com/loculus-project/loculus/issues/3633

preview URL: https://ingest-from-mirror.loculus.org/

### Summary

Allows the option to use the s3 datasets mirror for ingest configured by @corneliusroemer in https://github.com/loculus-project/loculus/pull/3709. 

### Breaking changes
Users should add 
```
ingest: 
  mirrorBucket: null
```
to their `values.yaml` if they are not using these new changes, and otherwise update the values.yaml with the values of our s3 instance, e.g. 
```
ingest:
  mirrorBucket: "https://hel1.your-objectstorage.com/loculus-public/mirror/"
```

### PR Checklist

- [x] Tested: old state (using ncbi datasets directly or via gateway still works
- [x] Tested: new state (using mirror works)


ingest/Snakefile
```diff
@@ -388,17 +410,33 @@ rule heuristic_group_segments:
     """Group segments based on heuristic, join with previous groups if available"""
     input:
         script="scripts/heuristic_group_segments.py",
-        metadata="results/metadata_ungrouped.ndjson" if GROUPS_OVERRIDE_JSON else "results/metadata_post_prepare.ndjson",
-        sequences="results/sequences_ungrouped.ndjson" if GROUPS_OVERRIDE_JSON else "results/sequences.ndjson",
-        metadata_grouped="results/metadata_grouped.ndjson" if GROUPS_OVERRIDE_JSON else "results/metadata_post_prepare.ndjson",
-        sequences_grouped="results/sequences_grouped.ndjson" if GROUPS_OVERRIDE_JSON else "results/sequences.ndjson",
+        metadata=(
```
anna-parker, line 413: sorry this a formatting change

ingest/Snakefile
```diff
@@ -34,7 +36,9 @@ dataset_server_map = {}
 dataset_name_map = {}
 
 if SEGMENTED:
-    GROUPS_OVERRIDE_JSON = config.get("grouping_override", None)  # JSON map from group to segments' insdcAccessionFull
+    GROUPS_OVERRIDE_JSON = config.get(
```
corneliusroemer, line 39: can you separate out the formatting changes? I.e. run `ruff format` on everything first, make that pr and then make this one?

ingest/Snakefile
```diff
@@ -274,6 +293,7 @@ if ALIGN:
 
 
 if not ALIGN:
+
```
corneliusroemer, line 296: Same here, format first, then make this PR

ingest/Snakefile
corneliusroemer: there are a bunch more, would be great to format once, make PR, we'll merge, then rebase onto that

ingest/Snakefile
corneliusroemer: sorry this is snakemake, so `snakefmt` it

corneliusroemer (commented):
- would be good to separate formatting changes out first
- we can just do: `||
curl [https://hel1.your-objectstorage.com/loculus-public/mirror/{{TAXONID}}.zip](https://hel1.your-objectstorage.com/loculus-public/mirror/%7B%7BTAXONID%7D%7D.zip) -o data/ncbi_dataset.zip` in the download rule which needs no secrets

ingest/Snakefile
```diff
@@ -92,13 +93,19 @@ rule fetch_ncbi_dataset_package:
         taxon_id=TAXON_ID,
         api_key=NCBI_API_KEY,
         gateway=f"--gateway-url {NCBI_GATEWAY_URL}" if NCBI_GATEWAY_URL else "",
+        use_mirror=bool(MIRROR_BUCKET), 
+        mirror_bucket=MIRROR_BUCKET,
```
corneliusroemer, line 97: I think it's clearer to construct the URL here as one doesn't actually see the lack of space in review mode:
image
```suggestion
        mirror_url=lambda w: f"{MIRROR_BUCKET}.strip('/')}/{TAXON_ID}.zip"
```

ingest/Snakefile
```diff
@@ -92,13 +93,19 @@ rule fetch_ncbi_dataset_package:
...
     shell:
         """
-        datasets download virus genome taxon {params.taxon_id} \
+        if [[ "{params.use_mirror}" == "True" ]]; then
+            curl {params.mirror_bucket}{params.taxon_id}.zip -o {output.dataset_package}
```
corneliusroemer, line 101:
```suggestion
            curl {params.mirror_url} -o {output.dataset_package}
```

kubernetes/loculus/values.yaml
```diff
@@ -1183,7 +1183,7 @@ defaultOrganismConfig: &defaultOrganismConfig
   ingest: &ingest
     image: ghcr.io/loculus-project/ingest
     configFile: &ingestConfigFile
-      taxon_id: 186540
+      taxon_id: 3052460
```
corneliusroemer, line 1186: Why are you changing this? I picked a different (more basal) taxon id for ebola sudan on purpose!

https://www.ncbi.nlm.nih.gov/taxonomy/?term=186540

anna-parker: We use the other taxon id for pathoplexus as well?

anna-parker: you mirrored the taxon id we also use for pathoplexus

corneliusroemer: Oh you're right, the one we used here is actually more distal, not more basal. We should use 3052460, thanks, my bad!

anna-parker: https://github.com/pathoplexus/pathoplexus/blob/main/loculus_values/values.yaml#L1329

ingest/Snakefile
corneliusroemer: Stripping means it doesn't need to end with trailing `/`, support both with and without trailing `/`

kubernetes/loculus/values.yaml
corneliusroemer: Thanks! :)

kubernetes/loculus/values.schema.json
```diff
@@ -1034,6 +1034,9 @@
       "properties": {
         "ncbiGatewayUrl": {
           "type": ["string", "null"]
+        },
+        "mirrorBucket": {
+          "type": ["string", "null"]
```
corneliusroemer, line 1039: Maybe add a comment that _any_ non-empty string enables the feature, even `"False"` does :)

kubernetes/loculus/values.schema.json
```diff
@@ -1034,6 +1034,10 @@
       "properties": {
         "ncbiGatewayUrl": {
           "type": ["string", "null"]
+        },
+        "mirrorBucket": {
+          "type": ["string", "null"],
+          "description": "Specifies the location of the public mirrored NCBI datasets bucket, where data is stored in the format {TAXON_ID}.zip. If provided with any non-null string, this setting takes priority over ncbiGatewayUrl."
```
corneliusroemer, line 1040: Nice!

corneliusroemer (approved):
Looks great, thanks!

PR #3759, opened 2025-02-25
fix(backend): Only use the string up to the first space as the sampleID

fhennig:
resolves #

preview URL:

### PR Checklist


backend/src/main/kotlin/org/loculus/backend/utils/FastaReader.kt
theosanderson, line 1: I realised this probably means we need to enforce that the `submissionId` doesn't contain any spaces (if we don't atm)

fhennig: I have added a check to the metadata file parser - that should technically be sufficient I think?

fhennig:
The core fix was made, I also added a check to the metadata file parsing, as well as docs and tests. IMO this is ready, but possibly we might need more docs or tests? Difficult to assess from my end, happy about suggestions!

backend/src/main/kotlin/org/loculus/backend/utils/FastaReader.kt
```diff
@@ -41,7 +41,7 @@ class FastaReader(inputStream: InputStream) :
                 if (sampleName != null) {
                     break
                 }
-                sampleName = nextLine!!.substring(1)
+                sampleName = nextLine!!.substring(1).substringBefore(" ")
```
rneher, line 44: should this be generalized to all whitespace?

backend/src/main/kotlin/org/loculus/backend/utils/FastaReader.kt
rneher: other characters than whitespace is probably not a primary concern for us since these ids are only used to match metadata and sequence. But if we started to include the submitter's ID into fasta headers, this would require more care.

docs/src/content/docs/reference/fasta-format.md
theosanderson, line 1: Slightly uncertain about this. I dunno if we want to become an encyclopedia. This is a very core bioinformatics concept which to me is akin to explaining what DNA is. But it's true that at the end you explain something that not everyone may know.

theosanderson: I think I would make this less about defining FASTA format and more about the format Loculus expects its FASTA files to have: so explaining that we expect them to have the id and then a trailing segment identifier for a multi-segmented FASTA.

docs/src/content/docs/reference/fasta-format.md
```diff
@@ -0,0 +1,30 @@
+---
+title: FASTA format
+---
+
+[_FASTA_](https://en.wikipedia.org/wiki/FASTA_format) is a file format to store sequence data.
+The FASTA format does not have an official definition.
```
theosanderson, line 6: IMO this isn't useful and is if anything confusing - there are a lot of aspects of FASTA that are fully agreed.

docs/src/content/docs/reference/fasta-format.md
```diff
@@ -0,0 +1,30 @@
...
+>test_NIHPAK-19_M
+GTGGATTGAGCATCTTAATTGCAGCATACTTGTCAACATCATGCATATATCATTGATGTATGCAGTTTTCTGCTTGCAGCTGTGCGGTCTAGGGAAAACTAACGGACTACACAATGGGACTGAACACAATAAGACACACGTTATGACAACGCCTGATGACAGTCGGAAGAGCTGTGAAATAGACAGTATCCCAGTTAAGTGTAGGCAGGGTTATTGCCTAAAAATCACTCAGGAAGGGAGGGGCCATGTGAAATTATCTAGAGGCTCAGAAGTTGTCTTGGATGTATGTGACTCAAGCTGT
+>test_NIHPAK-19_S
+GTGTTCTCTTGAGTGTTGGCAAAATGGAAAACAAAATCGAGGTGAACAACAAAGATGAGATGAACAAATGGTTTGAGGAGTTCAAGAAAGGAAATGGACTTGTGGACACTTTCACAAACTCNTATTCCTTTTGTGAAAGCGTNCCAAATCTGGACAGNTTTGTNAATGGAGAAAAGACATAGGCTTCCGTGTCAATGCAAATACGGCAGCTTTGAGTAACAAAGTCCTAGCAGAGTACAAAGTTCCTGGCGAGATTGTAATGTCTGTCAAAGAGATGTTGTCAGATATGATTAGAAGNAGGAACCTGATTCTCAACAGA
+```
+
+Every segment starts with a `>` character, and then an ID. After the ID can also follow an optional description, seperated by a space. This means the ID cannot contain spaces. Example:
```
theosanderson, line 19:
```suggestion
Every segment starts with a `>` character, and then an ID. After the ID can also follow an optional description, separated by a space. This means the ID cannot contain spaces. Example:
```

theosanderson (commented):
Thanks a lot for doing this!! Looks great!

docs/src/content/docs/reference/fasta-format.md
theosanderson: Not blocking

fhennig: Thanks! definitely agree that we shouldn't document the obvious. I have rewritten it to try and focus it more on the loculus specifics: https://github.com/loculus-project/loculus/pull/3759/commits/2259db011f42d3540ec457908552be823251c311

I'm also a bit unsure about what the best way is to document this, or if anything should be documented at all! If FASTA is actually well known etc, then we wouldn't have to document that the ID is only up to the first space. 

But you already said it's not blocking!

backend/src/main/kotlin/org/loculus/backend/utils/MetadataEntry.kt
```diff
@@ -32,6 +32,12 @@ fun metadataEntryStreamAsSequence(metadataInputStream: InputStream): Sequence<Me
             )
         }
 
+        if (submissionId.any { it.isWhitespace() }) {
+            throw UnprocessableEntityException(
+                "A value for $HEADER_TO_CONNECT_METADATA_AND_SEQUENCES contains whitespace: $record",
```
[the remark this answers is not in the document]
fhennig: You said it isn't blocking so I'm gonna leave it as is.

backend/src/main/kotlin/org/loculus/backend/utils/FastaReader.kt
fhennig: done [1935323](https://github.com/loculus-project/loculus/pull/3759/commits/193532345b25ae81c76e4ccb0ddd99bb42070ee8)

backend/src/main/kotlin/org/loculus/backend/utils/FastaReader.kt
fhennig: I'm marking this as 'resolved' - As you said for now it should be fine the way it is.

backend/src/main/kotlin/org/loculus/backend/utils/MetadataEntry.kt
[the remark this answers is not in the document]
theosanderson: 👍 made an issue to track

theosanderson (approved):
Great

PR #3761, opened 2025-02-25
feat!(website): allow selecting fields to download in metadata and default to a shorter selection

### LINKED ISSUE #850: Download panel: default metadata fields
The download panel (initially implemented in #757) currently downloads the whole set of metadata which might be very large (if we generate a lot of QC scores). Instead, it should be possible to define one or multiple default lists that people can choose.

### LINKED ISSUE #851: Download panel: metadata field selection
The user should be able to choose the metadata fields when they download the metadata file through the web interface.

### LINKED ISSUE #3695: Allow downloading a subset of fields
We should allow users to download a subset of the fields in the website metadata download. The important use case for Pathoplexus would be to restrict the number of fields in a default download to exclude a lot of the fields that are almost always blank.

theosanderson:
resolves #850, resolves #851, resolves #3695

preview URL: https://select-fields.loculus.org/ebola-sudan/search?

### Summary

Adds the capacity to select fields to download and defaults to a configured set of fields rather than to the full list.

The form to select these is quite nice. I haven't tried to merge it with the form that's used for selecting fields for the search fields and columns (or I did briefly but didn't succeed), but we can make an issue to do that in the future.

### Breaking changes

You need to select which fields should be selected by default by adding the `includeInDownloadsByDefault: true` property in the values yaml. We should decide what this should be for Pathoplexus and make a PR before we merge this.

I made a PR: https://github.com/pathoplexus/pathoplexus/pull/404

### Testing

I added integration tests to check the right number of columns are present. To achieve this I added a fixture. I would request relatively light touch reviewing on the integration tests.

### Screenshot

image

image

### PR Checklist

I tested that the downloads looked good. There are also some unit tests and an integration test.


theosanderson:
Chaoran pointed out the schema hasn't been updated - I remembered seeing something but clearly it didn't make it - we should figure out why CI didn't catch this and should ensure it's there

kubernetes/loculus/values.yaml
theosanderson, line 1: Chaoran pointed out the schema hasn't been updated - I remembered seeing something but clearly it didn't make it - we should figure out why CI didn't catch this and should ensure it's there

theosanderson: I have now added this - I will raise an issue to make the schema checking stricter

theosanderson: https://github.com/loculus-project/loculus/issues/3770

website/src/components/SearchPage/DownloadDialog/FieldSelector/FieldSelectorModal.tsx
chaoran-chen, line 1: Could you maybe also add two buttons for "Select all" and "Select none"? I think that would be quite important to have so that people have an easy way to download the whole dataset.

website/src/components/SearchPage/DownloadDialog/FieldSelector/FieldSelectorModal.tsx
```diff
@@ -0,0 +1,141 @@
...
+        return a.localeCompare(b);
+    });
+
+    return (
+        <BaseDialog title='Select Fields to Download' isOpen={isOpen} onClose={onClose}>
+            <div className='mt-4 max-h-[60vh] overflow-y-auto p-2'>
+                {sortedHeaders.map((header) => (
+                    <div key={header} className='mb-6'>
+                        <h3 className='font-medium text-lg mb-2 text-gray-700'>{header}</h3>
+                        <div className='grid grid-cols-1 md:grid-cols-2 gap-x-4 gap-y-2'>
```
chaoran-chen, line 102: Should we allow more columns on wider screens? On my screen, I can imagine 3-4 columns.

image

chaoran-chen: Alternatively, it would make sense to set a max. width for the modal.

theosanderson: 👍 I've put a hacky max width in - if you prefer to do it in another way I'd prefer we make it a follow up issue

website/src/components/SearchPage/DownloadDialog/FieldSelector/FieldSelectorModal.tsx
theosanderson: Done

fhennig:
I really like this feature! it's also cool that it becomes a single column when the screen is narrow.

I think the scrollbar could be permanently visible IMO, otherwise its not obvious that there are more fields further down.

theosanderson:
> I think the scrollbar could be permanently visible IMO, otherwise its not obvious that there are more fields further down.

I agree this is an issue, and we could try to think about something to do about it in a follow up to make this clearer, but I think it's primarily a general mac usability issue - macs have prioritised clean UI over indicating this and I don't think it's easy to fix

integration-tests/tests/specs/features/download.spec.ts
chaoran-chen, line 1: The download doesn't work for me if I filter for a lot of accession versions. I guess that something went wrong when constructing the POST request?

(It would be nice a test for this especially since it is easy to forget to test that case.)

theosanderson: I've added the test - thanks for spotting this. LAPIS behaviour here is a bit confusing: https://github.com/GenSpectrum/LAPIS/issues/1089

website/src/components/SearchPage/DownloadDialog/FieldSelector/FieldSelectorModal.tsx
```diff
@@ -0,0 +1,171 @@
+import { useState, type FC } from 'react';
```
fhennig, line 1: When selecting all fields, the option to copy the download URL is gone, because the URL probably got too long.

image

This is a regression, because it used to be possible.

It could still be implemented, by omitting the `fields` from the URL if _all_ fields are selected.

I can't say whether this is something to block on or not. If it isn't considered a blocker, could you kindly create a ticket for it and link it here?

website/src/components/SearchPage/DownloadDialog/FieldSelector/FieldSelectorModal.tsx
```diff
@@ -0,0 +1,171 @@
...
+                ))}
+
+                <div className='mt-6 flex justify-end'>
+                    <button
+                        type='button'
+                        className='inline-flex justify-center rounded-md border border-transparent bg-primary-600 px-4 py-2 text-sm font-medium text-white shadow-sm hover:bg-primary-700 focus:outline-none focus:ring-2 focus:ring-primary-500 focus:ring-offset-2'
+                        onClick={onClose}
+                    >
+                        Done
+                    </button>
```
fhennig, line 151: I know it's a small thing, but it'd be great if we ca reuse the classes from the DownloadButton (or another button). There is a `btn` class as well that covers probably most of this.

```
        <div className='flex items-center'>
            <div className={message && 'tooltip tooltip-open tooltip-right tooltip-info'} data-tip={message}>
                <a
                    className={`btn loculusColor ${disabled || handleClick === undefined ? 'btn-disabled' : ''} text-white`}
                    href={downloadUrl}
                    onClick={handleClick}
                    data-testid='start-download'
                >
                    Download
                </a>
            </div>
            {isGetRequest && !disabled && <CopyUrlButton url={downloadUrl} />}
        </div>
```

I can't directly suggest something because I'm not sure what exactly is needed, but I'd like to try and keep the UI uniform if possible.

website/src/components/SearchPage/DownloadDialog/FieldSelector/FieldSelectorModal.tsx
```diff
@@ -0,0 +1,171 @@
...
+    // Sort headers alphabetically, but keep "Other" at the end
+    const sortedHeaders = Object.keys(fieldsByHeader).sort((a, b) => {
+        if (a === 'Other') return 1;
+        if (b === 'Other') return -1;
+        return a.localeCompare(b);
+    });
+
+    return (
+        <BaseDialog title='Select Fields to Download' isOpen={isOpen} onClose={onClose} fullWidth={false}>
+            <div className='min-w-[950px]'></div>
```
fhennig, line 80: Hmm, this looks like it an unintentional left-over.

fhennig (changes requested):
Hey, thanks for implementing that! Quite useful.

I just have a few small remarks.

website/src/components/SearchPage/DownloadDialog/DownloadUrlGenerator.spec.ts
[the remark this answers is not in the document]
fhennig: I think there's no harm in keeping them! 👍

website/src/components/SearchPage/DownloadDialog/FieldSelector/FieldSelectorModal.tsx
theosanderson: I think this was actually intentional but let's see its effect after https://github.com/loculus-project/loculus/pull/3869 which I think may have changed the underlying situation

integration-tests/tests/specs/features/download.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+test('Download metadata and check number of cols', async ({ pageWithReleasedSequence: page }) => {
+    test.setTimeout(120000);
+    const searchPage = new SearchPage(page);
+
+    await page.goto('/');
+    await page.getByRole('link', { name: 'Crimean-Congo Hemorrhagic Fever Virus' }).click();
+
+    const loculusId = await searchPage.waitForLoculusId();
+    expect(loculusId).toBeTruthy();
+    console.log(`Found loculus ID: ${loculusId}`);
```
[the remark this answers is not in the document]
theosanderson: You don't get any columns at all unless there is data to download. I'll remove the logging which isn't needed.

website/src/components/SearchPage/DownloadDialog/FieldSelector/FieldSelectorModal.tsx
theosanderson: Thanks for flagging this - which I hadn't considered. IMO it's not something to block on because downloading all fields will be rare, and because generally this behaviour is better than the current behaviour which is super overwhelming. But it's a good suggestion https://github.com/loculus-project/loculus/issues/3870

website/src/components/SearchPage/DownloadDialog/FieldSelector/FieldSelectorModal.tsx
theosanderson: Done - I have somewhat mixed feelings on these things but agree that in this case it's likely an improvement!

website/src/components/SearchPage/DownloadDialog/FieldSelector/FieldSelectorModal.tsx
theosanderson: The purpose of this was to ensure this fully obscures the modal below to improve aesthetics. Other changes mean it now shorter than the modal below, which I'd prefer to tackle later if at all, but I think we can keep this to ensure this overlying modal is wider.

fhennig (approved):

PR #3762, opened 2025-02-26
feat(website): store previewed sequence ID in URL

### LINKED ISSUE #2025: Persist sequence opened in modal in URL
Related to #2024, I think that it would be very useful to persist the sequence entry ID in the URL when it is opened in a modal on the browse page. This would make it easier for people to send a link to an entry (now, you have to first open it in a new tab and not every user will know that). Further, it will re-open the same entry if the page is reloaded (which one might do on purpose but also, for example, upon restart of a browser).

theosanderson:
resolves #2025

preview URL: http://previewurl.loculus.org

### Summary
Stores the ID of the sequence currently previewed in a modal on the search page in the URL

1. Modified both the previewedSeqId and previewHalfScreen state
   to initialize from URL parameters
  2. Created callback functions to update both local state and
  URL state
  3. Added a useEffect hook to keep the local state in sync with
  URL parameters when they change
  4. Used the same pattern as the page parameter:
    - For selected sequence: selectedSeq=ABC123 in URL
    - For half screen mode: halfScreen=true in URL
    - Parameters are completely removed from URL when not active

  Now when users:
  1. Select a sequence for preview, it will be stored in the URL
  2. Change to half-screen view, that preference will be stored
  in the URL
  3. Reload the page or share the URL, both the selected sequence
   and view mode will be preserved


integration-tests/tests/specs/features/sequence-preview-url.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+    
+    await searchPage.clickOnSequence(0);
+    await expect(page.locator('[data-testid="sequence-preview-modal"]')).toBeVisible();
+    
+    const urlParams = await searchPage.getUrlParams();
+    const selectedSeqId = urlParams.get('selectedSeq');
+    
+    await (await searchPage.toggleHalfScreenButton()).click();
+    
+    await (await searchPage.closePreviewButton()).click();
```
fhennig, line 73: What is the effect of this part of the code on the test? It looks like it isn't affecting the part that comes afterwards.

fhennig (approved):
Hey, really nice! Also good tests, thanks!

PR #3775, opened 2025-02-28
feat(website): Append `&` to URLs that end in `*`.

### LINKED ISSUE #3518: Avoid ending URL with * but append &
In GenSpectrum / CoV-Spectrum, we made the experience that it is better to not end an URL with a `*` because some systems would not consider it still be part of the URL, see: https://github.com/GenSpectrum/dashboards/issues/454

Should we append a `&` to the URLs as well?

fhennig:
resolves #3518 

preview URL: https://avoid-star.loculus.org/

### PR Checklist


chaoran-chen (approved):
Looks great and nice test!

PR #3777, opened 2025-02-28
feat(website): active filter view above Search UI table

### LINKED ISSUE #2062: Show a list of active filters above the search result table
> Maybe we could show above the result table a list of active filters which the user can also use to easily remove a filter? Could be styled like this:

_Originally posted by @chaoran-chen in https://github.com/loculus-project/loculus/issues/82#issuecomment-1650106770_

fhennig:
resolves #2062 

preview URL: https://filters-in-search-UI.loculus.org

### Summary
- Show _active filters_ as badges above the table, with an 'x' button to remove the filter
- Small bugfix: `ins_4:A:G` was a recognized as a valid insertion, but caused a backend error. This has been fixed.

N.B.:
- when making a selection, that isn't shown as a filter - but on the download dialog that will then override the filters (just noted here as something that was explicitly considered. See video for demonstration).

#### automated tests
- Added a new test for the `ActiveFilters` component that checks that the button is rendered and calls the handler correctly if clicked
- Add integration test that checks that if a filter is set and the 'X' is clicked, the filter is removed.
- Add unit tests for mutation related functions.
- TODO: integration test that sets a value and checks that the active filter appears.

#### manual testing
- I have tested the accession, mutation and country fields: setting a value and removing them again.
- I have checked that when multiple accessions are searched for, that the label is eventually truncated.
- I have checked that if sequences are selected manually, the filter is still shown, but in the download dialog it just says "x sequences selected".
- I have checked that the active filters in the download dialog do not have the 'x'

### Screenshot

https://github.com/user-attachments/assets/0d6e5a47-d5e7-473c-82a3-c5336c81f3b5

### PR Checklist


fhennig:
TODO:
- try putting the filters above the download button
- take care that there's no regression on mobile
- limit max length of filter (i.e. filtering for lots of accessions)

theosanderson:
I like the new positioning- thanks!!

fhennig:
TODO: I found that the accession field content doesn't reset - looking into it.

fhennig:
Ok, so I think this is now almost done, Theo and I just want to add one more integration test that goes through the whole search (setting a value "manually").

website/src/components/common/ActiveFilters.tsx
```diff
@@ -4,25 +4,41 @@ import type { SequenceFilter } from '../SearchPage/DownloadDialog/SequenceFilter
...
+                >
+                    <span className='text-primary-900 font-light pr-1'>{label}:</span>
+                    <span className='text-primary-900 font-semibold'>{value}</span>
+                    {showXButton ? (
+                        <button
+                            aria-label='remove filter'
+                            className='inline ml-2 mt-0.5 pr-2'
+                            onClick={() => removeFilter(key)}
+                        >
+                            <svg className='w-3 h-4 text-primary-600' fill='currentColor' viewBox='0 0 20 20'>
```
theosanderson, line 29: We should refactor this sometime but I know it's in mine, let's leave it for now 👍

theosanderson (approved):
LGTM apart from the test - thanks for the refactoring!

PR #3799, opened 2025-03-05
fix(website): fasta was incorrectly generated with empty segments

### LINKED ISSUE #3798: Single sample Form submission: Error when only submitting 1 segment
When submitting via the form if I only upload one segment I get an error - this is a very likely scenario and should not cause an error.

Image

fhennig:
resolves #3798 

preview URL:

### Summary

the frontend code for the edit and single-submit form generated FASTAs that had segment headers with no data.
This PR adds a test for this case and fixes it, so empty segments aren't added to the FASTA anymore.

### Screenshot

n/A

### PR Checklist

- [x] Any manual testing that has been done is documented (i.e. what exactly was tested?) - @anna-parker: submitted a CCHF sequence with only the L segment and the submission was successful can be seen here: https://single-segment-bug.loculus.org/seq/LOC_000PERQ.1

### DIFF
--- a/website/src/components/Edit/InputForm.spec.tsx
+++ b/website/src/components/Edit/InputForm.spec.tsx
@@ -1,0 +1,24 @@
+import { describe, expect, test } from 'vitest';
+
+import { EditableSequences } from './InputForm';
+
+describe('InputForm', () => {
+    test('Empty editable sequences produces `undefined`', () => {
+        const emptyEditableSequences = EditableSequences.empty();
+        expect(emptyEditableSequences.getSequenceFasta('subId')).toBeUndefined();
+    });
+
+    test('Empty editable sequences (with names only) produces `undefined`', () => {
+        const emptyEditableSequences = EditableSequences.fromSequenceNames(['foo', 'bar']);
+        expect(emptyEditableSequences.getSequenceFasta('subId')).toBeUndefined();
+    });
+
+    test('GIVEN only a single segment has data THEN the fasta has only that segment', async () => {
+        let editableSequences = EditableSequences.fromSequenceNames(['foo', 'bar']);
+        editableSequences = editableSequences.update({ key: 'foo', value: 'ATCG' });
+        const fasta = editableSequences.getSequenceFasta('subId');
+        expect(fasta).not.toBeUndefined();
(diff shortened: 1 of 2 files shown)


website/src/components/Edit/InputForm.tsx
```diff
@@ -210,7 +210,16 @@ export class EditableSequences {
...
-                : sequences.map((sequence) => `>${submissionId}_${sequence.key}\n${sequence.value}`).join('\n');
+                : sequences
+                      .map((sequence) => {
+                          if (sequence.value.trim().length > 0) {
+                              return `>${submissionId}_${sequence.key}\n${sequence.value}`;
+                          } else {
+                              return null;
+                          }
+                      })
+                      .filter(Boolean)
```
anna-parker, line 221: hmm... doesn't filter(Boolean) remove `null` from the array?

anna-parker: ah ok this does work on a preview - so perhaps the backend sets them to `null`?

website/src/components/Edit/InputForm.spec.tsx
```diff
@@ -0,0 +1,24 @@
+import { describe, expect, test } from 'vitest';
+
+import { EditableSequences } from './InputForm';
+
+describe('InputForm', () => {
+    test('Empty editable sequences produces `undefined`', () => {
+        const emptyEditableSequences = EditableSequences.empty();
+        expect(emptyEditableSequences.getSequenceFasta('subId')).toBeUndefined();
```
anna-parker, line 8: Maybe outside the scope of this PR but I think it would be good to rename the `getSequenceFasta` function - in my understanding it produces fasta objects with the submissionId it is given - so `createFastaWithId` would be a better name.

Also to verify my understanding each editableSequences object only contain the contents of one sample? because this function will give all sequences in the editable sequence object the same submissionId - which I don't think we want? It might be good to rename or document that as well for maintainability.

website/src/components/Edit/InputForm.tsx
anna-parker: ah I see - it should be removed because this produces the fasta - please ignore this

PR #3801, opened 2025-03-05
feat(website): set global when searchfullui hydrated

theosanderson:
We need this for testing


fhennig (approved):
tested it, works. thanks!

PR #3804, opened 2025-03-05
disable-email

theosanderson:
resolves #

preview URL:

### PR Checklist

### DIFF
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -1611,7 +1611,7 @@ auth:
     replyTo: "noreply@loculus.org"
     from: "noreply@loculus.org"
     envelopeFrom: "noreply@loculus.org"
-  verifyEmail: true
+  verifyEmail: false
   resetPasswordAllowed: true
   registrationAllowed: true
   identityProviders:


kubernetes/loculus/values_preview_server.yaml
```diff
@@ -42,6 +42,7 @@ sequenceFlagging:
     repository: curation_reports
     issueTemplate: sequence-metadata-issue.md
 auth:
+  verifyEmail: false
```
theosanderson, line 45: hmm this is not working in the preview

fhennig: image


Do you also get this when trying to register?

theosanderson: Right sorry - so when I posted this there were two conflicting things specified: https://github.com/loculus-project/loculus/pull/3804/commits/11bf4405cb3179bdce99f1b9ecd63ffc9adbb41b and I was just getting a normal verification email. Then after I fixed that you got that error and I got the same when I tried - I think that was because we ran out of resources and the keycloak DB couldn't run. Everything is fine now.

fhennig (approved):
Works!

I think this is really great, I really like running the integration tests against previews. Thanks!

Issue #3631, opened 2025-02-05
Wrong interaction between data use terms search filter and data use terms download settings


corneliusroemer:
1. Select "restricted" data use terms as search filter
2. Select "only download open data" in download modal
3. Observe result. Actual: All open data is downloaded, in fact none should be downloaded, as there is no data that's "restricted" AND "Open"

This is happening on staging right now which uses latest loculus, which includes the PRs that were addressing:
- #3626
- #3624 

The PRs were:
- #3625 
- #3628

theosanderson:
Yes - I was aware of that behaviour - IMO it's not super unreasonable (you could argue you have selected something in the download box that supercedes your choice in the search box). But I also get your point. I think the only way we could deal with this would be to specifically detect the contradiction in the frontend and e.g. grey out the Download box with a message.

theosanderson:
(I do agree that's a better approach)

corneliusroemer:
>you could argue you have selected something in the download box that supersedes your choice in the search box

The user hasn't selected anything, it's the default choice :)

corneliusroemer:
Full matrix of current implementation:

Search: None, Download: Open: 
Search: None, Download: Open: 
Search: None, Download: Open: 
Search: None, Download: Open: 
Search: None, Download: Open:

fhennig:
I just ran into this as well and as a user it's definitely surprising that the dialog overrides my filter, but the filter is still displayed.

Maybe if the filter is selected, the radiobutton can be pre-set to "restricted"?

theosanderson:
No – at least not without other changes. Users have to explicitly make clear that they have opted in to downloading restricted sequences.

fhennig:
Ok! Then I agree, we should do what you said:

> I think the only way we could deal with this would be to specifically detect the contradiction in the frontend and e.g. grey out the Download box with a message.

theosanderson:
I was proposing that we potentially tackle this alongside the changes discussed here: https://loculus.slack.com/archives/C05F8GKR06R/p1739534268184379

PR #3809, opened 2025-03-05
feat(website): Make it possible to search revoked and non-latest sequences

### LINKED ISSUE #2816: Revoked and non-latest sequences are entirely inaccessible through the Search UI
We should make it possible to add a search field that lets you enable the display of revoked sequences and non-latest sequences

fhennig:
resolves #2816 

preview URL:

### PR Checklist

### DIFF
--- a/kubernetes/loculus/templates/_common-metadata.tpl
+++ b/kubernetes/loculus/templates/_common-metadata.tpl
@@ -25,6 +25,7 @@ fields:
     type: string
     header: Submission details
     enableSubstringSearch: true
+    {{/* This field maybe should be searchable */}}
   - name: isRevocation
     type: boolean
     notSearchable: true
@@ -101,6 +102,7 @@ fields:
       url: "__value__"
   {{- end}}
   {{- end}}
+    {{/* This field maybe should be searchable */}}
   - name: versionStatus
     type: string
     notSearchable: true
--- a/website/src/utils/search.ts
+++ b/website/src/utils/search.ts
@@ -184,6 +184,7 @@ export const getFieldValuesFromQuery = (
     hiddenFieldValues: FieldValues,
     schema: Schema,
 ): FieldValues => {
+    // here the hidden field values are there as the default - maybe if we make them settable, they will just override?
     const values: FieldValues = { ...hiddenFieldValues };
     const expandedSchema = getMetadataSchemaWithExpandedRanges(schema.metadata);
     for (const field of expandedSchema) {
--- a/website/src/components/SearchPage/SearchForm.tsx
+++ b/website/src/components/SearchPage/SearchForm.tsx
@@ -170,6 +170,8 @@ const SearchField = ({ field, lapisUrl, fieldValues, setSomeFieldValues, lapisSe
         }
     }
 
+    // TODO here we need to add support for boolean fields
+
     switch (field.type) {
         case 'date':
             return (
--- a/website/src/pages/[organism]/search/index.astro
+++ b/website/src/pages/[organism]/search/index.astro
@@ -10,6 +10,7 @@ import { getMyGroups } from '../../../utils/getMyGroups';
 import { getReferenceGenomesSequenceNames } from '../../../utils/search';
 import { performLapisSearchQueries } from '../../../utils/serversideSearch';
 
+// These values are defaults that we already set
 const hiddenFieldValues = {
     [VERSION_STATUS_FIELD]: versionStatuses.latestVersion,
     [IS_REVOCATION_FIELD]: 'false',


fhennig:
started to take notes of where things would need to be changed

website/src/utils/search.ts
```diff
@@ -56,7 +56,7 @@ const getFieldOrColumnVisibilitiesFromQuery = (
 ): Map<string, boolean> => {
     const visibilities = new Map<string, boolean>();
     schema.metadata.forEach((field) => {
-        if (field.hideOnSequenceDetailsPage === true || !visibilitySelectableAccessor(field)) {
```
fhennig, line 59: This change was necessary to be able to show the "is revocation" and "version status" field in the visibility selector - but now a bunch of other things are also visible. I think this isn't an issue, but I'll just have to add `notSearchable: true` to a bunch of other fields.

I also noticed for example this:

```yaml
  - name: submittedDate
    type: string
    hideOnSequenceDetailsPage: true
    generateIndex: true
    autocomplete: true
    displayName: Date submitted (exact)
```

This is currently live that way. `autocomplete` and `generateIndex` imply that someone wanted to have these searchable - but because of `hideOnSequenceDetailsPage` it's actually not searchable ... curious.

fhennig: some other fields actually _do_ have `notSearchable` and `hideOnSequenceDetailsPage` set explicitly. I am now assuming that the changes are made are actually fine without setting more `notSearchable` fields. Maybe someone with more knowledge about the data specifics can have a look and verify which fields should be searchable but not visible etc.

theosanderson: > someone wanted to have these searchable 

I wouldn't overthink this - especially in the Loculus config - there may be lots of accidental config

fhennig: Alright! Then I think It's fine the way it is. Thanks!

website/src/utils/search.ts
```diff
@@ -184,6 +184,7 @@ export const getFieldValuesFromQuery = (
     hiddenFieldValues: FieldValues,
     schema: Schema,
 ): FieldValues => {
+    // here the hidden field values are there as the default - maybe if we make them settable, they will just override?
```
theosanderson, line 187:
```suggestion
```

maybe?

website/src/components/SearchPage/DownloadDialog/SequenceFilters.tsx
```diff
@@ -39,6 +39,10 @@ export class FieldFilter implements SequenceFilter {
     private readonly hiddenFieldValues: FieldValues;
     private readonly schema: ConsolidatedMetadataFilters;
 
+    /**
+     * @param lapisSearchParamters The filters that are set
```
theosanderson, line 43:
```suggestion
     * @param lapisSearchParameters The filters that are set
```

website/src/components/SearchPage/DownloadDialog/SequenceFilters.tsx
```diff
@@ -39,6 +39,10 @@ export class FieldFilter implements SequenceFilter {
...
+     * @param hiddenFieldValues Key-value pairs that - if those are found - they will not be rendered.
```
theosanderson, line 44:
```suggestion
     * @param hiddenFieldValues Key-value pairs that - if even if they are found in the parameters - will not be rendered.
```

If this is what is meant

kubernetes/loculus/templates/_common-metadata.tpl
```diff
@@ -26,8 +26,9 @@ fields:
     header: Submission details
     enableSubstringSearch: true
   - name: isRevocation
+    displayName: Is revocation
```
theosanderson, line 29: This is a weird name to non-programmers but I think it's OK for now

website/src/components/SearchPage/SearchFullUI.tsx
theosanderson, line 1: I set up a revocation and searched for it:
image

image

If I download I get the non-revocations:
```
https://lapis-revocs-in-search.loculus.org/ebola-sudan/sample/details?downloadAsFile=true&downloadFileBasename=ebola-sudan_metadata_2025-03-13T1442&versionStatus=LATEST_VERSION&isRevocation=false&dataUseTerms=OPEN&dataFormat=tsv
```

theosanderson (changes requested):
Nice generally - but the download issue I've flagged is blocking

website/src/utils/search.ts
fhennig: Oh I think this was a note I left to myself 😅  I'll delete it.

website/src/components/SearchPage/DownloadDialog/SequenceFilters.tsx
fhennig: https://github.com/loculus-project/loculus/pull/3809/commits/32e73d368b190d9d5e5f23f362030a28963d8e88

website/src/utils/search.ts
fhennig: https://github.com/loculus-project/loculus/pull/3809/commits/32e73d368b190d9d5e5f23f362030a28963d8e88

kubernetes/loculus/templates/_common-metadata.tpl
fhennig: agreed but the only other boolean field I found was isLabHost (or something similar) and that was also like that.
Good of you to point out! I'll resolve for now.

website/src/components/SearchPage/SearchFullUI.tsx
chaoran-chen: This problem is similar to #3631, isn't it? The current behavior with the DUTs is that if a user filters for RESTRICTED sequences and goes to the download panel, by default, "only download open data" is selected, so nothing is downloaded. If the user selects "include restricted data", they would get only the RESTRICTED data as they wished. As we discussed in #3631, this behavior is absolutely **not ideal** but if, just for now, we are willing to accept this same behavior, a rather consistent solution would be:

- If the user filters for `is revocation = true` and "only download latest version" is selected, nothing gets returned.
- If the user filters for `is revocation = true` and "include older versions and revoked sequences" is selected, they get exactly what they would also find in the browse table.

This introduces the same confusing user experience as with the DUTs but would be consistent with the current behavior and if a user reads the questions/options in the download panel correctly, it also somewhat makes sense. We can then adapt the behavior when we also do #3631.

fhennig: but the `versionStatus` can also be filtered for, and that then adds further confusion/incompatible states.

chaoran-chen: Sorry, I'm not sure that I understand which further confusion/incompatible states you mean. In "Yes, include older versions and revoked sequences", "older versions" refers to versionStatus and "revoked sequences" refers to isRevocation, doesn't it?

fhennig: Update: In the stand-up today we decided to remove this selection from the download dialog, and instead leave it for users to apply the filters to include everything.

For completeness, the question was: What if a user selects to to not include revocations but _do_ include older versions. What about the radio-select then?

fhennig: I have removed the panel - also included a screenshot.

integration-tests/tests/specs/features/override-hidden-fields.ts
```diff
@@ -0,0 +1,110 @@
...
+
+    search = new SearchPage(page);
+    await search.ebolaSudan();
+
+    await search.select('Collection country', 'Uganda');
+    await search.enableSearchFields('Author affiliations');
+    await search.fill('Author affiliations', uuid);
+    await search.expectSequenceCount(1);
+    const revokedId = await page
+        .getByRole('link', { name: new RegExp('LOC_[A-Z0-9]{2,20}\.1') })
```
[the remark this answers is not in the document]
fhennig: It's sorted.

integration-tests/tests/specs/features/search.spec.ts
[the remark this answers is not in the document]
fhennig: I added a test

integration-tests/tests/test-1.spec.ts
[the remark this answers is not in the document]
fhennig: oops! removed.

website/src/types/config.ts
```diff
@@ -168,6 +168,9 @@ export const websiteConfig = z.object({
 });
 export type WebsiteConfig = z.infer<typeof websiteConfig>;
 
-export type FieldValues = Record<string, string | number | null>;
+export type FieldValues = {
+    mutation?: string;
+    acession?: string;
```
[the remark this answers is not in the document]
fhennig: https://github.com/loculus-project/loculus/pull/3809/commits/8f860b70984ae16bdedd4294c51351ee05d3a7c4

PR #3813, opened 2025-03-06
feat(website): Improve sequence details dropdown a little

theosanderson:
Removes somewhat confusing behaviour of a dropdown that is actually a button, with a more conventional dropdown that displays the current selected version.

image

preview URL: http://dropdown.loculus.org


fhennig:
https://github.com/user-attachments/assets/12258454-4d9a-46b1-a6c1-9f8404643fdf

fhennig (approved):
Thanks! I like it, good idea to move the 'All versions' into the dropdown.

PR #3833, opened 2025-03-11
refactor(deployment): make LAPIS its own deployment and allow number of replicas to be set

theosanderson:
resolves #

preview URL: https://separate-lapis.loculus.org/

Allows spinning up multiple replicas of LAPIS.

Note that this changes behaviour on roll-outs. Now if in a new rollout LAPIS cannot become healthy but SILO can we would end up with the old LAPIS pod, running against a new SILO service.

### DIFF
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -1694,7 +1694,7 @@ subdomainSeparator: "-"
 replicas:
   website: 1
   backend: 1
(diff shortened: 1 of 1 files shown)


chaoran-chen (commented):
Nice! I think that this makes sense in any case, even after fixing the memory issue.

chaoran-chen (approved):

PR #3839, opened 2025-03-11
fix(deployment): change lapis livenessProbe to readinessProbe and use a liveness probe based on health endpoint

theosanderson:
Before this when SILO saturated all LAPIS pods would be restarted

### DIFF
--- a/kubernetes/loculus/templates/lapis-deployment.yaml
+++ b/kubernetes/loculus/templates/lapis-deployment.yaml
@@ -56,6 +56,14 @@ spec:
             periodSeconds: 10
             failureThreshold: 2
             timeoutSeconds: 5
+          livenessProbe:
+            httpGet:
+              path: /actuator/health
+              port: 8080
+            initialDelaySeconds: 30
+            periodSeconds: 10
+            failureThreshold: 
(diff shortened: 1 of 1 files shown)


chaoran-chen (approved):

kubernetes/loculus/templates/lapis-deployment.yaml
```diff
@@ -48,14 +48,22 @@ spec:
...
             periodSeconds: 10
             failureThreshold: 2
             timeoutSeconds: 5
+          livenessProbe:
+            httpGet:
+              path: /actuator/health
+              port: 8080
+            initialDelaySeconds: 30
+            periodSeconds: 10
+            failureThreshold:
```
anna-parker, line 65: Just to check - I think as this is not set the failureThreshold will default to 3 - do we want this?

Kubernetes will check the health endpoint every 10 seconds (periodSeconds: 10),
If it fails 3 times in a row, it will consider the container unhealthy and restart it

theosanderson: Thanks for spotting – setting real values does seem good I'm not sure what happened ed here

PR #3842, opened 2025-03-12
feat(website): drop downs on edit form

fhennig:
resolves #

preview URL:

### PR Checklist


website/src/components/Edit/InputField.tsx
theosanderson, line 1: Ideally it would be nice to suppress browser autofill 
image

fhennig: just the country or everything?

https://developer.mozilla.org/en-US/docs/Web/Security/Practical_implementation_guides/Turning_off_form_autocompletion#disabling_autocompletion

everything would be easy, i think just one prop might be a values.yaml change.

might also be useful for some users to have suggestions for i.e. their uni or similar fields.

theosanderson: I guess everything with `options`

fhennig: I've added `autocomplete=off` - can you check if that helped?

theosanderson: I can see the HTML saying that but in Chrome I still get 
image

If I set `autocomplete="new-password"` in the Inspector that works..

anna-parker:
I love this so much, just adding some thoughts for a follow up PR - it would be great if users could check before uploading their sequences if their countries are named as we expect - could we add this to the docs? Something like a drop down with "find your country in INSDC accepted format"? Then once we add administrative regions we could also add them to the docs so users could check in advance what format we expect names to be in.

website/src/components/Edit/MetadataForm.tsx
```diff
@@ -278,21 +175,3 @@ export const SubmissionIdRow: FC<SubmissionProps> = ({ submissionId }) => (
         <td className='w-full'>{submissionId}</td>
     </tr>
 );
-
-const mapErrorsAndWarnings = (
```
anna-parker, line 282: Just as a general remark - when you are just moving functions from one file to another could you put that in a separate PR? This PR has a lot of actual code changes and adding in this refactoring (i.e. moving this function and moving EditableSequences) makes it harder to review.

website/src/components/Edit/DataRow.tsx
```diff
@@ -2,39 +2,67 @@ import { type FC } from 'react';
...
+                    />
+                </td>
+                <td className='text-right'>
+                    <div className='pr-2 flex flex-row items-center'>
+                        <ErrorAndWarningIcons row={row} />
+                    </div>
+                </td>
+                <td className='w-3/4'>
+                    <InputField row={row} onChange={onChange} colorClassName={colorClassName} options={options} />
+                </td>
```
anna-parker, line 45: Doesn't have to be in this PR but is there a way to remove the warnings/errors once the user has selected an option from the dropdown? In the case of the options the field should be valid once an option is chosen from the dropdown.

image

anna-parker (approved):
LGTM! This looks amazing I tested a bit it seems to be working as intended - I added a couple suggestions for future PRs :-)

website/src/components/Edit/DataRow.tsx
theosanderson: I think this is basically a more general point about warnings and errors: they apply to the old version not the current entered text. I think it would be a bit weird for them to immediately vanish, but I think we could add stuff to make clear they are from the submitted version.

website/src/components/Edit/MetadataForm.tsx
fhennig: Makes a lot of sense! I will try to do that more.

website/src/components/Edit/InputField.tsx
fhennig: I've set it to `none` https://github.com/loculus-project/loculus/pull/3842/commits/e8f8d8571db26e121d2d172777d16adf24e97ff3

PR #3846, opened 2025-03-12
feat(website): add sequence button to review cards

### LINKED ISSUE #3831: Show processed sequences on review page instead of edit page
At the moment, we show the processed sequences (i.e., aligned sequences and amino acid sequences) at the bottom of the edit page:

Image

This doesn't really make sense as we otherwise only show the original data on the edit page, and the aim of the edit page is for the user to modify their submission. The other processed data are shown on the review page itself and it would make more sense to show the processed sequences there as well. As an idea, we could have a button to open a modal that shows the sequences.

theosanderson:
resolves #3831

preview URL: http://sequence-button.loculus.org

### Summary
Removes display of processed sequences from edit page and adds a button to the review cards that pops up a modal for this. Covers this with an integration test.

### Screenshot
image

image

image

### PR Checklist


integration-tests/tests/pages/review.page.ts
```diff
@@ -17,4 +27,43 @@ export class ReviewPage {
         await this.page.getByRole('button', { name: /Release \d+ valid sequence/ }).click();
         await this.page.getByRole('button', { name: 'Release', exact: true }).click();
     }
+
+    async viewSequences() {
+        const button = this.viewSequencesButton();
+        await expect(button).toBeVisible({ timeout: 30000 }); // Increase timeout to allow for processing
```
fhennig, line 33: I think it'd be better if we had an explicit step in the test "wait for sequence processing to be finished" - this might well test for the presence of the button - but I think having this timeout here is unexpected.

I think we should keep timeouts to a minimum, and the ones we do have should be made quite explicit in the tests.

integration-tests/tests/pages/review.page.ts
```diff
@@ -1,8 +1,18 @@
...
 
+
+    private viewSequencesButton = () =>
+        this.page.getByTestId(/view-sequences-/).first();
+    private sequencesDialog = () =>
+        this.page.locator('div:has(> div > h2:text("Processed Sequences"))').first();
+    private sequencesDialogCloseButton = () =>
+        this.sequencesDialog().getByRole('button', { name: '✕' });
+    private sequenceViewerContent = () => this.page.getByTestId('fixed-length-text-viewer');
+    private sequenceTabs = () => this.page.locator('.tab');
```
fhennig, line 14: This selector is very broad. would `this.sequencesDialog().locator('.tab')` also work? Not blocking.

integration-tests/tests/pages/review.page.ts
```diff
@@ -17,4 +27,43 @@ export class ReviewPage {
         await this.page.getByRole('button', { name: /Release \d+ valid sequence/ }).click();
         await this.page.getByRole('button', { name: 'Release', exact: true }).click();
     }
+
+    async viewSequences() {
+        const button = this.viewSequencesButton();
+        await expect(button).toBeVisible({ timeout: 30000 }); // Increase timeout to allow for processing
+        await button.click();
+        await expect(this.sequencesDialog()).toBeVisible({ timeout: 10000 });
```
fhennig, line 35: Why is there a 10 second delay, is the API call behind the button so expensive?

integration-tests/tests/specs/features/sequence-view.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+            S: 'GTGTTCTCTTGAGTGTTGGCAAAATGGAAAACAAAATCGAGGTGAACAACAAAGATGAGATGAACAAATGGTTTGAGGAGTTCAAGAAAGGAAATGGACTTGTGGACACTTTC',
+        });
+        await submissionPage.acceptTerms();
+        await submissionPage.submitSequence();
+
+        await expect(
+            page.getByRole('heading', { name: 'Review current submissions' }),
+        ).toBeVisible();
+
+        console.log('Waiting for sequence processing to complete...');
```
fhennig, line 32: Can we use a logger here? Usually we do - but not blocking I think.

integration-tests/tests/specs/features/sequence-view.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+        await page.waitForTimeout(60000);
+
+        const { ReviewPage } = await import('../../pages/review.page');
```
fhennig, line 35: Why not import at the top of the file?

website/src/components/ReviewPage/ReviewCard.tsx
```diff
@@ -124,56 +139,76 @@ const ButtonBar: FC<ButtonBarProps> = ({
...
-                content={
-                    approvable
-                        ? 'Release this sequence entry'
-                        : sequenceEntryStatus.processingResult === errorsProcessingResult
-                          ? 'You need to fix the errors before releasing this sequence entry'
-                          : 'Still awaiting preprocessing'
-                }
-            />
-            {!sequenceEntryStatus.isRevocation && (
+        <div className='flex mb-auto pt-3.5 items-center'>
```
fhennig, line 142:
```suggestion
        <div className='flex mb-auto pt-2 items-center'>
```

website/src/components/ReviewPage/SequencesDialog.tsx
```diff
@@ -0,0 +1,84 @@
...
+    if (!isOpen || !dataToView) return null;
+
+    const processedSequences = extractProcessedSequences(dataToView);
+
+    if (processedSequences.length === 0) {
+        return null;
+    }
+
+    return (
+        <div className='fixed inset-0 flex items-center justify-center z-50 overflow-auto bg-black bg-opacity-30'>
```
fhennig, line 30: I think it'd be good to reuse the `BaseDialog.tsx` - then we have the same opacity, close button etc everywhere.

website/src/components/common/FixedLengthTextViewer.tsx
```diff
@@ -52,6 +52,7 @@ export const FixedLengthTextViewer: FC<FixedLengthTextViewerProps> = ({ text, ma
             <textarea
                 readOnly
                 value={formattedText}
+                data-testid='fixed-length-text-viewer'
                 className={`font-mono  resize-none select-none h-content flex-grow whitespace-pre bg-transparent border-0 focus:outline-none focus:ring-0 ${lineSpacing} p-0 pl-3 text-gray-800`}
```
fhennig, line 56:
```suggestion
                className={`font-mono resize-none select-none h-content flex-grow whitespace-pre bg-transparent border-0 focus:outline-none focus:ring-0 ${lineSpacing} p-0 pl-3 text-gray-800`}
```

fhennig (changes requested):
Haven't looked at the feature in a preview yet, but from the screenshots it looks good!

integration-tests/tests/specs/features/sequence-view.spec.ts
theosanderson: I removed these - they were hanging around from the LLM and I'd removed the other ones 👍

integration-tests/tests/specs/features/sequence-view.spec.ts
theosanderson: 👍  done!

website/src/components/ReviewPage/SequencesDialog.tsx
theosanderson: I'm not sure if we can style the width of that exactly as here. Happy for anyone to look into it in a follow up but IMO not blocking?

integration-tests/tests/pages/review.page.ts
theosanderson: I agree this could be good but would prefer not to tackle it in this PR

integration-tests/tests/pages/review.page.ts
theosanderson: Resolving this for now as I don't know the answer and would prefer to avoid another test cycle

integration-tests/tests/pages/review.page.ts
theosanderson: This was just written by an LLM that didn't know how long it would take -removed thanks

website/src/components/ReviewPage/SequencesDialog.tsx
fhennig: yes not blocking

anna-parker:
@theosanderson where does the split up segment picture come from? I just hopped on the preview and it doesn't seem split up there - in general I dont think we should split up the segment submission by default

theosanderson:
@anna-parker did you check the revise page?

theosanderson:
Ah I guess this was resolved by the Slack discussion

website/src/components/ReviewPage/ReviewCard.tsx
fhennig: I think it looks better with pt-2 but fine to drop it.

fhennig (approved):

fhennig:
Thanks, looks good! I think it's quite good to not have this on the edit form anymore, thanks!

PR #3859, opened 2025-03-16
Raw reads UI prototype

theosanderson:
resolves #

preview URL:

### PR Checklist


fhennig:
Recording for future reference if preview is down:

https://github.com/user-attachments/assets/83e82c57-d703-4968-9f2d-28e85a6b5f23

theosanderson:
> It would be more in line with how it currently works if we were to display the file structure, but only upload once the final submit button is clicked.

I guess it's useful to think about an extreme example in which it takes say an hour to do the upload. In my model I think that there's a clear point (after that) when you git the Submit button and everything goes through immediately and you get straight to the review page. I think in your model you'd hit the Submit button then have to make sure you stayed on the page. Then if I lost track of the page I wouldn't be sure if the upload had gone through or not (until I found the review page, and figured out if the stuff on it was the recently submitted stuff).

PR #3867, opened 2025-03-18
fix(website): Render timestamp as YYYY-MM-DD instead of UNIX timestamp

### LINKED ISSUE #1556: Timestamp range search's To-field should be inclusive
I was wondering why I was getting no search results for "submitted at from 2023-04-03 (today)"

When I put the timestamp that was populated into URL into a unix time stamp tool I noticed that the time was not 00:00 UTC but current time.

image

image

### LINKED ISSUE #3727: Download modal should display human formatted date not UNIX timestamp
Just noticed that we display UNIX timestamp here which is obviously not what we should do here:

Image

fhennig:
resolves #3727 , resolves #1556

preview URL: https://timestamp-format.loculus.org

### Summary
- Render timestamp fields as YYYY-MM-DD
- Check whether a timestamp field is 'From' or 'To' (or neither) and set the HH:MM component to 23:59 for 'To'.

#### Testing
- Added unit test to the `ActiveFilters` component to check that the format is correct for timestamps.
- Added unit test for the `TimestampField` to check if the correct unix timestamp gets set.

### Screenshot

image

### PR Checklist

### DIFF
--- a/website/src/components/SearchPage/DownloadDialog/SequenceFilters.tsx
+++ b/website/src/components/SearchPage/DownloadDialog/SequenceFilters.tsx
@@ -113,12 +113,19 @@ export class FieldFilter implements SequenceFilter {
                 .filter(({ filterValue }) => filterValue.length > 0)
                 .map(({ name, filterValue }): [string, [string, string]] => [
                     name,
-                    [this.findSchemaLabel(name), this.filterValueDisplayString(filterValue)],
+                    [this.findSchemaLabel(name), this.filterValueDisplayString(name, filterValue)],
                 ]),
         );
     }
 
-    private filterValueDisplayString(value: any): string {
+    private filterValueDisplayString(fieldName: string, value: any): string {
+        const filterValueType = this.schema
+            .flatMap((metadataFilter) => (metadataFilter.grouped ? metadataFilter.groupedFields : metadataFilter))
+            .find((metadataFilter) => metadataFilter.name === fieldName)?.type;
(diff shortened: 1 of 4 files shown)


fhennig:
image

The issue #3727 has now _technically_ been solved, but it turns out the timestamps are probably not what we want to filter for? It looks like the date picker spits out this date: `Thu Mar 27 2025 00:00:00 GMT+0100 (Central European Standard Time)` which is `Wed Mar 26 2025 23:00:00` in GMT

These are the functions we use as part of the date field:

        dateToValueConverter={(date) => {
            const initialValue = date ? String(Math.floor(date.getTime() / 1000)) : '';
            if (initialValue === 'NaN') {
                return '';
            } else {
                return initialValue;
            }
        }}
        valueToDateConverter={(value) => {
            const timestamp = Math.max(parseInt(value, 10));
            return isNaN(timestamp) ? undefined : new Date(timestamp * 1000);
        }}

We probably want to change something here?

What value is it that we actually want to send to the backend?

theosanderson:
There's an Issue for what you described somewhere (made by Cornelius). Ideally what I think we want is to set dateFrom to 0:00 UTC on date selected and set date to to 23:59 on the date selected.

fhennig:
I found the issue you mentioned: https://github.com/loculus-project/loculus/issues/1556

I'll tackle this in here as well.

theosanderson (approved):
LGTM!

PR #3869, opened 2025-03-18
Revert "chore(deps-dev): bump daisyui from 4.12.23 to 5.0.6 in /website"

theosanderson:
Reverts loculus-project/loculus#3866

https://revert-3866-dependabot-np.loculus.org/ebola-sudan/search?

The formatting of the download modal is a bit messed up - this seems to be the cause:

Main:
image

Vs this PR:
image

In due course we can merge bumping it whilst fixing the classes to give the original result (and we probably will want to check other forms carefully too)

### DIFF
--- a/website/package-lock.json
+++ b/website/package-lock.json
@@ -74,7 +74,7 @@
                 "@vitejs/plugin-react": "^4.3.4",
                 "astro-eslint-parser": "^1.2.1",
                 "cross-env": "^7.0.3",
-                "daisyui": "^5.0.6",
+                "daisyui": "^4.12.23",
                 "eslint": "^9.17.0",
                 "eslint-plugin-astro": "^1.3.1",
                 "eslint-plugin-import": "^2.31.0",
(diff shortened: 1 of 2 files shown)


chaoran-chen:
@JonasKellerer, you did the daisy ui upgrade for GenSpectrum, right? Did you encounter similar problems and do you happen to know an easy fix?

JonasKellerer:
Yeah. I did the upgrade. I have not encountered this specific issue. However, I think the upgrade of daisyui should be paired with an update of tailwind. When I see it correctly we still have tailwind v3 here. And it seems some tailwind classes are not used for the button. 
Secondly, I am not sure what elements you use here. When I recall correctly they changed the default width of input elements https://daisyui.com/docs/upgrade/#input. Maybe this contributes too?

Upgrading tailwind is not that hard, but will change the default color of  "border". So one has to take a look where it is used. Also the config changes slightly.

anna-parker (approved):
The changes don't bother me too much - but I'm ok with reverting this for now

PR #3871, opened 2025-03-19
feat(website): Visualize that a file upload component has data in it already if empty

### LINKED ISSUE #3850: Revise page: make clear that sequence input is optional
If I go to revise a sequence the metadata is already filled in:

Image

but the sequences contains no indication that sequences are already there (from the previous version), though they are

Image

I think probably it should appear something like after you upload something 

Image

(although we don't know the filename and that's fine)

fhennig:
resolves #3850

preview URL:

### Summary

- Allow defining 'initial data' for the file upload component. it is displayed as a file.
- add the 'undo' functionality to the file upload component so you can revert to the initial data.
- bugfix: empty segments now show an upload control on the edit form of a sequence.

#### Testing
- Added automated tests for the FileUploadComponent to check that it correctly discards and undos.
- Manually checked that it's possible to discard sequence data in the edit form.
- Add integration test to check that discarded or modified sequence data is submitted correctly in the edit form.

### Screenshot

https://github.com/user-attachments/assets/1fb2a998-38ed-4675-958a-1029fdea360f

### PR Checklist

### DIFF
--- a/website/src/components/Edit/EditPage.tsx
+++ b/website/src/components/Edit/EditPage.tsx
@@ -25,6 +25,7 @@ type EditPageProps = {
     organism: string;
     clientConfig: ClientConfig;
     dataToEdit: SequenceEntryToEdit;
+    segmentNames: string[];
     accessToken: string;
     groupedInputFields: Map<string, InputField[]>;
     submissionDataTypes: SubmissionDataTypes;
@@ -35,13 +36,16 @@ const logger = getClientLogger('EditPage');
 const InnerEditPage: FC<EditPageProps> = ({
     organism,
     dataToEdit,
+    segmentNames,
     clientConfig,
     accessToken,
     groupedInputFields,
     submissionDataTypes,
 }) => {
     const [editableMetadata, setEditableMetadata] = useState(EditableMetadata.fromInitialData(dataToEdit));
-    const [editableSequences, setEditableSequences] = useState(EditableSequences.fromInitialData(dataToEdit));
+    const [editableSequences, setEditableSequences] = useState(
+        EditableSequences.fromInitialData(dataToEdit, segmentNames),
+    );
 
     const isCreatingRevision = dataToEdit.status === approvedForReleaseStatus;
 
--- a/website/src/components/Edit/InputForm.tsx
+++ b/website/src/components/Edit/InputForm.tsx
@@ -275,7 +275,7 @@ export const SequencesForm: FC<SequenceFormProps> = ({ editableSequences, setEdi
                             small={true}
                             initialValue={
                                 field.initialValue.length > 0
-                                    ? new VirtualFile(field.initialValue, 'existing_data.txt')
+                                    ? new VirtualFile(field.initialValue, 'Existing data')
                                     : undefined
                             }
                             showUndo={true}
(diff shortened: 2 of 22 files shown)


fhennig:
TODO, I just noticed that when editing and a segment has no data, it's not possible to upload data for it because the control is missing:

image


(S segment missing)

will fix

website/src/components/Submission/FileUpload/FileUploadComponent.tsx
theosanderson, line 1: Thanks for the PR! Looking at the file I don't find this super intuitive. IMO it's unexpected that there is an "Upload" button if there is existing data. I would expect to have to remove the existing data before I could upload it, or at minimum would expect a "Replace" wording. Also I guess for a multisegmented organism this means it is impossible to remove the sequence data for one segment.

fhennig: I've updated the video, that's probably better now, right?

website/src/components/Edit/InputForm.tsx
```diff
@@ -259,6 +273,12 @@ export const SequencesForm: FC<SequenceFormProps> = ({ editableSequences, setEdi
                             ariaLabel={`${field.key} Segment File`}
                             fileKind={PLAIN_SEGMENT_KIND}
                             small={true}
+                            initialValue={
+                                field.initialValue.length > 0
+                                    ? new VirtualFile(field.initialValue, 'existing_data.txt')
```
theosanderson, line 278:
```suggestion
                                    ? new VirtualFile(field.initialValue, 'Existing data')
```
Would this still work? IMO it would look a bit better

fhennig: yes taht works! Done.

website/src/components/Edit/EditPage.tsx
theosanderson, line 1: I made a follow up for some more changes: https://github.com/loculus-project/loculus/issues/3877

theosanderson (approved):
LGTM! Thanks - made an issue for some follow ups

website/src/components/Edit/EditPage.tsx
fhennig: good idea, thanks for pointing it out, I added it to the board as well 👍

anna-parker:
From investigation this PR is the cause of the error in https://github.com/loculus-project/loculus/issues/3798#issuecomment-2748390256

website/src/components/Edit/EditPage.tsx
```diff
@@ -29,13 +30,16 @@ const logger = getClientLogger('EditPage');
...
+    segmentNames,
     clientConfig,
     accessToken,
     groupedInputFields,
     submissionDataTypes,
 }) => {
     const [editableMetadata, setEditableMetadata] = useState(EditableMetadata.fromInitialData(dataToEdit));
-    const [editableSequences, setEditableSequences] = useState(EditableSequences.fromInitialData(dataToEdit));
+    const [editableSequences, setEditableSequences] = useState(
+        EditableSequences.fromInitialData(dataToEdit, segmentNames),
```
anna-parker, line 41: I think the issue is here - we do this for all segments and previously it was only for all segments that were also in the input file

PR #3874, opened 2025-03-19
feat(integration-tests): use setup project to submit some data to speed up tests

fhennig:
A lot of tests do not edit the data, and creating a new group and submitting a sequence for each test slows down the tests a lot.

I added a base project dependency that submits a single sequence of ebola sudan to use in other tests.

I know this breaks the isolation of the tests, but I'd argue the test speed up is well worth it.

I also replaced a bunch of hard-coded timeouts with loops or waits for elements.

### PR Checklist

### DIFF
--- a/integration-tests/tests/readonly.setup.ts
+++ b/integration-tests/tests/readonly.setup.ts
@@ -1,0 +1,54 @@
+import { test as setup } from '@playwright/test';
+import { AuthPage } from './pages/auth.page';
+import { GroupPage } from './pages/group.page';
+import { createTestGroup } from './fixtures/group.fixture';
+import { SingleSequenceSubmissionPage } from './pages/singlesubmission.page';
+import { v4 as uuidv4 } from 'uuid';
+
+setup('init some stuff', async ({ page }) => {
+    setup.setTimeout(90000);
+    const authPage = new AuthPage(page);
+    await authPage.navigateToRegister();
+    const username = uuidv4().substring(0, 8);
+    const password = uuidv4().substring(0, 8);
+    await authPage.createAccount({
+        firstName: 'Foo',
+        lastName: 'Bar',
+        email: `${username}@foo.org`,
+        organization: 'Foo University',
+        password,
+        username,
+    });
+
+    const groupPage = new GroupPage(page);
+    await groupPage.navigateToCreateGroupPage();
+    await groupPage.createGroup(createTestGroup());
+
+    const submissionPage = new SingleSequenceSubmissionPage(page);
+    const reviewPage = await submissionPage.completeSubmission({
+        submissionId: 'foobar',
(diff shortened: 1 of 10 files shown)


fhennig:
before: https://github.com/loculus-project/loculus/actions/runs/13956364790/job/39068327452?pr=3871 (11:01 mins)

now: https://github.com/loculus-project/loculus/actions/runs/13958353623/job/39074714088?pr=3874 (4:16 mins)

integration-tests/tests/readonly.setup.ts
```diff
@@ -0,0 +1,60 @@
...
+import { AuthPage } from './pages/auth.page';
+import { GroupPage } from './pages/group.page';
+import { createTestGroup } from './fixtures/group.fixture';
+import { SingleSequenceSubmissionPage } from './pages/singlesubmission.page';
+import { v4 as uuidv4 } from 'uuid';
+
+setup('Initialize a single ebola sequence as base data', async ({ page }) => {
+    setup.setTimeout(90000);
+    const authPage = new AuthPage(page);
+    await authPage.navigateToRegister();
```
theosanderson, line 11:
```suggestion
```
Already covered in the `createAccount`

integration-tests/tests/readonly.setup.ts
```diff
@@ -0,0 +1,60 @@
...
+        firstName: 'Foo',
+        lastName: 'Bar',
+        email: `${username}@foo.org`,
+        organization: 'Foo University',
+        password,
+        username,
+    });
+
+    const groupPage = new GroupPage(page);
+    await groupPage.navigateToCreateGroupPage();
```
theosanderson, line 24:
```suggestion
```
Already included in `createGroup`

theosanderson (approved):
nice, lgtm! - as I mentioned on Slack I expect that we could have got the tests running pretty fast just by increasing parallelism but this still seems good. I guess the only downside is that with maximum parallelism potentially this could conceivably slow down total tests in that tests that _do_ require custom data can't start submitting that custom data until the read only data has been submitted and processed. But we can cross that bridge when/if it becomes relevant.

fhennig:
> I guess the only downside is that with maximum parallelism potentially this could conceivably slow down total tests in that tests that do require custom data can't start submitting that custom data until the read only data has been submitted and processed. But we can cross that bridge when/if it becomes relevant (likely never!).

Agreed! One could split the projects up more, so another group of tests runs without this first starter one. Splitting into projects can also be done by filename, or with tags, I read into it quite a bit, there's some pretty cool stuff available!

If we could increase parallelism in CI that would probably be cool. Now the next slowest thing is just waiting for the cluster to come up, and in the cluster it looks like lapis and the prepro pipeline are the slow-ish bits.

PR #3882, opened 2025-03-24
fix(website): File upload input gets reset if file is not valid

### LINKED ISSUE #3851: Single segment upload: unless filename changes re-upload is not attempted
Create a file like:

```
>abc
>def
ATGG
```

and upload it with the single-segment file upload

You get an error, as expected.

Now fix it to:

```
>def
ATGG
```

and try to reupload on the same page. Nothing happens. If the filename changes then it works OK.

fhennig:
resolves #3851 

preview URL:

### PR Checklist


theosanderson (approved):
code lgtm, thanks!!

Issue #3857, opened 2025-03-14
Clearing date filters from the 'active filters' doesn't reset the filter input fields.


fhennig:
Using the active filters, clearing a date field doesn't clear the value from the filter input field.

https://github.com/user-attachments/assets/0afdee10-3c11-4f85-8393-a8b04c99def2

I found this while working on the lineage field, where the issue also occurs but I'm implementing a fix here: https://github.com/loculus-project/loculus/pull/3811

A fix means adding a `useEffect` to react to changes in the `fieldValue`.

fhennig:
related

<img width="821" alt="Image" src="https://github.com/user-attachments/assets/c517f480-bb5e-469e-8b7a-def0814657d9" />

I think the issue is that we're visualizing the `lapisSearchParameters` - maybe we could instead visulize the `fieldValues`? Haven't really thought this through though.

Issue #3451, opened 2024-12-17
Website: fix `any` types


fengelniederhammer:
#3450 introduced a number of eslint ignore comments for the rule `@typescript-eslint/no-explicit-any`. `any` is harmful, because it essentially disables Typescript checks. You can access any method or field without Typescript complaining. In the long term it generates code that is harder to maintain (adding new features and refactoring becomes more dangerous because the type system can't guide you).

All of those should be fixed: 
* If the type is really unknown, we should use `unknown` and make sure that the variable has a certain type before using it.
* If the type can be improved, we should replace `any` by a concrete type that is more useful.

See TODO comments in the code that reference this issue.

corneliusroemer:
In terms of priority, I think this isn't super high up, while it's nice to have everything fully typed, if it's not easy to fix then I don't think we should invest a lot of time. 

I can't think of any bugs/runtime type errors we've had due to use of any.

Good point to consider `unknown` and narrowing as escape hatch though.

fhennig:
I had a look at it and saw some things that I touched recently and quickly fixed a few things here: https://github.com/loculus-project/loculus/pull/3464

The remaining things are almost all in the `SearchFullUI`, and related to the `lapisSearchParameters`. 

> I can't think of any bugs/runtime type errors we've had due to use of any.

maybe not, but I think this point is still quite valid:

> In the long term it generates code that is harder to maintain (adding new features and refactoring becomes more dangerous because the type system can't guide you).

I find reasoning about the `lapisSearchParameters` not so easy, because it's not really clear what's happening, there's even this:
https://github.com/loculus-project/loculus/blob/fix-some-any/website/src/utils/search.ts#L253 where the property changes type from `string` to `string[]` 😅 So you can't even be sure that a property you thought you knew the type of will have this same type at a different place in the code.

I can't speak to the priority of this, just wanted to emphasize the maintainability aspect again (as opposed to a strictly runtime-error-focused PoV).

The `unknown` type is cool, I didn't know about it before!

corneliusroemer:
Thanks for fixing some, that's definitely good!

[...]

PR #3883, opened 2025-03-24
refactor(SearchFullUI): Introduce `FilterSchema` class and consolidate scattered functionality in it

fhennig:
This is prep work for https://github.com/loculus-project/loculus/issues/3857

resolves some occurrences of `any` (part of https://github.com/loculus-project/loculus/issues/3451)

preview: https://refactor-filter-schema.loculus.org

### Summary
The `FieldFilter` shouldn't use the `lapisSearchParameters` as a base, instead it should use `fieldValues`. This is the core of what should be refactored. I then decided that the `FieldFilter` can generate the `lapisSearchParameters` instead, and consolidated some code in a new class.

- I moved `getReferenceGenomesSequenceNames` from `search.ts` to `config.ts` as it reads things from the config.
- I removed `addHiddenFilters` from `search.ts` as it wasn't used.
- I turned `type ConsolidatedMetadataFilters` into a class `FilterSchema` and put some functions into the class that were previously scattered elsewhere.
- the `FieldFilter` now doesn't take the `lapisSearchParameters` anymore, instead it takes `FieldValues` which is more in line with what it should show. the `tableFilter: FieldFilter` is now used to generate the `lapisSearchParameters`. - This is a key change to facilitate #3857 later on.

### PR Checklist


fhennig:
I think the broad strokes here are done, just some minor cleanup/polishing left. In case anybody wants to have a look if the structure in general seems ok @theosanderson

fhennig:
I've adressed the copilot nitpicks

website/src/utils/search.ts
```diff
@@ -101,155 +80,141 @@ export const getColumnVisibilitiesFromQuery = (schema: Schema, state: Record<str
...
-                    label: filter.label,
-                    initiallyVisible: filter.initiallyVisible,
-                };
-                fieldList.push(fieldForGroup);
-                groupsMap.set(filter.fieldGroup, fieldForGroup);
+export class FilterSchema {
+    public readonly filters: (MetadataFilter | GroupedMetadataFilter)[] = [];
+
+    constructor(metadataSchema: Metadata[]) {
+        // expand metadata fields for ranges and range overlap search
```
theosanderson, line 91: These things - expanding date ranges to a series of fields with a group property, and grouping anything with a group property together, used to be their own functions. IMO that made things easier to understand. We could still do it in some way with this design I presume?

website/src/utils/serversideSearch.ts
```diff
@@ -20,8 +20,10 @@ export const performLapisSearchQueries = async (
     hiddenFieldValues: FieldValues,
     organism: string,
 ): Promise<SearchResponse> => {
-    const fieldValues = getFieldValuesFromQuery(state, hiddenFieldValues, schema);
-    const lapisSearchParameters = getLapisSearchParameters(fieldValues, referenceGenomesSequenceNames, schema);
+    const filterSchema = new FilterSchema(schema.metadata);
+    const fieldValues = filterSchema.getFieldValuesFromQuery(state, hiddenFieldValues);
+    const fieldFilter = new FieldFilter(filterSchema, fieldValues, hiddenFieldValues, referenceGenomesSequenceNames);
```
theosanderson, line 25: I find this pretty confusing. IMO a `filter` (which I didn't love anyway) was roughly a field (maybe sometimes a grouped field). Here a filter seems to be all the fields?

website/src/utils/search.ts
fhennig: Yes we could, and I will put them back into their own functions, it's a good idea. at first I thought (and tried) to merge it all into one shorter function to make it easier, but that didn't work out.

website/src/utils/serversideSearch.ts
fhennig: It's not good if it's confusing! I wanted to make it less confusing (and I did make it less confusing for me) but I don't want it to be more confusing for others now.

(I totally agree that naming things is hard)

I wrote a docstring for the `FieldFilter` (not easy to see just in GitHub review):

> Filter sequences based on certain fields that have to match, i.e. 'country == China' or 'data use terms == OPEN'.

it's just a wrapper around `fieldValues`, so it's not all the fields (that exist) but it is the aggregate of fields that are currently used to filter.

IMO it makes sense to have such a wrapper object with associated functions because they group together nicely and I find this easier to reason about, but since we already have `MetadataFilter` this is - I see that now - possibly a bad name for it.

These two functions can - if we want to - be pulled out of their respective objects and can be available independently, if you think that is an improvement.

As you mentioned in your other comment, having these functions inside of the classes _does_ make things more restrictive. I did this, because I felt that there were too many moving parts in this part of the code and it lacked cohesion between its elements. I found it (and in part still find it) difficult to reason about. But there is no technical necessity for the functions now being inside the classes.

website/src/utils/serversideSearch.ts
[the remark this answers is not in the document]
fhennig: Thanks for raising your concerns!

I'm also not the biggest OOP fan in general, but (as I wrote in my other comment) I find it helpful to group things together with classes - although there are possibly other ways of achieving the same thing. One of my main concerns while doing this was achieving higher cohesion between parts of the code that I saw as related and that now are all found in the `SequenceFilters.tsx`, and that are now not individually referenced (coupled) anymore in `SearchFullUI.tsx`. I wanted to pull things out of there, because that component is (as you acknowledge) hard to reason about. I think classes come in pretty handy for achieving cohesion, but there are possibly other ways to do the same thing.

I appreciate your trust! Maybe we can also find a better way of approaching these refactorings next time, so less trust is required. Maybe we could do it in pair programming!

website/src/utils/search.ts
fhennig: [`3d75165` (#3883)](https://github.com/loculus-project/loculus/pull/3883/commits/3d75165e8df56b405605a3f24837cf23413f2321)

website/src/utils/search.ts
```diff
@@ -175,81 +148,82 @@ export const consolidateGroupedFields = (filters: MetadataFilter[]): Consolidate
...
-            if (s.includes('.')) {
-                return s.split('.')[0];
-            }
-            return s;
-        });
+/**
+ * Derives from the Metadata schema. For some metadata fields, they are expanded into multiple
+ * (grouped) filters.
+ */
+export class FilterSchema {
```
theosanderson, line 158: I'm wondering if this should have `Metadata` in the name too?

website/src/utils/serversideSearch.ts
theosanderson: Thanks yes the renaming is very helpful. As you say my confusion was because `MetadataFilter` is essentially one field and this was lots. (I'd missed that the addition of `FieldFilter` wasn't actually here).

website/src/components/SearchPage/DownloadDialog/SequenceFilters.tsx
```diff
@@ -34,19 +38,35 @@ export interface SequenceFilter {
  * Filter sequences based on certain fields that have to match, i.e. 'country == China' or
```
theosanderson, line 38: nit: IMO this docstring sounds like documentation for a function rather than a class - i.e. it suggests the class will actively do this thing whereas its more a means of storing data for doing this thing in LAPIS

website/src/utils/search.ts
fhennig: sure!

https://github.com/loculus-project/loculus/pull/3883/commits/b8348b1fabac1f66280c8f4ef6ccae2cfb002902

website/src/components/SearchPage/DownloadDialog/SequenceFilters.tsx
fhennig: https://github.com/loculus-project/loculus/pull/3883/commits/875217ae45774b10c5155ec28c67c37b77e987f2

fhennig: 👍

PR #3895, opened 2025-03-25
fix(website): fix handling of certain edgecases for the active filter display

### LINKED ISSUE #3857: Clearing date filters from the 'active filters' doesn't reset the filter input fields.

fhennig:
resolves #3857 

preview URL: https://improve-filters.loculus.org

### Summary
- "Is revocation" and "version status" were previously not shown if unset, but also not shown if set (because it's a hidden default). Now the fields are shown when unset, and clicking the "x" button will reset it to the hidden field value.
- for substring-search-enabled fields, the active filter showed "<fieldname>.regex: (i?)<value>" because the lapisSearchParameters were used. This is now not the case anymore.
- A small 'regression': previously, mutations were shown split into "nucleotide" and "aminoacid" etc, but now this isn't done anymore.
- date fields
	- use Monday as the start of the week. 
	- use `oneTap` -> no more clicking "OK" to confirm a date selection.
- date range field now also suppors 'x'-ing in the active filters.
- increase integration test worker count from default (2) to 4.

#### Manual testing

I checked that the _Author affiliations_ field (which uses substring search) renders correctly in the active filters. I tested that that field, as well as dates, date ranges, mutations, all are possible to remove via the active filters panel.

#### Added tests
- added integration tests to check that the fields can be removed
- added unit tests for date range field

#### Every field clears

https://github.com/user-attachments/assets/a711fb20-e88b-40b5-9def-b9d9cb044178

#### New handling of hidden fields

https://github.com/user-attachments/assets/8f0ed2e8-53df-464d-9b98-d155ac2e2278

### PR Checklist

### DIFF
--- a/website/src/components/SearchPage/fields/DateField.tsx
+++ b/website/src/components/SearchPage/fields/DateField.tsx
@@ -67,7 +67,7 @@ const CustomizedDatePicker: FC<CustomizedDatePickerProps> = ({
     valueToDateConverter,
     fieldValue,
 }) => {
-    const dateValue = fieldValue !== '' ? valueToDateConverter(fieldValue.toString()) : undefined;
+    const dateValue = fieldValue !== '' ? valueToDateConverter(fieldValue.toString()) : null;
     return (
         <div>
             <div className='flex justify-between items-center'>
@@ -77,7 +77,7 @@ const CustomizedDatePicker: FC<CustomizedDatePickerProps> = ({
                 <DatePicker
                     value={dateValue}
                     name={field.name}
-                    key={`${field.name}-${fieldValue}`}  // key set to force rerender when value changes
+                    key={field.name}
                     isoWeek={true}
                     oneTap={true}
                     onChange={(date) => {
--- a/website/src/components/SearchPage/fields/DateField.tsx
+++ b/website/src/components/SearchPage/fields/DateField.tsx
@@ -78,6 +78,7 @@ const CustomizedDatePicker: FC<CustomizedDatePickerProps> = ({
                     value={dateValue}
                     name={field.name}
                     key={field.name}
+                    isoWeek={true}
                     onChange={(date) => {
                         if (date) {
                             setSomeFieldValues([field.name, dateToValueConverter(date)]);
--- a/website/src/components/SearchPage/fields/DateField.tsx
+++ b/website/src/components/SearchPage/fields/DateField.tsx
@@ -20,10 +20,17 @@ export const DateField: FC<Omit<CustomizedDatePickerProps, 'dateToValueConverter
         {...props}
         dateToValueConverter={(date) => {
             if (!date) return '';
+            console.log(`foo: ${date}`);
             const isoDate = DateTime.fromJSDate(date).setZone('utc', { keepLocalTime: true }).toISODate();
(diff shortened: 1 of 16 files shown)


fhennig:
I found out that we could have the date picker be configured to be "one tap", then you wouldn't have to click "OK", just click a date and that's it. I think I'd prefer that - I might add it in. ([Example in the `rsuite` docs](https://rsuitejs.com/components/date-picker/#one-tap))

website/src/components/SearchPage/fields/DateField.tsx
```diff
@@ -67,16 +67,18 @@ const CustomizedDatePicker: FC<CustomizedDatePickerProps> = ({
...
     return (
         <div>
             <div className='flex justify-between items-center'>
                 <label htmlFor={field.name} className='block text-sm w-16 my-3 text-right mr-2 text-gray-400'>
                     {field.label}
                 </label>
                 <DatePicker
+                    value={dateValue}
                     name={field.name}
-                    defaultValue={fieldValue !== '' ? valueToDateConverter(fieldValue.toString()) : undefined}
```
theosanderson, line 78: Interesting - we had to add this here: https://github.com/loculus-project/loculus/pull/1991 but now we don't. Maybe it's `null` vs `undefined` that is the key.

website/src/components/common/ActiveFilters.tsx
```diff
@@ -19,7 +19,11 @@ export const ActiveFilters: FC<ActiveFiltersProps> = ({ sequenceFilter, removeFi
                     className='border-primary-600 rounded-sm border border-l-primary-600 bg-gray-100 border-l-8 pl-3 py-1 text-sm flex flex-row'
                 >
                     <span className='text-primary-900 font-light pr-1'>{label}:</span>
-                    <span className='text-primary-900 font-semibold'>{value}</span>
+                    {value === '' ? (
+                        <span className='text-primary-900 italic'>unset</span>
```
chaoran-chen, line 23: (I'm not sure whether unset is a very intuitive term, but I also don't know what's better, so I'm happy with leaving it as is but just bringing it up in case someone else has a better idea.)

fhennig: I also considered _(none)_

theosanderson: I think unset is better than none. `All` could work. But I don't think it matters too much as it's moderately niche.

theosanderson: Or "any"

fhennig: I think I like "any"!

chaoran-chen (approved):
Thanks! Looks all good to me.

> I found out that we could have the date picker be configured to be "one tap", then you wouldn't have to click "OK", just click a date and that's it. I think I'd prefer that

I'd prefer that too.

PR #3900, opened 2025-03-26
Add extra props for 'schema'

fhennig:
resolves #

preview URL:

### PR Checklist

### DIFF
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -1425,7 +1425,6 @@ defaultOrganisms:
     schema:
       image: "https://cdn.who.int/media/images/default-source/mca/mca-covid-19/coronavirus-2.tmb-1920v.jpg?sfvrsn=4dba955c_19"
       organismName: "Test organism (without consensus sequences)"
-      nucleotideSequences: []
       submissionDataTypes:
         consensusSequences: false
       metadata:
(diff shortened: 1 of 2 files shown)


theosanderson (approved):

kubernetes/loculus/values.yaml
```diff
@@ -1425,7 +1425,6 @@ defaultOrganisms:
     schema:
       image: "https://cdn.who.int/media/images/default-source/mca/mca-covid-19/coronavirus-2.tmb-1920v.jpg?sfvrsn=4dba955c_19"
       organismName: "Test organism (without consensus sequences)"
-      nucleotideSequences: []
```
theosanderson, line 1428: In the schema you say "Defaults to a list with just 'main'." - if so is this problematic?

fhennig: Ah yes, I think you're right.

fhennig: I'll put it back. Thanks for noticing.

kubernetes/loculus/values.yaml
theosanderson, line 1: As always - worth checking PPX values.yaml if you haven't

fhennig: thanks for pointing it out, I didn't do it before but now I did - all looking fine!

PR #3902, opened 2025-03-27
feat!(deployment): Standardize image config in Helm Chart

### LINKED ISSUE #3726: Helm: Standardize and expand image configuration
I would like to have a way to override specifically the backend image, but more generally probably all images that the Helm Chart uses, and I'd like to be able to do so in a unified way. Additionally I'd like to specify pull policies (i.e. not use 'Always') so I can easily use local images. I want to be able to do this to easily test my own versions of specific components.

Why do I want to do this? I want to use [tilt](https://tilt.dev/) for local testing of specifically the website and backend, and so I need to specify the settings above through the Helm Chart.

### Current situation

Lapis and silo images are specified like this:

```
images:
  lapisSilo: "ghcr.io/genspectrum/lapis-silo:0.5.3"
  lapis: "ghcr.io/genspectrum/lapis:0.3.14"
```

A website image can be specified as `customWebsiteImage`.

For the backend, this is hardcoded: `"ghcr.io/loculus-project/backend:{{ $dockerTag }}"` and the dockerTag can be specified as `loculus.dockerTag`. The imagePullPolicy is not configurable and set to `Always`.

### Proposed new structure

expand the `images` config section with more keys and add `website` and `backend` to it:

```
images:
  lapisSilo:
    repository: "ghcr.io/genspectrum/lapis-silo"
    tag: "0.5.3"
    pullPolicy: "IfNotPresent"

  lapis:
    repository: "ghcr.io/genspectrum/lapis"
    tag: "0.3.14"
    pullPolicy: "IfNotPresent"

  website:
    repository: ...
    tag: ...

  backend: ...

```

This would allow users to configure the settings mentioned above, and also make images be configured in one, unified way. Specifying just the dockerTag for the backend would still be possible, and also we can set imagePullPolicy `Always` as the default, so no functionality would be lost, only structure would change slightly.

It would be a breaking change, but we could also avoid breaking by making the old config settings still valid, just deprecated (to be decided).

### How would it be implemented?
- Adjust values.yaml and schema
- Adjust github Jobs (possibly)
- Maybe adjust the deploy.py script?

fhennig:
**BREAKING** - Changes the way images are configured. If your `values.yaml` specified `images.lapisSilo`, `images.lapis` or `customWebsiteImage`, you need to change those settings to use the new format. Otherwise, nothing needs to be done.

resolves #3726

preview URL (docs): https://docs-helm-refactor-image-conf.loculus.org/reference/helm-chart-config/

### Summary
Images for `lapis`, `lapisSilo`, `website` and `backend` are now all configured under an `images` key:

```yaml
images:
  lapisSilo:
    repository: "ghcr.io/genspectrum/lapis-silo"
    tag: "0.5.7"
    pullPolicy: IfNotPresent
  lapis:
    repository: "ghcr.io/genspectrum/lapis"
    tag: "0.4.1"
    pullPolicy: IfNotPresent
  website:
    repository: "ghcr.io/loculus-project/website"
    pullPolicy: Always
  backend:
    repository: "ghcr.io/loculus-project/backend"
    pullPolicy: Always
```

The `customWebsiteImage` key is no longer valid, use `images.website` instead.

Docs were adapted.

### Screenshot of the docs
image

---

image

### PR Checklist
- ~~Any manual testidng that has been done is documented (i.e. what exactly was tested?)~~

### DIFF
--- a/kubernetes/loculus/templates/silo-deployment.yaml
+++ b/kubernetes/loculus/templates/silo-deployment.yaml
@@ -29,7 +29,7 @@ spec:
         {{- include "loculus.configProcessor" (dict "name" "lapis-silo-database-config" "dockerTag" $dockerTag) | nindent 8 }}
       containers:
         - name: silo
-          image: "{{ $.Values.images.lapisSilo.repository }}:{{ $.Values.images.lapisSilo.tag | default "latest" }}"
+          image: "{{ $.Values.images.lapisSilo.repository }}:{{ $.Values.images.lapisSilo.tag }}"
           imagePullPolicy: "{{ $.Values.images.lapisSilo.pullPolicy }}"
           {{- include "loculus.resources" (list "silo" $.Values $key) | nindent 10 }}
           ports:
--- a/kubernetes/loculus/templates/silo-deployment.yaml
+++ b/kubernetes/loculus/templates/silo-deployment.yaml
@@ -29,7 +29,7 @@ spec:
         {{- include "loculus.configProcessor" (dict "name" "lapis-silo-database-config" "dockerTag" $dockerTag) | nindent 8 }}
       containers:
         - name: silo
-          image: {{ $.Values.images.lapisSilo.repository }}{{ $.Values.images.lapisSilo.tag }}
+          image: "{{ $.Values.images.lapisSilo.repository }}:{{ $.Values.images.lapisSilo.tag }}"
           {{- include "loculus.resources" (list "silo" $.Values $key) | nindent 10 }}
           ports:
             - containerPort: 8081
@@ -43,7 +43,7 @@ spec:
             - name: lapis-silo-shared-data
               mountPath: /data
         - name: silo-preprocessing
-          image: {{ $.Values.images.lapisSilo.repository }}{{ $.Values.images.lapisSilo.tag }}
+          image: "{{ $.Values.images.lapisSilo.repository }}:{{ $.Values.images.lapisSilo.tag }}"
           imagePullPolicy: IfNotPresent
           {{- include "loculus.resources" (list "silo-preprocessing" $.Values) | nindent 10 }}
           command:
--- a/kubernetes/loculus/templates/silo-deployment.yaml
+++ b/kubernetes/loculus/templates/silo-deployment.yaml
@@ -30,6 +30,7 @@ spec:
       containers:
         - name: silo
           image: "{{ $.Values.images.lapisSilo.repository }}:{{ $.Values.images.lapisSilo.tag }}"
+          imagePullPolicy: "{{ $.Values.images.lapisSilo.pullPolicy }}"
           {{- include "loculus.resources" (list "silo" $.Values $key) | nindent 10 }}
           ports:
             - containerPort: 8081
@@ -44,7 +45,7 @@ spec:
               mountPath: /data
         - name: silo-preprocessing
           image: "{{ $.Values.images.lapisSilo.repository }}:{{ $.Values.images.lapisSilo.tag }}"
-          imagePullPolicy: IfNotPresent
+          imagePullPolicy: "{{ $.Values.images.lapisSilo.pullPolicy }}"
           {{- include "loculus.resources" (list "silo-preprocessing" $.Values) | nindent 10 }}
           command:
             - sh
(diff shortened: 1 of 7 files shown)


kubernetes/loculus/values.schema.json
```diff
@@ -1117,19 +1112,15 @@
     },
     "images": {
       "type": "object",
-      "properties": {
-        "lapisSilo": {
-          "type": "string",
-          "description": "TODO",
-          "default": "ghcr.io/genspectrum/lapis-silo:0.5.2"
+      "additionalProperties": {
```
theosanderson, line 1115: I guess this means if I type "web-site" instead of "website" the schema won't catch it - would named properties be better? (not blocking)

fhennig: Ah yes you're right. This was a simple way of using a common schema for all values in the dict, but I think I can also do it with named keys and a defined type

fhennig: fixed.

fhennig:
TODO: schema docs

kubernetes/loculus/templates/silo-deployment.yaml
```diff
@@ -29,7 +29,8 @@ spec:
         {{- include "loculus.configProcessor" (dict "name" "lapis-silo-database-config" "dockerTag" $dockerTag) | nindent 8 }}
       containers:
         - name: silo
-          image: {{ $.Values.images.lapisSilo }}
+          image: "{{ $.Values.images.lapisSilo.repository }}:{{ $.Values.images.lapisSilo.tag | default "latest" }}"
```
theosanderson, line 32:
```suggestion
          image: "{{ $.Values.images.lapisSilo.repository }}:{{ $.Values.images.lapisSilo.tag }}"
```
doesn't seem necessary? You could always configure `latest`

fhennig: I think so at least

theosanderson: Well in that case you don't have one a bit below for `silo-preprocessing`. But I'm not really convinced. Someone could write a values.yaml with `tag: ""` and that would I guess yield something that doesn't start, but they could also write a values.yaml with `tag: "NonExistentTag"` which would also yield something that doesn't start

theosanderson: I guess it's a Q of whether we are seeing the schema as something to help external deployers or something to help us ensure our own starting `values.yaml` is compliant. I see it more as the former than the latter.

fhennig: Hmm, I'm fine with leaving it out!

But I think I'm not following what you're saying. Yes, one could specify an invalid tag. But I think that's different than omitting the tag (as the schema suggests you're allowed to do) and then it not starting.

I'm also not following what the distinction would be you're making between the target of the schema. I think the way I think about it is that, if I use a key that the schema doesn't require, my template needs to account for that and take measures to produce valid output. since the tag can be left out, and not providing a default would create an invalid image string, I think we would need to provide a default here.

fhennig: it's jsut that if you leave out the tag for the "website", the Chart will do something else and it will start. so it seemed to me that the same behaviour can be expected for the other images as well - hence the use of the default.

theosanderson: I'm saying that for an end-user-maintainer there isn't really an easy way they can omit the tag as it is specified in the base `values.yaml` from which everything is inherited.

kubernetes/loculus/values.yaml
theosanderson, line 1: Please can you make the corresponding PR for Pathoplexus before we merge this - probably even make a Pathopelxus preview targeting this

kubernetes/loculus/templates/silo-deployment.yaml
theosanderson: To be clear my understanding is that if the user specifies a repository and not an tag, the LAPIS tag in the default values.yaml will be used

fhennig: Ahhh I think I got it now. Yes the values.yaml is basically like a "default".

I'll go remove the marked bit.

kubernetes/loculus/values.yaml
fhennig: I've made a Pr here https://github.com/pathoplexus/pathoplexus/pull/432

(I did it to test if it's still deploying but I think we can undo the SHA change and then also merge that one probably)

anna-parker:
Question/comment: I noticed you havent moved the preprocessing image - is there a reason for this? Just curious

theosanderson (approved):

fhennig:
> is there a reason for this? Just curious

This was actually not on purpose, but what Theo said makes sense.

I was mainly concerned about the backend and website image, because I want to reconfigure those.

Thanks for approving Theo!

PR #3911, opened 2025-03-27
integration tests: increase CI worker count

fhennig:
testing 8 for now, and seeing if it's quick.
4 runs at about 11 minutes I think.

Maybe more workers doesn't even help, since it's all async anyways.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -76,7 +76,7 @@ jobs:
       - name: Run Integration test
         run: |
           set -o pipefail
-          cd integration-tests && npx playwright test --workers=8 2>&1 | tee output.txt
+          cd integration-tests && npx playwright test 2>&1 | tee output.txt
           EXIT_CODE=$?
           echo '```' >> $GITHUB_STEP_SUMMARY
           sed -n '/Running [0-9]\+ tests/,$p' output.txt >> $GITHUB_STEP_SUMMARY
--- a/integration-tests/playwright.config.ts
+++ b/integration-tests/playwright.config.ts
@@ -18,7 +18,9 @@ export default defineConfig({
     /* Fail the build on CI if you accidentally left test.only in the source code. */
     forbidOnly: !!process.env.CI,
     /* Retry on CI only */
-    retries: process.env.CI ? 2 : 0,
+    retries: process.env.CI ? 3 : 0,
+
+    workers: process.env.CI ? 8 : undefined,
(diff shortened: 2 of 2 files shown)


theosanderson:
> Maybe more workers doesn't even help, since it's all async anyways.

More workers should help _because_ it's async -i.e. not CPU-limited

theosanderson:
We could also make the tests within files run in parallel to speed things up further: https://playwright.dev/docs/test-parallel

fhennig:
I just checked, we already set `fullyParallel` to true in the playwright config.

> More workers should help because it's async -i.e. not CPU-limited

Hmm. but since it's async, each worker already can execute an infinite amout of tests if most of them are just waiting. It can just start executing 100 tests and context switch between them already, right? If we _weren't_ using async, but instead waiting would block the thread (classic multi-threading) then more workers would help.

The test also took 11 minutes this time, the same as with 4 workers.

fhennig:
Maybe we can go back down to 2 actually. One for chrome and one for firefox.

theosanderson:
> Hmm. but since it's async, each worker already can execute an infinite amout of tests if most of them are just waiting. It can just start executing 100 tests and context switch between them already, right? If we weren't using async, but instead waiting would block the thread (classic multi-threading) then more workers would help.

That's not how I was thinking about it - I would expect that each worker represents a single "browser window" - I don't think it can switch because that would mean losing the browser state. (Obvs I may be misunderstanding though)

theosanderson:
ChatGPT agreed that a single worker can only run one test at a time (not saying that proves anything, it could be wrong)

theosanderson:
Could we potentially still just be using 1 worker?
image

theosanderson:
The slowness will partly be that we enable retries in CI = 3x as long if something fails.

fhennig:
Ok I see, I didn't think much about the browser state, but I can see how it might be tied to the worker instead of the test. Then you're totally right!

I head a look at the report. Assuming we're running on a 2 core runner, I think we might be doing a lot of switching, and that then leads to individually longer test execution, and that's why we're seeing the timeouts now? Assuming that test times are measured in real time and not in execution time.

I'm just gonna try with a longer global timeout. I still think it should be possible to get a speedup with more workers then.

fhennig:
looking at the report, there are still 4 flaky tests, but time in the report went from 5.2 minutes to 4.1.

Not sure if this is worth it. maybe if we can really get a beefier runner we could get it faster.

But also just waiting for the cluster to spin up takes quite some time.

Issue #3903, opened 2025-03-27
Create buckets for development


chaoran-chen:
To get the development started quickly, we will first create a few buckets on Hetzner that we can use.

We need to create a public and a private one for each instance. For now, let's create one pair of buckets for each dev who's interested in working on it and a pair of buckets for the preview.

chaoran-chen:
I created the buckets. The credentials for the preview server are in our password manager under "Hetzner Object Store file-sharing preview buckets".

chaoran-chen:
I was originally working under the assumption that we need a public and a private bucket but it seems that it is easy, at least for min.io (I don't know yet how well this works across S3 platforms), to set a bucket to private by default but then make objects public for download: https://min.io/docs/minio/linux/reference/minio-mc/mc-anonymous-set.html. That would be much better and avoid the need to copy files from a public to a private bucket.

PR #3913, opened 2025-03-29
feat(backend): Add /files/request-uploads API

### LINKED ISSUE #3905: Implement `/files/request-uploads` API
The endpoint is to initiate an upload to S3. Users (=website/clients/preprocessing pipeline/etc) can call it to receive pre-signed URLs for uploading files.

The endpoint accepts the following parameters:

- numberFiles
- groupId

It returns:

```
[
  { fileId: string, url: string }
]
```

The endpoint will create entries in the `files` table (depends on #3904) and then pre-signed URLs. The files should be in the private bucket and use as object names `/files/{groupId}/{id}`.

### Tests

- Initially a smoke test on the controller-level that checks that the right number of URLs are returned

chaoran-chen:
_This PR is targeting the `file-sharing` feature branch and not the main branch. #3912 should be merged first._

resolves #3905

Preview: https://files-request-api.loculus.org

This implements the `/files/request-uploads` endpoint in the backend. To use it, the backend expects the following config:

```
    "s3": {
        "enabled": true,
        "bucket": {
            "endpoint": "hel1.your-objectstorage.com",
            "region": null,
            "bucket": "loculus-dev-preview-private",
            "accessKey": "....",
            "secretKey": "...."
        }
    },
```

`buckets` is only required if `enabled=true`.

Remaining tasks as part of this PR:

- [x] Add API Swagger documentation
- [x] Write a test as specified in the issue
- [x] Extend the Helm template for the config
- ~[ ] Configure the bucket (created in #3903) for the preview server~ The sealed secrets are not working, we'll do this in a later step.
- [x] Maybe change the config to only support one bucket? Or leave it as-is for now and decide later whether we need/want to use a public and a private bucket or one bucket for everything (see https://github.com/loculus-project/loculus/issues/3903#issuecomment-2763694670).


backend/src/test/kotlin/org/loculus/backend/controller/files/FilesClient.kt
```diff
@@ -0,0 +1,22 @@
...
+
+    fun requestUploads(groupId: Int?, numberFiles: Int?, jwt: String = jwtForDefaultUser): ResultActions {
+        var request = post("/files/request-uploads")
+            .withAuth(jwt)
+        if (groupId != null) {
+            request = request.param("groupId", groupId.toString())
+        }
+        if (numberFiles != null) {
+            request = request.param("numberFiles", numberFiles.toString())
+        }
```
chaoran-chen, line 19: Ah, just found out that `.param()` directly changes the object, so we can instead write

```suggestion
        groupId?.let { request.param("groupId", it.toString()) }
        numberFiles?.let { request.param("numberFiles", it.toString()) }
```

```suggestion
        val request = post("/files/request-uploads")
            .withAuth(jwt)
        groupId?.let { request.param("groupId", it.toString()) }
        numberFiles?.let { request.param("numberFiles", it.toString()) }
```

chaoran-chen:
ok, then let's change it now!

backend/src/test/kotlin/org/loculus/backend/controller/files/RequestUploadsEndpointTest.kt
```diff
@@ -0,0 +1,68 @@
...
+            .andExpect(status().isBadRequest)
+    }
+
+    @Test
+    fun `GIVEN a request with groupId of a different or non-existent group THEN fails`() {
+        val groupId = groupManagementClient.createNewGroup().andGetGroupId() + 1
+        client.requestUploads(groupId = groupId, numberFiles = 1)
+            .andExpect(status().is4xxClientError)
+    }
+}
```
fhennig, line 68: Should we maybe also test that you can't create a file for a group that you're not a part of?

backend/src/main/kotlin/org/loculus/backend/controller/FilesController.kt
```diff
@@ -0,0 +1,47 @@
...
+@Validated
+@SecurityRequirement(name = "bearerAuth")
+class FilesController(
+    private val filesDatabaseService: FilesDatabaseService,
+    private val s3Service: S3Service,
+    private val groupManagementPreconditionValidator: GroupManagementPreconditionValidator,
+) {
+
+    @Operation(description = "Requests S3 presigned URLs to upload files")
+    @PostMapping("/request-uploads")
```
fhennig, line 29: Should we maybe actually do `/request-upload` (not plural)?

backend/src/test/kotlin/org/loculus/backend/controller/files/RequestUploadsEndpointTest.kt
chaoran-chen: Sure, good idea!

backend/src/main/kotlin/org/loculus/backend/controller/FilesController.kt
chaoran-chen: okay, sure

kubernetes/loculus/values.schema.json
```diff
@@ -1083,6 +1083,39 @@
...
+            "endpoint": {
+              "groups": ["s3"],
+              "type": "string",
+              "description": "TODO"
+            },
+            "region": {
+              "groups": ["s3"],
+              "type": "string",
+              "description": "TODO"
+            },
```
chaoran-chen, line 1106: Is this correctly typed as optional? and the other fields as non-optional?

backend/src/main/kotlin/org/loculus/backend/controller/FilesController.kt
chaoran-chen: Done: 3a365999e5c047f06ade60c61ae1513dae8b2875

backend/src/test/kotlin/org/loculus/backend/controller/files/RequestUploadsEndpointTest.kt
chaoran-chen: Done: 3a365999e5c047f06ade60c61ae1513dae8b2875

fhennig (approved):

PR #3922, opened 2025-03-31
feat(backend): accept file mapping in /submit and /revise endpoints

### LINKED ISSUE #3906: Update submit and revise endpoints to support receiving file IDs
When users call the submit or revise endpoints, they should be able to include a file mapping in the body. For each "file column", it is possible to have one file or a list of files.

```
  {
    submissionId1: {
      raw_reads: [
         { name: 'raw_reads.r1.fastq', id 1},
         { name: 'raw_reads.r2.fastq', id: 2}
      ],
      celebration_dance_video: [
          { name: 'celebration_dance_video.mp4', id: 3}
      ]
    },
  }
```

- The endpoint will check that the ID exists and belongs to the right group.
- The file information will be stored in the `original_data` JSON blob: `{..., files: <the JSON object in submissionId1 from above>}`
- For now, the backend won't check whether the file really exists (i.e., the pipeline should expect 404)
- It also, for now, will not validate against the config.

**Edit:** For simplicity, I suggest initially only allowing lists of files. I adapted the code example above

chaoran-chen:
_This PR is targeting the `file-sharing` feature branch and not the main branch._

resolves #3906

To test this:

```bash
curl -X 'POST' \
  'http://localhost:8079/dummy-organism/submit?groupId=1&dataUseTermsType=OPEN' \
  -H 'accept: application/json' \
  -H 'Authorization: Bearer <token>' \
  -H 'Content-Type: multipart/form-data' \
  -F 'metadataFile=@metadata.tsv;type=text/tab-separated-values' \
  -F 'sequenceFile=@sequences.fasta' \
  -F 'fileMapping={"seq1":{"dance_video":[{"fileId":"550e8400-e29b-41d4-a716-446655440000","name":"dance.mp4"}]}}'
```

### Remaining tasks

- [x] Implement validation that file ids exist 
- [x] Implement validation that file ids belong to the right group
- [x] Implement validation that file submission IDs match metadata submission IDs
- [x] Fix tests
- [x] Add tests for all the validation steps

🚀 Preview: https://fs-submit-api.loculus.org

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
@@ -265,6 +265,10 @@ data class OriginalData<SequenceType>(
         description = "The key is the segment name, the value is the nucleotide sequence",
     )
     val unalignedNucleotideSequences: Map<SegmentName, SequenceType?>,
+    @Schema(
+        description = "TODO",
+    )
+    val files: FileColumnNameMap? = null,
 )
 
 data class AccessionVersionOriginalMetadata(
--- a/backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
@@ -8,6 +8,8 @@ import com.fasterxml.jackson.databind.JsonDeserializer
 import com.fasterxml.jackson.databind.JsonNode
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize
 import io.swagger.v3.oas.annotations.media.Schema
+import org.loculus.backend.model.SubmissionId
+import org.loculus.backend.service.files.FileId
 import org.loculus.backend.utils.Accession
 import org.loculus.backend.utils.Version
 import org.springframework.core.convert.converter.Converter
@@ -343,3 +345,9 @@ class CompressionFormatConverter : Converter<String, CompressionFormat> {
     }
         ?: throw IllegalArgumentException("Unknown compression: $source")
 }
+
+typealias SubmissionIdFilesMap = Map<SubmissionId, FileColumnNameMap>
+
+typealias FileColumnNameMap = Map<String, List<FileIdAndName>>
+
+data class FileIdAndName(val fileId: FileId, val name: String)
--- a/backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
@@ -352,9 +352,9 @@ class CompressionFormatConverter : Converter<String, CompressionFormat> {
 
 typealias SubmissionIdFilesMap = Map<SubmissionId, FileColumnNameMap>
 
-fun getAllFileIds(submissionIdFilesMap: SubmissionIdFilesMap): List<FileId> = submissionIdFilesMap.values.flatMap {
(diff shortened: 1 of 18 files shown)


fhennig:
I've changed the type of the RequestPart - the parsing is now done by Spring itself. To submit, add `type=application/json` to your `-F` param in curl (I've updated it in the PR description).

chaoran-chen:
Ah, nice, I was struggling with it and didn't know about "add `type=application/json` to your `-F` param in curl". How is it with the Swagger API docs? I believe to remember that the parameter disappeared when I changed the type away from String.

fhennig:
Grrr annoying. Yes you're right. It does disappear. I managed to get it to reappear by not using a typealias but a class instead, and then adding a "@Schema" annotation to that. But the annotation cannot be added to typealias. I tried a bunch of annotation combinations now and can#t get it to work. maybe it's easier to go back to how you did it. But I'm not very happy about this, why doesn't it just work?! having to manually do the object mapping on a string is a bit silly.

fhennig:
Ah! I can't request a review from you @chaoran-chen , but I think it'd still be good if you had a look

backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
```diff
@@ -343,3 +349,13 @@ class CompressionFormatConverter : Converter<String, CompressionFormat> {
     }
         ?: throw IllegalArgumentException("Unknown compression: $source")
 }
+
+typealias SubmissionIdFilesMap = Map<SubmissionId, FileColumnNameMap>
+
+fun SubmissionIdFilesMap.getAllFileIds(): Set<FileId> = this.values.flatMap {
+    it.values
+}.flatten().map { it.fileId }.toSet()
```
chaoran-chen, line 357: Embrace the extension functions of Kotlin! ;)

backend/src/main/kotlin/org/loculus/backend/model/SubmitModel.kt
```diff
@@ -400,6 +361,58 @@ class SubmitModel(
...
+        }
+
+        log.debug {
+            "Validating that submitted files belong to the group that their associated submission belongs to."
+        }
+        if (submissionParams is SubmissionParams.OriginalSubmissionParams) {
+            fileGroups.forEach {
+                if (it.value != submissionParams.groupId) {
+                    throw BadRequestException(
+                        "The File ${it.key} does not belong to group ${submissionParams.groupId}.",
```
chaoran-chen, line 395: I changed it to not print the group the file currently belongs to to avoid "leaking" this information.

chaoran-chen:
Nice! I made a few minor changes to the code (https://github.com/loculus-project/loculus/pull/3922/files/3decea37a5c522b013e6d536a2c0149daa5053a4..1de00044e20b4db51a9f35ab4176d0317e5f2503) but otherwise, everything looks good to me and I also manually tested the submission.

backend/src/main/kotlin/org/loculus/backend/model/SubmitModel.kt
fhennig: Oh, good idea!

backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
fhennig: Oh I really wanted to do something like that, but ChatGPT didn't tell me about it. Great stuff!!

fhennig:
Great! I'm merging.

PR #3924, opened 2025-04-01
fix preview bucket config secret

fhennig:
I didn't use the correct scope before when sealing the secret, I hope now it should work 🤞 

preview URL: https://fs-fix-preview-sealed-sec.loculus.org


fhennig:
I looked at the cluster, and this time the secret was unsealed! It worked. Now something else seems to be not working.

kubernetes/loculus/values.yaml
```diff
@@ -26,7 +26,10 @@ dataUseTerms:
     open: https://#TODO-MVP/open
     restricted: https://#TODO-MVP/restricted
 s3:
-  enabled: false
+  enabled: true
+  bucket:
+    endpoint: hel1.your-objectstorage.com
+    bucket: loculus-preview-private
```
chaoran-chen, line 32: I changed the schema to require accessKey and secretKey if the feature is enabled. Could that be causing issues?

fhennig: Ah probably also that. But also I think the config wouldn't have read the env vars anyways? Maybe with the latest two commits it will work. I set both of the settings, but also in such a way that the config preprocessor should handle the replacing.

fhennig:
It's working! I have requested a URL and uploaded a file to the bucket!

kubernetes/README.md
```diff
@@ -134,6 +134,18 @@ We do not currently support branch names containing characters that can't go in
 
 For preview instances this repo contains [sealed secrets](https://sealed-secrets.netlify.app/) that allow the loculus-bot to access the GitHub container registry and (separately) the GitHub repository. These are encrypted such that they can only be decrypted on our cluster but are cluster-wide so can be used in any namespace.
 
+### Adding a sealed secret
+
+Create a secret, for example like this:
+
+    kubectl create secret generic my-secret --from-literal=accessKey=ACC3SK3Y --from-literal=secretKey=S3Cr37K3Y --dry-run=client -o yaml > my-secret.yaml
```
chaoran-chen, line 141:
```suggestion
    kubectl create secret generic my-secret --from-literal=accessKey=<secret> --from-literal=secretKey=<secret> --dry-run=client -o yaml > my-secret.yaml
```

kubernetes/loculus/values.yaml
```diff
@@ -26,7 +26,12 @@ dataUseTerms:
...
+    accessKey: "[[s3accessKey]]"
+    secretKey: "[[s3secretKey]]"
```
chaoran-chen, line 34: Can we move this to the preview values? For the default values.yaml, we should disable it.

chaoran-chen (approved):
Nice!

kubernetes/loculus/values.yaml
fhennig: https://github.com/loculus-project/loculus/pull/3924/commits/603703496178b0dd7879e12469e0e1c717b61427

PR #3926, opened 2025-04-01
chore: remove comments from PR template

theosanderson:
We now have Squash merges defaulting to include PR description, which yields something like https://github.com/loculus-project/loculus/commit/1b944de935fe6b5b3dd9cd9895528d1de0f689b2 

image

which is very saturated with comments from the PR template. Obviously we had these for a reason but they could be something we describe in Nuclino or docs instead. My preference would be to remove them, as here, but I expect some may disagree. Just an idea.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,7 +1,5 @@
 resolves #
 
-preview URL: <branchname>.loculus.org
-
 ### Summary
 
 ### Screenshot
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,6 +1,6 @@
 resolves #
 
-preview URL:
+preview URL: <branchname>.loculus.org
 
 ### Summary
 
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -1,17 +1,12 @@
-<!-- Mention the issue that this PR resolves. Delete if there is no corresponding issue -->
 resolves #
 
-<!-- Add "preview" label in almost all cases to have testable deployment. Then lookup the deployment URL in argocd and paste it here (if you don't know how to look up the URL, ask here: https://loculus.slack.com/archives/C06JCAZLG14), it's something like `{REPLACE}.loculus.org` -->
 preview URL:
```
chaoran-chen, line 3: That might be helpful?

```suggestion
preview URL: <branchname>.loculus.org
```

chaoran-chen (approved):

theosanderson:
I'm going to merge this but we can still iterate in later PRs

PR #3936, opened 2025-04-02
Fix k8s config templating

fhennig:
resolves #

### PR Checklist


fhennig:
I've added Theos upload file component, but it uses an S3 lib and we can't directly use that.

theosanderson:
> I've added Theos upload file component, but it uses an S3 lib and we can't directly use that.

https://github.com/loculus-project/loculus/pull/3860 doesn't use an S3 lib in the client

fhennig:
Ah cool! Thanks! Can you teach your LLM to deal with our linter? It'd be neat if it could fix ther linter warnings on its own

fhennig:
I'm clocking out for today. I'm not quite sure what the best way to go about integrating the PoC into what we have here now - If you @theosanderson have a good idea, feel free to keep going!

Otherwise maybe tomorrow I'll first start by making the "request-uploads" api more available and then try to puzzle it together with the UI you already built.

theosanderson:
> Ah cool! Thanks! Can you teach your LLM to deal with our linter? It'd be neat if it could fix ther linter warnings on its own

Yeah Claude code can do this. My aim with these PRs was to show what was possible as we were deciding where to implement things rather than making something for merging

fhennig:
Ok yeah. I just thought since you were so quick to do this maybe you wanted to keep going.

theosanderson:
> I'm not quite sure what the best way to go about integrating the PoC into what we have here now

Yes, if I were going about this my approach would be to jump straight to multi-part uploads which is what the PoC uses (but I think Chaoran was less keen on that and that's all good!)

I have no problem with the PoC being discarded or being used to prompt an LLM to build something new. I think the UX of the PoC is pretty good and we should stick similar to it, or consciously consider the deviations from it, but in terms of implementation I haven't even looked at the code in much detail

theosanderson:
Yes, my aim here wasn't really to make something as a base (but if it's serving as one I'm glad, and I do recommend Claude Code to all :) ) Sorry I'm pretty tied up at this conference apart from snippets of time

fhennig:
Ok thanks for this additional info!

I also do like the UI, looks neat.

fhennig:
I just cobbled together the API calls for now, it's not actually uploading the folder, just the first file out of the folder.

I just wanted to get it through for one file at first.

fhennig:
@chaoran-chen It's not perfect yet, but I think it's good to merge for now! Revision isn't working yet - I think we need to think about how to best do the revision stuff, not sure yet.

website/src/components/Submission/DataUploadForm.tsx
```diff
@@ -229,6 +260,47 @@ const InputModeTabs = ({
...
+    group: Group;
+    fileFields: FileField[];
+    setFileMapping: Dispatch<SetStateAction<FileMapping | undefined>>;
+    onError: (message: string) => void;
+}) => {
+    return (
+        <div className='grid sm:grid-cols-3 gap-x-16 gap-y-4'>
+            <div>
+                <h2 className='font-medium text-lg'>Extra files</h2>
+                <p className='text-gray-500 text-sm'>Lorem Ipsum ...</p>
```
chaoran-chen, line 284:
```suggestion
```

fhennig:
@chaoran-chen can you review again?

chaoran-chen (approved):

PR #3964, opened 2025-04-08
feat(website): some small improvements to organism card (rev 2)

theosanderson:
Adds a shadow to organism cards and redesigns them in various small ways. Uses Felix's original number formatting.

image


fhennig:
Have you tried using the default shadow classes that come with tailwind? https://tailwindcss.com/docs/box-shadow IMO if we could to stick with what's there out of the box it'll be easier to stay consistent. (But since this is almost the only shadow we have, I'm fine with maybe not doing it)

theosanderson:
Yes - I started with tailwind classes then chose these because I thought they looked nicer (more subtle)

fhennig:
Alright then!

PR #3943, opened 2025-04-03
feat(backend): include file URLs in /extract-unprocessed-data

chaoran-chen:
_This PR is targeting the `file-sharing` feature branch and not the main branch._

resolves #3909

### PR Checklist
- [ ] Adapt preprocessing pipeline specification
- [ ] Add tests

PR #3967, opened 2025-04-09
Publish files for on approval

fhennig:
Adds a new function to fetch all the File IDs for files that were approved, and then iterates through them and sets them to public using the tag `public=true`.

For this to work, the S3 needs to be configured with ah policy:

```json
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect":     "Allow",
      "Action":     "s3:GetObject",
      "Resource":    "arn:aws:s3:::loculus-preview-private/*",
      "Principal":   "*",
      "Condition": {  "StringEquals": {"s3:ExistingObjectTag/public": "true" } }
    }
  ]
}
```


backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
theosanderson, line 1: Q not really related to this PR but just prompted by it [because I guess I was previously operating under a all-files-are-public-but-some-are-hidden-by-not-knowing-the-url model, or hadn't thought about it]: have we specced out how the preprocessing pipeline will download files to work on (eventually)? Will we use presigned URLs?

chaoran-chen: Yes, the pipeline will get presigned URLs and it's already implemented: https://github.com/loculus-project/loculus/pull/3943

theosanderson: Thanks!

PR #3970, opened 2025-04-09
Add backend endpoint to get a redirect to a presigned URL

fhennig:
resolves #

### PR Checklist

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/controller/FilesController.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/controller/FilesController.kt
@@ -3,13 +3,19 @@ package org.loculus.backend.controller
 import io.swagger.v3.oas.annotations.Operation
 import io.swagger.v3.oas.annotations.Parameter
 import io.swagger.v3.oas.annotations.security.SecurityRequirement
+import jakarta.servlet.http.HttpServletResponse
 import org.loculus.backend.api.FileIdAndUrl
 import org.loculus.backend.auth.AuthenticatedUser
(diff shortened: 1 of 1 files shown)


fhennig:
After all our discussions, I still think we want something like that (or in the future, without the groupId), right?

We can also ditch the PR; I just had it done already.

theosanderson:
(no strong feelings from me either way - I guess the change to remove the group id, but still do auth for private files, would be quite substantial so we likely won't use this as is in prod, but if it's useful for now, works for me)

fhennig:
I added the other endpoint as well, the one based on accession - that would the be something we would need in the future as well right?

I couldn't (quickly) figure out how to make the `AuthenticatedUser` optional. Just adding a `?` didn't work.

It might still be helpful for later.

backend/src/main/kotlin/org/loculus/backend/controller/FilesController.kt
```diff
@@ -61,4 +65,24 @@ class FilesController(
         response.sendRedirect(presignedUrl)
         return ResponseEntity.status(HttpStatus.FOUND).build() // 302 Redirect
     }
+
+    @GetMapping("/{accession}/{version}/{fileField}/{fileName}")
+    fun getFileUrl2(
+        @HiddenParam authenticatedUser: AuthenticatedUser,
```
theosanderson, line 71: I don't really know how the kotlin code works, but just to flag that we definitely need to make sure that this user can access this file at some point :) [not blocking for dev stuff, just important to keep track of very carefully before going to prod]

Issue #3907, opened 2025-03-27
Define file-sharing schema config values.yaml


chaoran-chen:
The values.yaml schema should be extended to contain config for file sharing. For the preview (during development), let's add for the dummy organism:

```yaml
schema:
  submissionFiles: --> What the user uploads
    - name: raw_reads
      multipleFiles: false
  publicFiles: --> What goes into SILO
    - name: raw_reads
      multipleFiles: false
    - name: aligned_reads
      multipleFiles: false
    - name: qc_plots
      multipleFiles: true
```

theosanderson:
We already have `submissionDataTypes` which was originally intended (in broad strokes) to include what is described as `submissionFiles` here.

I'm not clear on the specifics - in reality raw_reads probably would have multiple files? Do we definitely need this property at all?

theosanderson:
I guess `submissionDataTypes` in significant part is used to enable/disable the submission UI elements on the bulk page (and others), so that's the sense in which I imagine this fitting in there. Here essentially I imagine we are adding a "Raw reads" (we probably as ever need a display name) component to the submission form.

Would I imagine that segment data would be added to `submissionDataTypes`? Probably not, because the segment data is used in lots of other places too beyond submission and preprocessing, whereas I imagine this would essentially just be used there. (Some pipelines might assemble the read data and not make the underlying reads public, just the consensus).

chaoran-chen:
How would you then type `submissionDataTypes` exactly? Accepting boolean and a list of values would be a bit strange. Maybe something like the following?

```
submissionDataTypes:
    files:
        enabled: <boolean>
        fields:
            - name: raw_reads
               multipleFiles: false
    consensusSequences:
        enabled: <boolean>
        <would we envision to have more things here?>
```

theosanderson:
No strong feelings from me: your approach works for me. (Personally I don't mind the idea of something which can be a config object or can be `false/null` but I also don't prefer it to anything else, and it sounds like you do have a preference!)

fhennig:
What about the output files? If we do

```yaml
submissionDataTypes:
    files:
        enabled: <boolean>
        fields:
            - name: raw_reads
              multipleFiles: false
    consensusSequences:
        enabled: <boolean>
        <would we envision to have more things here?>
```

Where do we define

```yaml
  publicFiles: --> What goes into SILO
    - name: raw_reads
      multipleFiles: false
    - name: aligned_reads
      multipleFiles: false
    - name: qc_plots
[...]

PR #3978, opened 2025-04-11
File sharing

### LINKED ISSUE #3344: File/Raw data sharing MVP

chaoran-chen:
Resolves #3344

This PR adds the _file sharing_ functionality to Loculus in its initial implementation. The file sharing feature allows users to submit additional files on top of the usual metadata and sequence files. The files are stored in S3 and tracked in an additional database table. The files are passed to SILO as links, and then displayed on the sequence details page as such.

### Backend
- A new files controller is added, which supports generating presigned URLs, which can then be used to upload files. The `requestUpload` endpoint returns a file ID and a URL to upload a file content. The file ID is then used to pass the file in with other sequence data.
- Various endpoints have been adjusted to either accept or return file mappings. On submission, on giving data to the preprocessing pipeline as well as accepting data, and also for the `get-released-data` endpoint, they all now handle file mappings.
- The `get-released-data` endpoint contains a new generated metadata field for every file category, which has the same name as the file category. In it is a JSON list of the files that are associated with the sequence.

### Frontend
- A new `FolderUploadComponent` was added, which allows a user to upload a whole folder at once. The component requests presigned URLs from the backend, and then uploads the user submitted files. Afterwards, the file IDs are submitted to the backend together with the sequence metadata etc.
- In the sequence details view, the files are rendered with their original file name, and can be downloaded from there.

### Deployment/values.yaml
- A new section for S3 configuration was added. You can configure your own external S3, or use the `runDevelopmentS3` flag to deploy a MinIO instance. Some plumbing was added to support deploying MinIO.
- For every organism, you can now configure input and output files.

### Testing
- The backend has tests for the new endpoints, and some tests have been extended to check file related things.
- The backend can now run a MinIO Testcontainer which is also already used by some tests.
- An integration test has been added to test the whole file sharing feature end-to-end (upload to viewing an uploaded file).

---

### Manual testing

Test data: [test_data.zip](https://github.com/user-attachments/files/20205931/test_data.zip)

I've tested submission of test data through the single submission dialog as well as bulk submission. I've then checked that the files are actually displayed in the sequences, and that they can be downloaded as well (i.e. the S3 permissions work).

🚀 Preview: https://file-sharing.loculus.org


fhennig:
currently there are CORS errors again when trying to upload. MinIO is sending the correct response to an OPTIONS call, but traefik seems to not forward the headers. I researched this a bit, but wasn't satisfied with the solutions I found so far. possibly one needs to configure a middleware for this to work, but that seems weird to me.

backend/src/main/kotlin/org/loculus/backend/service/files/S3Service.kt
```diff
@@ -0,0 +1,107 @@
...
+    private var internalClient: MinioClient? = null
+
+    fun createUrlToUploadPrivateFile(fileId: FileId): String {
+        val config = getS3BucketConfig()
+        return getClient().getPresignedObjectUrl(
+            GetPresignedObjectUrlArgs.builder()
+                .method(Method.PUT)
+                .bucket(config.bucket)
+                .`object`(getFileName(fileId))
+                .expiry(PRESIGNED_URL_EXPIRY_SECONDS, TimeUnit.SECONDS)
```
theosanderson, line 26: IMO we should allow more time than this for upload. For upload I can imagine a scenario in which one requests upload URLs for e.g. 1000 files and then uploads them one by one. It might take more than 30 minutes to get to the 1000th file. In contrast for downloads, because IMO we should always use the canonical URL which generates a new presigned URL afresh this probably isn't important (although I guess it might be for resuming downloads, not sure).

kubernetes/loculus/templates/_config-processor.tpl
```diff
@@ -52,6 +52,16 @@
...
+    - name: LOCULUSSUB_s3accessKey
+      valueFrom:
+        secretKeyRef:
+          name: s3-bucket
+          key: accessKey
+    - name: LOCULUSSUB_s3secretKey
+      valueFrom:
+        secretKeyRef:
+          name: s3-bucket
+          key: secretKey
```
theosanderson, line 64:
```suggestion
    - name: LOCULUSSUB_s3accessKey
      valueFrom:
        secretKeyRef:
          name: s3-bucket
          key: accessKey
    - name: LOCULUSSUB_s3secretKey
      valueFrom:
        secretKeyRef:
          name: s3-bucket
          key: secretKey
```

IIRC we discussed that we don't need/use these

kubernetes/loculus/values.schema.json
```diff
@@ -245,14 +245,65 @@
                   "type": "boolean",
                   "default": true,
                   "description": "If false, the submission form will not allow submission of consensus sequences (i.e. the sequences file must be omitted). All consensus sequence related parts on the website will be hidden."
+                },
+                "files": {
+                  "type": "object",
+                  "properties": {
+                    "enabled": {
+                      "type": "boolean",
+                      "description": "TODO"
```
theosanderson, line 254: Should we make issues for these?

backend/src/main/kotlin/org/loculus/backend/service/files/S3Service.kt
```diff
@@ -0,0 +1,107 @@
...
+                args.extraQueryParams(
+                    mapOf(
+                        "response-content-disposition" to "attachment; filename=\"$downloadFileName\"",
+                    ),
+                )
+        }
+        return getClient().getPresignedObjectUrl(args.build())
+    }
+
+    fun createPublicUrl(fileId: FileId): String {
```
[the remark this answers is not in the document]
fhennig: yes, get seems more appropriate - renamed it: https://github.com/loculus-project/loculus/pull/3978/commits/94bc7a691cdd856d931f57f447e2cfb991d78326

kubernetes/loculus/templates/_config-processor.tpl
fhennig: yes! thanks. https://github.com/loculus-project/loculus/pull/3978/commits/7d26f497c88572095a793fdca60362ce190a8462

backend/src/main/kotlin/org/loculus/backend/service/files/S3Service.kt
fhennig: Valid point and good to revisit this number - I'm not sure what's best. I've added a note to our board about it, also tagging @chaoran-chen so he can thing about it as well.

I think maybe we can just leave it as is though - we might want to refine the whole upload interaction in general. We probably want multi-part upload and we probably also want batching. The client should maybe also request URLs in batches (would be my first intuition).

Either way, let's revisit and I'll resolve it for now.

kubernetes/loculus/values.schema.json
fhennig: Added the docs, previewable here: https://docs-file-sharing.loculus.org/reference/helm-chart-config/#organism-schema-type

backend/docs/db/schema.sql
```diff
@@ -266,6 +266,21 @@ CREATE VIEW public.external_metadata_view AS
 
 ALTER VIEW public.external_metadata_view OWNER TO postgres;
 
+--
+-- Name: files; Type: TABLE; Schema: public; Owner: postgres
+--
+
+CREATE TABLE public.files (
+    id uuid NOT NULL,
+    requested_at timestamp without time zone NOT NULL,
```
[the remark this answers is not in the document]
fhennig: good call! I've renamed it to "upload_requested_at" (in line with the "uploader" field for the requesting user).

backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
```diff
@@ -263,8 +270,16 @@ data class OriginalData<SequenceType>(
         description = "The key is the segment name, the value is the nucleotide sequence",
     )
     val unalignedNucleotideSequences: Map<SegmentName, SequenceType?>,
+    @Schema(
+        description = "TODO",
```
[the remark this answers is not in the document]
fhennig: oops, forgot something! will do now.

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
```diff
@@ -58,11 +62,12 @@ import org.springframework.web.bind.annotation.RequestBody
 import org.springframework.web.bind.annotation.RequestHeader
 import org.springframework.web.bind.annotation.RequestMapping
 import org.springframework.web.bind.annotation.RequestParam
+import org.springframework.web.bind.annotation.RequestPart
 import org.springframework.web.bind.annotation.ResponseStatus
 import org.springframework.web.bind.annotation.RestController
 import org.springframework.web.multipart.MultipartFile
 import org.springframework.web.servlet.mvc.method.annotation.StreamingResponseBody
-import java.util.UUID
+import java.util.*
```
rneher, line 70: do we really need a wildcard import here?

.github/workflows/e2e-k3d.yml
```diff
@@ -34,7 +34,7 @@ jobs:
     env:
       ALL_BROWSERS: ${{ github.ref == 'refs/heads/main' || github.event.inputs.all_browsers && 'true' || 'false' }}
       sha: ${{ github.event.pull_request.head.sha || github.sha }}
-      wait_timeout: ${{ github.ref == 'refs/heads/main' && 900 || 240 }}
+      wait_timeout: ${{ github.ref == 'refs/heads/main' && 900 || 600 }}
```
corneliusroemer, line 37: How come tests now take twice as long?

backend/docs/db/schema.sql
```diff
@@ -266,6 +266,21 @@ CREATE VIEW public.external_metadata_view AS
...
+--
+-- Name: files; Type: TABLE; Schema: public; Owner: postgres
+--
+
+CREATE TABLE public.files (
+    id uuid NOT NULL,
+    upload_requested_at timestamp without time zone NOT NULL,
+    uploader text NOT NULL,
+    group_id integer NOT NULL,
+    published_at timestamp without time zone
```
corneliusroemer, line 278: What does `published_at` mean? When the user uploaded? When it was made public, i.e. released?

theosanderson: When the file was made public - N.B. multiple sequence entry versions can refer to one file so there isn't an exact correspondence with sequence entry release status

backend/src/main/kotlin/org/loculus/backend/api/FileTypes.kt
```diff
@@ -0,0 +1,28 @@
...
+ */
+package org.loculus.backend.api
+
+import io.swagger.v3.oas.annotations.media.Schema
+import org.loculus.backend.service.files.FileId
+
+data class FileIdAndUrl(
+    @Schema(description = "The id of the file.", example = "8D8AC610-566D-4EF0-9C22-186B2A5ED793") val fileId: FileId,
+    @Schema(
+        description = "A presigned URL",
```
corneliusroemer, line 12: What actions are authorized by the presigned URL? Only read, or also write? Should the authorization information be a property of this class?

backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
```diff
@@ -263,8 +271,17 @@ data class OriginalData<SequenceType>(
...
     )
     val unalignedNucleotideSequences: Map<SegmentName, SequenceType?>,
+    @Schema(
+        example = """{"raw_reads": [{"fileId": "f1le-uuId-asdf", "name": "myfile.fastaq"]}""",
+        description = "A map from file field names, to lists of files. The files can also have URLs.",
+    )
+    val files: Map<String, List<FilesType>>? = null,
 )
 
+typealias OriginalData<SequenceType> = OriginalDataInternal<SequenceType, FileIdAndName>
```
corneliusroemer, line 281: I don't know much Kotlin but this looks strange to me, a generic type that is in fact more generic than it seems?

backend/src/main/kotlin/org/loculus/backend/config/S3Config.kt
```diff
@@ -0,0 +1,12 @@
+package org.loculus.backend.config
+
+data class S3Config(val enabled: Boolean = false, val bucket: S3BucketConfig?)
+
+data class S3BucketConfig(
+    val endpoint: String,
+    val internalEndpoint: String?,
+    val region: String?,
```
corneliusroemer, line 8: region can be null here, but you require it to be non-null here: https://github.com/loculus-project/loculus/pull/3978/files#diff-0d594f2e0976102924e662a789f1b90ad49f0e773d976da2304fc8f0e203472fR125

backend/src/main/kotlin/org/loculus/backend/config/BackendSpringConfig.kt
```diff
@@ -101,6 +109,25 @@ class BackendSpringConfig {
...
+        @Value("\${${BackendSpringProperty.S3_BUCKET_INTERNAL_ENDPOINT}:#{null}}") internalEndpoint: String? = null,
+        @Value("\${${BackendSpringProperty.S3_BUCKET_REGION}}") region: String? = null,
+        @Value("\${${BackendSpringProperty.S3_BUCKET_BUCKET}}") bucket: String? = null,
+        @Value("\${${BackendSpringProperty.S3_BUCKET_ACCESS_KEY}}") accessKey: String? = null,
+        @Value("\${${BackendSpringProperty.S3_BUCKET_SECRET_KEY}}") secretKey: String? = null,
+    ): S3Config {
+        if (!enabled) {
+            return S3Config(false, null)
+        }
+        if (endpoint != null && region != null && bucket != null && accessKey != null && secretKey != null) {
```
corneliusroemer, line 125: Region could be null per S3BucketConfig, see below, so could be removed here, or the nullability could be removed below

backend/src/main/kotlin/org/loculus/backend/controller/FilesController.kt
```diff
@@ -0,0 +1,92 @@
...
+        @HiddenParam authenticatedUser: AuthenticatedUser,
+        @Parameter(description = GROUP_ID_DESCRIPTION) @RequestParam groupId: Int,
+        @Parameter(description = "Number of URLs, default is 1") @RequestParam numberFiles: Int = 1,
+    ): List<FileIdAndUrl> {
+        groupManagementPreconditionValidator.validateUserIsAllowedToModifyGroup(
+            groupId,
+            authenticatedUser,
+        )
+        val response = mutableListOf<FileIdAndUrl>()
+        for (i in 0 until numberFiles) {
```
corneliusroemer, line 85:
```suggestion
        repeat(numberFiles) {
```

idiomatic Kotlin

backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
```diff
@@ -344,3 +361,25 @@ class CompressionFormatConverter : Converter<String, CompressionFormat> {
...
 }
+
+typealias SubmissionIdFilesMap = Map<SubmissionId, FileColumnNameMap>
+
+fun SubmissionIdFilesMap.getAllFileIds(): Set<FileId> = this.values.flatMap {
+    it.values
+}.flatten().map { it.fileId }.toSet()
+
+/**
+ * A map from fileFieldIds to lists of file IDs and names.
```
corneliusroemer, line 372: What are `fileFieldIds`? I can't find the term used anywhere else in the PR, is this the "type" of file like "raw_reads"?

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
```diff
@@ -108,12 +116,15 @@ open class SubmissionController(
                 innerDataUseTermsType = dataUseTermsType
             }
         }
-
+        val fileMappingParsed = fileMapping?.let {
```
corneliusroemer, line 119: Should probably throw BadRequestException here saying that the fileMapping could not be parsed.

backend/src/main/kotlin/org/loculus/backend/model/SubmitModel.kt
```diff
@@ -132,6 +150,9 @@ class SubmitModel(
         uploadDatabaseService.deleteUploadData(uploadId)
     }
 
+    /**
+     * Inserts the uploaded metadata (and sequence data) into the 'aux' tables in the database.
+     */
```
corneliusroemer, line 155: Good comment, when I saw the function name on its own I had no clue what it was doing. We should probably improve the name to something like `insertDataIntoAux`

backend/src/main/kotlin/org/loculus/backend/model/SubmitModel.kt
```diff
@@ -335,6 +361,58 @@ class SubmitModel(
         }
     }
 
+    private fun validateSubmissionIdSetsForFiles(metadataKeysSet: Set<SubmissionId>, filesKeysSet: Set<SubmissionId>) {
+        val filesKeysNotInMetadata = filesKeysSet.subtract(metadataKeysSet)
+        if (filesKeysNotInMetadata.isNotEmpty()) {
+            throw UnprocessableEntityException(
+                "Upload contains ${filesKeysNotInMetadata.size} submissionIds that are not present in the metadata " +
```
corneliusroemer, line 368:
```suggestion
                "File upload contains ${filesKeysNotInMetadata.size} submissionIds that are not present in the metadata " +
```

backend/src/main/kotlin/org/loculus/backend/service/files/FilesDatabaseService.kt
```diff
@@ -0,0 +1,52 @@
...
+
+    fun getGroupIds(fileIds: Set<FileId>): Map<FileId, Int> =
+        FilesTable.select(FilesTable.idColumn, FilesTable.groupIdColumn)
+            .where { FilesTable.idColumn inList fileIds }
+            .associate { Pair(it[FilesTable.idColumn], it[FilesTable.groupIdColumn]) }
+
+    /**
+     * Set the publishing date for all given file IDs to now, if they are not set already.
+     */
+    fun publish(fileIds: Set<FileId>) {
```
corneliusroemer, line 37: I wonder if we should use "release" here to not introduce new terminology: publish/public vs release/released? Is there a reason you chose to introduce a new term?

backend/src/main/kotlin/org/loculus/backend/service/files/FilesTable.kt
```diff
@@ -0,0 +1,17 @@
...
+typealias FileId = UUID
+const val FILES_TABLE_NAME = "files"
+
+object FilesTable : Table(FILES_TABLE_NAME) {
+
+    val idColumn = uuid("id")
+    val uploadRequestedAtColumn = datetime("upload_requested_at")
+    val uploaderColumn = text("uploader")
+    val groupIdColumn = integer("group_id")
+    val publishedAtColumn = datetime("published_at").nullable()
```
corneliusroemer, line 16: Technically this is denormalized if files are public when the corresponding entry is released - though that would require a bunch of unpleasant joining (finding out what the corresponding entry is and what its release date is)

backend/src/main/kotlin/org/loculus/backend/service/files/S3Service.kt
```diff
@@ -0,0 +1,115 @@
...
+import io.minio.GetPresignedObjectUrlArgs
+import io.minio.MinioClient
+import io.minio.SetObjectTagsArgs
+import io.minio.http.Method
+import org.loculus.backend.config.S3BucketConfig
+import org.loculus.backend.config.S3Config
+import org.springframework.stereotype.Service
+import java.util.concurrent.TimeUnit
+
+private const val PRESIGNED_URL_EXPIRY_SECONDS = 60 * 30
```
corneliusroemer, line 12: This should probably be configurable in values.yaml

backend/src/main/kotlin/org/loculus/backend/service/files/S3Service.kt
corneliusroemer: I'll make an issue so it doesn't become forgotten

backend/src/main/kotlin/org/loculus/backend/service/submission/CompressionService.kt
```diff
@@ -8,7 +8,7 @@ import org.loculus.backend.api.ProcessedData
 import org.loculus.backend.config.BackendConfig
 import org.springframework.stereotype.Service
 import java.nio.charset.StandardCharsets
-import java.util.Base64
+import java.util.*
```
corneliusroemer, line 11: replace with non-wildcard

corneliusroemer, line 11:
```suggestion
import java.util.Base64
```

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -83,9 +92,8 @@ import org.springframework.transaction.annotation.Transactional
 import java.io.BufferedReader
 import java.io.InputStream
 import java.io.InputStreamReader
-import java.util.Locale
+import java.util.*
```
corneliusroemer, line 95: replace wildcard

backend/src/test/kotlin/org/loculus/backend/controller/files/RequestUploadEndpointTest.kt
```diff
@@ -0,0 +1,102 @@
...
+
+    @Test
+    fun `GIVEN a request with no groupId THEN returns bad request`() {
+        client.requestUploads(groupId = null, numberFiles = 1)
+            .andExpect(status().isBadRequest)
+    }
+
+    @Test
+    fun `GIVEN a request with groupId of a non-existent group THEN returns not found`() {
+        val groupId = groupManagementClient.createNewGroup().andGetGroupId() + 100
```
corneliusroemer, line 91: Maybe test that the group really doesn't exist, this assumes groupId + 100 doesn't exist but who knows

corneliusroemer (changes requested):
I've read about half the code so far, some observations to date (if I don't submit the review it gets lost occasionally hence submitting now)

What does the upload folder need to look like so files are correctly mapped? Some example test data would be great!

How would it work for preprocessing to upload a processed version?  IIUC, files need to be owned by the group that the sequence entry belongs to. So prepro would have to submit a file on behalf of that group - does that work?

Small things I noticed:
- File links are not underlined on sequence details page in contrast to all other links on that page
- Download file name is the file id (uuid) and not the file name, which is what I would expect

kubernetes/loculus/templates/_siloDatabaseConfig.tpl
```diff
@@ -35,6 +35,12 @@ schema:
       name: {{ .name }}
   {{- end }}
   {{- end }}
+  {{- if .files }}
+  {{- range .files }}
+    - type: string
+      name: {{ .name }}
```
corneliusroemer, line 41: Does this mean the fileField names must be different from all other fields passed to silo? We should probably document this

kubernetes/loculus/values.schema.json
```diff
@@ -245,14 +245,72 @@
...
+                      "type": "array",
+                      "description": "The file fields that can be submitted for this organism.",
+                      "items": {
+                        "type": "object",
+                        "placeholder": "fileField",
+                        "properties": {
+                          "name": {
+                            "groups": ["schema"],
+                            "type": "string",
+                            "description": "The name of the file field, i.e. 'raw_reads'."
```
corneliusroemer, line 269:
```suggestion
                            "description": "The name of the file field, e.g. 'raw_reads'."
```

kubernetes/loculus/values.schema.json
```diff
@@ -245,14 +245,72 @@
...
+                  },
+                  "multipleFiles": {
+                    "groups": ["schema"],
+                    "type": "boolean",
+                    "description": "TODO"
+                  }
+                },
+                "additionalProperties": false
+              }
+            },
```
corneliusroemer, line 313: This looks like a duplication of the above files section, we should avoid having duplicates as they make our already verbose config even bigger - also things could go bad if one accidentally only changes one not the other of the duplicate parts.

kubernetes/loculus/values.yaml
```diff
@@ -1421,6 +1426,72 @@ defaultOrganisms:
...
+      submissionDataTypes:
+        consensusSequences: false
+        files:
+          enabled: true
+          fields:
+            - name: raw_reads
+              multipleFiles: false
+      files:
+        - name: raw_reads
+          multipleFiles: false
```
corneliusroemer, line 1443: We should get rid of this duplication here, see comment above.

preprocessing/specification.md
```diff
@@ -39,12 +39,14 @@ To retrieve unpreprocessed data, the preprocessing pipeline sends a POST request
...
-{"accession": 2, "version": 1, "data": {"metadata": {...}, "unalignedNucleotideSequences": {...}}, "submitter": john_smith, ...}
+{"accession": 1, "version": 1, "data": {"metadata": {...}, "unalignedNucleotideSequences": {...}, "files": {...}}, "submitter": insdc_ingest_user, ...}
+{"accession": 2, "version": 1, "data": {"metadata": {...}, "unalignedNucleotideSequences": {...}, "files": {...}}, "submitter": john_smith, ...}
 ```
 
 The `metadata` field contains a flat JSON object in which all values are strings. The fields and values correspond to the columns and values as provided by the submitter. Fields present in the metadata file header but left empty by the submitter will be returned as "". Only columns present in the submission metadata file will be keys in the JSON object - independent of the metadata schema.
 
+The `files` field contains a JSON object in which the keys are the names of the file fields and the values are lists of objects containing file IDs, names and URLs. This is only available if the Loculus instance and organism are configured to support files. The URLs are S3 presigned URLs that only valid for a limited amount of time.
```
corneliusroemer, line 48:
```suggestion
The `files` field contains a JSON object in which the keys are the names of the file fields and the values are lists of objects containing file IDs, names and URLs. This is only available if the Loculus instance and organism are configured to support files. The URLs are S3 presigned, read-only URLs that are only valid for a limited amount of time.
```

kubernetes/loculus/values.schema.json
theosanderson: I think what you are looking at above is the `submissionDataTypes`'s `files` section? If so, this is completely distinct: you could disable submitting files but enable outputting files, for example allowing the pipeline to take consensus sequences and generate e.g. PNGs with graphs about the consensus sequences, or you could enable inputting files but disable outputting files, i.e. use the raw reads to create a consensus sequence but not make them available for download.

kubernetes/loculus/templates/_s3-endpoint.tpl
```diff
@@ -0,0 +1,19 @@
...
+  {{- end -}}
+{{- end -}}
+
+{{- define "loculus.s3UrlInternal" -}}
+  {{- if $.Values.runDevelopmentS3 }}
+    {{- "http://loculus-minio-service:8084" -}}
+  {{- else -}}
+    {{- $.Values.s3.bucket.endpoint }}
+  {{- end -}}
+{{- end -}}
```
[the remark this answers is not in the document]
theosanderson: How come?

website/src/services/backendClient.ts
```diff
@@ -2,18 +2,16 @@ import axios, { AxiosError, type Method } from 'axios';
...
     dataUseTermsHistoryEntry,
     getSequencesResponse,
     info,
+    requestUploadResponse,
     sequenceEntryToEdit,
     type ProblemDetail,
 } from '../types/backend.ts';
 import { createAuthorizationHeader } from '../utils/createAuthorizationHeader.ts';
 
-const myLogger = getInstanceLogger('BackendClient');
```
corneliusroemer, line 15: What's the reason we're getting rid of the logger?

website/src/components/Submission/FileUpload/FolderUploadComponent.tsx
```diff
@@ -0,0 +1,447 @@
...
+                                                                        file.type === 'pending' &&
+                                                                        file.fileId === fileId
+                                                                    ) {
+                                                                        if (response.ok) {
+                                                                            return {
+                                                                                type: 'uploaded',
+                                                                                fileId: file.fileId,
+                                                                                name: file.name,
+                                                                                size: file.size,
+                                                                            };
```
corneliusroemer, line 177: Maybe break this into functions or multiple files, it's hard to review in Github UI as long lines get broken
image

corneliusroemer:
Quick brainstorming for a better name for `fileField` (the type of file, e.g. raw_reads, logs, cram etc) with Anya yielded:
- `fileKind`
- `fileCategory`

kubernetes/loculus/values.schema.json
```diff
@@ -245,14 +245,72 @@
...
+                      "items": {
+                        "type": "object",
+                        "placeholder": "fileField",
+                        "properties": {
+                          "name": {
+                            "groups": ["schema"],
+                            "type": "string",
+                            "description": "The name of the file field, i.e. 'raw_reads'."
+                          },
+                          "multipleFiles": {
```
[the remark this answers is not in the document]
fhennig: You're right it hasn't been implemented yet. I'd actually also be in favor of that! @chaoran-chen What do you say, can we just drop this for now?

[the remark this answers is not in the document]
chaoran-chen: Sure, happy with dropping it for now. We can change #4041 to add the feature from scratch and push it to the bottom of the backlog.

backend/src/main/kotlin/org/loculus/backend/config/S3Config.kt
chaoran-chen: Yes, you're right, that looks wrong! `region` should indeed be optional.

backend/src/main/kotlin/org/loculus/backend/service/files/FilesDatabaseService.kt
chaoran-chen: release sounds good to me

fhennig: I would then also change the "published at" column accordingly. Will go ahead with it.

backend/src/main/kotlin/org/loculus/backend/service/files/FilesTable.kt
chaoran-chen: A file can be associated with multiple sequence entries, so it's practical to record the timestamp explicitly. Also, a processed file can be created and released after the approval of the sequence entry when the pipeline reruns and decides to create new files (#4055).

backend/src/test/kotlin/org/loculus/backend/controller/files/RequestUploadEndpointTest.kt
chaoran-chen: I think that we are quite safe here because the database is wiped before each test. – At least, it should be, which reminds me that we forgot to add the new files table: https://github.com/loculus-project/loculus/pull/3978/files#r2063476852

backend/src/test/kotlin/org/loculus/backend/controller/EndpointTestExtension.kt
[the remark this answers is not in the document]
fhennig: good shout, done.

backend/src/main/kotlin/org/loculus/backend/config/S3Config.kt
theosanderson: (if we make it optional maybe we can not use it for MinIO, which iirc we do atm [which seems weird], but nbd)

website/src/components/Submission/FileUpload/FolderUploadComponent.tsx
fhennig: Oh yes, definitely. i've split out a function for "start uploading"

kubernetes/loculus/templates/_siloDatabaseConfig.tpl
fhennig: yes indeed, the metadata field names and file field names (name to be improved) should not have any overlaps. Yes we should document it.

website/src/services/backendClient.ts
fhennig: The logger.ts file only works server side and the backend client is now used client side as well. This is related to this PR: https://github.com/loculus-project/loculus/commit/da4e58f79259013cf83bcb1cbc4c556564466536

We got rid of the Zodius stuff, which could use the `logger` field out of the box (if I'm reading this right) but now the field is actually not used.

We could probably make an effort to enhance the logging, but I think it's not urgent.

kubernetes/loculus/values.schema.json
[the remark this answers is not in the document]
fhennig: I have updated the other issue. Should I then for now remove this flag here?

backend/src/test/kotlin/org/loculus/backend/controller/files/RequestUploadEndpointTest.kt
fhennig: agree, I've added the truncation now, we should be fine.

backend/src/main/kotlin/org/loculus/backend/service/files/S3Service.kt
fhennig: Possibly, although not at the moment in this PR; we will also touch the expiry time again in #4050, so I've added a sentence there for now

.github/workflows/e2e-k3d.yml
fhennig: It takes a little longer due to creating the MinIO instance I think.

backend/docs/db/schema.sql
fhennig: I've renamed it to released at, I hope this clarifies things as 'released' is well-ish defined concept. Feel free to unresolve if its still in need of improvement.

backend/src/main/kotlin/org/loculus/backend/api/FileTypes.kt
fhennig: The kind of URL inside of the data class can be different depending on context. The URL allows either GET or PUT, but not both. For uploading, only PUT is allowed.

> Should the authorization information be a property of this class?

I'm not entirely sure what you mean here, do you mean that the user gets extra info on what the URL can be used for? I'd say that this should be clear from context, i.e. depending on which endpoint you queried, you get a PUT or GET URL.

kubernetes/loculus/values.schema.json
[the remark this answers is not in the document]
chaoran-chen: Yes, that sounds good to me!

backend/src/main/kotlin/org/loculus/backend/service/submission/EmptyProcessedDataProvider.kt
```diff
@@ -20,6 +20,7 @@ class EmptyProcessedDataProvider(private val backendConfig: BackendConfig) {
             alignedAminoAcidSequences = referenceGenomes.genes.map { it.name }.associateWith { null },
             nucleotideInsertions = referenceGenomes.nucleotideSequences.map { it.name }.associateWith { emptyList() },
             aminoAcidInsertions = referenceGenomes.genes.map { it.name }.associateWith { emptyList() },
+            files = null,
```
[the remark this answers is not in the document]
corneliusroemer: I guess null is fine

backend/src/main/kotlin/org/loculus/backend/api/FileTypes.kt
corneliusroemer: I just thought presigned on its own is vague and differentiating between read vs write in the types might make things clearer in signatures, e.g. knowing that released data contains _read_ urls not _write_ urls

backend/src/main/kotlin/org/loculus/backend/config/SecurityConfig.kt
```diff
@@ -58,6 +58,7 @@ class SecurityConfig {
         "/get-seqset-cited-by-publication",
         "/get-author",
         "/*/get-released-data",
+        "/files/get/**",
```
corneliusroemer, line 61: Should `/files/get/**` be public also for not-yet-released files?

backend/src/main/kotlin/org/loculus/backend/controller/FilesController.kt
```diff
@@ -0,0 +1,92 @@
...
+    @GetMapping("/get/{accession}/{version}/{fileCategory}/{fileName}")
+    fun getFileDownloadUrl(
+        @HiddenParam user: User,
+        @PathVariable accession: Accession,
+        @PathVariable version: Long,
+        @PathVariable fileCategory: String,
+        @PathVariable fileName: String,
+    ): ResponseEntity<Void> {
+        val accessionVersion = AccessionVersion(accession, version)
+        val fileId = submissionDatabaseService.getFileId(accessionVersion, fileCategory, fileName)
```
corneliusroemer, line 49: This seems to require `fileName`s to be unique for a given `accessionVersion` and `fileCategory` - is this checked for by the backend at point of processing the submission?

Should one be able to just get a file by `fileId` as well? Not the complicated `{accession}/{version}/{fileCategory}/{fileName}`?

backend/src/main/kotlin/org/loculus/backend/controller/FilesController.kt
```diff
@@ -0,0 +1,92 @@
...
+                }
+            } else {
+                throw ForbiddenException("Authentication needed: the file is not public")
+            }
+        }
+        val presignedUrl = s3Service.createUrlToReadPrivateFile(
+            fileId,
+            fileName,
+        )
+        return ResponseEntity.status(HttpStatus.SC_TEMPORARY_REDIRECT)
```
corneliusroemer, line 68: Using a redirect means unless you inspect with curl or your browser dev tools or look at the download link in your browser one can't actually easily get that URL and share with others.

Every request to read a file, even a public file, has to go through backend

backend/src/main/kotlin/org/loculus/backend/service/files/FilesTable.kt
```diff
@@ -0,0 +1,17 @@
+package org.loculus.backend.service.files
+
+import org.jetbrains.exposed.sql.Table
+import org.jetbrains.exposed.sql.kotlin.datetime.datetime
+import java.util.*
```
corneliusroemer, line 5: another wildcard snuck in, maybe an IntelliJ setting?

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
```diff
@@ -108,12 +116,15 @@ open class SubmissionController(
...
+            objectMapper.readValue(it, object : TypeReference<SubmissionIdFilesMap>() {})
```
corneliusroemer, line 120: We should probably check in the backend that the submitted fileCategories are allowed fileCategories - the objectMapper doesn't do this, as it only tread `fileCategory` as a `String`

backend/src/main/kotlin/org/loculus/backend/config/S3Config.kt
theosanderson: Sorry, my message is maybe confusing. I was just remembering that when I read the minio config at present it says something like `us-east-1`, which seems weird given we're self-hosting (not in the US!) - so if by changing this we can also avoid that, that's great too!

backend/src/main/kotlin/org/loculus/backend/config/SecurityConfig.kt
theosanderson: Yes? In that we handle the authentication in the endpoint itself

backend/src/main/kotlin/org/loculus/backend/controller/FilesController.kt
theosanderson: > Every request to read a file, even a public file, has to go through backend

Yes - we discussed this in some detail on Slack: it means that each file will always be referenced with the same URL (which will be e.g. `http://instance.org/seq/PP_00140V0.1/raw_reads/forward.fastq.gz`). The fact that there is a single consistent URL that can persist over time, even if e.g. a new technology comes along that's not S3 or if we need to split storage across multiple regions is IMO a feature not a bug. It also means that files will download with the correct filename as we will generate a presigned URL that gives the file the right name in the attachment.

backend/src/main/kotlin/org/loculus/backend/controller/FilesController.kt
theosanderson: > Should one be able to just get a file by fileId as well? 

I don't see a need for this. At the moment file-ids are just a Loculus convenience feature - they are not something we plan to surface to users. That seems good to me. If you want a unique string to represent your file you have `{accession_version}/{fileCategory}/{fileName}`

backend/src/main/kotlin/org/loculus/backend/service/files/FilesTable.kt
fhennig: thanks, yes, I think I fixed it now.

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
fhennig: yes, we have an issue for that in #3994

fhennig: resolving this for now

backend/src/main/kotlin/org/loculus/backend/api/FileTypes.kt
fhennig: Ah I see, thanks for clarifying. So far the type didn't restrict which type of URL to use, but I see how making the name more specific to how it's currently used helps with understanding the data flow as well, so I've renamed it now. I was hesitating to put this constraint in because there isn't really a need, but it makes sense!

04f8f7888

kubernetes/loculus/values.schema.json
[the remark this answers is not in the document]
fhennig: It has now been removed.

kubernetes/loculus/values.yaml
fhennig: Adressed in comment above, let's keep the discussion there. resolving.

kubernetes/loculus/values.schema.json
fhennig: I've added some admin docs for the config, that also mention this: https://github.com/loculus-project/loculus/pull/3978/commits/cb2f35c115509d01dca1ea472615e6e13eec9139

kubernetes/loculus/templates/_siloDatabaseConfig.tpl
fhennig: I've documented it now! https://github.com/loculus-project/loculus/pull/3978/commits/cb2f35c115509d01dca1ea472615e6e13eec9139

backend/src/main/kotlin/org/loculus/backend/config/BackendSpringConfig.kt
fhennig: I've removed the null check here: ab92c6722

backend/src/main/kotlin/org/loculus/backend/config/SecurityConfig.kt
fhennig: Yes, in the endpoint we check if the `User` is an `AuthenticatedUser`, and that they are allowed to access the file. 

        val isPublic = filesDatabaseService.isFilePublic(fileId)!!
        if (!isPublic) {
            if (user is AuthenticatedUser) {
                accessionPreconditionValidator.validate {
                    thatAccessionVersionExists(accessionVersion)
                        .andThatUserIsAllowedToEditSequenceEntries(user)
                }
            } else {
                throw ForbiddenException("Authentication needed: the file is not public")
            }
        }
        
The endpoint is public in that you _can_ access it without authentication and in some cases you also will get a file back.

backend/src/main/kotlin/org/loculus/backend/config/BackendSpringConfig.kt
```diff
@@ -101,6 +109,25 @@ class BackendSpringConfig {
...
+        @Value("\${${BackendSpringProperty.S3_BUCKET_ACCESS_KEY}}") accessKey: String? = null,
+        @Value("\${${BackendSpringProperty.S3_BUCKET_SECRET_KEY}}") secretKey: String? = null,
+    ): S3Config {
+        if (!enabled) {
+            return S3Config(false, null)
+        }
+        if (endpoint != null && bucket != null && accessKey != null && secretKey != null) {
+            return S3Config(true, S3BucketConfig(endpoint, internalEndpoint, region, bucket, accessKey, secretKey))
+        }
+        throw Error("S3 bucket configurations are incomplete.")
```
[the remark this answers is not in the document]
theosanderson: (feel free to ignore)

backend/docs/file_sharing.md
```diff
@@ -0,0 +1,42 @@
...
+    id uuid NOT NULL,
+    upload_requested_at timestamp without time zone NOT NULL,
+    uploader text NOT NULL,
+    group_id integer NOT NULL,
+    released_at timestamp without time zone
+);
+```
+
+Entries in this table are created by the backend, then the `request-uploads` endpoint is called.
+From that request, it is recorded who made the request, for which group and when.
```
anna-parker, line 36: Maybe I misunderstand but shouldn't this be 
```suggestion
Entries in this table are created by the backend when the `request-uploads` endpoint is called.
The request records who made the request, for which group and when.
```

backend/docs/file_sharing.md
```diff
@@ -0,0 +1,42 @@
...
+
+![submission](./plantuml/sequenceFileSharingSubmission.svg)
+
+## Preprocessing
+
+![preprocessing](./plantuml/sequenceFileSharingPrepro.svg)
+
+## Releasing
+
+To release a file (i.e. make it public), the file only needs to be tagged with `public=true`.
```
anna-parker, line 18: This is a bit vague, could you still clarify a bit?

backend/docs/plantuml/sequenceFileSharingPrepro.puml
```diff
@@ -0,0 +1,26 @@
+@startuml
+    participant "Processing Pipeline" as pipeline #LightGreen
+    participant "Backend" as backend #Orange
+    participant "S3 / Object Storage" as s3 #LightGray
+    database "Database" as DB
+
+    pipeline -> backend: request entries needing processing
+    backend -> DB: fetch entries + file IDs
```
anna-parker, line 8: By file ID do you mean file name?

backend/docs/plantuml/sequenceFileSharingSubmission.puml
```diff
@@ -0,0 +1,16 @@
...
+
+    frontend -> backend: request pre-signed URLs and file IDs
+    backend -> DB: insert file records with file IDs and requesting user
+    backend -> frontend: return pre-signed URLs and file IDs
+
+    loop for each file
+        frontend -> s3: upload file using pre-signed URL
+    end loop
+
+    frontend -> backend: submit metadata, sequence data and file IDs per submission ID
```
anna-parker, line 15: we could maybe still add what release does

backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
```diff
@@ -164,6 +167,11 @@ data class ProcessedData<SequenceType>(
         description = "The key is the gene name, the value is a list of amino acid insertions",
     )
     val aminoAcidInsertions: Map<GeneName, List<Insertion>>,
+    @Schema(
+        example = """{"raw_reads": [{"fileId": "s0m3-uUiDd", "name": "data.fastaq"}], "sequencing_logs": []}""",
+        description = "The key is the file field name, the value is a list of files, with ID and name.",
```
anna-parker, line 172:
```suggestion
        description = "The key is the file category name, the value is a list of files, with ID and name.",
```

backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
```diff
@@ -263,8 +271,17 @@ data class OriginalData<SequenceType>(
         description = "The key is the segment name, the value is the nucleotide sequence",
     )
     val unalignedNucleotideSequences: Map<SegmentName, SequenceType?>,
+    @Schema(
+        example = """{"raw_reads": [{"fileId": "f1le-uuId-asdf", "name": "myfile.fastaq"]}""",
+        description = "A map from file field names, to lists of files. The files can also have URLs.",
```
anna-parker, line 276:
```suggestion
        description = "A map from file category names, to lists of files. The files can also have URLs.",
```

backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
anna-parker: sorry to bring this back up again - but I don't see where `OriginalData<SequenceType>` is used? Why is this required? oh apologies SequenceType is generic

backend/docs/plantuml/sequenceFileSharingPrepro.puml
fhennig: I've changed it to "fetch entries (including file IDs and file names)", does that clarify it?

backend/src/main/kotlin/org/loculus/backend/service/files/S3Service.kt
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    /**
+     * Use the internal client to make direct requests to S3.
+     */
+    private fun getInternalClient(): MinioClient {
+        if (internalClient == null) {
+            internalClient = createClient(getS3BucketConfig(), true)
+        }
+        return internalClient!!
```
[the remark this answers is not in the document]
fhennig: It's a good point! I've switched to eager initialization for the clients: 4d1675f40

anna-parker (changes requested):
thanks for adding docs - made this a lot faster! I am still reviewing the backend and website but thought I would already finish the review so you can see my suggestions on the kubernetes part :-)

backend/docs/file_sharing.md
anna-parker: sounds much clearer!

kubernetes/loculus/values.yaml
```diff
@@ -1421,6 +1426,70 @@ defaultOrganisms:
...
           sequence: "[[URL:https://corneliusroemer.github.io/seqs/artefacts/sars-cov-2/S.fasta]]"
+  dummy-organism-with-files:
+    schema:
+      image: "https://cdn.who.int/media/images/default-source/mca/mca-covid-19/coronavirus-2.tmb-1920v.jpg?sfvrsn=4dba955c_19"
+      organismName: "Test organism (with files)"
+      nucleotideSequences: []
+      submissionDataTypes:
+        consensusSequences: false
+        files:
+          enabled: true
```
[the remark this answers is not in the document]
fhennig: @chaoran-chen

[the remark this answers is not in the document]
theosanderson: This was discussed here: https://github.com/loculus-project/loculus/issues/3907 (I don't have strong feelings, but remember we generally need to be able to overwrite values so we probably need to at least be able to set it to null, so I'm not sure the `hasKey` would work great.

kubernetes/loculus/templates/minio-service.yaml
```diff
@@ -0,0 +1,19 @@
...
+    app: loculus
+    component: minio
+  ports:
+    - port: 8084
+      targetPort: 9000
+      {{- if ne $.Values.environment "server" }}
+      nodePort: 30084
+      {{- end }}
+      protocol: TCP
+      name: http
```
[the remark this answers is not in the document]
fhennig: external traffic uses HTTPS, but it is terminated at the ingress; cluster-internal traffic is HTTP for all our services, so this is in line with how the other services as configured.

kubernetes/loculus/templates/minio-deployment.yaml
```diff
@@ -0,0 +1,89 @@
...
+              valueFrom:
+                secretKeyRef:
+                  name: s3-bucket
+                  key: accessKey
+            - name: MINIO_ROOT_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: s3-bucket
+                  key: secretKey
+        - name: minio-init
```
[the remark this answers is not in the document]
theosanderson: I think you need to set `restartPolicy: onFailure` maybe? On the pod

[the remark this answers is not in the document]
fhennig: You cannot set that for deployment pods: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#pod-template

anna-parker (changes requested):
Wow a lot of work has gone into this! Great job! I'm not sure I was able to fully comprehend the FolderUploadComponent - but as soon as the page is available I will try out some more uploads. I added a few more comments now primarily for the backend

backend/src/main/kotlin/org/loculus/backend/config/SecurityConfig.kt
theosanderson: Because otherwise it would be impossible for any files to be accessed without auth

anna-parker:
Cool was able to test out the website code a bit now - it looks really good!

1. When I click on revise a sequence the files aren't listed - but I see there is a TODO for this
2. I just checked and when I download files that I uploaded they are not downloaded with the filename - this might be a bug.

theosanderson:
> I just checked and when I download files that I uploaded they are not downloaded with the filename - this might be a bug.

Will be resolved by https://github.com/loculus-project/loculus/pull/4064

backend/src/test/kotlin/org/loculus/backend/controller/files/RequestUploadEndpointTest.kt
```diff
@@ -0,0 +1,102 @@
...
+            .get(0)
+            .get("url")
+            .textValue()
+
+        val content = "test content".toByteArray()
+        val request = HttpPut(url)
+        request.entity = ByteArrayEntity(content, ContentType.TEXT_PLAIN)
+        val httpClient = HttpClients.createDefault()
+        val response = httpClient.execute(request)
+        Assertions.assertEquals(200, response.statusLine.statusCode)
```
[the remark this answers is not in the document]
theosanderson: There's pretty good coverage of this stuff in the integration tests I think

kubernetes/loculus/templates/minio-deployment.yaml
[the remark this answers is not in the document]
theosanderson: Options 2 and 3 here seem good: https://chatgpt.com/share/6812488d-d564-8005-81cd-4588f30657da (but nbd as is )

backend/src/test/kotlin/org/loculus/backend/controller/files/RequestUploadEndpointTest.kt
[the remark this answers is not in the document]
anna-parker: Ah yes - sorry I overlooked them!

corneliusroemer:
A few more things that don't fit any particular line in the code:
- It would be nice to add test data to the PR description, a zip folder of test metadata, sequences, raw data that one can unzip and upload directly
- We might want to track file size as a piece of metadata, we'd probably want to display this to users before they download 3TB 😂
- Documentation is presumably still a Todo
- Manual testing has not been documented in the description, as mentioned, it would be good to properly test things on prod data, i.e on staging. Ideally this would be done before merging here, as this is a big piece of code here with schema changes.

theosanderson:
> Manual testing has not been documented in the description, as mentioned, it would be good to properly test things on prod data, i.e on staging. 

When we discussed this synchronously you said that by this you meant just checking that Pathoplexus staging still works normally, rather than actually testing something about file sharing itself - just posting that here too because it wasn't clear to me

corneliusroemer:
> When we discussed this synchronously you said that by this you meant just checking that Pathoplexus staging still works normally, rather than actually testing something about file sharing itsel

Good point thanks for flagging for clarification.

By testing properly, I meant that existing functionality is not negatively impacted (no regressions). Not that  file sharing would work in prod, as that would require extra setup and is not necessary at this point.

backend/src/main/kotlin/org/loculus/backend/auth/AuthenticatedUser.kt
```diff
@@ -41,6 +46,9 @@ class UserConverter : HandlerMethodArgumentResolver {
         if (authentication is JwtAuthenticationToken) {
             return AuthenticatedUser(authentication)
         }
+        if (authentication is AnonymousAuthenticationToken) {
```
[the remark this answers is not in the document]
fhennig: I assumed that this is more something that Spring will make use of internally in it's routing etc, but good to clarify maybe - @chaoran-chen ?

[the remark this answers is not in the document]
chaoran-chen: @anna-parker, what exactly do you refer to with "this"? If you mean the `AnonymousUser` that we return, it is used in a controller.

backend/src/main/kotlin/org/loculus/backend/config/SecurityConfig.kt
anna-parker: Thanks! 

Adding more context for myself and anyone else is curious: `"/files/get/**"` is a get endpoint whereas `submit` is a post endpoint. In the security config (securityFilterChain) all POST requests fall into the `auth.anyRequest().authenticated()` bucket and are only let in with authentication whereas GET requests are handled with `auth.requestMatchers(HttpMethod.GET, *getEndpointsThatArePublic).permitAll()` - if  `"/files/get/**"` is not added here it will also fall through to `auth.anyRequest().authenticated()` bucket and require auth to be accessed

backend/src/main/kotlin/org/loculus/backend/auth/AuthenticatedUser.kt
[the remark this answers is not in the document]
chaoran-chen: In `FilesController.getFileDownloadUrl()`, we are taking a `User` as an argument. This is different than in most other controllers where we expect an `AuthenticatedUser`. Because in `getFileDownloadUrl()`, authentication is not always required, we are more general by allowing `User` and a `User` can be an `AuthenticatedUser` or an `AnonymousUser` which we check in the code.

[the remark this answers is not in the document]
chaoran-chen: (As an alternative, we could just return a `User` here instead of introducing the subtype `AnonymousUser`, which would work just as well. But I think that it is nicer to mark the user as anonymous clearly)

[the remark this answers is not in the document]
anna-parker: Thank you!

corneliusroemer:
@fhennig Here's a small thing you can do already :) There're some small merge conflicts that need to be resolved for the previews and actions and everything to work
image

theosanderson:
I think the preview is still working

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
```diff
@@ -132,12 +152,17 @@ open class SubmissionController(
         @Parameter(
             description = SEQUENCE_FILE_DESCRIPTION,
         ) @RequestParam sequenceFile: MultipartFile?,
+        @RequestPart fileMapping: String?,
     ): List<SubmissionIdMapping> {
+        val fileMappingParsed = fileMapping?.let {
+            objectMapper.readValue(it, object : TypeReference<SubmissionIdFilesMap>() {})
```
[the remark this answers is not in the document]
fhennig: yes, good point! I've made an issue for this: https://github.com/loculus-project/loculus/issues/4109

kubernetes/loculus/templates/minio-deployment.yaml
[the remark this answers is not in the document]
fhennig: I ruled out using a `Job` because I didn't see a good way to ensure it re-runs in case the deployment is destroyed and recreated (OTOH, maybe this isn't even a strict requirement?). But the third option (container lifecycle event) seems quite cool, I didn't know about container lifecycle events before!

kubernetes/loculus/values.yaml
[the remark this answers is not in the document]
fhennig: you can have output files without input files, so there can still be files listed in the UI even if files are disabled as a submission type.

But the backend doesn't currently check if files are enabled. it also doesn't check the categories yet, we have an issue for that: https://github.com/loculus-project/loculus/issues/3994 - I've added to this issue to also check whether submission is actually enabled.

kubernetes/loculus/templates/minio-deployment.yaml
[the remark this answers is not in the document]
fhennig: I've deployed it, and it works!

fhennig:
Thanks for your comment Cornelius!

> It would be nice to add test data to the PR description, a zip folder of test metadata, sequences, raw data that one can unzip and upload directly

I've added that now, it's very "dummy" but can be used with the _dummy organism (with files)_. It's only a metadata file and some dummy files (no sequences). IMO that's sufficient to just test the flow on a technical level.

> We might want to track file size as a piece of metadata, we'd probably want to display this to users before they download 3TB 😂

 I've created an issue for this: https://github.com/loculus-project/loculus/issues/4110

> Documentation is presumably still a Todo

There is some documentation in here, but there is also an issue to write the user docs: https://github.com/loculus-project/loculus/issues/4063. There is also the ADR PR still open, but interest in this seems to have waned.

> By testing properly, I meant that existing functionality is not negatively impacted (no regressions). Not that file sharing would work in prod, as that would require extra setup and is not necessary at this point.

I think the integration tests working show that pretty well!

kubernetes/loculus/values.yaml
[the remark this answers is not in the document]
fhennig: (I've now also actually added this check and a test for it to the submit endpoint; and also a check for the revise endpoint - the endpoint doesn't really work yet, so I didn't add a test yet.)

theosanderson (approved):
From my POV this is good to go into `main`, even though the feature isn't ready to enable in prod yet. Thanks for all the hard work.

PR #3983, opened 2025-04-14
feat(backend): add files canonical URL endpoint

### LINKED ISSUE #3977: Canonical URL for files (backend)
Similar to #3969 and as [discussed in Slack]([loculus.slack.com/archives/C05G172HL6L/p1744215684189449](https://loculus.slack.com/archives/C05G172HL6L/p1744215684189449)), we would like to have a backend endpoint at `/files/<accessionVersion>/<fileField>/<fileName>` for processed files that redirects to a presigned URL to the file with the correct attachment filename.

If the sequence entry has been approved, the endpoint should be public (i.e., not requiring authentication). If the sequence entry has not been approved, the endpoint must validate that the user is authenticated and has the needed rights.

chaoran-chen:
resolves #3977


backend/src/main/resources/db/migration/V1.11__file_sharing.sql
```diff
@@ -2,7 +2,8 @@ create table files (
     id uuid primary key,
     requested_at timestamp not null,
     uploader text not null,
-    group_id integer not null references groups_table(group_id)
+    group_id integer not null references groups_table(group_id),
+    public_at timestamp
```
chaoran-chen, line 6: Or something else? `published_at`? `made_public_at`? `became_public_at`?

backend/src/main/kotlin/org/loculus/backend/controller/FilesController.kt
```diff
@@ -23,8 +33,44 @@ class FilesController(
     private val filesDatabaseService: FilesDatabaseService,
     private val s3Service: S3Service,
     private val groupManagementPreconditionValidator: GroupManagementPreconditionValidator,
+    private val submissionDatabaseService: SubmissionDatabaseService,
+    private val accessionPreconditionValidator: AccessionPreconditionValidator,
 ) {
 
+    @GetMapping("/get/{accession}/{version}/{fileField}/{fileName}")
```
fhennig, line 40: I guess we need the `get` in here because otherwise an accession could clash with the `requestUpload`? It's not great to have the `get` in the path I think, but I can't think of another solution either.

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -1230,6 +1232,38 @@ class SubmissionDatabaseService(
...
+                ),
+                explicitStatementType = StatementType.SELECT,
+            ) { resultSet ->
+                if (resultSet.next()) {
+                    UUID.fromString(resultSet.getString("file_id"))
+                } else {
+                    null
+                }
+            }
+        }
```
fhennig, line 1265: Have you considered to just get the `processed_data`  for the `AccessionVersion`, parse it in Kotlin, and then get the FileID from there? That would have the advantage of type-safe JSON access, as well as not having the column and table names as strings in there (no literal SQL) with very little performance impact.

backend/src/main/kotlin/org/loculus/backend/service/files/FilesDatabaseService.kt
```diff
@@ -22,8 +24,26 @@ class FilesDatabaseService(private val dateProvider: DateProvider) {
...
     }
 
+    fun getGroupId(fileId: FileId): Int? = getGroupIds(setOf(fileId))[fileId]
+
     fun getGroupIds(fileIds: Set<FileId>): Map<FileId, Int> =
         FilesTable.select(FilesTable.idColumn, FilesTable.groupIdColumn)
             .where { FilesTable.idColumn inList fileIds }
             .associate { Pair(it[FilesTable.idColumn], it[FilesTable.groupIdColumn]) }
+
+    fun setPublicAtIfEmpty(fileIds: Set<FileId>) {
```
fhennig, line 34: I find `setPublicAtIfEmpty` _very_ hard to understand. How about "setPublishedAtNowIfNotPublished"? Or just "setPublishedAtNow" and leave the rest implied?

backend/src/main/kotlin/org/loculus/backend/controller/FilesController.kt
```diff
@@ -23,8 +33,44 @@ class FilesController(
...
+                }
+            } else {
+                throw ForbiddenException("Authentication needed: the file is not public")
+            }
+        }
+        val presignedUrl = s3Service.createUrlToReadPrivateFile(
+            fileId,
+            filesDatabaseService.getGroupId(fileId)!!,
+            fileName,
+        )
```
fhennig, line 68: We create a presignedUrl, even if the file is public here, right? We don't need to do that, right? IMO it would be cool to just link to the file normally if it is public, that means the link is valid for longer.

fhennig (changes requested):
Great!

Just a few things, nothing big, mostly optional comments. Just the comment about the literal SQL string is important-ish to me

I think it's great to have these URLs!

backend/src/main/kotlin/org/loculus/backend/controller/FilesController.kt
theosanderson: > We create a presignedUrl, even if the file is public here, right? We don't need to do that, right?

We need to so it gets the correct filename

fhennig: Ah yes, you're right, the link will be valid.

backend/src/main/resources/db/migration/V1.11__file_sharing.sql
chaoran-chen: Great, then let's go with `published_at`!

fhennig: I will make a commit to change it 👌

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
chaoran-chen: This was for me the simplest and most compact/readable way to implement it and it is probably the most efficient. But I also won't mind a different solution.

backend/src/main/kotlin/org/loculus/backend/service/files/FilesDatabaseService.kt
chaoran-chen: `setPublishedAtNow` sounds a bit misleading to me: the name indicates to me that it would overwrite the field with the date for now. What about just `publish()`?

backend/src/main/kotlin/org/loculus/backend/controller/FilesController.kt
chaoran-chen: (Agree with what Theo said)

backend/src/main/resources/db/migration/V1.11__file_sharing.sql
fhennig: https://github.com/loculus-project/loculus/pull/3983/commits/d93ceabca2e87a14a54c6f53e62274303499a8f1

backend/src/main/kotlin/org/loculus/backend/controller/FilesController.kt
fhennig: alright, I'd say let's just leave it the way it is!

backend/src/main/kotlin/org/loculus/backend/service/files/FilesDatabaseService.kt
fhennig: publish is good for me!

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
fhennig: I've changed it in this PR: 

https://github.com/loculus-project/loculus/pull/3983/commits/b62f201839607b5be749c89919f80d66b16bdcda

fhennig (approved):

Issue #3878, opened 2025-03-20
Copying from Loculus metadata copies with leading `<br>` in HTML format


theosanderson:
If you copy from the Loculus metadata page with a single word e.g.

<img width="353" alt="Image" src="https://github.com/user-attachments/assets/54d1ab61-840d-47db-9df4-e7d3a711626b" />

in Chrome

and inspect the clipboard with https://evercoder.github.io/clipboard-inspector/

you find the plain text version is fine, but the HTML version contains a rogue `<br />`

<img width="1030" alt="Image" src="https://github.com/user-attachments/assets/dbe5dec5-31a8-495e-9cfc-73e725b77eba" />

That linebreak then gets converted into `\r\n` on pasting which is the cause of the issue described in https://github.com/loculus-project/loculus/issues/3655 , but it also could affect pasting into other things. We should try to get rid of this rogue `<br />`

anna-parker:
I am looking into the code and this is happening where we use whitespace-wrap. I can't see this defined in https://tailwindcss.com/docs/white-space - so potentially it is adding the <br /> in front of the copied lines... yes I tried removing it and the empty space disappears:

![Image](https://github.com/user-attachments/assets/7fd5ec4d-b9ce-4c83-a15f-c51bb7e2335e)

anna-parker:
@theosanderson you are right - somehow I copied the text differently the second time (I think if I copy from back to front of a word there is no white space in front, but from front to back there is)

PR #3985, opened 2025-04-14
feat(website): Strip `\r` and `\n` on paste into text field

### LINKED ISSUE #3655: We should strip leading spaces in search queries
This is live on production, I cannot search by submissionId - can be tried by pasting the submissionId of any sequence into search:

e.g. https://main.loculus.org/seq/LOC_000PB42.1

Image

theosanderson:
resolves #3655.

N.B. This does not resolve the underlying issue https://github.com/loculus-project/loculus/issues/3878

It is unlikely we want linebreaks in singleline fields - it is quite easily to accidentally paste them and then they can cause confusion. This becomes very common due to https://github.com/loculus-project/loculus/issues/3878. Here we strip these characters from single line fields as a workaround. For now this only addresses singleline text fields.

I tested that this works

### PR Checklist

🚀 Preview: https://pastefix.loculus.org

### DIFF
--- a/website/src/components/SearchPage/fields/TextField.spec.tsx
+++ b/website/src/components/SearchPage/fields/TextField.spec.tsx
@@ -1,6 +1,5 @@
-import { render, screen } from '@testing-library/react';
+import { render, screen, fireEvent } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
-import type { ClipboardEvent } from 'react';
 import { describe, it, expect, vi } from 'vitest';
 
 import { TextField } from './TextField';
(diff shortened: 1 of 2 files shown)


fhennig:
Oh great! Also nice to now have better tests! Code looks good to me. I think it'd be awesome to also have a test where only a selection is replaced with a paste, but it's probably not easy to do and you're probably also quite busy so it's fine to merge without that.

Thanks for tacking this issue!

fhennig (approved):

theosanderson:
Thanks for the review - I will indeed merge - I have manually tested this fwiw

Issue #3987, opened 2025-04-15
Update with preview link fails for dependabot updates - we should just remove this for dependabot updates


anna-parker:
![Image](https://github.com/user-attachments/assets/585ed841-5cef-4e74-ba04-d828c16607b5)

PR #3988, opened 2025-04-15
fix(workflow): disable add preview url for PRs opened by dependabot

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/3987

### PR Checklist
- tested that this works by rebasing a dependabot PR on this branch: https://github.com/loculus-project/loculus/pull/3956 (PR passes this check, fails others

### DIFF
--- a/.github/workflows/add-preview-url.yml
+++ b/.github/workflows/add-preview-url.yml
@@ -14,6 +14,13 @@ jobs:
           github-token: ${{ secrets.LOCULUS_BOT_WRITE_PRS }}
           script: |
             const pr = context.payload.pull_request;
+
+            // Skip if PR author is Dependabot for security reasons
+            if (pr.user.login === 'dependabot[bot]') {
+              console.log("Skipping Dependabot PR.");
+              return;
+            }
+
             const branchName = pr.head.ref;
             const hasPreviewLabel = pr.labels.some(label => label.name === 'preview');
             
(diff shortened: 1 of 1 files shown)


chaoran-chen (approved):

.github/workflows/add-preview-url.yml
```diff
@@ -14,6 +14,13 @@ jobs:
           github-token: ${{ secrets.LOCULUS_BOT_WRITE_PRS }}
           script: |
             const pr = context.payload.pull_request;
+
+            // Skip if PR author is Dependabot for security reasons
```
theosanderson, line 18: Could you explain a bit more why we need this? You still have to add a label – all the action does is to add a link?

theosanderson: Ok, I understand now the check was failing. I would delete "security reasons" and say "because it doesn't work" or something but nbd

anna-parker: ah so it doesnt work because we cant give dependabot the keys due to how github enforces key access - I will try to make this clearer

theosanderson: In fact we could also just add this as a dependabot secret

image

but I've probably lost the secret now so this may be easier for now

anna-parker:
Closing as completed via https://github.com/loculus-project/loculus/pull/4290#event-17955996902

Issue #3666, opened 2025-02-11
We should strip starting and trailing `N`s in sequences


theosanderson:
https://pathoplexus.org/seq/PP_0011CQ4.2, the 2025 Ebola sequence, starts and ends with a short set of Ns, even in the unaligned sequence, because that's how it was submitted. In contrast, other Ebola sequences I examined (e.g. https://pathoplexus.org/seq/PP_000034P.4) have no starting or trailing Ns in the unaligned sequence (perhaps due to processing at INSDC? or perhaps not).

This has two consequences:
- it creates an inconsistency between this sequence (and any like it) and other sequences where these `N`s are trimmed that doesn't reflect any biological or technical difference
- it means that the "length" of this sequence doesn't reflect the real usable length (though in this case the two are pretty similar!) In an extreme case though, where only a short contiguous sequence was sequenced, and the rest of the genome filled in with `N`s there could be a very misleading "genome length" in this case.

IMO we should trim `N`s at the start and end of sequences early in the preprocessing pipeline.

dpark01:
My own opinion is that it would make sense to auto-trim continuous stretches of leading and trailing N's, as they are informationless, and this would be more consistent with how all other genome repositories handle things.

NCBI Genbank actually will reject submissions where the first or last "line" of the fasta exceeds 40% N content (not sure how long a "line" is... especially the last line of a fasta, but that's just what table2asn does). I don't recommend doing that per se, but there's no reason not to chop off pure-Ns at the edges.

theosanderson:
Thanks for the research

> Exception: circular genomes don't have "ends", the start/end is arbitrary.

> Trimming of Ns should only be done for non-circular genomes by default

Reading the doc you referenced it's not super clear to me that it is saying don't trim N's for these? in a way isn't the reason you need to write `gap at end` because you don't have Ns?

PR #4005, opened 2025-04-15
feat(preprocessing, integration tests): strip Ns in preprocessing and improve integration tests

theosanderson:
resolves https://github.com/loculus-project/loculus/issues/3666

This adapts the Nextclade preprocessing pipeline to trim starting and ending Ns.

It also:
- adds an integration test for this functionality
- separates integration tests into those that require the readonly test data first and those that don't to speed up running individual tests locally when they don't need this test data

I have mostly tested this by using the integration test but I think that is reasonable.

### PR Checklist

🚀 Preview: https://strip-ns.loculus.org


preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -728,7 +734,7 @@ def process_single(  # noqa: C901
         version=version_from_str(id),
         data=ProcessedData(
             metadata=output_metadata,
-            unalignedNucleotideSequences=unprocessed.unalignedNucleotideSequences,
+            unalignedNucleotideSequences=unaligned_nucleotide_sequences,
```
anna-parker, line 737: We need to trim this before the unaligned sequence is sent to nextclade - as it is we send the sequences with Ns to nextclade and this is reflected in the listed insertions and `Total unknown nucs` numbers - these then no longer match with the unaligned sequence shown on the webpage. 

I took the liberty of making a PR for this: https://github.com/loculus-project/loculus/pull/4079

integration-tests/README.md
```diff
@@ -23,10 +41,18 @@ Install dependencies:
     npm ci
     npx playwright install --with-deps
 
+### Option 1: Running local server
+
 Set up the cluster to test:
 
     ./start-server.sh
```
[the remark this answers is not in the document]
theosanderson: > Did I miss a step?

No - `start_server.sh` just deploys whatever is on `main` (because it doesn't build images locally). That works for adding new tests to main without changing code. For other stuff I typically use the preview server approach (Option 2). Should probably add a note on that.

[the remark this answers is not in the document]
anna-parker: Thanks! Otherwise the docs are very clear thanks for adding them!

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -192,7 +192,7 @@ def enrich_with_nextclade(  # noqa: C901, PLR0912, PLR0914, PLR0915
             elif len(unaligned_segment) == 1:
                 num_valid_segments += 1
                 unaligned_nucleotide_sequences[id][segment] = (
-                    entry.data.unalignedNucleotideSequences[unaligned_segment[0]]
+                    trim_ns(entry.data.unalignedNucleotideSequences[unaligned_segment[0]])
```
theosanderson, line 195: @anna-parker - Q: will this work for the non-nextclade case too? If not do you mind figuring out how we can since you know the code better?

anna-parker: Oh no good catch this will in fact not work for non-nextclade cases! (your approach would though so if need be we can combine them)

theosanderson: Maybe actually we don't need to worry too much about that case since it's a light touch pipeline in that case anyway

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
theosanderson: Thank you!

integration-tests/README.md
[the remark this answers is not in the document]
anna-parker: Thanks! I just ran using the second option and it looks good!

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
anna-parker: Guess we might as well fix this: https://github.com/loculus-project/loculus/pull/4081

preprocessing/nextclade/src/loculus_preprocessing/backend.py
```diff
@@ -82,14 +83,19 @@ def parse_ndjson(ndjson_data: str) -> Sequence[UnprocessedEntry]:
         except json.JSONDecodeError as e:
             error_msg = f"Failed to parse JSON: {json_str_processed}"
             raise ValueError(error_msg) from e
+        unaligned_nucleotide_sequences = json_object["data"]["unalignedNucleotideSequences"]
+        trimmed_unaligned_nucleotide_sequences = {}
+        for key, value in unaligned_nucleotide_sequences.items():
+            trimmed_unaligned_nucleotide_sequences[key] = trim_ns(value) if value else None
```
theosanderson, line 89:
```suggestion
        trimmed_unaligned_nucleotide_sequences = {key: trim_ns(value) if value else None for key, value in unaligned_nucleotide_sequences.items()}
```

theosanderson: @anna-parker maybe

anna-parker (approved):
Just tested and looks good, thanks @theosanderson!

PR #4008, opened 2025-04-17
fix(ingest): ensure when processing alignment that highest scoring al…

anna-parker:
1. ensure when processing alignment that highest scoring alignment is chosen

Currently we assume a sequence will only ever align to one segment and do not drop lower scoring results. (Use a script for this as it otherwise gets too complicated - this is nice as I then can remove tsv-utils from the dependencies)

This also adds a test (adds a small (mutated) segment of the M segment to an S segment in this tests and ensures that the highest scoring alignment - the S alignment is still chosen)

2. Allow non-segmented organisms to use nextclade-sort (this is useful for e.g. rsv)

resolves #

### PR Checklist

🚀 Preview: https://fix-ingest-alignment.loculus.org


corneliusroemer:
> for some reason now if we revoke an ingest sequence ingest fails with:

It fails in this section:

```
        try:
            entries = list(jsonlines.Reader(response.iter_lines()).iter())
        except jsonlines.Error as err:
            response_summary = response.text
            max_error_length = 100
            if len(response_summary) > max_error_length:
                response_summary = response_summary[:50] + "\n[..]\n" + response_summary[-50:]
            logger.error(f"Error decoding JSON from /get-original-metadata: {response_summary}")
            raise ValueError from err
```

It looks like this is emitted by the backend and not valid JSON (of course) `a.util.Map.get(Object)" because "metadata" is null` hence ingest fails.

So we now need to look at backend logs.

ingest/scripts/parse_alignment_output.py
```diff
@@ -0,0 +1,44 @@
...
+ - joins results of all alignments
+ - removes rows where there is no alignmentScore (i.e. alignment failed)
+ - group by seqName (=accession) and alignmentScore and keep the row with the highest score
+"""
+
+import click
+import pandas as pd
+
+
+def parse_files(input_files: list[str], output_file: str):
```
corneliusroemer, line 12: This does much more than parse which is a bit misleading hence, can call it main_main or something 😄

corneliusroemer (approved):
So we can now use minimizer for identifying segments XOR virus species, but not both, right?

anna-parker:
> So we can now use minimizer for identifying segments XOR virus species, but not both, right?

Partially yes, so when using the minimizer for segmented viruses you always need to add the segment - and so the minimizer should be added under segment_identification. But you can also add the virus subspecies (makes sense as within a subspecies you will also need to have a segment identifier for each subspecies). Now you can (as you mention) also use the minimizer index on non-segmented viruses for subspecies identification.

ingest/scripts/parse_alignment_output.py
anna-parker: Ah yes kind of silly to create a new function for this I will just move it into main - there is a description of all teh steps at the top of the file :-)

PR #4024, opened 2025-04-22
feat(config): clean up old failed/succeeded cronjobs

anna-parker:
I asked chatgpt how to ensure we don't store the logs of all failed/succeeded cronjobs and this was the result


theosanderson:
Apparently the defaults are not so different to this: 
image


I don't see an advantage to changing the default really?

anna-parker:
Ah so probably the issue was actually just all the old ingest cronjobs from main not being deleted - hopefully as I removed them using 
```
kubectl get jobs --all-namespaces --sort-by=.metadata.creationTimestamp \
  | grep -v 'Running' \
  | awk '$6 ~ /d/ && $6+0 > 30 {print $1, $2}' \
  | while read ns name; do kubectl delete job -n $ns $name; done
```
that will be enough to not run into the issue again :-)

PR #4025, opened 2025-04-22
file sharing: Add integration tests

### LINKED ISSUE #3990: Create some file sharing integration tests
Depends on #3934

fhennig:
resolves #3990

- make MinIO run when deploying locally
- Add a test that uploads some single submission data
- Add a test that uploads bulk data
- Add now `BulkSubmissionPage` page to playwright tests.

This also adds a new 'internal endpoint' s3 config setting. I ran into an issue with running the backend in k3s and using 'localhost' to access S3, which only works from outside the cluster, but caused connectivity issues from the backend Pod. Now there are two settings for the endpoint, one 'internal' which the backend will use when talking to S3 inside the cluster, and one external which will be used for generated links.

### PR Checklist

🚀 Preview: https://fs-integr-test.loculus.org


kubernetes/loculus/values.yaml
```diff
@@ -26,7 +26,12 @@ dataUseTerms:
     open: https://#TODO-MVP/open
     restricted: https://#TODO-MVP/restricted
 s3:
-  enabled: false
+  enabled: true
+  bucket:
+    region: us-east-1
+    bucket: loculus-preview-private
+    accessKey: "[[s3accessKey]]"
```
theosanderson, line 33: 🙌 
(while we're touching the code, we discussed deleting these, and same in the config-preprocessor)

fhennig: yes I remember - I wanted to do it seperately but can do it now 👍

fhennig: Done! (They were actually not referenced anymore anyways, so easy to remove!)

theosanderson: (sorry, of course would have been happy for it to be done separately)

integration-tests/tests/specs/features/file-sharing.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+    const reviewPage = await submissionPage.submitSequence();
+    await reviewPage.waitForZeroProcessing();
+
+    await reviewPage.releaseValidSequences();
+
+    await page.getByRole('link', { name: 'released sequences' }).click();
+    while (!(await page.getByRole('link', { name: /LOC_/ }).isVisible())) {
+        await page.reload();
+        await page.waitForTimeout(2000);
+    }
```
theosanderson, line 40: Optionally we could extract this into a function which I suspect would be quite generically useful - I actually think I maybe used to have one but it might have disappeared during some big refactoring

kubernetes/loculus/templates/_s3-endpoint.tpl
```diff
@@ -0,0 +1,23 @@
...
+    {{- end -}}
+  {{- else -}}
+    {{- $.Values.s3.bucket.endpoint }}
+  {{- end -}}
+{{- end -}}
+
+{{- define "loculus.s3UrlInternal" -}}
+  {{- if $.Values.runDevelopmentS3 }}
+    {{- if eq $.Values.environment "server" -}}
+        {{- (printf "https://s3%s%s" $.Values.subdomainSeparator $.Values.host) -}}
```
theosanderson, line 16: I don't immediately understand why this isn't still `http://loculus-minio-service:8084` - though I'm not saying it's wrong

fhennig: I think you're probably right! Just copy-pasted a bit haphazardly 😬 (either URL should be fine, but we can use the proper internal one 👍 )

fhennig:
Thanks for your review!

kubernetes/loculus/templates/_s3-endpoint.tpl
fhennig: https://github.com/loculus-project/loculus/pull/4025/commits/f62c2355d2bf632536aa5af215926906ccbbf732

integration-tests/tests/specs/features/file-sharing.spec.ts
fhennig: I agree. There is still a similar function but it's in search.page.ts (and now here we're not on the search page - even though they are very very similar). One thing I found difficult-ish to generalize, is that this won't work if there will be multiple matching things.

So yes we should probably do it, but I'll leave it for now - maybe in the future it'll be more obvious how to generalize it best. Thanks for bringing it up and making me look into it again!

Issue #3840, opened 2025-03-11
Adjust SILO readiness endpoint


theosanderson:
Previously we had SILO and LAPIS in one deployment. That meant that until all of them were healthy we kept using the old SILO. Now we've split them out. Now SILO can become "healthy" and therefore receive traffic before it has finished SILO import. We should only make it ready after SILO import to avoid data downtime during deployments.

fengelniederhammer:
Given https://github.com/GenSpectrum/LAPIS-SILO/issues/723 was fixed, then SILO should be have as follows:
* When no database has been loaded yet, `/info` returns status code 503.
* Once a database has been loaded, `/info` returns status code 200 along with some information
https://github.com/GenSpectrum/LAPIS-SILO/blob/faab6fbeb9d937d4060920032983a69b7ec51d77/src/silo/api/info_handler.cpp#L18-L26
https://github.com/GenSpectrum/LAPIS-SILO/blob/main/include/silo/database_info.h#L14-L20

It should be possible to run an empty SILO, so Loculus should probably not check the `count` in the info. I also think that the import job already imports an empty database when there is no data yet.

Does it make sense to consider SILO "live" as soon as the `/info` endpoint responds something (including errors) and consider it "ready" as soon as it returns 200?

PR #4027, opened 2025-04-22
chore(deployment): remove reloader

theosanderson:
1. Removed the reloader.stakater.com/auto: "true" annotation from
  all Kubernetes deployment files:
    - loculus-backend.yaml
    - loculus-website.yaml
    - silo-deployment.yaml
    - lapis-deployment.yaml
    - keycloak-deployment.yaml
    - ingest-deployment.yaml
    - ena-submission-deployment.yaml
    - docs-preview.yaml
    - keycloak-database-standin.yaml
    - loculus-preprocessing-deployment.yaml
  2. Removed the reloader installation code from deploy.py:
    - Removed install_reloader() function call from handle_cluster()
  function
    - Deleted the entire install_reloader() function

🚀 Preview: https://remove-reloader.loculus.org

### DIFF
--- a/deploy.py
+++ b/deploy.py
@@ -156,7 +156,6 @@ def handle_cluster():
             shell=True,
         )
     install_secret_generator()
-    install_reloader()
     while not is_traefik_running():
         print("Waiting for Traefik to start...")
         time.sleep(5)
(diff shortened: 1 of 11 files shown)


anna-parker:
Hi @theosanderson I'm not very familiar with the reloader I checked and it seems the affect of this removal will be:

- Pods will not restart when ConfigMaps or Secrets are updated.
- Our pods will continue running with stale configuration until they are manually restarted or resynced by something else (like Argo CD, Helm upgrade, or kubectl rollout restart).
- apps that rely on dynamic config (like feature toggles, connection strings, etc. - I believe this is the case for most of our pods as we get features from the config files), may misbehave or not reflect the new changes at all.

How do we normally do rollouts for PPX? Would we have to now do a rolling restart if there are config updates?

theosanderson:
The idea of the reloader was indeed to solve the issues that you describe here, however I'm not super confident that it ever actually worked. For Pathoplexus, we seemed to have continual issues with this.

Most components are restarted for Loculus updates because the image tag changes. For stuff where that's not the case I think we often don't see restarting anyway. We have an issue to make sure that everything is forced to restart on a roll-out with https://github.com/loculus-project/loculus/issues/3170 .

anna-parker:
Thanks! Just wanted to confirm - when we do a rolling restart will the website or search ever be temporarily unavailable?

theosanderson:
Yes, search becomes unavailable due to https://github.com/loculus-project/loculus/issues/3840

PR #4028, opened 2025-04-22
feat(config): add silo readinessProbe

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/3840

preview URL:

### Summary

Adds a SILO readinessProbe to ensure it doesnt "start" before it is healthy

### PR Checklist

🚀 Preview: https://add-silo-readiness-probe.loculus.org

### DIFF
--- a/kubernetes/loculus/templates/silo-deployment.yaml
+++ b/kubernetes/loculus/templates/silo-deployment.yaml
@@ -47,9 +47,6 @@ spec:
             httpGet:
               path: /info
               port: 8081
-              httpHeaders:
-                - name: Accept
-                  value: application/json
             initialDelaySeconds: 30
             periodSeconds: 10
             failureThreshold: 3
--- a/kubernetes/loculus/templates/silo-deployment.yaml
+++ b/kubernetes/loculus/templates/silo-deployment.yaml
@@ -43,6 +43,14 @@ spec:
           volumeMounts:
             - name: lapis-silo-shared-data
               mountPath: /data
+          readinessProbe:
+            httpGet:
+              path: /info
+              port: 8080
+            initialDelaySeconds: 30
+            periodSeconds: 10
+            failureThreshold: 3
+            timeoutSeconds: 5
         - name: silo-preprocessing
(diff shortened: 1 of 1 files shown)


kubernetes/loculus/templates/silo-deployment.yaml
```diff
@@ -43,6 +43,25 @@ spec:
           volumeMounts:
             - name: lapis-silo-shared-data
               mountPath: /data
+          livenessProbe:
```
theosanderson, line 46: > consider SILO "live" as soon as the /info endpoint responds something (including errors)

I doubt the current implementation does this - I think an error would manifest as unhealthiness. IMO for now we could not add the livenessProbe if we can't figure out how to do this. I think it would be better not to add it than to add it like this

theosanderson: (thanks for taking on the PR!)

anna-parker: True - even once the database is loaded argocd still shows SILO as degraded.

anna-parker: ah ok now it looks ok - it just seems a bit delayed

theosanderson: Yeah but my point still stands (I think?)

image

theosanderson: To be clear, I'm happy with the readinessProbe you've implemented!

anna-parker: Cool! Was just going to suggest we only add the readiness probe. I'm not sure though - do we need the 
```
httpHeaders:
    - name: Accept
      value: application/json
```
part? will errors also be seen as errors for the readiness probe?

theosanderson: Yeah I wasn't clear why that was added - I wouldn't bother unless you added it because something wasn't working

> will errors also be seen as errors for the readiness probe?

yes

theosanderson (approved):
Thanks!

PR #4034, opened 2025-04-22
feat(deployment): timestamp pods on deployment to force restart of all pods on every deployment

### LINKED ISSUE #3170: Figure out how to ensure Pathoplexus restarts appropriately on config update

theosanderson:
resolves #3170

This adds timestamp annotations to all pods that basically describe the time on which helm was run to make a deployment. The effect of this should be that all pods have to restart when a new deployment is made, because this value will change. This should ensure that config-only updates are still made.

Testing:
- Tested that the product still works
- Tested that I can see the annotations in the pod

In terms of whether it fixes the issue - it should but IMO we can wait and see

🚀 Preview: https://timetamp-now.loculus.org

### DIFF
--- a/kubernetes/loculus/templates/docs-preview.yaml
+++ b/kubernetes/loculus/templates/docs-preview.yaml
@@ -8,7 +8,6 @@ metadata:
   name: loculus-docs
   annotations:
     argocd.argoproj.io/sync-options: Replace=true
-    timestamp: {{ now | quote }}
 spec:
   replicas: 1
   selector:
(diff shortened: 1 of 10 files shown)


anna-parker (approved):
LGTM!

PR #4040, opened 2025-04-22
fix(backend,ingest): handle revoked sequences in /get-original-metadata, they have null metadata

corneliusroemer:
resolves #4036 

co-authored by @anna-parker

Fix backend to deal with `null` original metadata.

Now we get this for revoked sequences in `/get-original-metadata` (we add the `isRevocation` field to the output):

```
{"accession":"LOC_000M2HN","version":2,"submitter":"superuser","originalMetadata":null, "isRevocation": True}
```

Ingest also works: we remove the key sort and additionally handle revocations separately (as they do not contain the INSDC accession in the originalMetadata as it is empty) 

We now emit isRevocation in `/get-original-metadata` to allow ingest to handle revocations properly.

### Testing

- Unit test added that reproduces the bug and no longer does so with the fix
- Manual in preview: Revoke a sequence, rerun ingest without problems
- Manual in preview: Revoke two sequences, rerun ingest without problems (Anya noticed this failed and she fixed it)

### Ideas for future

- Nice to have: Integration test for reingest with a revoked sequence

### PR Checklist
- ~[ ] All necessary documentation has been adapted.~
- [x] The implemented feature is covered by appropriate, automated tests.
- [x] Any manual testing that has been done is documented (i.e. what exactly was tested?)

🚀 Preview: https://fix-null-metadata.loculus.org

PR #4042, opened 2025-04-23
fix(ingest): handle revocations correctly in call_loculus

anna-parker:
resolves https://github.com/loculus-project/loculus/pull/4040#issuecomment-2823260215

### Screenshot

If we now revoke 2 sequences that were created by the insdc_ingest_user we get another error. Original metadata is empty for revocations and without this the call_loculus script maps the revoked sequences to insdc_accession=NULL. We throw an error if multiple sequences map to the same insdc_accession - it also good to make sure this are mapped correctly so we never try to revise a revoked sequence. 

### PR Checklist

🚀 Preview: https://original-metadata-endpoin.loculus.org

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
@@ -269,6 +269,7 @@ data class AccessionVersionOriginalMetadata(
     override val accession: Accession,
     override val version: Version,
     val submitter: String,
+    val isRevocation: Boolean,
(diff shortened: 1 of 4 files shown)


corneliusroemer (approved):
Thanks! Good idea to pass along whether something is revocation.

I don't fully grok the ingest logic anymore but I trust it's better now! We can iterate on better revocation handling, it's hardly worse now than before

anna-parker:
@corneliusroemer thanks! I was still having some errors but now it looks nice locally so I will merge :-)

PR #4045, opened 2025-04-24
file sharing: Fix incorrectly specified service in Ingress

fhennig:
The Ingress didn't have the correct service name specified, now it does.

All CORS related stuff was removed (wasn't needed).

🚀 Preview: https://fs-fix-cors.loculus.org


kubernetes/loculus/templates/ingressroute.yaml
```diff
@@ -45,6 +45,18 @@ spec:
     replacement: "https://$1"
     permanent: true
 ---
+apiVersion: traefik.containo.us/v1alpha1
+kind: Middleware
+metadata:
+  name: cors-forward
```
theosanderson, line 51: forward doesn't seem the best name if as I understand it's not forwarding the CORS bits (just adding them afresh)

theosanderson: Ah, but maybe this does something different sorry!

theosanderson: (we do it differently in https://github.com/loculus-project/loculus/blob/6ff96d1eb5ada563dcc46ad3d96189fe60716956/kubernetes/loculus/templates/lapis-ingress.yaml#L6 )

theosanderson (approved):

PR #4046, opened 2025-04-24
feat(ingest): upload tzst compressed ncbi datasets packages to reduce traffic

corneliusroemer:
The default compression NCBI datasets uses zip which compresses long sequences only poorly, e.g. mpox only 3x. Zstd does much better, compressing up to 100x, i.e. 30x better.

This PR adds .tzst upload alongside the existing native .zip

This reduces s3 bandwidth a lot (at least 10x) at the cost of slightly higher actions usage (35min per run every 2hr rather than 20min)

### Testing

- Ran edited action, checked it completed. Overall minute usage goes up from 20min to 35min, but that seems ok
- Locally tested ingest works for both mirror and direct ncbi datasets

### PR Checklist

🚀 Preview: https://tzst-compress.loculus.org


ingest/Snakefile
```diff
@@ -101,45 +101,61 @@ rule clean:
...
-            --filename {output.dataset_package} \
-            --api-key {params.api_key} \
-            {params.gateway}
+                --no-progressbar \
+                --filename results/dataset.zip \
+                --api-key {params.api_key} \
+                {params.gateway} \
+            && {params.unzip} \
+                -o results/dataset.zip \
+                -d results
```
theosanderson, line 125: Could you explain what this does differently?

theosanderson: Is it that it does the extraction itself?

ingest/Snakefile
```diff
@@ -101,45 +101,61 @@ rule clean:
...
             --output {output.ncbi_dataset_tsv}
         """
 
+rule format_ncbi_dataset_sequences:
+    input:
+        dataset_package="results/ncbi_dataset",
+    output:
+        ncbi_dataset_sequences="results/sequences.fasta",
+    params:
+        unzip=unzip,
```
theosanderson, line 151: Is this needed?

corneliusroemer: Yeah we chop everything from the header line except for the id with ` -i, --only-id               print IDs instead of full headers`

corneliusroemer: it was already there just moved up to a more logical place, see the deleted lines bottom left

ingest/Snakefile
corneliusroemer: Yeah because we now download a different format from the mirror we need different types of decompression depending on the type of data source used. Hence we do the extraction straight in the rule that does the download to keep things simple.

ingest/Snakefile
theosanderson: I meant this specific line with the unzip parameter - (but I don't really know snakemake)

corneliusroemer: Ah yeah necessary ffor things to work on macOS, see above line 71-76


```
if os.uname().sysname == "Darwin":
    # Don't use conda-forge unzip on macOS
    # Due to https://github.com/conda-forge/unzip-feedstock/issues/16
    unzip = "/usr/bin/unzip"
else:
    unzip = "unzip"
```

theosanderson: Thanks!

corneliusroemer: guess we could name it `unzip_binary` to be more explicit

corneliusroemer: aaah now I got it, you're right, this is no longer needed in this place!

ingest/Snakefile
```diff
@@ -99,48 +99,69 @@ rule clean:
...
         """
 
 
+rule format_ncbi_dataset_sequences:
+    input:
+        dataset_sequences="results/ncbi_dataset/data/genomic.fna",
+    output:
+        ncbi_dataset_sequences="results/sequences.fasta",
+    params:
+        unzip=unzip,
```
corneliusroemer, line 156:
```suggestion
```

PR #4064, opened 2025-04-30
[BREAKING] file sharing: use canonical links

### LINKED ISSUE #3969: Canonical URL for files
We discussed canonical URLs for accessing files and the idea that 

`http://instance.org/seq/PP_00140V0.1/raw_reads/forward.fastq.gz`

Might be a good pattern for this, where `raw_reads` is the file field and `forward.fastq.gz` is the filename.

This would either query or redirect to a backend endpoint, which would return a presigned URL to the file with the correct attachment filename (after checking for auth).

fhennig:
**BREAKING** - the `website.runtimeConfig.public` section has been moved to the top level, i.e. is now directly at `public`.

---

resolves #3969 

This PR adds a new endpoint to the frontend to get a link to a file with the pattern: `$websiteUrl/seq/$accession.$version/$fileCategory/$fileName`. There is a new config setting (`filesSharing.outputFileUrlType`) that allows configuring whether this kind of link should be output by `get-released-data` or whether to use a direct link to S3.

The website URL uses the filename, and so files downloaded are also using the filename.

An integration test was added to test this.

**Manual testing**: I've manually checked that the link looks ok, and I've also tested `website` vs `s3` URLs in the preview to check that the Helm Chart works.

We plan to use the new `fileSharing` top level config object later on for more file sharing related settings, such as the expiry time of a pre signed URL, or garbage collector related things.

### Screenshot

https://github.com/user-attachments/assets/6bde9882-868c-4cab-8aa3-efbdf42e5201

### PR Checklist
- [x] All necessary documentation has been adapted. -> the schema was expanded.

🚀 Preview: https://file-sharing-canonical-ur.loculus.org


fhennig:
note to self: I think the test is failing because `Values.host` isn't set in the integration test run. Looking into it

fhennig:
ok, that worked _kinda_ but fetch still failed. maybe HTTPS is the issue, tried disabling that now (or just remove it from the link, rather).

fhennig:
Ok, somehow there is still a connection issue. I can't reproduce it either in the preview or locally - Maybe @theosanderson you have an idea? I think maybe it's something network related, since it's with 2 redirects, first to the backend, then to the S3 instance.

fhennig:
This is based on the `file-sharing` branch but I'd rather not merge more stuff in there; instead I want to merge this into main after `file-sharing` is merged. But we can already review.

backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
```diff
@@ -199,6 +206,24 @@ open class ReleasedDataModel(
...
+        version: Version,
+        filesMap: FileCategoryFilesMap,
+    ): Map<String, List<FileIdAndNameAndReadUrl>> = filesMap.addUrls {
+            fileCategory,
+            fileId,
+            fileName,
+        ->
+        when (backendConfig.fileSharing.outputFileUrlType) {
+            FileUrlType.WEBSITE -> {
+                val encodedName = URLEncoder.encode(fileName, StandardCharsets.UTF_8)
```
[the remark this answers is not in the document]
theosanderson: See line below - good point - I think we can just harcode HTTPS

website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName]/index.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import { getRuntimeConfig } from '../../../../../config';
+import { parseAccessionVersionFromString } from '../../../../../utils/extractAccessionVersion';
+
+export const GET: APIRoute = ({ params }) => {
+    const runtimeConfig = getRuntimeConfig();
+    const { accessionVersion, fileCategory, fileName } = params;
+    const { accession, version } = parseAccessionVersionFromString(accessionVersion!);
+
+    const redirectUrl = `${runtimeConfig.public.backendUrl}/files/get/${accession}/${version}/${fileCategory}/${fileName}`;
```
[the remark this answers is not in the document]
Comment (theosanderson) [reply website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName]/index.ts:11]: seems fair - we still haven't decided what to do in terms of allowed characters, but can't hurt?

backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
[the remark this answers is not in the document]
chaoran-chen: If we hardcode HTTPS, wouldn't we have a problem with local/development deployment? Maybe we need to make it configurable as well. At the moment, if someone tries to access HTTP in a production instance, they would get redirected to HTTPS, so that's maybe not too much of a problem.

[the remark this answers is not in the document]
theosanderson: Looks like Felix has a nice solution

kubernetes/loculus/templates/_backend-url.tpl
```diff
@@ -0,0 +1,10 @@
+{{- define "loculus.backendUrl" -}}
+{{- $publicRuntimeConfig := (($.Values.website).runtimeConfig).public }}
```
theosanderson, line 2: I don't currently understand this

website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName]/index.ts
Comment (theosanderson) [inline website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName]/index.ts:1]: I might have expected this file to be called `website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts` - not sure if there's a functional difference?

Comment (theosanderson) [reply website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName]/index.ts:1]: (I might imagine you'd need a trailing slash as is but may well be wrong)

backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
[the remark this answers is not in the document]
fhennig: Hmm, I don't have a solution (I think?) - I added the `backendUrl` part but that doesn't really solve it. We could maybe get `websiteUrl` (with protocol) instead of `websiteHost`, but then we still need to get the protocol from somewhere.

Also currently the `backendUrl` ist kind of burried under `websiteConfig`. Maybe we'd want to refactor that.

website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName]/index.ts
Comment (theosanderson) [reply website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName]/index.ts:1]: Yeah I don't think that it does add the `.html`, at least in Next.Js but I also don't think it matters (to me it is slightly more intuitive my way, but who cares)

Comment (fhennig) [reply website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName]/index.ts:1]: done!

backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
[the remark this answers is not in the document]
fhennig: I'm using this now:

```
{{- define "loculus.websiteUrl" -}}
  {{- if eq $.Values.environment "server" -}}
    {{- (printf "https://%s" $.Values.host) -}}
  {{- else -}}
    {{- "http://localhost:3000" -}}
  {{- end -}}
{{- end -}}
```

kubernetes/loculus/values.schema.json
```diff
@@ -1144,6 +1158,21 @@
...
+    "fileSharing": {
+      "type": "object",
+      "groups": ["general"],
+      "description": "Settings related to the file sharing feature.",
+      "properties": {
+        "outputFileUrlType": {
+          "type": "string",
+          "groups": ["general"],
+          "enum": ["website", "backend", "s3"],
+          "description": "Whether the backend should supply file URLs as links using the website, or directly to the S3 objects."
```
corneliusroemer, line 1170: What's the meaning of the enum value backend? I don't see it described in description in contrast to website and s3

corneliusroemer: Would be good to add to `description`

theosanderson: The ideal place to make such suggestions is in an issue or a PR :)

PR #4070, opened 2025-05-01
chore!(deployment): Remove `dateToSortBy`, `partitionBy` as now deprecated in SILO (and the wrapper `silo` value)

theosanderson:
These options will be deprecated when SILO is bumped to https://github.com/GenSpectrum/LAPIS-SILO/releases/tag/v0.6.0 so I thought we might as well remove the fluff while I remember.

Breaking!: ensure the `silo` section is removed from `values.yaml` to avoid helm lint errors


chaoran-chen (approved):
thanks!

PR #4072, opened 2025-05-01
feat(website): use consistent and better field selector

### LINKED ISSUE #2064: Group fields by category in the search field selector (and ultimately the column selector)
Use some sort of collapsible tree view with the categories from the sequence details page instead of: 

image

theosanderson:
resolves #2064

We used to have two components for selecting fields, the old one for column/search selection:

image

And the new one for Download field selection:

image

This removes the first one and makes everything use a beefed up version of the second

It also moves a test from E2E tests to integration tests as part of our general aim to eventually remove the e2e tests entirely

🚀 Preview: https://fields2.loculus.org


website/src/components/common/FieldSelectorModal.tsx
theosanderson, line 1: Most of this is from the original `DownloadDialog/FieldSelector`

anna-parker:
I love the look of this - sadly I found a bug when I select None all check boxes are still clicked

https://github.com/user-attachments/assets/e6786bf0-ad85-456d-ad07-516ffb454fdd

theosanderson:
Thank you @anna-parker  - that is fixed now!

website/src/components/SearchPage/SearchFullUI.tsx
```diff
@@ -84,6 +84,20 @@ export const InnerSearchFullUI = ({
...
 
+    const columnFieldItems: FieldItem[] = useMemo(
+        () =>
+            schema.metadata
+                .filter((field) => !(field.hideInSearchResultsTable ?? false))
+                .map((field) => ({
+                    name: field.name,
+                    displayName: field.displayName ?? field.name,
+                    header: field.header,
+                    alwaysSelected: field.name === schema.primaryKey,
```
anna-parker, line 95: Weirdly AccessionVersion should be the primary key but it isn't marked as `alwaysSelected` on the Add Search Field section

website/src/components/SearchPage/SearchFullUI.tsx
```diff
@@ -84,6 +84,20 @@ export const InnerSearchFullUI = ({
 
     const [isColumnModalOpen, setIsColumnModalOpen] = useState(false);
 
+    const columnFieldItems: FieldItem[] = useMemo(
+        () =>
+            schema.metadata
+                .filter((field) => !(field.hideInSearchResultsTable ?? false))
+                .map((field) => ({
+                    name: field.name,
+                    displayName: field.displayName ?? field.name,
```
anna-parker, line 93: It would be nice to have the same displayName for each field in the search and in the download options box, in search we convert displayName additionally using 
`displayName: filter.displayName ?? filter.label ?? sentenceCase(filter.name),`
this means that accessionVersion becomes Accession version in search but stays accessionVersion in the download box.

website/src/components/SearchPage/DownloadDialog/FieldSelector/FieldSelectorModal.spec.tsx
```diff
@@ -92,12 +92,6 @@ describe('FieldSelectorModal', () => {
             expect(screen.getByText('Field 2')).toBeInTheDocument();
             expect(screen.getByText('Field 3')).toBeInTheDocument();
             expect(screen.getByText('Field 4')).toBeInTheDocument(); // Now should be rendered
-            expect(screen.getByText('Accession Version')).toBeInTheDocument();
```
anna-parker, line 95: I think then this test will still hold

anna-parker (approved):
One thing to address but approving as overall this looks great :-)

website/src/components/SearchPage/SearchFullUI.tsx
theosanderson: Thanks - I agree we should make this consistent. I think the use of `sentenceCase` on a technical name is really not a good pattern and the right way to do this is to remove the `sentenceCase` and to add `displayNames` for everything. Since the current state is not a regression, just maintains current behaviour, I will make an issue for this. (We already have more displayNames on PPX so this looks better there, but we need more).

theosanderson: https://github.com/loculus-project/loculus/issues/4089

theosanderson: (For an example of how we need the display names anyway, see the review cards:

image


)

Issue #2885, opened 2024-09-26
Handling suppression / withdrawal of sequences by INSDC


theosanderson:
Sequences on INSDC may be `suppressed` (often due to errors) or in rare circumstances `withdrawn` https://ena-docs.readthedocs.io/en/latest/faq/release/data-availability-policy.html . Presumably in such cases they will no longer appear in our NCBI datasets download.

**Suppression**

The desired behaviour should be revocation of these sequences in Loculus. In cases of temporary suppression this should be followed by re-instatement. Suppression, in general, is likely to be a relatively common occurrence.

**Withdrawal**

The desired behaviour is a bit more debatable. At minimum sequences should be revoked but they should likely be fully deleted. (This would be high risk, so would likely involve manual intervention by database admins, and these events are likely to be rare so this need not be very automated).

theosanderson:
I've added this to Priority 1, because I was reminded of it and think it is relatively important. It could also be "Priority 2" if we had a priority 2..

corneliusroemer:
I've just come across a sequence that has been suppressed _before_ we ingested it, another indication that this will sooner or later happen to some sequences we've ingested. See: https://loculus.slack.com/archives/C05F8GKR06R/p1742480860636979

anna-parker:
This has now happened, see details here: https://loculus.slack.com/archives/C05G172HL6L/p1745926352130099 the sequences: https://www.ncbi.nlm.nih.gov/nuccore/HG994351.1, HG994352.1, HG994353.1 and HG994354.1 have all been suppressed. Currently ingest does not handle suppression - we could revoke them for now.

PR #4075, opened 2025-05-04
feat(ingest): notify on potential suppression

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/2885

Send a notification on slack if re-ingested sequences do not include insdc-accessions that were previously ingested (and are not in a blocked, error, revoked state).

### PR Checklist
Test:
1. ingest from ebola zaire and set db to persistent
2. switch to ingest from ebola sudan
3. ensure receive notification:
image
admittedly its not very easy to search this message - but I assume we will generally not receive a notification of 3k sequences that are missing and I would need to give ingest the slack token to upload files
4. revoke some sequences (e.g. ebola zaire: OK572988.1), re-ingest ensure not notified about these sequences 
image

=> I am now reverting the persistent DB changes to stop us getting notifications every 30min on slack.

🚀 Preview: https://notify-on-suppression.loculus.org

### DIFF
--- a/ingest/scripts/compare_hashes.py
+++ b/ingest/scripts/compare_hashes.py
@@ -86,10 +86,12 @@ def process_hashes(
           hash: abcd
           status: APPROVED_FOR_RELEASE
           submitter: insdc_ingest_user
+          jointAccession: insdc_accession.seg1/insdc_accession.seg2
         - version: 2
           hash: efg
           status: HAS_ERRORS
           submitter: curator
+        jointAccession: insdc_accession.seg1/insdc_accession.seg2
     """
     if ingested_insdc_accession not in submitted:
         update_manager.submit.append(fasta_id)
@@ -172,6 +174,7 @@ def main(
         config.debug_hashes = True
 
     submitted: dict = json.load(open(old_hashes, encoding="utf-8"))
+    ingested_insdc_accessions = set(submitted.keys())
 
     update_manager = SequenceUpdateManager(
         submit=[],
@@ -197,7 +200,10 @@ def main(
             )
             if config.debug_hashes:
                 update_manager.hashes.append(hash_float)
-            process_hashes(insdc_accession_base, fasta_id, record["hash"], submitted, update_manager)
+            process_hashes(
+                insdc_accession_base, fasta_id, record["hash"], submitted, update_manager
+            )
+            ingested_insdc_accessions.discard(insdc_accession_base)
             continue
 
         insdc_keys = [f"insdcAccessionBase_{segment}" for segment in config.nucleotide_sequences]
@@ -215,6 +221,8 @@ def main(
                 if record[key]
             ]
         )
+        insdc_accessions = [record[key] for key in insdc_keys if record.get(key)]
+        ingested_insdc_accessions.difference_update(insdc_accessions)
         hash_float, update_manager.sampled_out = sample_out_hashed_records(
             insdc_accession_base, subsample_fraction, update_manager.sampled_out, fasta_id
         )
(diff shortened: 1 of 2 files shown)


theosanderson (approved):
LGTM! Thanks!

ingest/scripts/compare_hashes.py
```diff
@@ -265,6 +283,21 @@ def main(
         else:
             logger.info(f"{text}: {len(value)}")
 
+    if len(ingested_insdc_accessions) > 0:
```
corneliusroemer, line 286: Maybe it would be cleaner to use a different variable and not mutably change one without changing the name.

ingest/scripts/compare_hashes.py
```diff
@@ -215,6 +231,8 @@ def main(
                 if record[key]
             ]
         )
+        insdc_accessions = [record[key] for key in insdc_keys if record.get(key)]
+        ingested_insdc_accessions.difference_update(insdc_accessions)
```
corneliusroemer, line 235: Here we should create a new variable with a good name rather than mutate, makes it easier to review 😀

anna-parker: Oh I created the `ingested_insdc_accessions` variable just to mutate it - should I maybe choose a different name so that that is clearer? 

ingested_insdc_accessions_not_yet_reingested?

theosanderson: I think the idea would be `result = ingested_insdc_accessions.difference(insdc_accessions)` (or something)

anna-parker: but I update the variable ingested_insdc_accessions on every iteration over the new entries - after the first difference I would need to take the difference from the potentially_revoked list

anna-parker: An alternative would be to create two sets and check their difference
```
missing_records = previous_records - set(new_records)
if missing_records:
```

theosanderson: Yep, sorry, `ingested_insdc_accessions_not_yet_reingested` makes sense to me.

theosanderson: > An alternative would be

That's also great - and probably easier to follow

anna-parker: Ok I changed to use 2 lists now (and made a couple small changes to make ruff happy ;-) ) - is this clearer now?

anna-parker: @corneliusroemer is it ok if I merge?

ingest/scripts/compare_hashes.py
```diff
@@ -185,6 +186,7 @@ def main(
 
     submitted: dict = json.load(open(old_hashes, encoding="utf-8"))
     ingested_insdc_accessions = get_approved_submitted_accessions(submitted)
+    reingested_insdc_accessions = set()
```
theosanderson, line 189: To understand, does this store just reingested sequences or all sequences ingested this time?(including new ones)

anna-parker: including new ones :-)

theosanderson: So to me this is odd (inaccurate) variable naming, I would call this variable `ingested_insdc_accessions`, or `current_ingested_accessions` and the variable above one `already_ingested_accessions` or something. (But not blocking, just since we already started discussing it)

ingest/scripts/compare_hashes.py
anna-parker: Merging as @corneliusroemer told me offline that was ok

PR #4086, opened 2025-05-08
chore(github): remove special codespace configuration

theosanderson:
Some time ago I made a special configuration that was intended to start a backend and a frontend in codespaces. No one is using this at the moment, and it is very slow, and we don't even know if it still works. Codespaces are now used as part of GitHub's Agentic CoPilot and so we should remove this as it makes it very slow to start a codespace at all. We can always add it back if desired.


anna-parker (approved):
Generally fine with removing this just wanted to check how do you know noone is using it?

theosanderson:
> Generally fine with removing this just wanted to check how do you know noone is using it?

I mean because I'm the only person I think has ever really used it and I'm not anymore

PR #4123, opened 2025-05-16
feat(prepro): add nextclade sort option to check if sequence aligns better to another organism

anna-parker:
resolves #

### PR Checklist

### DIFF
--- a/preprocessing/nextclade/src/loculus_preprocessing/config.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/config.py
@@ -28,6 +28,7 @@ class Config:
     nextclade_dataset_tag: str | None = None
     nextclade_dataset_server: str = "https://data.clades.nextstrain.org/v3"
     nextclade_dataset_server_map: dict[str, str] | None = None
+    nextclade_sort: bool = False
     config_file: str | None = None
     log_level: str = "DEBUG"
     genes: list[str] = dataclasses.field(default_factory=list)
--- a/preprocessing/nextclade/src/loculus_preprocessing/prepro.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/prepro.py
@@ -14,6 +14,7 @@ from tempfile import TemporaryDirectory
 from typing import Any, Literal, TypeVar
 
 import dpath
+import pandas as pd
 from Bio import SeqIO
 
 from .backend import fetch_unprocessed_sequences, submit_processed_sequences
@@ -118,6 +119,68 @@ def parse_nextclade_json(
         id = result["seqName"]
         nextclade_metadata[id][segment] = result
     return nextclade_metadata
+
+
+def parse_sort(id: str, result_file: str, input_file: str, config: Config) -> dict:
+    warning_dict: dict[AccessionVersion, list[ProcessingAnnotation]] = {}
+    # TODO: currently this only works for datasets that are in the official nextclade_datasets repo
+    command = [
+        "nextclade3",
+        "sort",
+        f"-r {result_file}",
+        "--max-score-gap=0.3--min-score=0.05",
+        "--min-hits=2--all-matches",
+        "--",
+        input_file,
+    ]
+    logger.debug(f"Running nextclade sort: {command}")
+
+    exit_code = subprocess.run(command, check=False).returncode  # noqa: S603
+    if exit_code != 0:
+        msg = f"nextclade sort failed with exit code {exit_code}"
+        raise Exception(msg)
+
+    df = pd.read_csv(result_file, sep="\t", dtype={"index": "Int64"})
+
+    # Drop rows where 'score' is NaN - i.e. no hits
+    df["score"] = pd.to_numeric(df["score"], errors="coerce")
+    df = df.dropna(subset=["score"])
+
+    if df.empty:
+        # Sort failed - sequence will likely not align, do not give error twice
+        return warning_dict
+    df_sorted = df.sort_values(["index", "score"], ascending=[True, False])
+    # TODO: fix this for mutli-segmented case
+    if df_sorted.shape[0] > 1 and (
+        df_sorted["dataset"].iloc[0] != config.nextclade_dataset_name
+        or df_sorted["score"].iloc[0] < 0.5
+    ):
+        other_dataset = set(df_sorted["dataset"].unique()) - set(config.nextclade_dataset_name)
+        warning_dict[id] = warning_dict.get(id, [])
+        warning_dict[id].append(
+            ProcessingAnnotation(
+                unprocessedFields=(
+                    AnnotationSource(
+                        name="alignment",
+                        type=AnnotationSourceType.NUCLEOTIDE_SEQUENCE,
+                    ),
+                ),
+                processedFields=(
+                    (
+                        AnnotationSource(
+                            name="alignment",
+                            type=AnnotationSourceType.NUCLEOTIDE_SEQUENCE,
+                        ),
+                    )
+                ),
+                message=(
+                    f"This sequence aligns to a different reference (${' '.join(other_dataset)}) "
+                    "than expected - check you are submitting your sequence to the correct database."
+                ),
+            )
+        )
+        return warning_dict
+    return warning_dict
 
 
 def enrich_with_nextclade(  # noqa: C901, PLR0912, PLR0914, PLR0915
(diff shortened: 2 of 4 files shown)


preprocessing/nextclade/src/loculus_preprocessing/config.py
```diff
@@ -113,6 +116,8 @@ def get_config(config_file: str | None = None) -> Config:
 
     # Overwrite config with CLI args
     for key, value in args.__dict__.items():
+        # TODO: Defining boolean values as bool in the Config will lead to them getting overwritten
```
anna-parker, line 119: @corneliusroemer I think I found a bug in the config!

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -120,6 +121,80 @@ def parse_nextclade_json(
...
+                            ),
+                        )
+                    ),
+                    message=(
+                        f"This sequence aligns to a different reference (${' '.join(other_dataset)}) "
+                        "than expected - check you are submitting your sequence to the correct database."
+                    ),
+                )
+            )
+            return warning_dict
```
anna-parker, line 194:
```suggestion

```

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -251,6 +327,11 @@ def enrich_with_nextclade(  # noqa: C901, PLR0912, PLR0914, PLR0915
             if is_empty:
                 continue
 
+            if config.nextclade_sort:
+                warning_dict = parse_sort(result_dir_seg, input_file, warning_dict, config)
+            else:
+                warning_dict = {}
```
anna-parker, line 333:
```suggestion
```

kubernetes/loculus/values.yaml
```diff
@@ -1234,13 +1234,15 @@ defaultOrganisms:
         version: 1
         configFile:
           <<: *preprocessingConfigFile
+          nextclade_sort: True
```
corneliusroemer, line 1237: The name might not be obvious as to what happens here. Let's use something like `require_nextclade_sort_match`?

preprocessing/nextclade/src/loculus_preprocessing/config.py
corneliusroemer: interesting, I'll have a look

corneliusroemer: found an elegant fix by looking at python docs :D turns out even Gemini 2.5 didn't realize this in multiple iterations!

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -139,7 +241,7 @@ def enrich_with_nextclade(  # noqa: C901, PLR0912, PLR0914, PLR0915
     unaligned_nucleotide_sequences: dict[
         AccessionVersion, dict[SegmentName, NucleotideSequence | None]
     ] = {}
-    error_dict: dict[AccessionVersion, list[ProcessingAnnotation]] = {}
+    alerts: Alerts = Alerts(errors={}, warnings={})
```
corneliusroemer, line 244: We should just initialize alerts with empty lists then we save a lot of the `.get(id,[])` stuff

corneliusroemer: done

corneliusroemer:
We probably don't want to abort on mismatch, but still do the metadata stuff to show some context in review

Bottom is better than top
image

kubernetes/loculus/values.yaml
```diff
@@ -1234,13 +1234,23 @@ defaultOrganisms:
         version: 1
         configFile:
           <<: *preprocessingConfigFile
+          require_nextclade_sort_match: True
+          minimizer_path: https://raw.githubusercontent.com/nextstrain/nextclade_data/refs/heads/ebola/data_output/minimizer_index.json
```
corneliusroemer, line 1238: We should not use the githubusercontent links due to it getting rate limited, let's create a loculus resources repo and use GitHub pages to avoid rate limits?

also let's call it URL not path? as path is only part of a URL

kubernetes/loculus/values.yaml
```diff
@@ -1234,13 +1234,23 @@ defaultOrganisms:
         version: 1
         configFile:
           <<: *preprocessingConfigFile
+          require_nextclade_sort_match: True
```
corneliusroemer, line 1237: Let's nest all the sort params under its own yaml dict?

preprocessing/nextclade/src/loculus_preprocessing/backend.py
```diff
@@ -166,3 +166,19 @@ def submit_processed_sequences(
...
+
+def download_minimizer(url, save_path):
+    try:
+        response = requests.get(url, timeout=10)
+        response.raise_for_status()
+
+        json_bytes = json.dumps(response.json()).encode("utf-8")
+        Path(save_path).write_bytes(json_bytes)
+
+        print(f"Minimizer downloaded successfully and saved to '{save_path}'")
```
corneliusroemer, line 179: Log not print :)

preprocessing/nextclade/src/loculus_preprocessing/backend.py
```diff
@@ -166,3 +166,19 @@ def submit_processed_sequences(
...
+
+    except requests.exceptions.RequestException as e:
+        msg = f"Failed to download minimizer: {e}"
```
corneliusroemer, line 182: Log here as well

anna-parker:
Ah I should probably change the location of the minimize so that we dont see this warning in the logs:
```
2025-05-20 13:30:19.668 [W] dataset_download.rs:203: When reading dataset: The following files are present in the dataset archive, but are not declared in its pathogen.json:
  - minimizer.json
Context:
Files declared in pathogen.json:
  - CHANGELOG.md
  - README.md
  - genome_annotation.gff3
  - pathogen.json
  - reference.fasta
  - sequences.fasta
  - tree.json
Files present in the archive:
  - CHANGELOG.md
  - README.md
  - genome_annotation.gff3
  - minimizer.json
  - pathogen.json
  - reference.fasta
  - sequences.fasta
  - tree.json
This is not an error. Nextclade ignores unknown file declarations and undeclared files. But this could be a mistake by the dataset author. For example, there could be a typo in pathogen.json file declaration, or a file could have been added to the dataset, but not declared in the pathogen.json. In this case, Nextclade analysis could be missing some of the features intended by the author. Contact the author to resolve this. It could also be that the dataset contains files for a newer version of Nextclade, and that the currently used version does not recognize these files. In which case try to upgrade Nextclade.
```

preprocessing/nextclade/src/loculus_preprocessing/backend.py
anna-parker: ah wupps - that is what I get for using chatgpt for that subfunction

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -617,8 +736,8 @@ def process_single(  # noqa: C901
                 )
             )
 
+        # Break if there are sequence related errors
         if errors:
-            # Break early
             return ProcessedEntry(
```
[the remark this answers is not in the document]
anna-parker: Im worried this might break some things as this has been in the code for a while but I can try it out

preprocessing/nextclade/src/loculus_preprocessing/backend.py
```diff
@@ -166,3 +166,19 @@ def submit_processed_sequences(
...
         raise RuntimeError(msg)
     logging.info("Processed data submitted successfully")
+
+
+def download_minimizer(url, save_path):
+    try:
+        response = requests.get(url, timeout=10)
+        response.raise_for_status()
+
+        json_bytes = json.dumps(response.json()).encode("utf-8")
```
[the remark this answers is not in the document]
anna-parker: This way we do check the file is valid json but fair point we will see that when we run sort

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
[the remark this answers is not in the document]
anna-parker: seems to be working!
image

corneliusroemer (approved):

PR #4129, opened 2025-05-17
s/genomial/genomic/g

theosanderson:
Genomial isn't really a word I think, or definitely not a commonly used one. Somehow in my mind I always read this as "genomic" :)


chaoran-chen (approved):
Oh right, never realized! I think that I started to believe that it is an actual word because it's written here.

PR #4130, opened 2025-05-17
Fix NDJSON comparison in ingest tests

theosanderson:
resolves https://github.com/loculus-project/loculus/issues/4132 (but therefore breaks CI at present)

Hey @anna-parker, @corneliusroemer. This seems a pretty nasty bug that codex found whereby a function used to check tests wasn't working (was missing a `return` statement, so one would always compare `None` == `None`). I will leave it in your hands as to how to take it on.

## Summary
- return the generated dictionary in `create_dict_from_ndjson`
- rename local variable to avoid shadowing `dict`


anna-parker:
Thanks so much for fixing this! I just checked and the difference is in the `hash` function - I probably did not update this the last time there was a change in prepro, I will do that now :-)

theosanderson:
Nice, lgtm!

anna-parker (approved):

Issue #1400, opened 2024-03-19
Truncate long strings in review page cards


corneliusroemer:
Some strings, like authors, can get very long and overflow in review page cards.

They should be  truncated and shown with ellipsis `...`

<img width="2517" alt="image" src="https://github.com/loculus-project/loculus/assets/25161793/f393b1b9-b26a-488d-9d65-b39604b44ee6">

PR #4139, opened 2025-05-17
feat(website): truncate review cards

theosanderson:
resolves https://github.com/loculus-project/loculus/issues/1400

- truncate long metadata values in ReviewCard
- show tooltip when text is truncated

image

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ temp/
 
 logs/
 __pycache__/
+website/agents.md
--- a/.gitignore
+++ b/.gitignore
@@ -25,4 +25,3 @@ temp/
 
 logs/
 __pycache__/
-website/agents.md
--- a/website/src/components/ReviewPage/ReviewCard.tsx
+++ b/website/src/components/ReviewPage/ReviewCard.tsx
@@ -1,4 +1,4 @@
-import { type FC, useState } from 'react';
+import { type FC, useState, useRef, useEffect } from 'react';
 
 import { SequencesDialog } from './SequencesDialog.tsx';
 import { backendClientHooks } from '../../services/serviceHooks.ts';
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -25,3 +25,4 @@ temp/
 
 logs/
 __pycache__/
+website/agents.md
```
chaoran-chen, line 28: This looks like it was committed by mistake?

website/src/components/ReviewPage/ReviewCard.tsx
```diff
@@ -403,19 +403,31 @@ const KeyValueComponent: FC<KeyValueComponentProps> = ({
...
     const noteTooltipId = 'note-tooltip-' + keyName + accessionVersion;
 
+    const textRef = useRef<HTMLSpanElement>(null);
+    const [isTruncated, setIsTruncated] = useState(false);
+
+    useEffect(() => {
+        if (textRef.current) {
+            setIsTruncated(textRef.current.scrollWidth > textRef.current.clientWidth);
+        }
+    }, [value]);
```
chaoran-chen, line 413: I'm not sure here but I am just wondering: do we need to add `textRef.current` as a dependency as well?

.gitignore
theosanderson: THanks - yes this was a hack before another thing was merged

website/src/components/ReviewPage/ReviewCard.tsx
theosanderson: References are kind of outside the React change-detection logic. This will just run immediately after render and that's ok here

theosanderson: (thanks for checking!)

chaoran-chen (approved):

Issue #2598, opened 2024-08-26
Detect metadata TSVs that don't contain tabs


theosanderson:
If you try to prepare a metadata TSV in e.g. VS code it uses spaces and not tabs.

This leads to a confusing message:
<img width="787" alt="image" src="https://github.com/user-attachments/assets/5ce7e1e5-6552-4126-94f7-0c09efbe3204">

This has confused some of us so it might confuse users. It would be nice to display a message saying "No tabs detected" if the first line contains spaces and not tabs.

PR #4146, opened 2025-05-17
feat(backend): Detect missing tabs in metadata TSV

theosanderson:
Resolves https://github.com/loculus-project/loculus/issues/2598

## Summary
- raise a specific error in the backend when metadata files don't have tabs and so are not TSVs
- detect this case for regular and revision metadata
- add unit tests for both metadata types

🚀 Preview: https://codex-detect-metadata-tsv.loculus.org

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ temp/
 
 logs/
 __pycache__/
+website/agents.md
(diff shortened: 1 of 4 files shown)


theosanderson:
Closing this for now as the buffered thing seems to open a can of potential worms that don't feel worth it this second

PR #4147, opened 2025-05-17
Fix duplicate DataUseTerms type

theosanderson:
## Summary
- prevent duplicate `type` field in DataUseTerms by ignoring the interface property

## Testing
- `./gradlew ktlintCheck` *(fails: Could not resolve dependencies)*
- `./gradlew test --no-daemon --stacktrace` *(fails: build did not complete)*

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/api/DataUseTerms.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/api/DataUseTerms.kt
@@ -57,6 +57,7 @@ enum class DataUseTermsType {
 )
 @JsonPropertyOrder(value = ["type", "restrictedUntil"])
 sealed interface DataUseTerms {
+    @get:JsonIgnore
     val type: DataUseTermsType
 
     @JsonTypeName("OPEN")


fhennig (approved):
wow, quite obscure but looks sensible to me!

chaoran-chen (commented):
Thanks! This is solving the output but it has some side effect on the Swagger documentation. On main, the [Swagger docs](https://backend-main.loculus.org/swagger-ui/index.html#/submission-controller/getSequenceList) example says:

```
      "dataUseTerms": {
        "type": "OPEN"
      }
```

and the schema:

image

With this change, the [Swagger](https://backend-codex-fix-duplicate-data.loculus.org/swagger-ui/index.html#/submission-controller/getSequenceList) example just say

```
    "dataUseTerms": {
        "type": "string"
      }
```

and the schema also just types it as `string`.

The reason why we had duplicated types was that [`@JsonTypeInfo`](https://fasterxml.github.io/jackson-annotations/javadoc/2.4/com/fasterxml/jackson/annotation/JsonTypeInfo.html) also adds the type. It is not obvious to me why we need `@JsonTypeInfo` and `@JsonSubTypes`, I wonder whether we could remove them instead.

I'm happy though to merge this as is and investigate and possibly refactor in a future PR.

fhennig:
This all seems quite niche and complicated - I wanted to suggest that maybe we document the annotations a bit better so we know what's there for which reason in the future?

chaoran-chen:
(I can try removing `@JsonTypeInfo` and `@JsonSubTypes` later or tomorrow if no one gets to it first.)

chaoran-chen:
ok, we can't just remove the two annotations as they are needed for deserialization. Let's then merge this PR as-is and create an issue to investigate whether we can improve the Swagger documentation?

PR #4160, opened 2025-05-19
chore(website): Use npm ci in website Dockerfile

theosanderson:
## Summary
- use `npm ci` in Dockerfile for reproducible dependency installs


website/Dockerfile
chaoran-chen, line 1: I think that `chore(website)` or `ci(website)` would make more sense as it does not change any feature of the software.

Issue #3470, opened 2024-12-19
Config: rename `order` to something more specific


fengelniederhammer:
https://github.com/loculus-project/loculus/blob/a7e76d573056e15f0f3b37003492fc9f8059139b/kubernetes/loculus/values.yaml#L76

I don't find `order` very intuitive. The name is quite out of context what kind of "order" this refers to. I suggest renaming it to `searchTableColumnOrder` or similar (maybe a bit lengthy but at least gives some hint to what it's used for).

PR #4161, opened 2025-05-19
Preserve column order in metadata downloads

theosanderson:
## Summary
- keep metadata download fields ordered per search table columns
- update download form logic accordingly
- keep field selector logic unchanged

## Testing
- `npm run test`
- `npm run check-types`
- `npm run format`


website/src/components/SearchPage/DownloadDialog/DownloadForm.tsx
chaoran-chen, line 1: It would also be nice to add a test. We have already a number of tests in `DownloadDialog.spec.tsx`, maybe we can add one there?

corneliusroemer:
How did this close #3470? I didn't see any renaming, so maybe we just decided to close as not planned? Did codex maybe misunderstand that issue? 😀

theosanderson:
That issue was saying "order" was too broad for something only used in the search table. Order is no longer used only in the search table so it seems a relatively reasonable name now.

PR #4163, opened 2025-05-19
Reformat prepro

anna-parker:
partially resolves https://github.com/loculus-project/loculus/issues/4189

### Screenshot
@corneliusroemer showed me how to enable type checking for python in VSCode - this PR cleans up prepro. It also adds a mypy type check to the cli to ensure that we do not regress again.

All changes are type related, main changes:
- ArgValue is now correctly typed as `ArgValue = list[str] | str | bool | None` and the type is checked before being used in functions
- `FunctionArgs` is now always a dictionary and never None (it was always used as a dict which caused type errors)
- In `ProcessingAnnotation` `unprocessedFields` and `processedFields` are now `list[AnnotationSource]` instead of tuples (they were typically used as lists)

### PR Checklist
- sign in as super user check that warnings and errors are being displayed correctly

🚀 Preview: https://reformat-prepro.loculus.org

### DIFF
--- a/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
@@ -41,10 +41,15 @@ def standardize_option(option):
     return " ".join(option.lower().split())
 
 
-def invalid_value_annotation(input_datum, output_field, input_fields, value_type) -> ProcessingAnnotation:
+def invalid_value_annotation(
+    input_datum, output_field, input_fields, value_type
+) -> ProcessingAnnotation:
     return ProcessingAnnotation(
         processedFields=[AnnotationSource(name=output_field, type=AnnotationSourceType.METADATA)],
-        unprocessedFields=[AnnotationSource(name=field, type=AnnotationSourceType.METADATA) for field in input_fields],
+        unprocessedFields=[
+            AnnotationSource(name=field, type=AnnotationSourceType.METADATA)
+            for field in input_fields
+        ],
         message=f"Invalid {value_type} value: {input_datum} for field {output_field}.",
     )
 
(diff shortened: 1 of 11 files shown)


theosanderson (approved):

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -582,16 +587,33 @@ def concatenate(
 
         number_fields = len(input_data.keys()) + 1
 
-        accession_version = args["accession_version"]
+        if not isinstance(args["accession_version"], str):
```
corneliusroemer, line 590: Ugh this is quite horribly verbose - I don't think we need to add that much code here?

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -582,16 +587,33 @@ def concatenate(
...
         order = args["order"]
         type = args["type"]
 
-        # Check accessionVersion only exists once in the list:
-        if number_fields != len(order):
-            logging.error(
-                f"Concatenate: Expected {len(order)} fields, got {number_fields}. "
-                f"This is probably a configuration error. (accession_version: {accession_version})"
-            )
+        def add_errors():
```
corneliusroemer, line 616: That's a weird function - rather than calling a magic `add_errors` function, we should just define the error item once, then do `errors.append(standard_error)` - that's a lot more explicit and doesn't use closure

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -605,28 +627,61 @@ def concatenate(
                     "This may be a configuration error, please contact the administrator.",
                 )
             )
+
```
corneliusroemer, line 630: Verbosity alarm

PR #4167, opened 2025-05-19
Reduce type-check hints

theosanderson:
## Summary
- remove unused variables and adjust imports
- inline redirect helpers
- prefix variables to avoid type hints

## Testing
- `npm run format`
- `npm run check-types`
- `npm run test -- --run`


fhennig (approved):
thanks for that, has been bugging me for a while!

PR #4169, opened 2025-05-19
fix(backend): fix disabling of S3

chaoran-chen:
The backend cannot start at the moment if S3 is disabled because it still needs to initialize an `S3Service` and the service is creating an S3 client while being initialized, for which it needs S3 connection data. This PR fixes the problem by offering two S3Service implementations: one real implementation that is used when S3 is enabled and another implementation that is used when S3 is disabled. The latter will always initialize but only throw an error when its functions are being called.

### PR Checklist
    - I verified in a local deployment that it is now possible to set `s3.enabled=false`


backend/src/main/kotlin/org/loculus/backend/service/files/S3ServiceNoop.kt
theosanderson, line 1: When you say "S3 is disabled" you mean minio is disabled? But we would still need an S3 Service in that case?

theosanderson: (e.g. for AWS S3)

theosanderson: (sorry by "minio" here I mean 'our minio server')

chaoran-chen: I mean if `s3.enabled=false`, i.e., no S3 service is configured. Then, the code still needs an instance of the `S3Service`: At the moment, the `/files` endpoint would still exist, and users could call it, but would then get an error message. We should disable the controller, but I'd do it separately, and it's not causing direct problems if it is around. Apart from the `/files`-controller, other parts of the code, like the very broadly used `SubmissionDatabaseService`, also have a reference to the `S3Service`, but they should only call a function of the `S3Service` when files are actually used.

backend/src/main/kotlin/org/loculus/backend/service/files/S3Service.kt
chaoran-chen, line 1: @fhennig (and anyone else), what do you think of this implementation?

backend/src/main/kotlin/org/loculus/backend/service/files/S3ServiceNoop.kt
theosanderson: Ah - sorry I had confused "runDevelopmentS3" and "S3.enabled"

backend/src/main/kotlin/org/loculus/backend/service/files/S3Service.kt
theosanderson: (My instinct is that it feels a bit weird to create a special version of a service that doesn't work, but I don't know Kotlin so it probably is the only/best way - and regardless it would be good to have something!)

theosanderson (approved):

PR #4170, opened 2025-05-19
feat(website): limit max number of displayed autocomplete options

chaoran-chen:
The auto-completion of the search is crashing my browser tab if there are too many options. As a solution, this PR proposes to only show a limited number of options. I observed a problem in a field with ~8000 values while it worked well for ~700 values, so I chose a cutoff of 1000.

### PR Checklist
    - Up to 1000 options work well for me (if we observe that it is too much for some devices, we can further adjust)


website/src/components/SearchPage/fields/AutoCompleteField.tsx
chaoran-chen, line 1: @theosanderson, does this sound like a reasonable solution to you?

theosanderson: Yes – thanks

theosanderson (approved):
Great!

PR #2451, opened 2024-08-19
Ingest add revocation comment

anna-parker:
### Summary

1. Add description of why sequence has been revoked if revoked due to grouping changes
2. Only revoke a sequence due to a regrouping if there is some uncertainty on how to regroup, e.g. if due to new sequences a previous grouping must be split up or two groups have to be joined. If a new segment can be added to an existing group (i.e. was missing before) just add it to the group using a revision
3. Small bug fixes: give revoke_and_regroup cron-job proper permissions to actually be able to do its job
4. Small bug fixes: use the latest `jointAccession` for comparing - otherwise if a new segment is added we can get into a revision loop.

### PR Checklist
Use the new [CCHF nextclade dataset](https://github.com/nextstrain/nextclade_data/pull/265) for testing which improves alignment, resulting in many M segments that were previously lost now being aligned and ingested.

- ingest using current dataset
- set db to persistent
- ingest and processing using new dataset
- run revoke_and_regroup cronjob
- check results

### Screenshots
- Message is sent with sequences that should be revoked and regrouped:
image
- Message after sequence has been revoked is clear:
image

🚀 Preview: https://ingest-add-revocation-com.loculus.org

PR #4177, opened 2025-05-20
feat(ingest): refactor call_loculus and compare_hashes

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/4183

### Screenshot
As discussed with @corneliusroemer the regrouping logic is quite complex and adding to this it has been split across two files, this file moves most of the logic into compare_hashes and refactors call_loculus to just return the function call results.

This should keep all current ingest functionality and logic.

1. call_loculus.get_submitted no longer returns a parsed json but the raw ndjson (with added status)
2. get_submitted output is now parsed in one location: compare_hashes
- code is refactored and types and classes are added
- edge case: two loculus accessions for one INSDC accession no longer results in a value error if one of those accessions has been revoked (this is important for regrouping)
- the argument `debug_hashes` is removed as it is no longer used

### PR Checklist
- [x] The implemented feature is covered by appropriate, automated tests:
edge case added to tests: state where two sequences were previously grouped and then that grouping was revoked and two new sequences created
- set db to persistent
- make curation and confirm ingest does not revise: LOC_00003Y0.2 (ebola), LOC_000M14G.2 (revoked), LOC_000PBCL.2 (cchf)
- use new CCHF dataset, check output -> same results as in https://github.com/loculus-project/loculus/pull/2451
- run cronjob -> succeeded
- check reingest is working well, logs look good curated sequences are also not overwritten, no duplicate accession error
- [x] test on staging that can revoke and regroup and doesn't lead to any future re-ingest errors: https://github.com/loculus-project/loculus/pull/4177

🚀 Preview: https://start-ingest-refactor.loculus.org


ingest/scripts/compare_hashes.py
```diff
@@ -62,60 +71,37 @@ def md5_float(string: str) -> float:
...
-    fasta_id: str,
+    fasta_id: JointInsdcAccession,
     ingested_hash: str,
-    submitted: dict[InsdcAccession : dict[str, Any]],
+    submitted: dict[InsdcAccession, LatestVersion],
     update_manager: SequenceUpdateManager,
 ):
     """
     Decide if fasta_id should be submitted, revised, or noop
-    The ingested_insdc_accession corresponds to the INSDC accession base
```
anna-parker, line 86: this was actually inaccurate - I have corrected the types

theosanderson: My codexing identified a few clear type issues that I assume would have been clear to a linter: wondering if we'd want to add CI to check these?

anna-parker: that sounds like a great idea! I didn't have the python linter enabled properly before and missed quite a few of these, so would be good to enable in a follow up PR :-)

kubernetes/loculus/templates/loculus-database-standin.yaml
```diff
@@ -5,7 +5,7 @@ kind: Deployment
 metadata:
   name: loculus-database
   annotations:
-    argocd.argoproj.io/sync-options: Replace=true
+    argocd.argoproj.io/sync-options: Replace=false
```
anna-parker, line 8: revert before merging
```suggestion
    argocd.argoproj.io/sync-options: Replace=true
```

kubernetes/loculus/values.yaml
```diff
@@ -1670,8 +1670,8 @@ defaultOrganisms:
         taxon_id: 3052518
         segment_identification:
           method: "align"
-          nextclade_dataset_server: https://raw.githubusercontent.com/nextstrain/nextclade_data/cornelius-cchfv/data_output
-          nextclade_dataset_name: nextstrain/cchfv/linked
+          nextclade_dataset_name: community/pathoplexus/cchfv
+          nextclade_dataset_server: https://raw.githubusercontent.com/anna-parker/nextclade_data/cchfv_updates/data_output
```
anna-parker, line 1674: revert before merging
```suggestion
          nextclade_dataset_name: nextstrain/cchfv/linked
          nextclade_dataset_server: https://raw.githubusercontent.com/nextstrain/nextclade_data/cornelius-cchfv/data_output
```

ingest/scripts/compare_hashes.py
anna-parker: I created an issue for this: https://github.com/loculus-project/loculus/issues/4189

ingest/scripts/compare_hashes.py
```diff
@@ -62,60 +76,34 @@ def md5_float(string: str) -> float:
 
 
 def sample_out_hashed_records(
-    joint_insdc_accession: str,
+    joint_insdc_accession: JointInsdcAccession,
     subsample_fraction: float,
-    sampled_out: dict[str, str],
-    fasta_id: str,
-) -> tuple[float, list[dict[str, str]]]:
+) -> float:
```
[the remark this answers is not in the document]
anna-parker: ah yes wupps should be bool

ingest/scripts/call_loculus.py
```diff
@@ -426,24 +427,10 @@ def get_sequence_status(config: Config):
     return result
 
 
-def get_submitted(config: Config):
-    """Get previously submitted sequences
+def get_submitted(config: Config, output: str):
```
corneliusroemer, line 430: Might be nice to add return types and/or the shape of the output in the doc comment

corneliusroemer (approved):

PR #4184, opened 2025-05-21
Remove S3 Noop service

theosanderson:
## Summary
- drop `S3ServiceNoop`
- update `S3ServiceImpl` to always exist and lazily create clients
- adjust tests for the new behaviour

## Testing
- `./backend/gradlew -p backend bootJar`


kubernetes/loculus/values.yaml
theosanderson, line 1: revert changes

theosanderson:
@chaoran-chen if you have a chance to consider if this looks reasonable would appreciate it :)

backend/src/main/kotlin/org/loculus/backend/service/files/S3ServiceImpl.kt
```diff
@@ -6,17 +6,15 @@ import io.minio.SetObjectTagsArgs
...
 
 private const val PRESIGNED_URL_EXPIRY_SECONDS = 60 * 30
 
 @Service
-@ConditionalOnProperty(name = ["loculus.s3.enabled"], havingValue = "true", matchIfMissing = false)
 class S3ServiceImpl(private val s3Config: S3Config) : S3Service {
-    private val client: MinioClient = createClient(getS3BucketConfig())
-    private val internalClient: MinioClient = createClient(getS3BucketConfig(), true)
+    private val clientValue: MinioClient by lazy { createClient(getS3BucketConfig()) }
+    private val internalClientValue: MinioClient by lazy { createClient(getS3BucketConfig(), true) }
```
chaoran-chen: Was it intentional to change the variable names??

chaoran-chen (approved):
Thanks very much, @theosanderson! `by lazy` is indeed a nice solution. I just adapted the tests and removed the now unneeded separation between the S3Service interface and S3ServiceImpl.

backend/src/main/kotlin/org/loculus/backend/service/files/S3ServiceImpl.kt
theosanderson: You'd have to ask the LLM - I guess it would be to distinguish a variable as the output of `getClient` with the value used within it, but I've reverted

chaoran-chen: Interesting! I'll take a look later.

theosanderson: OK, Claude explained it to me. Kotlin will automatically generate a method `getClient` (or something) for something named `Client` and that will clash with the manually declared `getClient`.

chaoran-chen: Thanks, interesting, I didn't know. So, we could instead rename it to `minioClient`? For me, that's a better name than `clientValue` as the "value" confuses me. But no big deal!

PR #4186, opened 2025-05-21
chore(backend): backend helpers for codex and and allow testing without docker

theosanderson:
This PR allows OpenAI Codex to work on the backend by allowing it to run tests. Codex is currently unable to use docker, which we use for testcontainers for minio and postgres. This PR allows the optional use of `minio` and `postgres` binaries instead of docker, enabled by setting `USE_NONDOCKER_INFRA` to `true`. It maintains all existing functionality and keeps this special logic to use the binaries separate so that it could easily be removed in future if this becomes irrelevant.

It also adds a `downloadDependencies` function which can be used to preinstall all Kotlin dependencies.


fhennig (approved):
looks generally fine! I'm assuming nothing will change for "normal" usage

PR #4200, opened 2025-05-24
Pass pod priority to ENA submission cronjob

theosanderson:
## Summary
- make the ENA submission cronjob use podPriorityClassName if provided

## Testing
- `helm` or `yamllint` not available


anna-parker (approved):
Thanks! I added this for the revoke and regroup cronjob as well :-)

PR #4218, opened 2025-05-27
feat(db): add index on organism in backend DB

theosanderson:
This seems like low-hanging fruit that we should have done a long time ago

### DIFF
--- a/backend/src/main/resources/db/migration/V1.22__organism_index.sql
+++ b/backend/src/main/resources/db/migration/V1.22__organism_index.sql
@@ -1,0 +1,1 @@
+CREATE INDEX IF NOT EXISTS sequence_entries_organism_idx ON public.sequence_entries USING btree (organism);


anna-parker (approved):

backend/src/main/resources/db/migration/V1.22__organism_index.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX IF NOT EXISTS sequence_entries_organism_idx ON public.sequence_entries USING btree (organism);
```
chaoran-chen, line 1:
```suggestion
CREATE INDEX IF NOT EXISTS sequence_entries_organism_idx ON sequence_entries (organism);

```

- We don't specify a schema (`public`).
- The default is `btree`, so we don't need to specify it.

chaoran-chen (approved):

backend/src/main/resources/db/migration/V1.22__organism_index.sql
chaoran-chen, line 1: Why V1.22? It should be V1.12?

Issue #4110, opened 2025-05-14
file sharing: track and display file size


fhennig:
As a user I want to see the size of the file in advance, before I hit download, so I know what to expect and how long the download might take.

fhennig:
@chaoran-chen any implementation preferences?

theosanderson:
(sorry that I missed this question)

chaoran-chen:
(No worries, after looking at the code, I found your suggestion easier to implement :))

PR #4227, opened 2025-05-28
feat(backend): validate file existence and fetch file size

### LINKED ISSUE #3995: Validate file existence
When the backend receives a file mapping (currently, this happens in `/submit`, `/revise`, `/submit-edited-data`, `/submit-processed-data`), it currently only checks that the file ID exists but it doesn't check whether a file has actually been uploaded. It should check it.

fhennig:
resolves #3995 

- Adds the `size` field to the file table which tracks the size of the files. This will be used in #4110 as well, but is also used to track whether a file was uploaded already or not.
- Add the check for file existence in the four relevant endpoints. 
- Add backend tests for the check

### PR Checklist
- ~~Any manual testing that has been done is documented (i.e. what exactly was tested?)~~ IMO The automated tests are sufficient

🚀 Preview: https://fs-file-existence.loculus.org


backend/src/main/kotlin/org/loculus/backend/service/submission/FileMappingPreconditionValidator.kt
```diff
@@ -45,6 +53,19 @@ class FileMappingPreconditionValidator(private val backendConfig: BackendConfig)
...
+
+    fun validateFilesExist(fileIds: Set<FileId>): FileMappingPreconditionValidator {
+        val uncheckedFileIds = filesDatabaseService.getUncheckedFileIds(fileIds)
+        uncheckedFileIds.forEach { fileId ->
+            val fileSize = s3Service.getFileSize(fileId)
+            if (fileSize == null) {
+                throw UnprocessableEntityException("No file uploaded for file ID $fileId.")
+            } else {
+                filesDatabaseService.setFileSize(fileId, fileSize)
+            }
```
chaoran-chen, line 65:
```suggestion
            }
            filesDatabaseService.setFileSize(fileId, fileSize)
```

backend/src/main/kotlin/org/loculus/backend/service/files/FilesDatabaseService.kt
```diff
@@ -49,4 +47,21 @@ class FilesDatabaseService(private val dateProvider: DateProvider) {
         .map { it[FilesTable.releasedAtColumn] }
         .first()
         .let { it != null }
+
+    /**
+     * Return the subset of file IDs that either do not exist or have not been checked yet.
```
chaoran-chen, line 52: Is the comment accurate? The function doesn't return file IDs that do not exist, but only those for which the size is not known, right?

backend/src/test/kotlin/org/loculus/backend/controller/submission/SubmissionConvenienceClient.kt
```diff
@@ -516,4 +521,19 @@ class SubmissionConvenienceClient(
                 .contentAsString
         return objectMapper.readValue(content)
     }
+
+    /**
+     * Upload a file (content) to a presigned write URL (S3).
+     */
+    fun uploadContent(presignedWriteUrl: String, content: String) {
```
chaoran-chen, line 528: Why not just name it `uploadFile` - i.e., why introduce the term `content`? I think that we're not using this term otherwise and I found it a bit confusing.

chaoran-chen (approved):
Looks great! I also manually tested it in Swagger and the errors are thrown as expected.

backend/src/test/kotlin/org/loculus/backend/controller/submission/SubmissionConvenienceClient.kt
fhennig: Hmm, not sure why the name was chosen. But I've changed it now! https://github.com/loculus-project/loculus/pull/4227/commits/9d72abd25922018fdf547ddc58385b2d0ce2cf31

Issue #4236, opened 2025-05-28
Sequence details modal flashes wrong data for a split second causing jarring double flash of content


corneliusroemer:
I've just noticed consciously that upon clicking a sequence, we first show sequence details in the modal, then remove the sequence data and flash `Loading...` before showing the sequence details again.

If you slow things down by a factor of 20 it's easier to see (but the bare eye can see it as well if carefully focussing on it). I'm not sure why we're doing this. Playing around a bit more, I noticed we actually show potentially _wrong_ data first (if you change the sequence you click at every time)

Notice how the details are first for a different sequence. E.g. the coverage percentage is different, frameshifts etc.

https://github.com/user-attachments/assets/1a8dc961-18e9-4515-be96-252c8ea87e37

First screen:

<img width="513" alt="Image" src="https://github.com/user-attachments/assets/22160078-7e7a-40b6-accc-ca83db300a05" />

Second screen:

<img width="640" alt="Image" src="https://github.com/user-attachments/assets/5b159776-ddc9-4fa1-a59f-c177eded2590" />

PR #4237, opened 2025-05-28
fix(website): reset sequence preview state when switching

theosanderson:
## Summary
- ensure Sequence Preview modal resets when selecting a new sequence

Resolves https://github.com/loculus-project/loculus/issues/4236

I couldn't see the issue in the preview

🚀 Preview: https://codex-fix-sequence-detail.loculus.org


corneliusroemer (approved):
Nice, this seems to work - I wonder if we have more such bugs, I remember seeing these double flashes in search dropdowns for example.

Also, wondering if we might want to not show `Loading...` for the first 500ms as it's unnecessary flicker to show text for only a fraction of a second. But that might require deeper changes

PR #4238, opened 2025-05-28
feat(website): only show "Loading..." text after 500ms of loading

corneliusroemer:
Reduce flicker, only show loading if it's actually taking a while to load.

### PR Checklist

Mostly written by GPT-4.1 in VS Code in Agent mode

🚀 Preview: https://delay-loading.loculus.org


theosanderson:
Hm. When I tried this I was connecting through my phone internet (accidentally) and the experience was actually significantly worse in that there was some jarring as the Loading appeared and then more jarring when it disappeared after sometimes like 50 ms. But I guess not many of our users will be on phones.

theosanderson:
Yes, what you describe, and yes I've used the fading-in technique in the past

PR #4246, opened 2025-05-30
chore(k8s): align revoke-regroup cronjob names

corneliusroemer:
Job and template metadata were out of sync.
Now they are both the same as they usually are and also more informative
about what's actually happening.
We no longer do ingest via cronjob, only revoke, hence good to be
specific

resolves #

### PR Checklist


anna-parker (approved):
wupps! thanks for catching this!

PR #4248, opened 2025-05-30
feat(website): add metadata fields and organisms to seqsets table

theosanderson:
Now SeqSets are getting use we fairly urgently need to make them look less basic as they will be many people's first entrypoint to Pathoplexus. In particular we need to make sure they credit submitters effectively without further clicks. This implements that in a fairly hacky way where we lazily query LAPIS once we get a list of accessions. It's also not tested for now - we need to bring the SeqSets over to the integration tests but I don't think this is blocking and can't take that on now.

image

## Summary

This PR enhances the seqsets table by:
1. Making the it display metadata fields configurable via Helm values
2. Adding an organism column that shows user-friendly display names

## Changes

- Added `fieldsToDisplay` configuration under `seqSets` in `values.yaml` with default fields:
  - Country (`geoLocCountry`)
  - Collection Date (`sampleCollectionDate`) 
  - Authors (`authors`)
- Updated Helm templates to pass this configuration to the website
- Made the seq-sets table query LAPIS for these metadata when it receives the list of accessions

🚀 Preview: https://seqset-metadata.loculus.org

### DIFF
--- a/kubernetes/loculus/templates/lapis-deployment.yaml
+++ b/kubernetes/loculus/templates/lapis-deployment.yaml
@@ -43,11 +43,16 @@ spec:
             - name: lapis-silo-database-config-processed
               mountPath: /workspace/reference_genomes.json
               subPath: reference_genomes.json
+          startupProbe:
+            httpGet:
+              path: /actuator/health
+              port: 8080
+            periodSeconds: 5
+            failureThreshold: 36 # 3 minutes to start
           readinessProbe:
             httpGet:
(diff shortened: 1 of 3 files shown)


corneliusroemer (approved):
Not tested? Manual testing is also testing - I'm sure you did that :)

Works well, just tested myself!

Went briefly through the code but not in detail. Looks great!

image

theosanderson:
There's a screenshot above :)

PR #4252, opened 2025-05-30
fix(backend): fix large seqset bug

### LINKED ISSUE #4251: Creating very large SeqSets fails
If I paste a very large set of sequences into the create seqset form I get this error

```
Internal Server Error. org.postgresql.util.PSQLException: ERROR: stack depth limit exceeded Hint: Increase the configuration parameter "max_stack_depth" (currently 2048kB), after ensuring the platform's stack depth limit is adequate.
```

The error is before I even try to make the SeqSet I think it is when it is attempting to validate the accessions

theosanderson:
appears to resolve #4251 by batching checks to avoid SQL issues

I created a SeqSet with all WNV sequences which I seemed unable to do otherwise

🚀 Preview: https://seqset-sql.loculus.org


anna-parker (approved):
LGTM - I generated a small seqSet and deleted it - I trust this fixes the issue for larger seqSets

PR #4253, opened 2025-05-30
fix(website): move download button to file upload component on edit page

### LINKED ISSUE #3877: Move download sequence button to be part of upload component on revise/edit page
Currently (after Felix's PR) we have a button for Downloading the sequence on the edit page which I think is in a very unintuitive place ( I find myself wanting to hit it to submit)

Image

I think we should change this to like this:

Image

We should probably put a border around this component too

theosanderson:
## Summary
- Moves the download button from the bottom action buttons to be integrated within each file upload component

image

## Motivation
Fixes #3877 

The download button was previously positioned at the bottom of the edit page next to the submit button, which was confusing for users as it felt like it was where a submit button would be. This change moves the download functionality to be part of each file upload component, and makes it download each segment's data separately.

🚀 Preview: https://fix-3877-move-download-bu.loculus.org

### DIFF
--- a/website/src/components/Edit/SequencesForm.tsx
+++ b/website/src/components/Edit/SequencesForm.tsx
@@ -105,7 +105,7 @@ export const SequencesForm: FC<SequenceFormProps> = ({
 }) => {
     const singleSegment = editableSequences.rows.length === 1;
     return (
-        <div className='border border-gray-300 rounded-lg p-4'>
+        <>
             <h3 className='subtitle'>{`Nucleotide sequence${singleSegment ? '' : 's'}`}</h3>
             <div className='flex flex-col lg:flex-row gap-6'>
                 {editableSequences.rows.map((field) => (
@@ -159,6 +159,6 @@ export const SequencesForm: FC<SequenceFormProps> = ({
                     </div>
                 ))}
             </div>
-        </div>
+        </>
     );
 };
--- a/website/src/components/Edit/SequencesForm.tsx
+++ b/website/src/components/Edit/SequencesForm.tsx
@@ -94,11 +94,18 @@ export class EditableSequences {
 type SequenceFormProps = {
     editableSequences: EditableSequences;
     setEditableSequences: Dispatch<SetStateAction<EditableSequences>>;
+    dataToEdit?: SequenceEntryToEdit;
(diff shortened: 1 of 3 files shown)


website/src/components/Edit/SequencesForm.tsx
```diff
@@ -126,6 +133,28 @@ export const SequencesForm: FC<SequenceFormProps> = ({ editableSequences, setEdi
                                     : undefined
                             }
                             showUndo={true}
+                            onDownload={
```
anna-parker, line 136: Is it possible to keep this code in a function generateAndDownloadFastaFile? - it might be easier to read

theosanderson: Thanks - now done!

fhennig (approved):
Had a go at it in the preview, works! Code looks good too, I like it that way!

I think having a test would be good, but I appreciate that there wasn't a test before and you're basically just moving the button. So a test is appreciated but not required.

Thanks!

website/src/components/Edit/SequencesForm.tsx
```diff
@@ -126,6 +157,20 @@ export const SequencesForm: FC<SequenceFormProps> = ({ editableSequences, setEdi
...
+                                field.initialValue.length > 0 && dataToEdit
+                                    ? () => {
+                                          const accessionVersion = `${dataToEdit.accession}.${dataToEdit.version}`;
+                                          generateAndDownloadFastaFile(
+                                              accessionVersion,
+                                              field.initialValue,
+                                              field.key,
```
anna-parker, line 167: From the code above it looks like field.key always exists -> I think we can then simplify the `generateAndDownloadFastaFile` function.

Also as a side note - where is the key defined to be the segmentName? It might be nice to add that somewhere as a comment at least

theosanderson: Typescript would generally tell us if we were doing a pointless comparison - I expect we are passing on something that could be undefined or null

theosanderson: The key thing does seem confusing but it's not something I've introduced here and I'd prefer if we tackle it separately as I'm not sure exactly where it comes from and where we should document this

PR #4255, opened 2025-05-31
fix(website): Fix separator used in processing authors so that Show more / Show less works correctly

theosanderson:
We changed the separator but forgot to update this so at the moment we truncate to just the first name / initial of last author, not their full name.

image

### DIFF
--- a/website/src/components/SequenceDetailsPage/getDataTableData.ts
+++ b/website/src/components/SequenceDetailsPage/getDataTableData.ts
@@ -70,7 +70,7 @@ export function getDataTableData(listTableDataEntries: TableDataEntry[]): DataTa
         ) {
             result.topmatter.authors = entry.value
                 .toString()
-                .split(',')
+                .split(';')
                 .map((x) => x.trim());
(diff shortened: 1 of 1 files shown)


anna-parker (changes requested):
This is somehow converting all the `;` into `,`, making the split between authors hard to seeimage

theosanderson:
Ah, I'm actually used to seeing authors formatted like that so it didn't throw me. Hmm.. this is how e.g. GenBank formats stuff - I think at some point we should consider if semicolons are a bad way to display authors - but we can separate that question out!

theosanderson:
(I fixed that – thanks for flagging)

anna-parker (approved):
Looks good thanks!
image

theosanderson:
> Should something like 5 names be enough to make it look a bit nicer?

IMO no it's good to give credit pretty widely by default unless it's a really unmanageable number of names. We were inspired by NPG with this _View more_ thing and they often wrap across multiple lines.

PR #4274, opened 2025-06-02
fix(website): fallback to `name` if `displayName` not available

### LINKED ISSUE #4273: Some search filter boxes are empty
We seem to have some search filter boxes have no text at all (on main loculus at least) 

Maybe we used to show something there in place of `displayName` and the `displayName` is missing here?

Image

theosanderson:
resolves #4273 at least for now - we should do some thinking about whether we use `sentenceCase` for these sorts of purposes or not (we currently inconsistently sometimes do and sometimes don't - but it would be great to merge this as the current state is bad!)

### Screenshot
image

🚀 Preview: https://namefallback.loculus.org

### DIFF
--- a/website/src/components/SearchPage/SearchForm.tsx
+++ b/website/src/components/SearchPage/SearchForm.tsx
@@ -160,7 +160,7 @@ const SearchField = ({ field, lapisUrl, fieldValues, setSomeFieldValues, lapisSe
         } else {
             return (
                 <div key={field.name} className='flex flex-col border p-3 mb-3 rounded-md border-gray-300'>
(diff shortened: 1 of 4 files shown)


corneliusroemer (approved):
We could theoretically use a utility function here to simplify updates in the future, we'd only need to make things sentence case in one place then. on the other hand, if these are the only 3 places, it might not be worth it. but i suspect we have more places where we do (or should be doing) `field.displayName ?? field.name`

theosanderson:
:+1: there are other places where we are currently using `sentenceCase` - we could standardise on a utility function. We could also potentially make everything have a `displayName` which is just taken from the `name` if unset

corneliusroemer:
I've made an issue: #4276

PR #4298, opened 2025-06-03
feat(website): Add dialog to view processed files in 'review' window (before approving)

### LINKED ISSUE #3968: Make it possible to view processed files in the Review overview
As a reviewer I want to see the files that the pipeline generated when deciding whether to approve a sequence for release or not, so I can make sure that the files look good.

At that point, the files are still private. To allow the user to see them, we need to generate pre-signed URLs. We should have a button on a review card that the user can click, and in the popup we can link to the files associated with the sequence. To get the links, we need a new endpoint.

fhennig:
resolves #3968 

Adds a button similar to the 'view sequences' button to view files that opens a dialog with the file links.

The /seq/...filename.txt route is refactored to directly return a redirect to S3, instead of a redirect to the backend (which could then redirect to S3), so one network hop is saved.

### Manual testing

I've made a submission with files, waited for them to process, and then downloaded them through the dialog. works! See also the video below.

### Screenshot

https://github.com/user-attachments/assets/4af8f542-823f-4e4f-8d2f-0be38b5d3480

### PR Checklist

🚀 Preview: https://review-files.loculus.org

### DIFF
--- a/integration-tests/tests/specs/features/file-sharing.spec.ts
+++ b/integration-tests/tests/specs/features/file-sharing.spec.ts
@@ -31,6 +31,14 @@ test('submit a single sequence with two files', async ({ pageWithGroup, page }) 
     const reviewPage = await submissionPage.submitSequence();
     await reviewPage.waitForZeroProcessing();
 
+    // check that files can be seen after processing
+    const filesDialog = await reviewPage.viewFiles();
+    expect(filesDialog.getByText('hello.txt')).toBeVisible();
+    await checkFileContent(page, 'hello.txt', 'Hello');
+    expect(filesDialog.getByText('world.txt')).toBeVisible();
+    await checkFileContent(page, 'world.txt', 'World');
+    await reviewPage.closeFilesDialog();
+
     await reviewPage.releaseValidSequences();
 
     await page.getByRole('link', { name: 'released sequences' }).click();
--- a/integration-tests/tests/specs/features/file-sharing.spec.ts
+++ b/integration-tests/tests/specs/features/file-sharing.spec.ts
@@ -33,9 +33,9 @@ test('submit a single sequence with two files', async ({ pageWithGroup, page }) 
 
     // check that files can be seen after processing
     const filesDialog = await reviewPage.viewFiles();
-    expect(filesDialog.getByText('hello.txt')).toBeVisible();
+    await expect(filesDialog.getByText('hello.txt')).toBeVisible();
     await checkFileContent(page, 'hello.txt', 'Hello');
-    expect(filesDialog.getByText('world.txt')).toBeVisible();
+    await expect(filesDialog.getByText('world.txt')).toBeVisible();
     await checkFileContent(page, 'world.txt', 'World');
     await reviewPage.closeFilesDialog();
 
--- a/website/src/components/ReviewPage/FilesDialog.tsx
+++ b/website/src/components/ReviewPage/FilesDialog.tsx
@@ -1,38 +1,75 @@
 import { type FC } from 'react';
+import { type SequenceEntryToEdit } from '../../types/backend.ts';
 
 type FilesDialogProps = {
     isOpen: boolean;
     onClose: () => void;
+    dataToView: SequenceEntryToEdit | undefined;
 };
 
-const exampleFiles = [
-    { name: 'Example 1', url: 'https://example.com/file1.txt' },
-    { name: 'Example 2', url: 'https://example.com/file2.txt' },
-    { name: 'Example 3', url: 'https://example.com/file3.txt' },
-];
+export const FilesDialog: FC<FilesDialogProps> = ({ isOpen, onClose, dataToView }) => {
+    if (!isOpen || !dataToView) return null;
 
-export const FilesDialog: FC<FilesDialogProps> = ({ isOpen, onClose }) => {
-    if (!isOpen) return null;
+    const originalFilesByCategory = dataToView.originalData.files;
+    const processedFilesByCategory = dataToView.processedData.files;
 
     return (
         <div className='fixed inset-0 flex items-center justify-center z-50 overflow-auto bg-black bg-opacity-30'>
             <div className='bg-white rounded-lg p-6 max-w-xl mx-3 w-full max-h-[90vh] flex flex-col'>
                 <div className='flex justify-between items-center mb-4'>
                     <h2 className='text-xl font-semibold'>Files</h2>
-                    <button className='text-gray-500 hover:text-gray-700' onClick={onClose}>
-                        ✕
-                    </button>
+                    <button className='text-gray-500 hover:text-gray-700' onClick={onClose}>✕</button>
                 </div>
 
-                <ul className='list-disc pl-5 space-y-2'>
-                    {exampleFiles.map(({ name, url }) => (
-                        <li key={url}>
-                            <a href={url} className='text-blue-600 hover:underline' target='_blank' rel='noopener noreferrer'>
-                                {name}
-                            </a>
-                        </li>
-                    ))}
-                </ul>
+                <div className='overflow-auto space-y-6'>
+
+                    <div>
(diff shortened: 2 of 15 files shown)


website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts
```diff
@@ -3,17 +3,52 @@ import type { APIRoute } from 'astro';
...
-export const GET: APIRoute = ({ params }) => {
+function getCookieValue(cookieHeader: string | null, cookieName: string): string | null {
+    if (!cookieHeader) return null;
+    const cookies = cookieHeader.split(';').map((c) => c.trim());
+    for (const cookie of cookies) {
+        const [name, ...rest] = cookie.split('=');
+        if (name === cookieName) return rest.join('=');
+    }
+    return null;
+}
```
Comment (theosanderson) [inline website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts:14]: Have a look at how we handle access tokens in the groups pages maybe? I think that's a good pattern to follow [although I don't remember exactly how it works, so do correct me!]

Comment (theosanderson) [reply website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts:14]: https://github.com/loculus-project/loculus/blob/88efaf0b7a792d89e2e8fa952d1ac0f12de67da8/website/src/pages/user/createGroup.astro#L7

Comment (fhennig) [reply website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts]: yeah that bit was ChatGPT. I've used the 'cookie' prop from the context now: https://github.com/loculus-project/loculus/pull/4298/commits/8d3b1fa4a65e027b93b931ab0a28cd805dc7cf37

To check if it really works, this needs to be merged: https://github.com/loculus-project/loculus/pull/4301

Comment (theosanderson) [reply website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts]: Thanks sorry for the too-early comment!

Comment (fhennig) [reply website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts]: But is the session available? Because the route code is running in the backend, and I believe the code you linked is running in the client. But if it's available it'd make sense, agreed. I'll check it soon!

Comment (theosanderson) [reply website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts]: Yeah if you see screenshot of docs above I think you need to do it a bit differently for API routes but believe it's accessible

website/src/components/ReviewPage/FilesDialog.tsx
```diff
@@ -0,0 +1,48 @@
...
+                    <h2 className='text-xl font-semibold'>Files</h2>
+                    <button className='text-gray-500 hover:text-gray-700' onClick={onClose}>
+                        ✕
+                    </button>
+                </div>
+
+                <div>
+                    {Object.entries(dataToView.processedData.files ?? []).map(([category, files]) => (
+                        <div key={category} className='mb-4'>
+                            <h3 className='font-medium'>{category}</h3>
```
theosanderson, line 27: Sometime (feel free to make an issue) we should add a `displayName` for file categories

website/src/components/ReviewPage/FilesDialog.tsx
```diff
@@ -0,0 +1,48 @@
...
+                            <ul className='list-disc pl-5 space-y-1'>
+                                {files.map((file) => (
+                                    <li key={file.fileId}>
+                                        <a
+                                            href={`/seq/${dataToView.accession}.${dataToView.version}/${category}/${file.name}`}
+                                            className='text-blue-600 hover:underline'
```
theosanderson, line 33: We should probably use primary colors for this as I don't think we use blue elsewhere

theosanderson (commented):
Looking great generally! The endpoint is really cool

website/src/components/ReviewPage/ReviewCard.tsx
```diff
@@ -128,19 +134,36 @@ const ButtonBar: FC<ButtonBarProps> = ({
...
+            <div className='flex space-x-4'>
+                {viewFiles && (
+                    <>
+                        <button
+                            className={buttonBarClass(notProcessed)}
+                            onClick={viewFiles}
+                            data-tooltip-id={'view-files-tooltip' + sequenceEntryStatus.accession}
+                            data-testid={`view-files-${sequenceEntryStatus.accession}`}
+                            key={'view-files-button-' + sequenceEntryStatus.accession}
+                            disabled={notProcessed}
```
theosanderson, line 157: All the checking of `notProcessed` here doesn't make sense as the button doesn't appear at all if `notProcessed`

website/src/components/ReviewPage/ReviewCard.tsx
```diff
@@ -128,19 +134,36 @@ const ButtonBar: FC<ButtonBarProps> = ({
...
     const approvable =
         sequenceEntryStatus.status === processedStatus &&
         !(sequenceEntryStatus.processingResult === errorsProcessingResult);
     const notProcessed = sequenceEntryStatus.status !== processedStatus;
 
     return (
         <div className='flex mb-auto pt-3.5 items-center'>
-            <div className='flex space-x-1'>
+            <div className='flex space-x-4'>
+                {viewFiles && (
```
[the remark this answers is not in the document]
fhennig: Ah yeah makes sense to only show for organisms that actually have files configured. For the other part: Before processing is finished, we don't know if there are files or not, but the greying out part sounds good.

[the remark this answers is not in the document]
theosanderson: Before processing is finished we already don't show the button at all - which I think is good - then I think it could appear grayed out after processing finishes if there are none (but files are configured)
image

website/src/components/ReviewPage/ReviewCard.tsx
```diff
@@ -28,6 +29,7 @@ import Locked from '~icons/fluent-emoji-high-contrast/locked';
 import Unlocked from '~icons/fluent-emoji-high-contrast/unlocked';
 import EmptyCircle from '~icons/grommet-icons/empty-circle';
 import RiDna from '~icons/mdi/dna';
+import Files from '~icons/mdi/files';
```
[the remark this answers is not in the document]
fhennig: I agree that oulines would be nicer, this is what it looks like now:

image

website/src/types/backend.ts
```diff
@@ -203,6 +203,19 @@ export const unprocessedData = accessionVersion.merge(
 );
 export type UnprocessedData = z.infer<typeof unprocessedData>;
 
+// The struct to store the files per category for one submission
+const categoryFileMapping = z.record(
```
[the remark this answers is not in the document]
theosanderson: (I think I do find its suggestion more intuitive, fwiw)

website/src/components/ReviewPage/ReviewCard.tsx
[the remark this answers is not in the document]
fhennig: Done, I've updated the video as well

website/src/components/ReviewPage/ReviewCard.tsx
fhennig: removed!

website/src/components/ReviewPage/FilesDialog.tsx
fhennig: Created, #4315

integration-tests/tests/specs/features/file-sharing.spec.ts
```diff
@@ -31,6 +31,14 @@ test('submit a single sequence with two files', async ({ pageWithGroup, page })
     const reviewPage = await submissionPage.submitSequence();
     await reviewPage.waitForZeroProcessing();
 
+    // check that files can be seen after processing
+    const filesDialog = await reviewPage.viewFiles();
+    expect(filesDialog.getByText('hello.txt')).toBeVisible();
```
theosanderson, line 36: I'm not saying this will fix your issue - but we normally `await` `expects` I think

PR #4312, opened 2025-06-04
fix(website): prevent text selection from triggering row click

### LINKED ISSUE #4311: Selecting text on search table to copy shouldn't trigger click
It's useful to be able to copy directly from the metadata table, but atm highlighting stuff triggers a click which makes this difficult

theosanderson:
resolves #4311
- avoid opening the sequence  when the user is just trying to select text in the search results table

------
https://chatgpt.com/codex/tasks/task_e_684024fd9cfc8325ac24d80868b61e15

🚀 Preview: https://codex-prevent-click-on-te.loculus.org


anna-parker (approved):
Just tested this on a preview and it is working well!

PR #4320, opened 2025-06-04
fix(website): avoid unnecessary url params

### LINKED ISSUE #4319: Hitting "select none" on column selection means that the field-visible specification becomes too long to fit in URL
On Pathoplexus if you hit "Select none" in the table column selector it stops encoding the tables to display in the URL. This will be because it tries to have a URL which specifies all fields as invisible. Instead we only need to specify fields in the URL if they are distinct from the default.

theosanderson:
Resolves #4319

- stop storing default column visibility states in the URL

I manually tested that I can hit "Select none" and get no columns displayed with the state stored appropriately in the URL - and that "Select all" also still works, and various things in between

------
https://chatgpt.com/codex/tasks/task_e_684053a648788325a903def53889b7b7

### DIFF
--- a/website/src/components/SearchPage/SearchFullUI.tsx
+++ b/website/src/components/SearchPage/SearchFullUI.tsx
@@ -233,10 +233,17 @@ export const InnerSearchFullUI = ({
     };
 
     const setAColumnVisibility = (fieldName: string, visible: boolean) => {
-        setState((prev: any) => ({
-            ...prev,
-            [`${COLUMN_VISIBILITY_PREFIX}${fieldName}`]: visible ? 'true' : 'false',
-        }));
+        setState((prev: any) => {
+            const newState = { ...prev };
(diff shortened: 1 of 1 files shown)


emmahodcroft:
Hahaha I'm sorry I don't think this was very urgent at all and didn't mean to be complaining! it just struck me when I was copying URLs the other day, that even after I cleared time filters the visibility was still in the URL 🙃

theosanderson:
> Hahaha I'm sorry I don't think this was very urgent at all and didn't mean to be complaining! it just struck me when I was copying URLs the other day, that even after I cleared time filters the visibility was still in the URL 

Oh I'm not sure I even noticed you say it! - the reason it's bad (in a semi urgent way) is that if you want to choose a set of columns to create a view to send to someone and you hit "Select none" and then select a couple of fields, then in the current version the URL doesn't capture that and so you can't send your view to someone else (because it's too many settings to fit in a URL).

fhennig (approved):
code looks sensible

PR #4322, opened 2025-06-04
fix(backend): prevent open sequence being set to Open

### LINKED ISSUE #4272: Prevent duplicate "Open" entry in data terms history
In PPX prod I found a duplicate Open entry in data terms history. Would be good to find out why (maybe it's a bug, maybe it's working as intended). If working as designed, we should probably improve how we annotate the history so it's not confusing.

Image

theosanderson:
resolves #4272

This (codex) revealed various test cases which didn't make a lot of sense!

When you first do it you get this:

image

(that was already the case)

now when you try again you get this (minus the typo which I will have fixed by the time you read this)

image


fhennig (approved):
Code looks good to me; I see you had a preview going and I'm assuming you had a go at it manually to see that it works (there are screenshots) - so I trust that you checked that it works. Screencast would be cool next time but I'm not fussed about it.

Thanks!

corneliusroemer (commented):
Nice, thanks!

Issue #4293, opened 2025-06-03
Author formatting on sequence details page


theosanderson:
We expect author information to be submitted in the format `LastName, FirstInitialOrName; LastName, FirstInitialOrName`. This makes sense as it is the only way to effectively disambiguate surnames and first / middle names. But in terms of displaying author data this is a very uncommon and somewhat ugly format.

More common formats are:

1. Similar but with all commas, when using just initials:

<img width="765" alt="Image" src="https://github.com/user-attachments/assets/edf2748c-b09c-48ea-bb64-8d6c97bdc99f" />

2. Using `FirstnameOrInitial Lastname, FirstNameOrInitial Lastname`

<img width="576" alt="Image" src="https://github.com/user-attachments/assets/f0420819-19f5-42e1-a3fc-1b934788f5f2" />

IMO we should use one of these to display authors on the sequence details page to make this more aesthetically pleasing.

chaoran-chen:
For Pathoplexus, I also prefer option 2.

For Loculus, we should leave some flexibility. For example, in Chinese, one typically writes `<last name><first name>` (without space).

PR #4331, opened 2025-06-04
fix(website): display authors as "FirstName LastName" in topmatter

theosanderson:
Relates to https://github.com/loculus-project/loculus/issues/4293

image

- display authors as `FirstName Lastname` in the topmatter, to avoid ugly semicolons
- uses "&" for last author
- other places remain as before, but that inconsistency makes sense: in a tabular form you do probably want to start with the surname
- @chaoran-chen suggested we at some point make the format customisable for Loculus. I have nothing against anyone doing that whenever they want, but I don't think it blocks us implementing something that works well for our current deployments.

I think we should consider if we think this looks better than what we have atm: I am a bit uncertain - I think it will look better with full first names, but with initials it's a bit meh.

------
https://chatgpt.com/codex/tasks/task_e_6840bdc93a9c83258b381f3c9af02094

🚀 Preview: https://codex-format-authors-as-f.loculus.org

### DIFF
--- a/website/src/components/SequenceDetailsPage/getDataTableData.ts
+++ b/website/src/components/SequenceDetailsPage/getDataTableData.ts
@@ -12,9 +12,9 @@ export type DataTableData = {
 };
 
 function formatAuthorName(author: string): string {
-    const [last, first] = author.split(',').map((x) => x.trim());
-    if (first !== undefined && last !== undefined && first.length > 0 && last.length > 0) {
-        return `${first} ${last}`;
+    const parts = author.split(',').map((x) => x.trim());
+    if (parts.length >= 2 && parts[0].length > 0 && parts[1].length > 0) {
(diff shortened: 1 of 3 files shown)


website/src/components/SequenceDetailsPage/getDataTableData.ts
```diff
@@ -71,7 +79,9 @@ export function getDataTableData(listTableDataEntries: TableDataEntry[]): DataTa
             result.topmatter.authors = entry.value
                 .toString()
                 .split(';')
-                .map((x) => x.trim());
+                .map((x) => x.trim())
+                .filter((x) => x.length > 0)
```
chaoran-chen, line 83:
```suggestion
                .filter((x) => x.trim().length > 0)
```

since we are also trimming in `formatAuthorName`

Issue #4030, opened 2025-04-22
Nextclade preprocessing: upload sequence annotation files using file share


theosanderson:
We want to use Nextclade to annotate sequences with protein coordinates (it outputs a GFF3).

Update: We now decided to have preprocessing use the annotations from nextclade to create embl flat files/gff3 and embl flatfiles and then upload them using the file sharing feature. 

@anna-parker already created code to generate the embl flat files from the nextclade results and the metadata - can be copied over from the ena deposition package here: https://github.com/loculus-project/loculus/pull/4076

rneher:
should have gone here
https://github.com/loculus-project/loculus/issues/4029#issuecomment-2825175864

anna-parker:
## Steps required for this feature:
1. upload dummy files in prepro using file share, using `group_id`. Add functions to request pre-signed URL and upload. Confirm this works on a preview
2. Get annotations for each processed entry from nextclade (see how I did this in https://github.com/loculus-project/loculus/pull/4076)
3. After processing all entries create embl flat files and upload (this should be added to the main function here: https://github.com/loculus-project/loculus/blob/main/preprocessing/nextclade/src/loculus_preprocessing/prepro.py#L968). To do this, for each entry
- check if there are errors (if yes break)
- request presigned URL
- create embl flat file from processed data by copying over code from https://github.com/loculus-project/loculus/blob/7d2e1d69f35eb41a9132c16803e07bdf58e051ad/ena-submission/src/ena_deposition/ena_submission_helper.py#L447 (potentially add a dictionary of which metadata fields are used to the values.yaml e.g. sampleCollectionDate, country, authors,...). Note: this code compresses the file, we might not want this.
- upload file
4. confirm file can be downloaded from webpage and replace `create_flatfile` request with call to S3. 
5. Add tests to prepro that embl flat file creation works well (see examples I added in https://github.com/loculus-project/loculus/pull/4076)

### Changes to the values.yaml
- `scientific_name` and `molecule_type` need to be added to the preprocessing config: https://github.com/loculus-project/loculus/blob/7d2e1d69f35eb41a9132c16803e07bdf58e051ad/kubernetes/loculus/values.yaml#L1291
- potentially add a dictionary to the prepro config with 
```
- file: name
  type: embl_flat_file
  metadata_map:
    - qualifier1: [loculus_metadata_fields...]
    - qualifier2: [loculus_metadata_fields...]
```

PR #4076, opened 2025-05-04
Automate annotations

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/3739

### Summary

Update: architecture will change but mapping described in EMBL flat file annotation will be used

Adds a new field to the processedData schema called `annotations`. This is populated by the preprocessing pipeline with the `annotation` dictionary generated by running `nextclade` in preprocessing (following suggestion in https://github.com/loculus-project/loculus/issues/4029). The preprocessing pipeline sends the processed data to the backend. 

The backend now returns these annotations in the `get-released-data` endpoint (should we change this?). The `get-ena-submission-list` cronjob processes the data from the `get-released-data` endpoint and (after approval) this is sent to the ena-submission pod where it is used to annotate embl flat files: main code here: https://github.com/loculus-project/loculus/pull/4076/files#r2079351986

### EMBL flat file annotations

Currently we only annotate CDS and gene regions following the defined [WebFeat](https://www.ebi.ac.uk/ena/WebFeat/). 

#### Ranges: 
In nextclade.json the CDS ranges are 0-indexed with exclusive ends.

EMBL files must be 1-indexed ranges with inclusive ends for the embl flat files, due to the [feature_table docs](http://www.insdc.org/files/feature_table.html): 
_Sequence spans are indicated by the starting base number and the ending base number separated by two periods (e.g., '34..456')_ - note that Bio does this conversion for us.

Additionally, the [\codon_start](https://www.ebi.ac.uk/ena/WebFeat/qualifiers/codon_start.html) qualifier `has valid value of 1 or 2 or 3, indicating the offset at which the first complete codon of a coding
feature can be found, relative to the first base of that feature;`

This is inline with all sequences I compared in west-nile and cchf. It also looks like generally ENA embl files include the stop codon in the CDS range but not in the translated sequence.

#### Translations

For CDS only I create translation from the unaligned sequence using ranges in annotation, note that these will **INCLUDE** CDS-insertions although these are excluded on the website translations (as these must align for SILO).

### Testing

As this PR touches many pods I wanted to simplify testing of the embl flat file output.
To do this I added a bash script which when given a loculus `ACCESSION` (of this preview) and `ORGANISM` will create an embl flat file using our approach and also download the corresponding embl flat file from the ENA (genbank only has genbank files).

To run this test simply checkout this branch and run:
```
cd ena-submission
micromamba activate loculus-ena-submission
bash scripts/compare_embl.bash ACCESSION ORGANISM
```
[...]

PR #4344, opened 2025-06-05
feat(preprocessing): Make nextclade pipeline submit annotations as embl files

fhennig:
resolves https://github.com/loculus-project/loculus/issues/4030

builds on work done in #4076 

- backend: add 'files' section to the config, where output files are configured
- backend: fix a bug where input and output files where mixed up during validation
- prepro: bump nextclade from 3.9 to 3.13
- prepro: `process_all` returns `SubmissionData` (new type) instead of `ProcessedEntry` which groups a `ProcessedEntry` together with annotations and group ID, which are needed to create the embl file later on.

### Tests

- backend: only minor changes, which are covered by altering existing tests
- ena deposition: the 'create flatfile' part was moved out and now there is only a 'download flatfile' which isn't really worth doing any tests with
- prepro: 
  - backend.py has new upload/download functions, but since it's only IO I don't think it's worth testing - we have it covered in the integration tests
  - embl.py - test cases added with test files for regression. One example west nile, one example cchf
- integration test: A new test that checks on a submission of the annotations file is there.

### Manual testing
I made my own submissions, but also had a look at automatically ingested sequences. Whenever I looked at the sequence details, the file section was there, the file was there and it could be opened. I also confirmed the file contents based on references from annas PR, to make sure that the files match.

### Screenshot

image

### PR Checklist

🚀 Preview: https://pipeline-file-submission.loculus.org


fhennig:
Ah, of course the test is failing because it cannot call the endpoints .. hmm..

preprocessing/nextclade/tests/factory_methods.py
```diff
@@ -36,6 +36,7 @@ def create_unprocessed_entry(
             accessionVersion=f"LOC_{accession_id}.1",
             data=UnprocessedData(
                 submitter="test_submitter",
+                group_id=1,
```
anna-parker, line 39: it might be good to use group_id=2 as 1 will always be the insdc_ingest group

fhennig: done!

preprocessing/nextclade/src/loculus_preprocessing/embl.py
```diff
@@ -0,0 +1,226 @@
...
+
+    for seq_name, sequence_str in unaligned_nucleotide_sequences.items():
+        if not sequence_str:
+            continue
+        reference = Reference()
+        reference.authors = authors
+        sequence = SeqRecord(
+            Seq(sequence_str),
+            id=f"{accession}_{seq_name}" if multi_segment else accession,
+            annotations={},  # TODO
```
anna-parker, line 187: btw pylance throws a type error for the annotations dict but you can ignore it

fhennig: ah good to know, thanks!

fhennig:
Alright! So current state is:
- the embl file creation itself is working from what I can tell, and it's enabled now for all organisms
- the ena deposition is switched over to downloading, but I didn't test that yet
- no automated testing was implemented yet at all

ena-submission/src/ena_deposition/ena_submission_helper.py
```diff
@@ -305,72 +306,21 @@ def post_webin(config: ENAConfig, xml: dict[str, Any]) -> requests.Response:
     )
 
 
-def create_chromosome_list(list_object: AssemblyChromosomeListFile, dir: str | None = None) -> str:
```
anna-parker, line 308: this function is still required ;-)

ena-submission/src/ena_deposition/ena_submission_helper.py
```diff
@@ -379,57 +329,9 @@ def create_flatfile(
...
-
-        with tempfile.NamedTemporaryFile(delete=False, suffix=".embl") as temp_seq_file:
-            SeqIO.write(sequence, temp_seq_file.name, "embl")
-
-        with open(temp_seq_file.name, encoding="utf-8") as temp_seq_file:
-            embl_content.append(temp_seq_file.read())
-
-    final_content = "\n".join(embl_content)
-
     gzip_filename = filename + ".gz"
```
corneliusroemer, line 332: Doesn't this make it `sequences.embl.gz.gz`?

preprocessing/nextclade/src/loculus_preprocessing/backend.py
```diff
@@ -173,6 +180,44 @@ def submit_processed_sequences(
     logger.info("Processed data submitted successfully")
 
 
+def request_upload(group_id: int, number_of_files: int, config: Config) -> Sequence[FileUploadInfo]:
+    # we need to parse this here, because we don't want the organism in there.
```
corneliusroemer, line 184: What is `this`, `here`, and `there`?

corneliusroemer: `backend_host` seems to be a misnomer - if it has organism after the proper hostname, it should be name `backend_url` or something like that not `backend_host` so we might want to fix this now

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -819,6 +832,15 @@ def process_single(  # noqa: C901
...
         )
+    annotations = {}
+    if unprocessed.nextcladeMetadata is not None:
+        for segment in config.nucleotideSequences:
+            if segment in unprocessed.nextcladeMetadata:
+                annotations[segment] = None
+                if unprocessed.nextcladeMetadata[segment]:
+                    annotations[segment] = unprocessed.nextcladeMetadata[segment].get(
+                        "annotation", None
+                    )
```
corneliusroemer, line 843:
```suggestion
    annotations = {}

    if unprocessed.nextcladeMetadata is not None:
        for segment in config.nucleotideSequences:
            segment_metadata = unprocessed.nextcladeMetadata.get(segment)
            if segment_metadata is not None:
                annotations[segment] = segment_metadata.get("annotation")
```

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -965,6 +988,29 @@ def run(config: Config) -> None:
...
+                    processed_entry.data.annotations,
+                )
+                processed_entry.data.annotations = None  # remove it so it's not submitted
+                upload_info = request_upload(group_id, 1, config)[0]
+                file_id = upload_info.fileId
+                url = upload_info.url
+                upload_string_to_presigned_url_zipped(file_content, url)
+                processed_entry.data.files = {
+                    "annotations": [
+                        FileIdAndName(fileId=file_id, name="sequences.embl.gz")
```
corneliusroemer, line 1010: Do we want to upload as .gz? Is that because webin expects that? It might be nicer to have it uncompressed maybe for people to just view directly after download.

corneliusroemer (commented):
Nice that it works! We might want to change the name to be `ACCESSION.embl` instead of always `sequences.embl`

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -727,11 +736,15 @@ def process_single(  # noqa: C901
...
 
         submitter = unprocessed.inputMetadata["submitter"]
+        group_id = int(str(unprocessed.inputMetadata["groupId"]))
         unaligned_nucleotide_sequences = unprocessed.unalignedNucleotideSequences
     else:
         submitter = unprocessed.submitter
+        group_id = unprocessed.group_id
         unaligned_nucleotide_sequences = unprocessed.unalignedNucleotideSequences
 
+    output_metadata["groupId"] = group_id
```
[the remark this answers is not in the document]
fhennig: Yeah it isn't necessary to have it in the metadata, I also wrote in the PR description:

> the way the group_id and annotations are pulled through is a bit ugly - maybe there's a better way to structure the code.

I did this to get things to a first working Draft and still want to iterate from there - I'm not working on this code a lot so I wasn't sure how to best structure it yet.

preprocessing/nextclade/src/loculus_preprocessing/backend.py
fhennig: I've changed the comment to 

> we need to parse the backend URL, to extract the API path without the organism component

I agree that `backend_host` isn't the right name, I'd call it `backend_base_url` but it'd be a breaking change because it'd require config changes in the `values.yaml` for the nextclade pipeline, so maybe it's not worth it? I've also added as a comment above the field def: "The backend host is the base API URL including the organism"

ena-submission/src/ena_deposition/ena_submission_helper.py
```diff
@@ -305,72 +306,21 @@ def post_webin(config: ENAConfig, xml: dict[str, Any]) -> requests.Response:
...
-    authors = get_authors(metadata.get("authors", ""))
-    moleculetype = get_molecule_type(organism_metadata)
+    version = metadata["version"]
+    category = "annotations"
+    filename = "sequences.embl.gz"
+
+    base_url = config.backend_url.rstrip("/")
+    url = f"{base_url}/files/get/{accession}/{version}/{category}/{filename}"
+
+    response = requests.get(url, timeout=10)
```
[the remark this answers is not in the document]
fhennig: I've changed it to 60 seconds

ena-submission/src/ena_deposition/ena_submission_helper.py
fhennig: The full definition of the filename is:

```python
    if dir:
        os.makedirs(dir, exist_ok=True)
        filename = os.path.join(dir, "sequences.embl")
    else:
        with tempfile.NamedTemporaryFile(delete=False, suffix=".embl") as temp:
            filename = temp.name

    gzip_filename = filename + ".gz"

    Path(gzip_filename).write_bytes(response.content)

    return gzip_filename
```

so it's just one `.gz` - maybe you just didn't see it so easily in the GH change display.

ena-submission/src/ena_deposition/ena_submission_helper.py
fhennig: oops haha, added it back!

preprocessing/nextclade/src/loculus_preprocessing/backend.py
fhennig: (marking this as resolved so it's easier to see what's going on, but feel free to un-resolve it again)

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
fhennig: The backend doesn't do anything to the files that get uploaded, it's also uploaded with a presigned URL directly to S3, so no additional processing there. Then I'll go ahead and not compress it, and instead compress it before it gets submitted!

anna-parker: Generally I'm happy with that approach - I just wonder will it increase the file upload time? We should probably check if it starts taking much longer (but we should see this on the preview without custom measurements if it is really a problem)

preprocessing/nextclade/src/loculus_preprocessing/backend.py
```diff
@@ -173,6 +180,44 @@ def submit_processed_sequences(
...
+        raise RuntimeError(msg)
+
+
+def upload_string_to_presigned_url(content: str, url: str) -> None:
+    _upload_bytes(content.encode("utf-8"), url)
+
+
+def upload_string_to_presigned_url_zipped(content: str, url: str) -> None:
+    _upload_bytes(_gzip_string(content), url)
+
```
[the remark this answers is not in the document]
fhennig: much simplified now - resolving.

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
fhennig: I've changed this, I'm marking this as resolved, haven't noticed a slower upload time yet

preprocessing/nextclade/src/loculus_preprocessing/embl.py
```diff
@@ -0,0 +1,210 @@
...
+    version,
+    metadata,
+    unaligned_nucleotide_sequences,
+    annotation_object: dict[str, Any] | None = None,
+) -> str:
+    collection_date = metadata.get("sampleCollectionDate", "Unknown")
+    country = get_country(metadata)
+    description = get_description(accession, version, config.db_name)
+    organism = config.scientific_name
+    authors = get_authors(metadata.get("authors", ""))
```
anna-parker, line 158: These fields are all created using fields in the Loculus metadata - I think it would be good to add them to the config to make this clear (as if we ever change the name of these fields the embl flat file creation will fail and now that we are doing this in prepro we really want to prevent any errors)

fhennig: Ah yeah. So you mean add a config like "sample_collection_data_field_name = sampleCollectionDate" ?

fhennig: I implemented that here: https://github.com/loculus-project/loculus/pull/4344/commits/320aa86bb1ca33dab4bfbf0c4c32c31c88e21471

anna-parker: Thank you!

anna-parker: maybe we could make it a bit clearer that we use these fields for the embl flat file, I wonder if we could make a subsection embl_flat_file_fields and then give this dictioanry of values

fhennig: Ah yes that sounds like a good idea

ena-submission/src/ena_deposition/ena_submission_helper.py
```diff
@@ -354,24 +352,29 @@ def get_authors(authors: str) -> str:
...
-    collection_date = metadata.get("sampleCollectionDate", "Unknown")
-    country = get_country(metadata)
-    organism = organism_metadata.get("scientific_name", "Unknown")
     accession = metadata["accession"]
-    description = get_description(config, metadata)
-    authors = get_authors(metadata.get("authors", ""))
-    moleculetype = get_molecule_type(organism_metadata)
+    version = metadata["version"]
+    category = "annotations"
+    filename = f"{accession}.embl"
```
anna-parker, line 371: I think it would be good to rename the file as 
```suggestion
    filename = f"{accession}.{version}.embl"
```

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -965,8 +1000,47 @@ def run(config: Config) -> None:
...
+                    try:
+                        file_content = create_flatfile(
+                            config,
+                            submission_data.processed_entry.accession,
+                            submission_data.processed_entry.version,
+                            submission_data.processed_entry.data.metadata,
+                            submission_data.processed_entry.data.unalignedNucleotideSequences,
+                            submission_data.annotations,
+                        )
+                        file_name = f"{submission_data.processed_entry.accession}.embl"
```
anna-parker, line 1015: here as well
```suggestion
                        file_name = f"{submission_data.processed_entry.accession}.{submission_data.processed_entry.version}.embl"
```

preprocessing/nextclade/src/loculus_preprocessing/embl.py
```diff
@@ -0,0 +1,261 @@
...
+
+
+def reformat_authors_from_loculus_to_embl_style(authors: str) -> str:
+    authors_list = [author for author in authors.split(";") if author]
+    ena_authors = []
+    for author in authors_list:
+        last_names, first_names = author.split(",")[0].strip(), author.split(",")[1].strip()
+        initials = "".join([name[0] + "." for name in first_names.split() if name])
+        ena_authors.append(f"{last_names} {initials}".strip())
+    return ", ".join(ena_authors) + ";"
```
anna-parker, line 87: The ";" is now duplicated in the embl-flat files because Bio is adding a ";" as well - as the authors are only used for the embl flat file here we should just remove the ";"
```suggestion
    return ", ".join(ena_authors)
```

preprocessing/nextclade/tests/test_processing_functions.py
```diff
@@ -661,5 +667,53 @@ def test_parse_date_into_range() -> None:
...
+        if len(records) == 1:
+            unaligned_nucleotide_sequences = {"main": str(records[0].seq)}
+        else:
+            unaligned_nucleotide_sequences = {}
+            for record in records:
+                # Use the part after the last underscore as the key
+                key = record.id.split("_")[-1]
+                unaligned_nucleotide_sequences[key] = str(record.seq)
+
+    # Load annotation object
```
anna-parker, line 698: These comments aren't really needed
```suggestion
```

preprocessing/nextclade/tests/test_processing_functions.py
```diff
@@ -661,5 +667,53 @@ def test_parse_date_into_range() -> None:
...
+
+    # Read metadata from metadata.tsv
+    metadata_tsv_path = os.path.join(test_data_dir, "metadata.tsv")
+    metadata = {}
+    with open(metadata_tsv_path, encoding="utf-8") as f:
+        header = f.readline().strip().split("\t")
+        values = f.readline().strip().split("\t")
+        metadata = dict(zip(header, values, strict=False))
+
+    # Load unaligned nucleotide sequences
```
anna-parker, line 686:
```suggestion
```

integration-tests/tests/specs/features/annotations.dependent.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+        const fileUrl = await page
+            .getByRole('link', { name: /LOC_\w{6,9}\.embl/ })
+            .getAttribute('href');
+
+        await Promise.all([
+            page.waitForResponse(
+                async (resp) =>
+                    resp.status() === 200 && (await resp.text()).startsWith('ID   LOC_'),
+            ),
+            page.evaluate((url) => fetch(url), fileUrl),
```
anna-parker, line 33: it would be good to have a checkFileContent here - this way if nextclade ever updates their annotations.json in a way that affects our pipeline we will see it in a test

fhennig: ok

backend/src/main/kotlin/org/loculus/backend/config/Config.kt
corneliusroemer, line 1: All the backend changes should be turned into a separate PR as they are not trivial and not directly related to the ena-submission/prepro changes.

This PR can be (re)based on that new backend PR branch.

backend/src/main/kotlin/org/loculus/backend/service/submission/FileMappingPreconditionValidator.kt
```diff
@@ -75,8 +64,29 @@ class FileMappingPreconditionValidator(
...
-                        "Allowed categories are: ${allowedCategoriesName.joinToString(", ")}.",
+                    "The category $category is not part of the configured submission categories for ${organism.name}." +
+                        " Allowed categories are: ${allowedFileCategories.joinToString(", ")}.",
+                )
+            }
+        }
+        return this
+    }
+
+    fun validateCategoriesMatchOutputSchema(
```
corneliusroemer, line 75: Move this back up to the original place of the `fun validateCategoriesMatchProcessedSchema(` to make the diff smaller and the PR easier to review

backend/src/main/kotlin/org/loculus/backend/service/submission/FileMappingPreconditionValidator.kt
```diff
@@ -44,17 +44,6 @@ class FileMappingPreconditionValidator(
         return validateCategoriesMatchSchema(fileCategoriesFilesMap, allowedFileCategories, organism)
     }
 
-    fun validateCategoriesMatchProcessedSchema(
```
anna-parker, line 47: @fhennig and @chaoran-chen I'm getting a bit confused with these changes - it seems like this is a rename of the functions added in https://github.com/loculus-project/loculus/pull/4385 where `validateCategoriesMatchProcessedSchema` has been renamed to `validateCategoriesMatchOutputSchema` and the error message has been changed. Otherwise the functions do the same thing - can we condense the functions instead?

ena-submission/src/ena_deposition/ena_submission_helper.py
```diff
@@ -380,57 +383,9 @@ def create_flatfile(
...
-            SeqIO.write(sequence, temp_seq_file.name, "embl")
-
-        with open(temp_seq_file.name, encoding="utf-8") as temp_seq_file:
-            embl_content.append(temp_seq_file.read())
-
-    final_content = "\n".join(embl_content)
-
     gzip_filename = filename + ".gz"
 
-    with gzip.open(gzip_filename, "wt", encoding="utf-8") as file:
```
anna-parker, line 432: the simple gzip function I had before is all you need (we anyways need to write the files to disk for the webin-cli to upload to ena)

ena-submission/src/ena_deposition/ena_submission_helper.py
```diff
@@ -380,57 +383,9 @@ def create_flatfile(
...
-        file.write(final_content)
+    Path(gzip_filename).write_bytes(_gzip_string(response.content))
```
corneliusroemer, line 388:
```suggestion
    Path(gzip_filename).write_bytes(
        gzip.compress(response.text.encode("utf-8"))
    )
```

preprocessing/nextclade/src/loculus_preprocessing/backend.py
```diff
@@ -173,6 +178,28 @@ def submit_processed_sequences(
...
+    headers = {"Authorization": "Bearer " + get_jwt(config)}
+    response = requests.post(url, headers=headers, params=params, timeout=10)
+    if not response.ok:
+        msg = f"Upload request failed: {response.status_code}, {response.text}"
+        raise RuntimeError(msg)
+    return [FileUploadInfo(**item) for item in response.json()]
+
+
+def upload_embl_file_to_presigned_url(content: str, url: str) -> None:
+    headers = {"Content-Type": "chemical/x-embl"}
```
anna-parker, line 196: I think the correct type is maybe "chemical/x-embl-dl-nucleotide" based on https://en.wikipedia.org/wiki/Chemical_file_format

```suggestion
    headers = {"Content-Type": "chemical/x-embl-dl-nucleotide"}
```

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -26,6 +32,8 @@
     AminoAcidSequence,
     AnnotationSource,
     AnnotationSourceType,
+    FileIdAndName,
+    SubmissionData,
```
anna-parker, line 36: this import list is unsorted

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -819,8 +834,17 @@ def process_single(  # noqa: C901
                 message=("No segment aligned."),
             )
         )
+    annotations: dict[str, Any] = {}
```
anna-parker, line 837: I guess if we are not returning embl flat files (i.e. if embl_file is False in the config) we could also just set the annotations to None to not have to carry them around

corneliusroemer (changes requested):
Integration tests appear to be flaky, especially a newly added one

https://github.com/loculus-project/loculus/actions/runs/15763001950/job/44433724960#step:18:144

preprocessing/nextclade/src/loculus_preprocessing/backend.py
```diff
@@ -27,7 +29,7 @@
 class JwtCache:
     def __init__(self) -> None:
         self.token: str = ""
-        self.expiration: dt.datetime = dt.datetime.min
+        self.expiration: dt.datetime = dt.datetime.min  # noqa: DTZ901
```
corneliusroemer, line 32:
```suggestion
        self.expiration: dt.datetime = dt.datetime.min.replace(tzinfo=datetime.UTC)
```

preprocessing/nextclade/src/loculus_preprocessing/config.py
```diff
@@ -42,6 +57,8 @@ class Config:
     pipeline_version: int = 1
     multi_segment: bool = False
     alignment_requirement: str = "ALL"
+    embl_file: bool = False
+    embl_info: EmblInfo = dataclasses.field(default_factory=EmblInfo)
```
corneliusroemer, line 61: Unclear variable name

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -965,8 +1000,47 @@ def run(config: Config) -> None:
...
+                        file_id = upload_info.fileId
+                        url = upload_info.url
+                        upload_embl_file_to_presigned_url(file_content, url)
+                        submission_data.processed_entry.data.files = {
+                            "annotations": [
+                                FileIdAndName(fileId=file_id, name=file_name)
+                            ]
+                        }
+                    except Exception as e:
+                        submission_data.processed_entry.errors.append(
```
anna-parker, line 1026: wupps these errors will never be sent to the user as that only happens in `submit_processed_sequences` - we will see that in the logs but I think that is not enough. We need to somehow be notified if the files cannot be uploaded.

I think we should retry on a failed upload but if that still doesn't work we need to think about how we handle this case

@fhennig or @corneliusroemer do you have ideas on how to ensure we always know when the file upload fails?

My suggestions are:
1. raise error in prepro (have a failing pod that we should get notified about -> but also bad for other users
2. send a slack notification
3. return the entry in `submit_processed_sequences` with this error so that users will contact us

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -727,9 +740,11 @@ def process_single(  # noqa: C901
             )
 
         submitter = unprocessed.inputMetadata["submitter"]
+        group_id = int(str(unprocessed.inputMetadata["group_id"]))
```
corneliusroemer, line 743:
```suggestion
        group_id = int(unprocessed.inputMetadata["group_id"])
```

Is there a reason for the str first? I can't think of one.

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -835,41 +859,52 @@ def process_single(  # noqa: C901
...
-                    AnnotationSource(name="unknown", type=AnnotationSourceType.METADATA)
-                ],
-                message=(
-                    f"Failed to process submission with id: {id} - please review your submission "
-                    "or reach out to an administrator if this error persists."
-                ),
-            )
-        ],
-        warnings=[],
+        group_id=-1,
```
corneliusroemer, line 898:
```suggestion
        group_id=None,
```

preprocessing/nextclade/tests/test_processing_functions.py
```diff
@@ -661,5 +667,49 @@ def test_parse_date_into_range() -> None:
...
+        metadata = dict(zip(header, values, strict=False))
+
+    with open(os.path.join(test_data_dir, "sequence.fa"), encoding="utf-8") as f:
+        records = list(SeqIO.parse(f, "fasta"))
+        if len(records) == 1:
+            unaligned_nucleotide_sequences = {"main": str(records[0].seq)}
+        else:
+            unaligned_nucleotide_sequences = {}
+            for record in records:
+                # Use the part after the last underscore as the key
```
corneliusroemer, line 692:
```suggestion
```

Let's remove prompt comments

preprocessing/nextclade/src/loculus_preprocessing/embl.py
```diff
@@ -0,0 +1,261 @@
...
+        sequence.features.append(source_feature)
+        if annotation_object and annotation_object.get(seq_name, None):
+            seq_feature_list = get_seq_features(annotation_object[seq_name], sequence_str, config)
+            for feature in seq_feature_list:
+                sequence.features.append(feature)
+
+        buffer = io.StringIO()
+        SeqIO.write(sequence, buffer, "embl")
+        buffer.seek(0)
+        embl_content.append(buffer.read())
```
corneliusroemer, line 259:
```suggestion
        embl_content.append(sequence.format("embl"))
```

preprocessing/nextclade/tests/test_processing_functions.py
```diff
@@ -661,5 +667,49 @@ def test_parse_date_into_range() -> None:
...
+)
+def test_create_flatfile(config: Config, case_dir):
+    test_data_dir = os.path.join(os.path.dirname(__file__), "test_data", case_dir)
+
+    metadata_tsv_path = os.path.join(test_data_dir, "metadata.tsv")
+    metadata = {}
+    with open(metadata_tsv_path, encoding="utf-8") as f:
+        header = f.readline().strip().split("\t")
+        values = f.readline().strip().split("\t")
+        metadata = dict(zip(header, values, strict=False))
```
anna-parker, line 683: we can use pandas for this :-) 
```suggestion
metadata = (
    pd.read_csv(Path(test_data_dir) / "metadata.tsv", sep="\t", nrows=1)
      .iloc[0]
      .to_dict()
)
```

preprocessing/nextclade/tests/test_processing_functions.py
```diff
@@ -661,5 +667,49 @@ def test_parse_date_into_range() -> None:
     ), "dateRangeLower: empty date should be returned as None."
 
 
+@pytest.mark.parametrize(
+    "case_dir",
+    ["cchf_1", "west_nile_1"],
+    ids=["cchf_1", "west_nile_1"],
+)
+def test_create_flatfile(config: Config, case_dir):
```
anna-parker, line 675: maybe add a comment that this test requires that there is only one sample (single or multi-segmented) in the files

anna-parker:
This looks great! I added a lot of comments but most are just for code formatting/structure and not actual content issues. The tests are also sadly failing after the `;;` change. 

I still want to test 2 additional things if that is ok:
1. Actually upload to ENA dev - at the moment the ena-deposition pod points to the test branch - we should point it to another branch and add data of files we have uploaded to this preview there so we can watch the submission process on ENA dev with the new files
2. Check if we still need to make the changes suggested here: https://github.com/loculus-project/loculus/pull/4076#issuecomment-2907878600

anna-parker (changes requested):

backend/src/main/kotlin/org/loculus/backend/config/Config.kt
chaoran-chen: I agree, the backend changes should be moved to one or maybe even two separate PRs: if the changes are closely related, they can be one PR, but if one part is a bug fix and one part rather a new feature or an independent refactoring or a very different bug, two PRs would be better.

backend/src/main/kotlin/org/loculus/backend/service/submission/FileMappingPreconditionValidator.kt
chaoran-chen: (I'll review the changes once they are in their own PRs: https://github.com/loculus-project/loculus/pull/4344/files#r2157417223)

preprocessing/nextclade/tests/test_processing_functions.py
fhennig: Ah, neat! much better, thanks.

preprocessing/nextclade/tests/test_processing_functions.py
fhennig: Ah, makes sense; now that you mention it I can see how there could be multiple different ones in the same file.

preprocessing/nextclade/src/loculus_preprocessing/embl.py
```diff
@@ -0,0 +1,261 @@
...
+        GENOMIC_RNA = "GENOMIC_RNA"
+        VIRAL_CRNA = "VIRAL_CRNA"
+
+        def __init__(self, value): self.value = value
+        def __str__(self): return self.value
+    try:
+        return MoleculeType(molecule_type)
+    except Exception as err:
+        msg = f"Invalid molecule type: {molecule_type}"
+        raise ValueError(msg) from err
```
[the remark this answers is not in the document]
fhennig: I see that the enaDeposition config also has a similar type and setting:

```
class MoleculeType(Enum):
    GENOMIC_DNA = "genomic DNA"
    GENOMIC_RNA = "genomic RNA"
    VIRAL_CRNA = "viral cRNA"

    def __str__(self):
        return self.value
```

I think there might be potential for confusion in the values.yaml:

image

I'm not sure what the best way to resolve this is; or is it fine to leave it like that?

kubernetes/loculus/values.yaml
```diff
@@ -1199,9 +1201,12 @@ defaultOrganismConfig: &defaultOrganismConfig
         - "prepro"
       replicas: 2
       configFile: &preprocessingConfigFile
+        scientific_name: "Sudan ebolavirus"
```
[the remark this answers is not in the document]
fhennig: Yes this setting appears in the prepro config section as well as the ena config section.

IMO if we want to keep the Helm Chart pretty decoupled from the specific prepro pipeline as well as the ena submission, I'd leave it the way it is; otherwise we'd have to have these sections influence each other, or have a third config section elsewhere that's only affecting our specific prepro implementation, right?

preprocessing/nextclade/src/loculus_preprocessing/embl.py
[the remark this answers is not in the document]
anna-parker: Actually I'm now even more confused - why did you modify my previous function? This is how it was previously defined.

```
from enum import Enum


class MoleculeType(Enum):
    GENOMIC_DNA = "genomic DNA"
    GENOMIC_RNA = "genomic RNA"
    VIRAL_CRNA = "viral cRNA"

    def __str__(self):
        return self.value

def get_molecule_type(organism_metadata: dict[str, str]) -> MoleculeType:
    try:
        moleculetype = MoleculeType(organism_metadata.get("molecule_type"))
    except ValueError as err:
        msg = f"Invalid molecule type: {organism_metadata.get('molecule_type')}"
        raise ValueError(msg) from err
    return moleculetype
```

Your current function lets you add invalid types without a warning `def __init__(self, value): self.value = value`
image

Lets just keep it like my function in ENA deposition this works as intended and as you mention doesn't require any changes in what we give to the values.yaml

image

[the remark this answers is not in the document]
anna-parker: No worries - lets add a test maybe for these two examples to ensure it won't break again in future :-)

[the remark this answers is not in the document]
anna-parker: This should be good: https://github.com/loculus-project/loculus/pull/4344/commits/2cd641a63fb0b18b772758b29701ac8f4beb58ec

integration-tests/tests/specs/features/annotations.dependent.spec.ts
fhennig: implemented here, please comment there for further feedback: https://github.com/loculus-project/loculus/pull/4490/files#diff-e75658feb714e49aaf56b62014ca9e401e98b3219185b168a4f94e2da71797f7

backend/src/main/kotlin/org/loculus/backend/config/Config.kt
```diff
@@ -51,7 +53,7 @@ data class Schema(
     val externalMetadata: List<ExternalMetadata> = emptyList(),
     val earliestReleaseDate: EarliestReleaseDate = EarliestReleaseDate(false, emptyList()),
     val submissionDataTypes: SubmissionDataTypes = SubmissionDataTypes(),
-    val files: List<FileCategory> = emptyList(),
+    val files: List<OutputFileCategory> = emptyList(),
```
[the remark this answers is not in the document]
fhennig: I've undone this change in the new PR: https://github.com/loculus-project/loculus/pull/4490

PR #4354, opened 2025-06-05
ci(integration-tests): run different browser tests in different containers

theosanderson:
This drops the "run integration tests" step of the integration tests from about 4 mins 40 to 2 mins 30.
(And when tests fail it should drop from potentially 8 to 4)

Seems worth it? 

------
https://chatgpt.com/codex/tasks/task_e_6841a367f36c8325ac445bab0c072522

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -22,6 +22,12 @@ jobs:
       checks: read
       actions: read # Required by workflow-telemetry-action
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        browser:
+          - chromium
+          - firefox
     timeout-minutes: 45
     env:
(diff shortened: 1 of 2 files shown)


corneliusroemer (commented):
Are the tests able to run in parallel? When we did E2E the worry was that if tests run in parallel they'd interfere. If that's no longer an issue great - it had always bothered me that we couldn't run things in parallel.

theosanderson:
(A) I don't think the question actually relates to this PR very directly - this actually makes them less parallel in one sense (i.e. previously chrome ran next to firefox in the same github job, now it runs as a separate job)
(B) Yes, the integration tests are designed around parallel tests

corneliusroemer:
Nice, sorry for not grokking Integration tests yet, sounds great to speed
up like this.

chaoran-chen (approved):
Sounds good, thanks!

PR #4357, opened 2025-06-05
chore(tests): various tweaks to integration tests

corneliusroemer:
Grab bag of tweaks to integration tests. Notable ones:
- We're using quite old browsers due to old playwright, need dependabot to keep us up to date
- Dependabot validate is no longer necessary as Github does that on its own it seems
- Make `./start-server.sh` executable (it had shebang but wasn't)
- Make start server fail fast (it kept executing even though an early step failed for me)
- Fix race condition in file sharing integration test helper function (due to lack of `await`)


.github/dependabot.yml
```diff
@@ -71,3 +71,12 @@ updates:
...
           - "patch"
+  - package-ecosystem: npm
+    directory: integration-tests/
+    schedule:
+      interval: weekly
+    groups:
+      minorAndPatch:
+        update-types:
+          - "minor"
+          - "patch"
```
corneliusroemer, line 82:
```suggestion
          - "patch"

```

integration-tests/start-server.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
- ../deploy.py cluster --delete
- ../deploy.py cluster
- ../deploy.py helm --for-e2e --enablePreprocessing
- python ../.github/scripts/wait_for_pods_to_be_ready.py
\ No newline at end of file
+../deploy.py cluster --delete \
+&& ../deploy.py cluster \
+&& ../deploy.py helm --for-e2e --enablePreprocessing \
+&& python ../.github/scripts/wait_for_pods_to_be_ready.py
```
corneliusroemer, line 5:
```suggestion
&& python ../.github/scripts/wait_for_pods_to_be_ready.py

```

theosanderson (approved):
Thanks for sorting out flakes!

integration-tests/tests/specs/features/file-sharing.spec.ts
```diff
@@ -58,24 +58,19 @@ test('submit two sequences with one file each', async ({ pageWithGroup, page })
...
+    await page.getByRole('heading', { name: 'Extra files' }).scrollIntoViewIfNeeded();
+
+    await submissionPage.uploadExternalFiles('raw_reads', {
+        sub1: {
+            'foo.txt': 'Foo',
+        },
+        sub2: {
+            'bar.txt': 'Bar',
+        },
+    });
```
chaoran-chen, line 70: I agree with @fhennig's comment in [Slack](https://loculus.slack.com/archives/C05G172HL6L/p1749167902426519) that cleaning up is a good practice and the code also doesn't seem overly complex to me. If you think that the cleanup should be removed, let's maybe discuss it in a separate PR?

PR #4360, opened 2025-06-06
PREVIEW: Wastewater Instance - DO NOT MERGE

gordonkoehn:
Kind request to set up a preview instance to help us integrate the new pre-signed S3 URL flow into sr2silo for the Wastewater Loculus deployment.

The preview would be to address:
- https://github.com/cbg-ethz/sr2silo/issues/214

I configured the S3 as described. I am a rookie on all things Kubernetes and loculus configs.

@fhennig @chaoran-chen, may I kindly ask for some help to configure this preview –  I’m new to this 🙏
Would it be possible to run a preview for a few days max?

### DIFF
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -37,8 +37,8 @@ fileSharing:
 s3:
   enabled: true
   bucket:
-    region: us-east-1
-    bucket: loculus-preview-private
+    region: eu-central-1
+    bucket: sr2silo01
 name: "Loculus"
 logo:
   url: "/favicon.svg"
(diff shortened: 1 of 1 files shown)


kubernetes/loculus/values.yaml
```diff
@@ -1787,8 +1787,8 @@ secrets:
   s3-bucket:
     type: raw
     data:
-      accessKey: "dummyAccessKey"
-      secretKey: "dummySecretKey"
+      accessKey: "QUtJQTZBQjVFRkszTjZLREpKNTI="
+      secretKey: "SDBGTnVLTG03dXVMYTVsczFJNjdjcmppRFJYdm1rd1BrQ1lGTTdhcw=="
```
corneliusroemer, line 1791: @gordonkoehn is that your real access key?  😬 you might want to revoke it as it would have been exposed publicly now

gordonkoehn: Thank you for flagging. Yes, I messed up when committing. It's revoked. Just spoke with Chaoran that we shall set this preview instance up differently anyway.

PR #4379, opened 2025-06-07
fix(backend): file upload for revisions

chaoran-chen:
There was a bug that prevented files from being attached to revisions. We need to associate revision entries with the existing sequence entries and their groups before checking the group ownership of files. I added a test to verify that attaching a file mapping now works.

### PR Checklist
- ~[ ] Any manual testing that has been done is documented (i.e. what exactly was tested?)~

🚀 Preview: https://fix-files-revisions.loculus.org


backend/src/main/kotlin/org/loculus/backend/model/SubmitModel.kt
```diff
@@ -147,6 +141,13 @@ class SubmitModel(
             )
         }
 
+        submissionParams.files?.let { submittedFiles ->
+            val fileSubmissionIds = submittedFiles.keys
+            validateSubmissionIdSetsForFiles(metadataSubmissionIds, fileSubmissionIds)
+            // the check below reads the DB, so needs to be after 'insertDataIntoAux'
```
theosanderson, line 147: unrelated, but this comment feels unncessary

PR #4384, opened 2025-06-08
fix(backend): don't emit files twice in released data

chaoran-chen:
We were outputting the whole `ProcessedData`, with additionally computed metadata, in `/get-released-data`, the `files` were emitted as a top-level field. This was not desired as SILO (only) takes the file mapping as a (stringified) metadata field. This PR removes the redundant `files` field.

### PR Checklist
    - Called `/get-released-data` manually

### DIFF
--- a/backend/src/test/kotlin/org/loculus/backend/controller/submission/GetReleasedDataEndpointTest.kt
+++ b/backend/src/test/kotlin/org/loculus/backend/controller/submission/GetReleasedDataEndpointTest.kt
@@ -25,8 +25,10 @@ import org.hamcrest.CoreMatchers.`is`
 import org.hamcrest.MatcherAssert.assertThat
 import org.hamcrest.Matchers
 import org.hamcrest.Matchers.equalTo
+import org.hamcrest.Matchers.everyItem
 import org.hamcrest.Matchers.greaterThan
 import org.hamcrest.Matchers.hasSize
+import org.hamcrest.Matchers.`in`
 import org.hamcrest.Matchers.matchesPattern
 import org.hamcrest.Matchers.not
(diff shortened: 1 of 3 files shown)


backend/src/test/kotlin/org/loculus/backend/controller/submission/GetReleasedDataEndpointTest.kt
```diff
@@ -114,13 +108,34 @@ class GetReleasedDataEndpointTest(
...
-        val responseBody = response.expectNdjsonAndGetContent<ProcessedData<GeneticSequence>>()
+        val responseBody = response.expectNdjsonAndGetContent<ReleasedData>()
         assertThat(responseBody, `is`(emptyList()))
         response.andExpect(status().isOk)
             .andExpect(header().string(ETAG, notNullValue()))
             .andExpect(header().string("x-total-records", `is`("0")))
     }
 
+    @Test
+    fun `Given released data THEN does not have unknown top-level fields`() {
```
theosanderson, line 119: This feels (to someone like me who doesn't know much) like it should be the job of the typechecker rather than a test - but I assume that's not possible

theosanderson (approved):
lgtm!

backend/src/test/kotlin/org/loculus/backend/controller/submission/GetReleasedDataEndpointTest.kt
chaoran-chen: I wanted to have an explicit check to avoid having new fields being "leaked" without noticing, as happened previously, where we added a field to a class and, because the tests reused the class when parsing the response, the tests didn't fail.

PR #4403, opened 2025-06-10
chore(backend): use BOM to manage JUnit dependency

chaoran-chen:
In #4393, dependabot was not able to update a JUnit dependency. The issue is that we import JUnit dependencies through different paths: `org.junit.platform:junit-platform-launcher` is imported directly, while other packages are imported indirectly through other dependencies, and in case of #4393, this is causing conflicting packages to be imported. To solve this, we can use a BOM (Bill of Material) package to manage the versions.

For more information about the problem and the BOM-packages, [here](https://reflectoring.io/maven-bom/) is a good article.


theosanderson (approved):
(I haven't read up but if tests pass I'm happy!)

PR #4443, opened 2025-06-18
chore(deps): pin node types to nvmrc version (22)

corneliusroemer:
Avoids #4430 coming back

### DIFF
--- a/.github/workflows/website-image.yml
+++ b/.github/workflows/website-image.yml
@@ -71,7 +71,7 @@ jobs:
       - name: Get node version build arg
         if: env.CACHE_HIT == 'false'
         run: |
-          NODE_VERSION=$(cat website/.nvmrc | tr -cd [:digit:].)
+          NODE_VERSION=$(grep -v "^[[:space:]]*#" website/.nvmrc | tr -d 'v')
           echo "NODE_VERSION=$NODE_VERSION" >> $GITHUB_ENV
       - name: Build and push image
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -17,6 +17,8 @@ updates:
     - dependency-name: "change-case"
     # see https://github.com/loculus-project/loculus/issues/4020
     - dependency-name: "flowbite-react"
+    - dependency-name: "@types/node"
+      versions: [">=23"] # Increase when we update node version in .nvmrc
```
anna-parker, line 21: I could see that this will be potentially forgotten - can we add a comment to the .nvmrc file to update this as well if it changes (I assume that will be a manual update)

corneliusroemer: ha, i had the same thought and no we can't add comments to the nvmrc 🤷‍♂️

theosanderson: Apparently you can now https://github.com/nvm-sh/nvm/commit/29dce5edfd0976f9a1728c5746715c24061fd404
(obvs would need recent nvm version)

corneliusroemer:
Good find! I must have used outdated information

.github/dependabot.yml
corneliusroemer: Done

anna-parker (approved):
Oh awesome! Thanks for this!

PR #4460, opened 2025-06-19
feat(deployment): Make it possible to use local IP instead of `localhost`

fhennig:
based on #4448, #4450, #4453

In #4344, #4298 we ran into issues with the integration tests running in CI. After some research we found that the issues were caused by the backend giving pre-signed URLs to the prepro pipeline, that only work from outside the cluster.

This PR fixes that, by making it possible to use the local IP of the host instead of the literal `localhost`. The IP will work inside and outside of the cluster.

- Introduces new `localHost` value in Helm Chart to override `localhost` literal with something else.
- Introduces `./deploy.py helm --use-localhost-ip` flag to use the IP instead of the `localhost` literal.
- Introduces `./deploy.py cluster --bind-all` flag to bind to all interfaces to reach bound ports via the IP.
- Uses the new flags in integration tests CI.

### PR Checklist
- ~~Any manual testing that has been done is documented (i.e. what exactly was tested?)~~

### DIFF
--- a/deploy.py
+++ b/deploy.py
@@ -271,7 +271,6 @@ def get_local_ip() -> str:
         raise RuntimeError(msg)
 
     except (subprocess.CalledProcessError, FileNotFoundError, IndexError) as e:
-        # Fallback if hostname command fails or is not available
         msg = "Could not determine local IP address (hostname -I failed)."
         raise RuntimeError(msg) from e
 
--- a/deploy.py
+++ b/deploy.py
@@ -196,7 +196,7 @@ def cluster_exists(cluster_name):
     return cluster_name in result.stdout
 
 
-def handle_helm():
+def handle_helm():  # noqa: C901
     if args.uninstall:
         run_command(["helm", "uninstall", HELM_RELEASE_NAME])
         return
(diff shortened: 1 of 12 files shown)


deploy.py
```diff
@@ -253,6 +256,24 @@ def handle_helm_upgrade():
...
+    try:
+        result = subprocess.run(['hostname', '-I'],
+                              capture_output=True,
+                              text=True,
+                              check=True)
+
+        ip_addresses = result.stdout.strip().split()
+        if ip_addresses:
+            return ip_addresses[0]
+        return "127.0.0.1"
```
theosanderson, line 270: Maybe actually we should not fallback here – we definitely shouldn't silently fallback which would be really confusing. I would just give an error.

deploy.py
```diff
@@ -253,6 +256,24 @@ def handle_helm_upgrade():
...
+
+    except (subprocess.CalledProcessError, FileNotFoundError, IndexError):
+        # Fallback if hostname command fails or is not available
```
theosanderson, line 273: (same)

theosanderson (approved):
Great

fhennig:
I'm skipping the failing firefox test and created an issue here: https://github.com/loculus-project/loculus/issues/4461

Issue #4084, opened 2025-05-08
Flaky test (timezones?): ""From" field renders date correctly"


theosanderson:
Currently (at 13:46 BST on 8 May) I get this test failure locally on `main` in website unit tests:

```
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯

 FAIL  src/components/SearchPage/fields/DateField.spec.tsx > TimestampField > "From" field renders date correctly
Error: expect(element).toHaveValue(17/03/2025)

Expected the element to have value:
  17/03/2025
Received:
  16/03/2025
 ❯ src/components/SearchPage/fields/DateField.spec.tsx:29:23
     27|         expect(input).toBeInTheDocument();
     28| 
     29|         expect(input).toHaveValue('17/03/2025');
       |                       ^
     30|     });
     31| 

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯


 Test Files  1 failed | 36 passed (37)
      Tests  1 failed | 234 passed (235)
   Start at  13:44:50
   Duration  5.88s (transform 1.72s, setup 4.60s, collect 11.49s, tests 6.22s, environment 5.26s, prepare 1.82s)
```

PR #4477, opened 2025-06-22
fix(website): make rsuite date picker work acceptably with keyboard control

### LINKED ISSUE #4425: Date field bugs
See https://github.com/pathoplexus/pathoplexus/issues/530

**Describe the bug**
The date widgets in the search side panel is impossible to use when typing a date

**To Reproduce**
Click on the 'day' field, type (say), 0-1-0-1-2-0-2-2 (to get 1st Jan 2022) and you actually get: `31-12-1022`

If you type 01, 01, that works but as soon you start typing in the year field it changes the day, month to 31-12 and then messes up the year.

**Desktop (please complete the following information):**
 - OS: Mac OS
 - Browser: Safari & Chrome

I don't think this used to be quite as bad.

corneliusroemer:
This PR reverts back to using rsuite's date picker but with key changes that fix various bugs (I'm not categorically opposed to using a different date picker, or as in this case rolling our own but there's a lot of edge cases that need handling and the current version doesn't work on mobile, for example - so it's best in my view to go back to the tried and tested way with key fixes and develop the other without pressure).

The PR targets the "revert-custom-date" branch which has the custom date picker reverted to make the diff here easier to review. The custom date picker came with test improvements that are worth keeping in either case.

The switch to `yyyy-MM-dd` is in separate PR #4480 which targets this branch.

I switch back to uncontrolled mode as controlled mode plays badly with keyboard input, possibly because that mode is not well tested, especially when partial/early dates are visible in the field, such as `11-01-0019` which is inevitable when using the keyboard.

Lots of keyboard entry bugs (#4425, #4084, unreported inability to type for timestamp field) were introduced when we switched to controlled mode in https://github.com/loculus-project/loculus/pull/3895. The switch to controlled mode was presumably made to allow clearing the date field using the search facet's x:
image

This PR uses conditional rerendering of the DatePicker to control _only_ clearing when we want it: when the fieldValue changes from upstream.

I also rewrite the JSDate <-> IsoDate logic in the DateField component to allow round tripping. This was essential to fix #4425 while using controlled date picker. It's no longer essential for the uncontrolled date picker - but nice to have.

### Manual testing

- Verify dates can be typed in properly in the date picker in both date range and timestamp modes (both didn't work before)
- Clearing works: via cross on date picker, and facet above the search ui
- Input doesn't lose focus on backspace
- Value is correctly set from URL params and vice versa

🚀 Preview: https://rsuite-fix.loculus.org


theosanderson:
(the timezone thing you mention is possibly related to https://github.com/loculus-project/loculus/issues/4084  / https://github.com/loculus-project/loculus/pull/4152 )

theosanderson:
Let's chat in Monday meeting about what direction we want to go on this

website/src/components/SearchPage/fields/DateField.tsx
```diff
@@ -69,30 +137,60 @@ const CustomizedDatePicker: FC<CustomizedDatePickerProps> = ({
...
+        internalClear.current = true;
+        setSomeFieldValues([field.name, '']);
+    };
+
     return (
         <div>
-            <div className='flex justify-between items-center'>
-                <label htmlFor={field.name} className='block text-sm w-16 my-3 text-right mr-2 text-gray-400'>
+            <div className='flex items-center justify-between'>
+                <label htmlFor={field.name} className='block w-16 my-3 mr-2 text-sm text-right text-gray-400'>
```
corneliusroemer, line 181:
```suggestion
            <div className='flex justify-between items-center'>
                <label htmlFor={field.name} className='block text-sm w-16 my-3 text-right mr-2 text-gray-400'>
```

website/src/components/SearchPage/fields/DateField.tsx
```diff
@@ -69,30 +137,60 @@ const CustomizedDatePicker: FC<CustomizedDatePickerProps> = ({
...
-                        if (date) {
-                            setSomeFieldValues([field.name, dateToValueConverter(date)]);
-                        } else {
-                            setSomeFieldValues([field.name, '']);
-                        }
-                    }}
-                    onClean={() => {
-                        setSomeFieldValues([field.name, '']);
-                    }}
+                    defaultValue={defaultDate}
```
corneliusroemer, line 188:
```suggestion
                    defaultValue={defaultDate}
                    key={resetKey}
```

website/src/components/SearchPage/fields/DateRangeField.spec.tsx
```diff
@@ -1,7 +1,6 @@
 import { render, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
-import { Settings } from 'luxon';
-import { useCallback, useState } from 'react';
+import { useState, useCallback } from 'react';
```
corneliusroemer, line 3:
```suggestion
import { useCallback, useState } from 'react';
```

website/src/components/SearchPage/fields/DateField.tsx
```diff
@@ -69,30 +137,60 @@ const CustomizedDatePicker: FC<CustomizedDatePickerProps> = ({
...
+
     return (
         <div>
             <div className='flex justify-between items-center'>
                 <label htmlFor={field.name} className='block text-sm w-16 my-3 text-right mr-2 text-gray-400'>
                     {field.displayName ?? field.name}
                 </label>
                 <DatePicker
-                    value={dateValue}
+                    defaultValue={defaultDate}
```
corneliusroemer, line 185:
```suggestion
                    format='dd/MM/yyyy'
                    defaultValue={defaultDate}
```

theosanderson (approved):
Thanks - looks good! The fact that we need so much code to workaround stuff does suggest it may make sense to have our own version at some point (LLMs change the effort-calculus on these things) but this seems better than what's on main atm

PR #4507, opened 2025-06-24
Deduplicate inherently coupled ena deposition related configs

corneliusroemer:
It doesn't make sense to duplicate the scientific name. It's:
- Annoying for maintainers
- Prone to cause bugs
- The annotation file is inherently coupled to ENA deposition, that's the reason we're putting those annotations in there

Without this, we'd need to add 40 lines of yaml to our PPX production yaml: 10 organisms x 2 processing versions x 2 lines per version... that's just silly

### DIFF
--- a/kubernetes/loculus/templates/loculus-preprocessing-config.yaml
+++ b/kubernetes/loculus/templates/loculus-preprocessing-config.yaml
@@ -5,6 +5,9 @@
 {{- $nucleotideSequencesList := (eq $rawNucleotideSequences nil | ternary (list "main") $rawNucleotideSequences) }}
 {{- range $processingIndex, $processingConfig := $organismConfig.preprocessing }}
 {{- if $processingConfig.configFile }}
+{{- /* Use the enaDepositionConfig as the base config */}}
+{{- $enaDepositionConfig := dig "enaDeposition" "configFile" dict $organismConfig }}
+{{- $mergedConfigFile := merge $processingConfig.configFile $enaDepositionConfig }}
 ---
 apiVersion: v1
 kind: ConfigMap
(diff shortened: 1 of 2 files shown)


kubernetes/loculus/templates/loculus-preprocessing-config.yaml
```diff
@@ -5,6 +5,9 @@
 {{- $nucleotideSequencesList := (eq $rawNucleotideSequences nil | ternary (list "main") $rawNucleotideSequences) }}
 {{- range $processingIndex, $processingConfig := $organismConfig.preprocessing }}
 {{- if $processingConfig.configFile }}
+{{- /* Use the enaDepositionConfig as the base config for scientific_name and molecule_type */}}
```
anna-parker, line 8: We should probably only import these fields if the enaDeposition container is also configured e.g. `{{- if not .Values.disableEnaSubmission }}`

corneliusroemer: There's no benefit to making it conditional on not being disabled. Disabling shouldn't change the flatfile.

anna-parker (approved):

kubernetes/loculus/templates/loculus-preprocessing-config.yaml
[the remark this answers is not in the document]
corneliusroemer: Good question - for one that's where the fields are set at the moment. So this requires the last changes. Could do it the other way round but there are multiple preprocessing versions so it's not clear which one to use as the basis - probably the highest one? Also the prepro config is massive whereas ena deposition is small so there's less baggage carried this way round.

fhennig (approved):

PR #4508, opened 2025-06-24
fix(ci): parse nvmrc with comments correctly for keycloakify image as well

corneliusroemer:
I forgot to update the keycloakify image action in this PR #4443 causing keycloakify image builds to fail, see https://github.com/loculus-project/loculus/actions/runs/15854243044/job/44695382181#step:10:11

Testing: Keycloakify builds again

### DIFF
--- a/.github/workflows/keycloakify-image.yml
+++ b/.github/workflows/keycloakify-image.yml
@@ -48,7 +48,7 @@ jobs:
       - name: Generate files hash
         id: files-hash
         run: |
-          DIR_HASH=$(echo -n ${{ hashFiles('./keycloak/keycloakify/**','.github/workflows/keycloakify-image.yml') }})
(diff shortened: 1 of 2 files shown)


.github/workflows/keycloakify-image.yml
theosanderson, line 1: I guess in theory we should update such that in future we'd have caught this in CI, but nbd

theosanderson (approved):

.github/workflows/keycloakify-image.yml
corneliusroemer: good point! i'll add the nvmrc file as an update condition for the image

corneliusroemer: we should have seen it - but might have gotten lost among the ena-submission failures

PR #4511, opened 2025-06-24
fix(integration-tests): add missing await that otherwise causes spurious test failures

corneliusroemer:
Lack of await caused weird failures in "afterHooks" which made no sense to me until @anna-parker and I stumbled over this together looking through the Playwright UI and tests.

@fhennig remember to sprinkle `await` liberally - linters will readily tell when the `await` is not needed. They often don't tell you when `await`s are missing.


theosanderson (approved):

corneliusroemer:
Actually this eslint PR will error in the future #4512 :D

PR #4512, opened 2025-06-24
chore(integration-tests): set up eslint 9 to catch test bugs early

corneliusroemer:
Adds proper eslint setup to prevent bugs like missing awaits.


theosanderson (approved):
lgtm

Issue #4063, opened 2025-04-29
file sharing: Add user docs


fhennig:
I want to do this after the UI has been a bit more polished

Documentation should contain that file names will become identifiers later on and need to be unique.

Issue #3935, opened 2025-04-02
Build a garbage collector


fhennig:

fhennig:
We have this task in the backend:

```kotlin
@Component
class UseNewerProcessingPipelineVersionTask(
    private val submissionDatabaseService: SubmissionDatabaseService,
    private val auditLogger: AuditLogger,
) {

    @Scheduled(fixedDelay = 10, timeUnit = TimeUnit.SECONDS)
    fun task() {
        submissionDatabaseService.useNewerProcessingPipelineIfPossible()
    }
}

```

I'm imagining the garbage collector to be implemented with such a task as well.

- Index the files in the bucket, cross check with the files in the files table. Any rows in the table that are old and no files were submitted, we can delete the rows in the table.
- Then we should fetch all the IDs that are assigned in some original data somewhere. Any ID that is older than X but not part of any submission can also be deleted.

How do we communicate this to users? That they need to make their submission within timeframe X of the file upload?

theosanderson:
> How do we communicate this to users? That they need to make their submission within timeframe X of the file upload?

We don't need to be agressive with the deletions - so in terms of web UI I don't think we need to communicate this except with a sensible error message if the final submission fails for something that is likely this reason. For programmatic upload, include it  in the docs.

anna-parker:
some recommendations from claude: 

 The cleanest fixes are:                                                                                               
  
  - S3 lifecycle rule (no code needed): configure an                                                                    
  https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpu-abort-incomplete-mpu-lifecycle-config.html rule to
  auto-abort stale incomplete uploads after e.g. 1 day. For completed-but-orphaned objects, a lifecycle rule expiring   
  objects under files/ after N days unless tagged (submitted objects could get a "keep" tag on submission) would work.
  - Backend cleanup endpoint: add a DELETE /files/{fileId} that the frontend calls on discard, which calls
  s3Client.deleteObject(...) and removes the DB row. Straightforward but requires the frontend to track which file IDs  
  were uploaded so it can send them on discard.
  - Scheduled cleanup job: a background task that periodically deletes DB file entries (and their S3 objects) that were 
  created more than X days ago and never referenced by a submission.                                                    
   
  The lifecycle rule for incomplete multipart uploads is a no-brainer to add regardless. For orphaned completed objects 
  the scheduled cleanup is probably the most robust since it also catches abandoned submissions, not just discards. Want
[...]

PR #4515, opened 2025-06-25
fix(backend): publish approved files after pipeline update

### LINKED ISSUE #4055: Make new processed files of approved entries public after pipeline update
Currently, processed files are made public when a sequence entry that they are linked to is approved. When a pipeline is updated, it may create new files. Files associated with an approved entry should be made public when the new pipeline is switched in.

(The garbage collector #3935 will be responsible for removing unused files.)

chaoran-chen:
resolves #4055

When submitting processed data of an already approved entry, associated files will be made public immediately. 

### PR Checklist
    - The automated tests cover the fix well.

### DIFF
--- a/backend/src/test/kotlin/org/loculus/backend/controller/submission/SubmitProcessedDataEndpointTest.kt
+++ b/backend/src/test/kotlin/org/loculus/backend/controller/submission/SubmitProcessedDataEndpointTest.kt
@@ -1,9 +1,11 @@
 package org.loculus.backend.controller.submission
 
+import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.databind.node.BooleanNode
 import com.fasterxml.jackson.databind.node.DoubleNode
 import com.fasterxml.jackson.databind.node.IntNode
 import com.fasterxml.jackson.databind.node.TextNode
+import com.fasterxml.jackson.module.kotlin.readValue
 import org.hamcrest.MatcherAssert.assertThat
 import org.hamcrest.Matchers.containsString
 import org.hamcrest.Matchers.hasEntry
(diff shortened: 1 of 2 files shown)


corneliusroemer (dismissed):
Can you briefly explain at high level how publicness is managed for files?

Should files of old processing versions remain public after bumping of processing version? If not, and people use direct s3 links that would break the links - I guess that's a reason not to expose those direct links.

Maybe we can document the intended behavior somewhere in README or file sharing dev docs.

Have you manually tested it? If so add a short blurb pls to PR.

chaoran-chen:
Thanks for the review, @corneliusroemer.

> @chaoran-chen backend tests failing apparently (you might be aware, just in case)

It's fixed now.

> Can you briefly explain at high level how publicness is managed for files?

A file is publicly accesible in S3 if it has the "public" tag. It is made public when (1) a sequence entry is approved which includes the file in its processed data and (2) (added in this PR) when the pipeline submits processed data for an already approved entry that includes the file.

> Should files of old processing versions remain public after bumping of processing version? If not, and people use direct s3 links that would break the links - I guess that's a reason not to expose those direct links.

The current plan is that old files will be removed. This will be implemented in #3935.

> Maybe we can document the intended behavior somewhere in README or file sharing dev docs.

We'll write up the documentation as part of #4063 (some part will be enduser facing and some admin facing).

> Have you manually tested it? If so add a short blurb pls to PR.

As already written in the PR checklist, the automated test covers the feature sufficiently in my opinion. (I wrote the test (semi-)manually...)

backend/src/test/kotlin/org/loculus/backend/controller/submission/SubmitProcessedDataEndpointTest.kt
```diff
@@ -615,6 +628,71 @@ class SubmitProcessedDataEndpointTest(
             )
     }
 
+    @Test
+    fun `WHEN I submit file for already approved entry THEN new file is public`() {
```
fhennig, line 632: The title of the test is a bit odd to me - as a user I cannot submit files for already approved entries, I have to make a revision. IMO it'd be good to mention the pipeline and pipeline version aspect. Maybe: 

```suggestion
    fun `WHEN the preprocessing pipeline submits files for already approved entries THEN new file is public`() {
```

(line might no be too long and cause lint failures)

fhennig (changes requested):
Nice! I like the step by step comments. Just a few things I noted, but looks generally good.

fhennig (approved):
great, thanks!

PR #4523, opened 2025-06-26
feat(int-tests): speed up read-only dependent tests by using pre-existing sequences

corneliusroemer:
When doing Playwright debugging locally, it was annoying that for dependent tests, there would be new sequence creation every time one runs a test. This took 15+ seconds.

This PR now checks for existence of a sequence and skips submission of sequence if there already is a sequence in a group. This takes only ~2 seconds allowing faster iteration on dependent tests.

### DIFF
--- a/integration-tests/tests/pages/submission.page.ts
+++ b/integration-tests/tests/pages/submission.page.ts
@@ -14,7 +14,16 @@ class SubmissionPage {
 
     async navigateToSubmissionPage(organism: string = 'Ebola Sudan') {
         await this.page.getByRole('link', { name: 'Submit', exact: true }).click();
-        await this.page.getByRole('link', { name: organism }).click();
+
+        // Dependign on current state (organism selected or not), we need to switch
+        const organismSwitchLink = this.page.getByRole('link', { name: organism });
+        const organismLocator = this.page.locator('label').filter({ hasText: organism });
+
+        await Promise.race([
+            organismLocator.waitFor({ state: 'visible', timeout: 5000 }),
+            organismSwitchLink.click(),
+        ]);
+
         await this.page.getByRole('link', { name: 'Submit Upload new sequences.' }).click();
     }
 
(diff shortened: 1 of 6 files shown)


integration-tests/tests/pages/submission.page.ts
```diff
@@ -13,8 +13,17 @@ class SubmissionPage {
     }
 
     async navigateToSubmissionPage(organism: string = 'Ebola Sudan') {
-        await this.page.getByRole('link', { name: 'Submit' }).click();
-        await this.page.getByRole('link', { name: organism }).click();
+        await this.page.getByRole('link', { name: 'Submit', exact: true }).click();
+
+        // Dependign on current state (organism selected or not), we need to switch
```
corneliusroemer, line 18:
```suggestion
        // Depending on current state (organism selected or not), we need to switch
```

integration-tests/tests/pages/submission.page.ts
```diff
@@ -101,15 +110,25 @@ export class SingleSequenceSubmissionPage extends SubmissionPage {
...
             submissionId: string;
             collectionCountry: string;
             collectionDate: string;
             authorAffiliations: string;
+            groupId?: string;
         },
         sequenceData: Record<string, string>,
     ): Promise<ReviewPage> {
         await this.navigateToSubmissionPage();
+        // Switch to group the url if a groupId is provided
```
corneliusroemer, line 124:
```suggestion
```

integration-tests/tests/readonly.setup.ts
theosanderson, line 1: I don't feel super strongly because it wasn't set up ideally before anyway, but I don't think these decisions about which direction to go are clear cut: ideally one wants the test's outcome not to depend on any past runs, and this moves us further away from that

corneliusroemer: The DX is bad as is. And readonly only requires there to be a sequence so it doesn't matter how that sequence comes to exist. The setup already uses that same sequence - this doesn't change anything.

theosanderson: I realise that message is insufficiently clear as I don't describe how one would go about having this work: searching based on the specific group that created this round of read only data - happy to approve this for now to save further discussion and we can always change back in future

corneliusroemer: It doesn't matter because if we require certain things we can make sure of that in the future.

corneliusroemer: Yes - but none of the reuse here matters for CI, only for local testing. One could make this reuse more optional but let's cross that bridge once we get there.

theosanderson: I would like to be able to test reproducibly locally too!! :) (I agree we can tackle it later if needed)

PR #4539, opened 2025-06-30
feat(website): add show more for long values

### LINKED ISSUE #1797: Use "show more" for more fields with potentially very large input: insertions, deletions, ...
Ingesting mpox shows that we need to clip by default and add "show more" for more fields on the seq details page:

Brave Browser 2024-05-07 15 01 27

Any of the sequence fields can get potentially large for both amino acids and nucs: insertions/deletions/substitutions.

The reason AA are not blowing up is that we've artificially only added one gene for mpox rather than all 200, so don't limit the show more to nucs.

Update: mpox also has a looong list of genes - on the amino acid view it would also be good to have the option to collapse the list of genes and their mutations into a "show more" option as this otherwise takes up a full page.

corneliusroemer:
resolves #1797

### Summary
- handle long string fields by truncating them and providing a "Show more" toggle
- the truncation happens preferentially on `,` or `;`, followed by space. Only if none of these occur is termination done inside a word/number
- Max length is currently hard coded, this could be turned into a setting eventually.
- The truncation applies to the PlainDisplayValue - which is now no longer quite plain. We could change the name and possibly in the future add a new real plain one that never truncates. I think truncating by default is a good default - we don't want to add custom display to dozens of fields

### Screenshot

image

------
https://chatgpt.com/codex/tasks/task_e_6862782bf26883258c65a17300b0a76b

🚀 Preview: https://codex-adapt-plainvaluedis.loculus.org


website/src/components/SequenceDetailsPage/DataTableEntryValue.tsx
```diff
@@ -91,13 +91,51 @@ const CustomDisplayComponent: React.FC<Props> = ({ data, dataUseTermsHistory })
...
     if (typeof value === 'boolean') {
         return <span>{value ? 'True' : 'False'}</span>;
     }
 
+    if (typeof value === 'string' && value.length > MAX_PLAIN_STRING_LENGTH) {
+        const preview = computePreview(value);
+        return (
+            <span>
+                {showMore ? value : `${preview}...`}{' '}
+                <button onClick={() => setShowMore(!showMore)} className='underline'>
```
theosanderson, line 130: IMO this is better as a `className='block'` element in the `Show less` case to give some separation so you can e.g. screenshot the list of insertions and just for aesthetics

theosanderson (approved):

kubernetes/loculus/values.yaml
theosanderson, line 1: to revert

corneliusroemer: revert what?

website/src/components/SequenceDetailsPage/DataTableEntryValue.tsx
```diff
@@ -91,13 +91,51 @@ const CustomDisplayComponent: React.FC<Props> = ({ data, dataUseTermsHistory })
...
```
corneliusroemer, line 130:
```suggestion
                <button onClick={() => setShowMore(!showMore)} className='block'>
```

website/src/components/SequenceDetailsPage/DataTableEntryValue.tsx
```diff
@@ -91,13 +91,51 @@ const CustomDisplayComponent: React.FC<Props> = ({ data, dataUseTermsHistory })
...
+
 const PlainValueDisplay: React.FC<{ value: TableDataEntry['value'] }> = ({ value }) => {
+    const [showMore, setShowMore] = React.useState(false);
+
     if (typeof value === 'boolean') {
         return <span>{value ? 'True' : 'False'}</span>;
     }
 
+    if (typeof value === 'string' && value.length > MAX_PLAIN_STRING_LENGTH) {
+        const preview = computePreview(value);
```
chaoran-chen, line 126: It would be good to cache/memo this as the calculation is not very cheap.

chaoran-chen (commented):
Thanks, @corneliusroemer! I just have a few minor comments, looks good otherwise!

website/src/components/SequenceDetailsPage/DataTableEntryValue.tsx
corneliusroemer: Done

website/src/components/SequenceDetailsPage/DataTableEntryValue.tsx
```diff
@@ -91,13 +91,51 @@ const CustomDisplayComponent: React.FC<Props> = ({ data, dataUseTermsHistory })
...
 };
 
+const MAX_PLAIN_STRING_LENGTH = 400;
+const SHOW_MORE_LENGTH = 10; // 'Show more' text length
+
+// Preview ends at the last comma or semicolon in the last 50 characters preceding the display limit.
+// If there's no comma or semicolon, it will cut off at the last space before the display limit.
+// If no space is found, it will cut off at the display limit.
+// We reserve 13 characters for the 'Show more' text, so the preview is limited to 387 characters.
+const computePreview = (value: string): string => {
```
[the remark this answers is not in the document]
corneliusroemer: I've renamed to `computePreviewString`

PR #4541, opened 2025-06-30
fix(ingest): fix bare string in ingest

theosanderson:
This looks unintended? (Found by codex)


anna-parker (approved):
Good catch!

PR #4545, opened 2025-06-30
chore(k8s): add mpox to default values.yaml but only ingest 1%

corneliusroemer:
Mpox poses unique challenges with its large genome and many genes that current organisms don't cover. See e.g. #4539

To not use too much RAM/CPU, we only ingest 1% of mpox sequences.

resolves #

### PR Checklist


kubernetes/loculus/values.yaml
theosanderson, line 1: I was going to suggest removing another organism but it's not clear which - happy to merge this for now and see what happens

corneliusroemer: Could remove West Nile if we wanted to - I'm not sure it does anything that mpox doesn't.

theosanderson: WNV has been where we discovered bugs relating to number of sequences, e.g. timeouts in `get-released-data`

theosanderson: (but I guess now we have RSV in PP, so WNV is insufficient to test the number of seqs we have in production anyway)

PR #4555, opened 2025-06-30
feat(integration-tests): Fail immediately on console warning instead of at the end of a test

fhennig:

### DIFF
--- a/integration-tests/tests/fixtures/console-warnings.fixture.ts
+++ b/integration-tests/tests/fixtures/console-warnings.fixture.ts
@@ -1,36 +1,23 @@
 import { test as base, ConsoleMessage, expect } from '@playwright/test';
 
-type ConsoleWarningsFixtures = {
-    consoleWarnings: string[];
-};
-
-export const test = base.extend<ConsoleWarningsFixtures>({
-    consoleWarnings: [
-        async ({ page, browserName }, use) => {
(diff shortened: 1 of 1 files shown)


integration-tests/tests/fixtures/console-warnings.fixture.ts
```diff
@@ -1,36 +1,23 @@
...
+            if (msg.type() === 'warning' || msg.type() === 'error') {
+                const messageText = msg.text();
+                // Filter out known harmless warnings
+                if (
+                    !messageText.includes(
+                        'Form submission canceled because the form is not connected',
+                    ) &&
+                    browserName === 'chromium'
+                ) {
+                    expect(false, `Unexpected console ${msg.type()}: ${messageText}`).toBe(true);
```
theosanderson, line 15: idk if `throw` is better than this? but nbd either way

theosanderson (approved):
Nice - I think this will really help actually - we're seeing lots of these!

integration-tests/tests/fixtures/console-warnings.fixture.ts
fhennig: I don't know either, I wanted to use something that's part of the framework, usually I'd expect there to be a sort of "fail" method but I couldn't find one. I'll leave it for now.

PR #4572, opened 2025-07-01
chore(ci): run ena integration tests on all ena-submission file changes

corneliusroemer:
The action was set to only run when the test file itself changed - we want it to run in general when there are relevant changes to the ena package as a whole.

### DIFF
--- a/.github/workflows/ena-submission-workflow-tests.yml
+++ b/.github/workflows/ena-submission-workflow-tests.yml
@@ -8,9 +8,6 @@ on:
     branches:
       - main
   workflow_dispatch:
-concurrency:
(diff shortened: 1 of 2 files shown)


.github/workflows/ena-submission-workflow-tests.yml
```diff
@@ -2,15 +2,12 @@ name: ena-submission-workflow-tests
...
   pull_request:
     paths:
-      - "ena-submission/scripts/test_ena_submission_integration.py"
+      - "ena-submission/**"
       - ".github/workflows/ena-submission-workflow-tests.yml"
   push:
     branches:
       - main
   workflow_dispatch:
-concurrency:
```
anna-parker, line 11: maybe we should still keep this to not send too many requests to ENA

anna-parker (approved):

PR #4574, opened 2025-07-01
fix(deposition): fix all `CreationResult(results=` typos, allow setting user/pass in .env file

corneliusroemer:
I used find/replace to replace the remaining CreationResult kwarg typos (there were 2 left)

Also, allow using `.env` file to keep username/password for deposition (more convenient than env variables that need to be explicitly set).

### DIFF
--- a/ena-submission/src/ena_deposition/ena_submission_helper.py
+++ b/ena-submission/src/ena_deposition/ena_submission_helper.py
@@ -61,6 +61,8 @@ def get_ena_config(
     ena_submission_url_default: str,
     ena_reports_service_url_default: str,
 ) -> ENAConfig:
+    import dotenv
+    dotenv.load_dotenv()
     ena_submission_username = os.getenv("ENA_USERNAME")
     if not ena_submission_username:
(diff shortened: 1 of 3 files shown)


ena-submission/src/ena_deposition/ena_submission_helper.py
```diff
@@ -61,6 +61,8 @@ def get_ena_config(
     ena_submission_url_default: str,
     ena_reports_service_url_default: str,
 ) -> ENAConfig:
+    import dotenv
```
anna-parker, line 64: why is this needed?

anna-parker (approved):
Thanks for finding these!!

ena-submission/src/ena_deposition/ena_submission_helper.py
corneliusroemer: You also need to set the environment variables `ENA_USERNAME` and `ENA_PASSWORD` to your ENA test account credentials. You can use the `.env` file in the root of this repository to set these variables.

PR #4581, opened 2025-07-02
fix(backend): Fix file publishing for pipeline update race condition

### LINKED ISSUE #4580: Files not public if approval is between being processed and the pipeline version update
As @anna-parker noticed, files may sometimes still not be public after a pipeline update despite #4515. I think that the following series of events can cause problems:

1. Submission of an entry
2. Entry is processed by pipeline version 1
3. Entry is processed by pipeline version 2
4. Entry is approved -> this only releases the files of pipeline version 1 at the moment
5. Pipeline version 2 is put in place

fhennig:
resolves #4580

This PR changes the file publishing code to not only publish files for the current pipeline version, but also for all larger pipeline versions (even if they are not the current one yet). This means that if the current pipeline version is switched later on, the file will already be public.

The PR includes a test - I checked that the test failed without the change applied.

**manual testing** - If I'm not mistaken, this requires quite a bit of luck to get the timing right, so it's not trivial to test manually. IMO the automated test makes me very confident in the fix.

### PR Checklist

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
@@ -423,7 +423,7 @@ class SubmissionDatabaseService(
                     sequenceEntries.accessionVersionIsIn(accessionVersionsChunk)
                 }
                 .flatMap {
-                    it[SequenceEntriesPreprocessedDataTable.processedDataColumn]?.files?.values.orEmpty()
+                    it[preproData.processedDataColumn]?.files?.values.orEmpty()
                 }
                 .flatten()
                 .forEach { result.add(it.fileId) }


backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -399,15 +399,31 @@ class SubmissionDatabaseService(
...
+                    },
+                )
+                .select(preproData.processedDataColumn)
                 .where {
-                    SequenceEntriesView.accessionVersionIsIn(accessionVersionsChunk)
+                    sequenceEntries.accessionVersionIsIn(accessionVersionsChunk)
                 }
                 .flatMap {
-                    it[SequenceEntriesView.processedDataColumn]?.files?.values.orEmpty()
+                    it[SequenceEntriesPreprocessedDataTable.processedDataColumn]?.files?.values.orEmpty()
```
chaoran-chen, line 426:
```suggestion
                    it[preproData.processedDataColumn]?.files?.values.orEmpty()
```

chaoran-chen (approved):
Looks great, thanks!

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -619,7 +635,7 @@ class SubmissionDatabaseService(
             }
         }
 
-        val filesToPublish = this.selectFilesForAccessionVersions(accessionVersionsToUpdate)
+        val filesToPublish = this.selectFilesToPublishForAccessionVersions(accessionVersionsToUpdate)
```
anna-parker, line 638: sorry I only had a look at this now - what is the transaction level here?

Is the pipeline able to upload new prepro versions in parallel? could we have the case where prepro version 2 gets uploaded at the same time (or after) the user clicks approve? `selectFilesToPublishForAccessionVersions` doesn't see the new prepro version when it starts (this looks like just a read operation so it won't stop concurrent writes) and thus the second version will not be published

backend/src/test/kotlin/org/loculus/backend/controller/submission/SubmitProcessedDataEndpointTest.kt
```diff
@@ -693,6 +693,71 @@ class SubmitProcessedDataEndpointTest(
...
+
+        submissionControllerClient.submitProcessedData(
+            PreparedProcessedData.withFiles(
+                accession,
+                mapOf("myFileCategory" to listOf(FileIdAndName(fileIdAndUrlV2.fileId, "v2.txt"))),
+            ),
+            pipelineVersion = 2,
+        ).andExpect(status().isNoContent)
+
+        // Step 4: approve entry
```
anna-parker, line 737: I think it would be good to add a test for what happens when step 3 and step 4 are inverted (maybe you have already covered this)

fhennig: I think that's the test right above

fhennig: yep it is. We have that one covered, that was the issue we had before.

anna-parker: Awesome thanks!

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
fhennig: thanks for pointing it out again!

PR #4582, opened 2025-07-02
WIP

fhennig:
resolves #

### PR Checklist


preprocessing/dummy/lineage-alternative.yaml
```diff
@@ -0,0 +1,33 @@
...
+B:
+  aliases: []
+  parents: []
+B.1:
+  aliases: []
+  parents:
+  - B
+B.1.1:
+  aliases:
+  - C
```
theosanderson, line 26:
```suggestion
C:
  aliases:
  - B.1.1
```
I think from previously this is the intended set up?

fhennig: nono, it was deliberately done the way it is. @chaoran-chen told me that it is supposed to look like that.

fhennig: you only use aliases once you have a number. So you wouldn't use "C" as a lineage def on its own. That was my understanding.

preprocessing/dummy/lineage-alternative.yaml
```diff
@@ -0,0 +1,33 @@
...
+  parents:
+  - B.1
+C.1:
+  aliases:
+  - B.1.1.1
+  parents:
+  - C
```
chaoran-chen, line 33:
```suggestion
  - C

```

I don't like GitHub's

image

symbol ;)

PR #4585, opened 2025-07-02
chore(deps): bump kotlinx-datetime in backend

fhennig:
changelog: https://github.com/Kotlin/kotlinx-datetime/blob/master/CHANGELOG.md#070
Migration info: https://github.com/Kotlin/kotlinx-datetime#deprecation-of-instant

There was a breaking change, where the `kotlinx` library deprecated their own `Clock` and `Instant` in favor of the std lib implementations. Using the std lib implementations however still requires [`ExperimentalTime`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.time/-experimental-time/) - I think. That's why I added the compiler opt in option.
Without it, all places that use time now raise this comile error:

    e: file:///...SeqSetCitationsDatabaseService.kt:314:32 This declaration needs opt-in.
    Its usage must be marked with '@kotlin.time.ExperimentalTime' or '@OptIn(kotlin.time.ExperimentalTime::class)'

Also, I tried to not use the `-0.6.x-compat` version of the `0.7.0`, but I got a bunch of runtime errors. However, this compat version will not be supplied for `0.8.0`, so in the meantime we'd need to see that we migrate away from it. We need to find exactly the libraries that might be causing issues.

I think, the issue is `exposed-kotlin-datetime`: https://github.com/JetBrains/Exposed/blob/89e0be3139a36df8106a6c20b28960cc6cd44408/gradle/libs.versions.toml#L40

Might be good for someone else to have a look too if I understood the Changelog correctly.

If I got it right, I'd recommend we create an issue to check back on bumping away from the compat version.

### PR Checklist


backend/build.gradle
```diff
@@ -26,6 +26,8 @@ kotlin {
     jvmToolchain(21)
     // Added to dismiss warnings, see: https://github.com/loculus-project/loculus/issues/4529
     compilerOptions.freeCompilerArgs.add("-Xannotation-default-target=param-property")
+    // required since kotlinx-datetime:0.7.0
```
chaoran-chen, line 29: This comment is not really useful in the future and I think that it is easy enough to find out what it does. I'd also suspect that once the API is stable and the flag is not needed, the compiler will show a corresponding warning.

```suggestion
```

chaoran-chen (approved):
Thanks! I understand the changelog as you did.

> I'd recommend we create an issue to check back on bumping away from the compat version.

Good idea, would you like to create it?

PR #4605, opened 2025-07-04
chore(docs): typo fixes in file sharing docs

corneliusroemer:
A few typo fixes to the file sharing docs

### DIFF
--- a/docs/src/content/docs/for-administrators/configuring-extra-files.md
+++ b/docs/src/content/docs/for-administrators/configuring-extra-files.md
@@ -111,5 +111,5 @@ my-organism:
 ```
 
 :::note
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/for-administrators/configuring-extra-files.md
```diff
@@ -111,5 +111,5 @@ my-organism:
 ```
 
 :::note
-The name (`rawReads` in the example above) needs to not be a name that is also used by a metadata field!
+The name (`rawReads` in the example above) must to not be a name that is also used by a metadata field!
```
corneliusroemer, line 114:
```suggestion
The name (`rawReads` in the example above) must not be a name that is also used by a metadata field!
```

PR #4616, opened 2025-07-07
fix(backend): handle error edge cases correctly in file endpoints

### LINKED ISSUE #4602: Passing invalid group ID as superuser to /files/request-upload causes Internal server error
```
 curl -s -D - -w "%{http_code}" 'https://backend-main.loculus.org/files/request-upload?groupId=0&numberFiles=1' \
  -X 'POST' \
  -H 'accept: application/json, text/plain, */*' \
  -H 'authorization: Bearer <REDACTED>' \

HTTP/2 500
cache-control: no-cache, no-store, max-age=0, must-revalidate
content-type: application/json
date: Thu, 03 Jul 2025 20:50:53 GMT
expires: 0
pragma: no-cache
strict-transport-security: max-age=31536000 ; includeSubDomains
vary: Accept-Encoding
vary: origin,access-control-request-method,access-control-request-headers,accept-encoding
x-content-type-options: nosniff
x-frame-options: DENY
x-request-id: 7ac1e7b4-b504-4225-ab01-964a2efdb518
x-robots-tag: noindex, nofollow
x-xss-protection: 0

{"type":"about:blank","title":"Internal Server Error","status":500,"detail":"org.postgresql.util.PSQLException: ERROR: insert or update on table \"files\" violates foreign key constraint \"files_group_id_fkey\"\n  Detail: Key (group_id)=(0) is not present in table \"groups_table\".","instance":"/files/request-upload"}
```

Other ideas for potential bugs: 0 or negative number of files. Very high number of files. Non-integer for both fields.

corneliusroemer:
resolves #4602

- if user isn't authenticated throw Unauthorized not Forbidden
- check group exists before executing sql
- validate number of files that were requested can't be < 1
- include 'WWW-Authenticate: Bearer' header for 401/403 per spec
- fix grammar in error message where single group can exist

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/controller/ExceptionHandler.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/controller/ExceptionHandler.kt
@@ -119,7 +119,7 @@ class ExceptionHandler : ResponseEntityExceptionHandler() {
         detail: String?,
     ): ResponseEntity<ProblemDetail> = ResponseEntity
         .status(httpStatus)
-        .contentType(MediaType.APPLICATION_JSON)
+        .contentType(MediaType.APPLICATION_PROBLEM_JSON)
         .body(
             ProblemDetail.forStatus(httpStatus).also {
                 it.title = title
--- a/backend/src/main/kotlin/org/loculus/backend/controller/ExceptionHandler.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/controller/ExceptionHandler.kt
@@ -1,8 +1,5 @@
 package org.loculus.backend.controller
 
-import io.swagger.v3.oas.annotations.headers.Header
-import io.swagger.v3.oas.annotations.media.Schema
-import io.swagger.v3.oas.annotations.responses.ApiResponse
 import jakarta.validation.ConstraintViolationException
 import mu.KotlinLogging
 import org.springframework.http.HttpHeaders
@@ -17,6 +14,7 @@ import org.springframework.web.bind.annotation.ExceptionHandler
 import org.springframework.web.bind.annotation.ResponseStatus
 import org.springframework.web.context.request.WebRequest
 import org.springframework.web.servlet.mvc.method.annotation.ResponseEntityExceptionHandler
+import java.net.URI
 
 private val log = KotlinLogging.logger {}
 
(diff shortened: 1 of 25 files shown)


backend/src/main/kotlin/org/loculus/backend/controller/ExceptionHandler.kt
```diff
@@ -45,21 +43,19 @@ class ExceptionHandler : ResponseEntityExceptionHandler() {
         )
     }
 
-    @ExceptionHandler(DummyUnauthorizedExceptionToMakeItAppearInSwaggerUi::class)
```
anna-parker, line 48: oh interesting - this function apparently is not actually used at runtime and will only show up in the swaggerUI

backend/src/test/kotlin/org/loculus/backend/controller/files/GetFilesEndpointTest.kt
```diff
@@ -38,7 +38,7 @@ class GetFilesEndpointTest(
         )
 
         filesClient.getFile(data.accession, data.version, "myFileCategory", "hello.txt")
-            .andExpect(status().isForbidden())
+            .andExpect(status().isUnauthorized())
```
anna-parker, line 41: do we have a test for the Forbidden case?

backend/src/main/kotlin/org/loculus/backend/controller/ExceptionHandler.kt
```diff
@@ -104,10 +100,15 @@ class ExceptionHandler : ResponseEntityExceptionHandler() {
     fun handleForbiddenException(e: ForbiddenException): ResponseEntity<ProblemDetail> {
         log.info { "Caught forbidden exception: ${e.message}" }
 
-        return responseEntity(
-            HttpStatus.FORBIDDEN,
-            e.message,
-        )
+        return ResponseEntity.status(HttpStatus.FORBIDDEN)
```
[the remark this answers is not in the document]
corneliusroemer: indeed :D

backend/src/test/kotlin/org/loculus/backend/controller/files/GetFilesEndpointTest.kt
corneliusroemer: I don't know - but the error code was wrong so this just fixes that

Issue #3984, opened 2025-04-14
Organisms which have sub-organisms with different references from a single LAPIS instances


theosanderson:
Some viruses have subclades with very different reference sequences. We would like to support these under one grouped organism and be able to query all the data together.

For these we probably want:
- a single submission page
- preprocessing to use nextclade to figure out which subclade / suborganism a sequence belongs to
- a metadata column that captures this suborganism/subclade
- selecting a suborganism would be necessary to enable the mutation search
- in terms of LAPIS implementation, the idea I recall hearing is that for the aligned sequences you would have a segment name like `[segment_name]_[subclade]` (for segmented viruses) or `main_[subclade]` (for single-segment viruses).

PR #4641, opened 2025-07-09
Multi-pathogen solution design

fengelniederhammer:
https://github.com/loculus-project/loculus/blob/multiPathogenSolutionDesign/backend/docs/multi_pathogen.md

This is a document to sketch out some details of the multi-pathogen feature (https://github.com/loculus-project/loculus/issues/3984). I don't know whether we want to merge this at some point, but I think it's good to have some version controlled overview over what needs to be done. Feedback is appreciated!


backend/docs/multi_pathogen.md
```diff
@@ -0,0 +1,205 @@
...
+  "version": 1,
+  "data": {
+    "metadata": {...},
+    "unalignedNucleotideSequences": {
+      "main": "ACTG"
+      // TODO: or do we also need multiple segments here?
+    },
+    "alignedNucleotideSequences": {
+      "matchingSuborganism": "ACTG",
+      "otherSuborganism": null
```
chaoran-chen, line 94: We should avoid the need to submit and store `null` or empty arrays.

backend/docs/multi_pathogen.md
```diff
@@ -0,0 +1,205 @@
...
+### Sequence entry details page
+
+Don't show sequences that are not there anyway.
+Only show the "segments" of the relevant suborganism.
+Only show the relevant genes.
+Strip the suborganism name for the sequence names.
+![img.png](sequenceDisplay.png)
+
+Do we need all the references here?
+![img_1.png](mutationReferences.png)
```
chaoran-chen, line 177: No, it should only show the reference for the corresponding suborganism.

backend/docs/multi_pathogen.md
```diff
@@ -0,0 +1,205 @@
...
+
+TODO: Is there more?
+
+### Search Page
+
+Mutation and lineage search should only show up once the search has been narrowed down to a specific suborganism.
+
+Mutation filter: Also strip the suborganism name for the user?
+User inputs `A123T`, we send `suborganism1:A123T` to LAPIS.
+(Similar for amino acid mutations).
```
chaoran-chen, line 190: This also holds for insertions (if we do not make changes to LAPIS): the query to LAPIS would be `ins_<suborganism>:123:AATTGG` or `ins_<suborganism>-<gene>:123:LP`.

fengelniederhammer:
Also feel free to push to this branch. This is not supposed to be my private work, but something collaborative.

backend/docs/multi_pathogen.md
```diff
@@ -0,0 +1,335 @@
...
+
+The purpose of this feature to enable Organisms that are divided into several "suborganisms".
+It should roughly work as follows:
+* The user uploads sequences as usual
+* The preprocessing pipeline will figure out which suborganism the sequence belongs to
+* There will be a LAPIS instance for that organism that has a nucleotide segment for each suborganism
+
+## Helm Chart Configuration
+
+TODO: how do we make an organism multi-pathogen?
```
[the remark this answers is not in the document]
fengelniederhammer: You mean we could do something like:
```yaml
defaultOrganisms:
  entero:
    <<: *defaultOrganismConfig
    schema:
      <<: *schema
      organismName: "Entero Viruses"
    referenceGenomes:
      CV-A10:
        nucleotideSequences:
          - name: main
            sequence: "..."
            insdcAccessionFull: ...
        genes: [...]
      EV-A71:
        nucleotideSequences:
          - name: main
            sequence: "..."
            insdcAccessionFull: ...
        genes: [...]
      ...
```

and that already tells all of the rest that we're in the multi pathogen case for this organism? Or do you think we would still need to configure "multi pathogen" somewhere else for this organism?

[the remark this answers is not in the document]
theosanderson: Yes, thanks - that is what I was proposing.

> and that already tells all of the rest that we're in the multi pathogen case for this organism? Or do you think we would still need to configure "multi pathogen" somewhere else for this organism?

I don't have strong feelings about this: one can argue it both ways. I'd maybe start off by just inferring from the number.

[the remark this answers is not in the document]
chaoran-chen: How would it look like for a single pathogen? Would it not change or would one have to define a sub-organism or write something like `main`?

I favor the latter: although it would be a breaking change and add nesting for the common, single-pathogen case, it keeps the type structure of the config simple(r) and is easier to implement.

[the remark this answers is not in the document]
theosanderson: 👍  this will also solve the fact that this is called `referenceGenomes` when it has until now only been a single reference genome :)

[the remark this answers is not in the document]
corneliusroemer: We should definitely not call it main as we then have the same name for 2 different things (the only segment and the only suborganism)

If we use a placeholder nonetheless we should make sure it remains as seamless/hidden to end users as possible as it's otherwise confusing/not pretty.

References aren't the only thing that need to be "per suborganism" - we haven't got a term of art for it yet so I'll use that until we have something better. Off the top of my head those will require a "per suborganism" as well:
- taxon id and maybe more for ena deposition
- nextclade datasets (links/names/servers)
- some sort of mapping from suborganism to name of reference in minimizer index (unless we require minimizer keys to be identical to suborganism keys for convenience as minimizer will have to be handmade anyways at least for enteros)
- maybe more (@anna-parker might think of a few more)

Bit annoying to have to repeat the same keys in many places - the other option is to tear apart the current top level organisation (preprocessing, deposition, ...) and put "per suborganism" config as one group (containing a repeat of the top level structure preprocessing/deposition/...):

It's probably easiest to add persub where it arises for now and decide whether to refactor/invert later once we have an overview of what will be per sub

```
prepro:
  sharedc:
  sharedd:
  per sub:
    suba:
    subb:
prepro:
  sharedc:
  sharedd:
  per sub:
    suba:
    subb:
referencegenomes:
  persub:
    ...
```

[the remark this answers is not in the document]
fengelniederhammer: > We should definitely not call it main as we then have the same name for 2 different things (the only segment and the only suborganism)

Any suggestions what we could use instead? `single`? Or we just repeat the organism name?

backend/docs/multi_pathogen.md
```diff
@@ -0,0 +1,423 @@
...
+
+TODO: Is the `fileMapping` relevant here?
+
+### INSDC Ingest
+
+No changes here.
+
+## Storage of unprocessed sequences
+
+```json
```
anna-parker, line 129: this has to be changed for multi-pathogen organisms - we discussed in the Monday meeting that we want to add the possibility of the dictionary of `unalignedNucleotideSequences` into a list - this is required as the segments will not be assigned yet

backend/docs/multi_pathogen.md
```diff
@@ -0,0 +1,205 @@
...
+
+Mutation and lineage search should only show up once the search has been narrowed down to a specific suborganism.
+
+Mutation filter: Also strip the suborganism name for the user?
+User inputs `A123T`, we send `suborganism1:A123T` to LAPIS.
+(Similar for amino acid mutations).
+
+Do we still show the segments in the download modal?
+![img.png](downloadModal.png)
+
```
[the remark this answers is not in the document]
emmahodcroft: Good question - we should probably ensure we revisit this. If you can send to Nextclade in multi-path mode this should work I guess, since all Datasets will be there? But should ensure this works in practice not just in theory :)

For the map, I guess it is ok to look at this both for all/multiple EVs and for just one...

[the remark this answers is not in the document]
theosanderson: > Good question - we should probably ensure we revisit this. If you can send to Nextclade in multi-path mode this should work I guess, since all Datasets will be there? But should ensure this works in practice not just in theory :)

At the moment we select a dataset so it wouldn't work in the way we are currently doing it

But I guess my take atm is that we can defer thinking about this until we have implemented the main thing and just e.g. not add any tools until we have something in place for this that we are happy with

backend/docs/multi_pathogen.md
```diff
@@ -0,0 +1,454 @@
...
+```json
+{
+  "accession": "key1",
+  "version": 1,
+  "data": {
+    "metadata": {...},
+    "unalignedNucleotideSequences": {
+      "matchingSuborganism": "ACTG",
+      // omit the non-matching suborganisms
```
anna-parker, line 209: We currently send a list of all segments to the backend where a segment is None when it does not exist in the data

fengelniederhammer: As discussed: We'll keep the requirement that the preprocessing sends all required segments.

backend/docs/multi_pathogen.md
fengelniederhammer: As discussed: We'll probably do that somewhen later (sending a map `fasta header -> sequence` to preprocessing). But for now we stick with having the segment names here.

backend/docs/multi_pathogen.md
```diff
@@ -192,20 +199,39 @@ Multi-segments:
...
 
 Similar to what is stored in the DB.
 
 ## Preprocessing
 
-TODO: Minimizer index, etc?
+The preprocessing pipeline will need to assign each sequence to a segment, if there are multiple segments
+(including validation and useful error messages for the submitter).
+
+TODO: details how to do that - minimizer index, etc.?
```
anna-parker, line 234: Higher level: 
- receive a group of metadata and sequences
- assign suborganism using nextclade sort (will require a minimizer index and potentially a map from match name to suborganism
- once suborganism is assigned run nextclade using the correct nextclade dataset - must be configured per organism 
- must validate that: input does not have multiple segments for one sample - for EV case could also add a config option to allow or disallow co-infections
- as currently, either have hard or soft enforcement of all or any of segments aligning

fengelniederhammer: Would you mind adding this yourself to that section? You have a better overview what to write there and I don't claim any ownership on this PR.

PR #4646, opened 2025-07-10
feat(cli): Command-line interface for Loculus

### LINKED ISSUE #2801: Command line interface (CLI)
We would like to add a CLI for Loculus to download and submit sequences.

Any suggestions on how the interface should look like?

theosanderson:
resolves #2801

Adds an initial non-production-ready command-line interface which we will continue to improve.

Installation instructions in `cli/README.md`

To test it start with:
`loculus instance add cli.loculus.org`

----------

- **Python CLI** (`cli/` package) built with Click framework
- **Core commands**: `auth`, `submit`, `get`, `status`, `release`, `config`, `instance`, `organism`, `group`, `schema`
- (revise is not implemented yet, for no particular reason)
- **Authentication**: Secure credential storage with keyring, OAuth2/Keycloak integration
- **Multiple output formats**: Table, JSON, TSV, FASTA, NDJSON for search results
- **Developer-friendly**: Type checking, rich terminal output
- **Reasonable testing**: Integration tests with Playwright 
- **Website**: New `/loculus-info` endpoint for CLI auto-discovery

## Key Features

- Submit sequences with metadata from FASTA/TSV files
- Search and retrieve sequences with filters
- Monitor submission status and release sequences
- Manage multiple Loculus instances and organisms
- Generate metadata templates for easy submission

## Technical Details

- Python 3.10+ with modern tooling (uv, mypy, black, ruff)
- API clients for both backend and LAPIS endpoints
- Environment variable and config file support
- CI/CD workflows for testing and linting

🚀 Preview: https://cli.loculus.org


corneliusroemer (commented):
Poetry wants to install numpy from source and then fails when I `poetry install`

Maybe we should use uv instead. I guess the LLM hasn't seen much UV yet as it's much newer but there's lots of momentum switching to uv

theosanderson:
Thanks it's now using `uv`!

theosanderson:
> I was a but surprised that uv run loculus --organism cchf schema fields works but uv run loculus schema fields --organism cchv doesn't. I guess this has to do with global vs scoped options.

This is indeed to do with that but I agree it's really not good and I thought I had a hack such that it would now work - I'll look into that

theosanderson:
(all @rneher's points now should have been addressed)

theosanderson:
I would plan to merge this in ~24 hours

PR #4652, opened 2025-07-11
feat(deposition): don't revise assembly if no change, remove all mypy type errors

anna-parker:
1. When revising a sequence check if there is a difference in the assembly, by looking for differences in the `DEFAULT_EMBL_PROPERTY_FIELDS` or the actual sequence - if there is no change do not revise the sequence just copy over the results from the previous version. If there has been a change revise the sequence as normal
1. Adds an integration test that the assembly is not revised if there is no change

### PR Checklist
tested on preview

🚀 Preview: https://dont-revise-assembly-if-n.loculus.org

### DIFF
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -1756,7 +1756,6 @@ enaDeposition:
   enaDbName: Loculus
   enaUniqueSuffix: Loculus
   enaIsBroker: False
-  enaGithubTestUrl: "https://raw.githubusercontent.com/pathoplexus/ena-submission/test_branch/test/approved_ena_submission_list.json"
 subdomainSeparator: "-"
 replicas:
   website: 1


ena-submission/src/ena_deposition/create_assembly.py
```diff
@@ -429,6 +429,90 @@ def can_be_revised(config: Config, db_config: SimpleConnectionPool, entry: dict[
...
+        error_msg = f"Last version {version_to_revise} not found in submission_table"
+        raise RuntimeError(error_msg)
+
+    if (
+        entry["unaligned_nucleotide_sequences"]
+        != last_version_data[0]["unaligned_nucleotide_sequences"]
+    ):
+        logger.debug(
+            f"Unaligned nucleotide sequences have changed for {entry['accession']}, "
+            f"from {version_to_revise} to {entry['version']} - should be revised"
```
corneliusroemer, line 452:
```suggestion
            f"from {version_to_revise} to {entry['version']} - should be revised. "
            "(Metadata maybe also changed.)"
```

ena-submission/scripts/test_ena_submission.py
```diff
@@ -60,6 +61,10 @@ def mock_config():
     config.metadata_mapping_mandatory_field_defaults = defaults[
         "metadata_mapping_mandatory_field_defaults"
     ]
+    config.embl_property_fields = EmblPropertyFields()
```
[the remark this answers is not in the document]
corneliusroemer: I see we now use those fields in multiple places - so good to use central place, but a constant would do, doesn't need to be configurable?

ena-submission/src/ena_deposition/create_assembly.py
```diff
@@ -429,6 +429,90 @@ def can_be_revised(config: Config, db_config: SimpleConnectionPool, entry: dict[
...
+        config.embl_property_fields.authors_property,
+        *config.embl_property_fields.admin_level_properties,
+    ]
+    for field in fields:
+        last_entry = last_version_data[0]["metadata"].get(field)
+        new_entry = entry["metadata"].get(field)
+        if last_entry != new_entry:
+            logger.debug(
+                f"Field {field} has changed from {last_entry} to {new_entry} "
+                f"for {entry['accession']}"
```
corneliusroemer, line 468:
```suggestion
                f"for {entry['accession']}. (Maybe other fields changed as well)"
```

ena-submission/src/ena_deposition/create_assembly.py
```diff
@@ -429,6 +429,90 @@ def can_be_revised(config: Config, db_config: SimpleConnectionPool, entry: dict[
...
+            )
+            return False
+    logger.debug(
+        f"No changes detected for {entry['accession']} from version {version_to_revise} "
+        f"to {entry['version']}"
+    )
+    return True
+
+
+def update_assembly_results_with_latest_version(
```
corneliusroemer, line 478: This could use the retry refactor changes (once merged)

ena-submission/src/ena_deposition/create_assembly.py
```diff
@@ -429,6 +429,90 @@ def can_be_revised(config: Config, db_config: SimpleConnectionPool, entry: dict[
     return True
 
 
+def no_change(config: Config, db_config: SimpleConnectionPool, entry: dict[str, Any]) -> bool:
```
corneliusroemer, line 432: Maybe name could be slightly more self-documenting :D 

`is_flatfile_data_changed`

ena-submission/scripts/test_ena_submission.py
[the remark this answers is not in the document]
anna-parker: I thought it might be easiest to keep all the defaults in the defaults.yaml - otherwise all the metadata fields that we use are defined there. Its not very important I just wanted it to be easier to refactor if we change the name of a field

ena-submission/src/ena_deposition/create_assembly.py
anna-parker: done :-)

ena-submission/src/ena_deposition/ena_submission_helper.py
```diff
@@ -399,22 +400,22 @@ def get_authors(authors: str) -> str:
...
-    country = metadata.get("geoLocCountry", "Unknown")
-    admin_levels = ["geoLocAdmin1", "geoLocAdmin2", "geoLocCity", "geoLocSite"]
-    admin_values = [val for level in admin_levels if (val := metadata.get(level))]
-    admin = ", ".join(admin_values)
-    return f"{country}: {admin}" if admin else country
+    country = metadata.get(DEFAULT_EMBL_PROPERTY_FIELDS.country_property, "Unknown")
+    admin_values = ", ".join(
+        filter(None, map(metadata.get, DEFAULT_EMBL_PROPERTY_FIELDS.admin_level_properties))
+    )
+    return f"{country}: {admin_values}" if admin_values else country
```
corneliusroemer, line 407:
```suggestion
    admin = ", ".join(
        filter(None, map(metadata.get, DEFAULT_EMBL_PROPERTY_FIELDS.admin_level_properties))
    )
    return f"{country}: {admin}" if admin_values else country
```

corneliusroemer (approved):

kubernetes/loculus/values.yaml
```diff
@@ -1756,6 +1756,7 @@ enaDeposition:
   enaDbName: Loculus
   enaUniqueSuffix: Loculus
   enaIsBroker: False
+  enaGithubTestUrl: "https://raw.githubusercontent.com/pathoplexus/ena-submission/test_branch/test/approved_ena_submission_list.json"
```
anna-parker, line 1759: revert before merging
```suggestion
```

PR #4692, opened 2025-07-16
feat!(kubernetes): [CFG] add support for multiple reference genomes

### LINKED ISSUE #4691: Multi pathogen support: multiple reference genomes for a single organism
See https://github.com/loculus-project/loculus/pull/4641#discussion_r2204387061

Make the reference genomes a dictionary that takes multiple reference genomes

fengelniederhammer:
resolves #4691 

Breaking change in the config:

The reference genomes became a dictionary `[suborganismName, referenceGenome]`. Before:
```yaml
    referenceGenomes:
        nucleotideSequences:
          - name: 'main'
            sequence: 'NNN'
        genes: []
```

After:
```yaml
    referenceGenomes:
      singleReference:
        nucleotideSequences:
          - name: 'main'
            sequence: 'NNN'
        genes: []
```

The key must be `singleReference` if there only one suborganism. (Maybe we'll bind logic to that specific name and it's easier to drop that requirement later compared to adding it)

This is supposed to be minimally invasive and change as few place in the code as possible. For now this simply assumes that each organism has exactly one reference genome configured (handling multiple ones in the code will be the next step).

### PR Checklist

🚀 Preview: https://4691-multi-pathogen-suppo.loculus.org

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/service/submission/CompressionService.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/service/submission/CompressionService.kt
@@ -181,6 +181,7 @@ class CompressionService(private val backendConfig: BackendConfig) {
         backendConfig
             .getInstanceConfig(organism)
             .referenceGenomes
+            .values.first()
             .getNucleotideSegmentReference(
                 segmentName,
             )
@@ -189,6 +190,7 @@ class CompressionService(private val backendConfig: BackendConfig) {
     private fun getDictionaryForAminoAcidSequence(geneName: String, organism: Organism): ByteArray? = backendConfig
         .getInstanceConfig(organism)
         .referenceGenomes
+        .values.first()
         .getAminoAcidGeneReference(
             geneName,
         )
--- a/backend/src/main/kotlin/org/loculus/backend/service/submission/EmptyProcessedDataProvider.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/service/submission/EmptyProcessedDataProvider.kt
@@ -11,15 +11,16 @@ import org.springframework.stereotype.Component
 class EmptyProcessedDataProvider(private val backendConfig: BackendConfig) {
     fun provide(organism: Organism): ProcessedData<GeneticSequence> {
         val (schema, referenceGenomes) = backendConfig.getInstanceConfig(organism)
+        val referenceGenome = referenceGenomes.values.first()
 
-        val nucleotideSequences = referenceGenomes.nucleotideSequences.map { it.name }.associateWith { null }
+        val nucleotideSequences = referenceGenome.nucleotideSequences.map { it.name }.associateWith { null }
         return ProcessedData(
             metadata = schema.metadata.map { it.name }.associateWith { NullNode.instance },
             unalignedNucleotideSequences = nucleotideSequences,
             alignedNucleotideSequences = nucleotideSequences,
-            alignedAminoAcidSequences = referenceGenomes.genes.map { it.name }.associateWith { null },
-            nucleotideInsertions = referenceGenomes.nucleotideSequences.map { it.name }.associateWith { emptyList() },
-            aminoAcidInsertions = referenceGenomes.genes.map { it.name }.associateWith { emptyList() },
+            alignedAminoAcidSequences = referenceGenome.genes.map { it.name }.associateWith { null },
+            nucleotideInsertions = referenceGenome.nucleotideSequences.map { it.name }.associateWith { emptyList() },
+            aminoAcidInsertions = referenceGenome.genes.map { it.name }.associateWith { emptyList() },
             files = null,
         )
     }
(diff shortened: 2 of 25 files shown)


backend/src/test/resources/backend_config.json
```diff
@@ -5,22 +5,24 @@
...
-                    {
-                        "name": "someLongGene",
-                        "sequence": "AAAAAAAAAAAAAAAAAAAAAAAAA"
-                    },
-                    {
-                        "name": "someShortGene",
-                        "sequence": "MADS"
-                    }
-                ]
+                "dummyOrganism": {
```
chaoran-chen, line 8: To me, it is quite confusing to have the same name here as the organism (also applies to the configs below), raising questions whether this needs to be the same and is somehow used for linking things (which are not the case). If we don't like `main`, can we find another term? E.g., `theOnlyReferenceGenome`, `referenceGenome1`, `mainReferenceGenome`?

fengelniederhammer: The key doesn't really matter for now. I don't know yet whether it will matter, but I could imagine that the logic becomes simpler if e.g. "is single pathogen" (which we will need in some places I assume) is equivalent to `"mainReferenceGenome" in referenceGenomes`.

Should I change them to `mainReferenceGenome`?
* changing the exact name later should be relatively easy
* if we notice that the key does no matter we can still silently remove the restriction. Going from specific to generic is not a breaking change (compared to the other way round when we suddenly need to enforce a specific name in those cases)

fengelniederhammer:
I checked that for CCHF `INSDC accession L: PV696553.1`, the ingested Loculus entry looks similar on Loculus main and on my branch (on the sequence detail page):
* I checked some the mutation badges. All of them were equal.
* The insertions and deletions are equal
* I checked some of the sequences, I didn't see any difference

fengelniederhammer:
Also submitting some dummy organism sequences works.

backend/src/test/resources/backend_config.json
theosanderson: Agree it doesn't matter much. I think for a single name my preference would be `singleReference`. `main` to me implies the existence of something else that is not main.

chaoran-chen: `singleReference` sounds great to me!

corneliusroemer: I feel like it's better to call what we're doing "allowing subOrganisms" rather than "multiple reference genomes". As a result I think we should use `singleSuborganism`

backend/src/main/kotlin/org/loculus/backend/service/submission/EmptyProcessedDataProvider.kt
```diff
@@ -11,15 +11,16 @@ import org.springframework.stereotype.Component
 class EmptyProcessedDataProvider(private val backendConfig: BackendConfig) {
     fun provide(organism: Organism): ProcessedData<GeneticSequence> {
         val (schema, referenceGenomes) = backendConfig.getInstanceConfig(organism)
+        val referenceGenome = referenceGenomes.values.first()
 
-        val nucleotideSequences = referenceGenomes.nucleotideSequences.map { it.name }.associateWith { null }
+        val nucleotideSequences = referenceGenome.nucleotideSequences.map { it.name }.associateWith { null }
```
corneliusroemer, line 16: I think this is confusing - rather than renaming what we call "referenceGenomes" we should iterate over "subOrganisms" higher up

backend/src/main/kotlin/org/loculus/backend/service/submission/CompressionService.kt
```diff
@@ -181,6 +181,7 @@ class CompressionService(private val backendConfig: BackendConfig) {
         backendConfig
             .getInstanceConfig(organism)
             .referenceGenomes
+            .values.first()
```
corneliusroemer, line 184:
```suggestion
            .subOrganisms
            .values.first()
            .referenceGenomes
```

kubernetes/loculus/templates/_common-metadata.tpl
```diff
@@ -371,8 +371,11 @@ organisms:
       {{- end }}
       {{- end }}
     referenceGenomes:
-      {{ $referenceGenomes:= include "loculus.generateReferenceGenome" $instance.referenceGenomes | fromYaml }}
-      {{ $referenceGenomes | toYaml |nindent 8}}
+      {{- range $subOrganismName, $referenceGenome := $instance.referenceGenomes }}
+      {{ $subOrganismName }}:
```
chaoran-chen, line 375: `subOrganism` or `suborganism`? In most of the other places, we now have `suborganism` / `Suborganism`.

backend/src/test/resources/backend_config.json
fengelniederhammer: As discussed: We stick with `singleReference` for now.

chaoran-chen (approved):
Looks great, thanks!

Issue #2893, opened 2024-09-27
Deposition: update accessions in external metadata as soon as available


corneliusroemer:
When we deposit to ENA we get accessions immediately, but we don't currently surface it until we have assembly accessions.

Related to #2430

anna-parker:
Bumping priority on this as: If we upload sequences as private we will not get the gca accession (only released when public) so the pipeline will wait until the sequence is public which is not the intention.

Issue #2430, opened 2024-08-14
feat(ena-submission, website): return bioproject and biosample as soon as they are known


anna-parker:
ENA submission: add some sort of notification to submitters when their accessions have returned, also return bioproject and biosample as soon as they are known so that users do not have to wait for the assembly accessions to return.

PR #4637, opened 2025-07-08
feat(deposition): upload new accessions incrementally

theosanderson:
PR by @anna-parker 

Resolves https://github.com/loculus-project/loculus/issues/2893, https://github.com/loculus-project/loculus/issues/2430

This code was started with codex
https://chatgpt.com/codex/tasks/task_e_686d2d12afb0832599bf0701616ddac2 but I since rewrote most of it. The submission was tested on a preview by uploading west-nile sequences and pointing the deposition pod to a test branch where I uploaded the cronjobs results.

Errors in external metadata update are logged correctly:
```
09:44:00     INFO (upload_external_metadata_to_loculus.py: 129) - Partial external metadata update failed for LOC_000PK9J: Error: 422 - {"type":"about:blank","title":"Unprocessable Entity","status":422,"detail":"Accession versions are not in one of the states [APPROVED_FOR_RELEASE]: LOC_000PK9J.1 - RECEIVED","instance":"/west-nile/submit-external-metadata"}
```
Correct partial update also logged well:
```
09:52:13    DEBUG (     call_loculus.py:  93) - Submitting external metadata for organism: west-nile, metadata: {'accession': 'LOC_000PK9J', 'version': 1, 'externalMetadata': {'bioprojectAccession': 'PRJEB94055', 'biosampleAccession': 'SAMEA131656770'}}
09:52:13    DEBUG (   connectionpool.py: 544) - https://backend-codex-deposit-accessions.loculus.org:443 "POST /west-nile/submit-external-metadata?externalMetadataUpdater=ena HTTP/1.1" 204 0
09:52:13    DEBUG (     call_loculus.py: 117) - External metadata submitted successfully for organism: west-nile, metadata: {'accession': 'LOC_000PK9J', 'version': 1, 'externalMetadata': {'bioprojectAccession': 'PRJEB94055', 'biosampleAccession': 'SAMEA131656770'}}
09:52:13     INFO (upload_external_metadata_to_loculus.py: 119) - Partial external metadata update for LOC_000PK9J succeeded
```
image

🚀 Preview: https://codex-deposit-accessions.loculus.org

PR #4719, opened 2025-07-18
feat(deposition): Upload external metadata as soon as available, not just when complete

corneliusroemer:
Resolves https://github.com/loculus-project/loculus/issues/2893, resolves https://github.com/loculus-project/loculus/issues/2430

This is a refactoring of #4637 that also adds incremental processing of external metadata to integration test: calling it after each step of simple submission to see it doesn't error and asserting on the exact data sent in the mock call.

### General point for future

Rather than polling the db if results have changed, it might make more sense to call external metadata update only when we actually got an update from ENA - it's silly to ask the db all the time whether something ha changed when we know exactly when something has changed.

We should really be triggering a call to update external metadata here:
https://github.com/loculus-project/loculus/blob/d47f4f3d4d628b49112dc0d568d3c004f93bd917/ena-submission/src/ena_deposition/create_project.py#L341-L350

Rather than having a separate "service" that keeps polling every 5 minutes creating a lot of db requests for nothing really. But that's for a different refactor I guess.

🚀 Preview: https://cr-eager-accession-upload.loculus.org

### DIFF
--- a/ena-submission/scripts/test_ena_submission_integration.py
+++ b/ena-submission/scripts/test_ena_submission_integration.py
@@ -522,8 +522,6 @@ def multi_segment_submission(
     assert payload["version"] == 1
     assert set(payload["externalMetadata"]) == {"bioprojectAccession"}
     assert payload["externalMetadata"]["bioprojectAccession"].startswith("PRJEB")
-    assert mock_submit_external_metadata.call_args[0][0]["accession"] == "LOC_0001TLY"
-    assert mock_submit_external_metadata.call_args[0][0]["version"] == 1
 
     _test_successful_sample_submission(db_config, config, sequences_to_upload)
     get_external_metadata_and_send_to_loculus(db_config, config)
--- a/ena-submission/scripts/test_ena_submission_integration.py
+++ b/ena-submission/scripts/test_ena_submission_integration.py
@@ -11,7 +11,6 @@ flyway -url=jdbc:postgresql://localhost:5432/loculus -schemas=ena_deposition_sch
 # ruff: noqa: S101 (allow asserts in tests))
 import json
 import logging
-import pprint
 import re
 import uuid
 from datetime import datetime, timedelta
@@ -495,7 +494,6 @@ def simple_submission(
     args = mock_submit_external_metadata.call_args_list
     assert len(args) == 3  # noqa: PLR2004
     payload = args[2][0][0]  # first positional argument of third call
-    logger.info(f"Payload: {pprint.pformat(payload)}")
     assert payload["accession"] == "LOC_0001TLY"
     assert payload["version"] == 1
     assert set(payload["externalMetadata"]) == {
--- a/ena-submission/scripts/test_ena_submission_integration.py
+++ b/ena-submission/scripts/test_ena_submission_integration.py
@@ -455,13 +455,19 @@ def simple_submission(
     mock_submit_external_metadata.return_value = mock_requests_post()
     sequences_to_upload: Final = get_sequences()
 
+    get_external_metadata_and_send_to_loculus(db_config, config)
+
     upload_sequences(db_config, sequences_to_upload)
     check_sequences_uploaded(db_config, sequences_to_upload)
+    get_external_metadata_and_send_to_loculus(db_config, config)
 
     _test_successful_project_submission(db_config, config, sequences_to_upload)
-    _test_successful_sample_submission(db_config, config, sequences_to_upload)
-    _test_successful_assembly_submission(db_config, config, sequences_to_upload)
+    get_external_metadata_and_send_to_loculus(db_config, config)
 
+    _test_successful_sample_submission(db_config, config, sequences_to_upload)
+    get_external_metadata_and_send_to_loculus(db_config, config)
+
+    _test_successful_assembly_submission(db_config, config, sequences_to_upload)
     get_external_metadata_and_send_to_loculus(db_config, config)
     check_sent_to_loculus(db_config, sequences_to_upload)
 
--- a/ena-submission/scripts/test_ena_submission_integration.py
+++ b/ena-submission/scripts/test_ena_submission_integration.py
@@ -11,6 +11,8 @@ flyway -url=jdbc:postgresql://localhost:5432/loculus -schemas=ena_deposition_sch
 # ruff: noqa: S101 (allow asserts in tests))
 import json
 import logging
+import pprint
+import re
 import uuid
 from datetime import datetime, timedelta
 from typing import Any, Final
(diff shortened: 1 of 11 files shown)


ena-submission/src/ena_deposition/upload_external_metadata_to_loculus.py
```diff
@@ -26,101 +26,202 @@
...
+
+    if gca := result.get("gca_accession"):
+        data["gcaAccession"] = gca
     else:
-        raise Exception
-    return data
+        all_present = False
+
+    segment_names = result.get("segment_order", [])
+    for segment in segment_names:
```
anna-parker, line 106: this will be incorrect if we upload a multi-segmented organism where only 1 segment was provided - my function handled this correctly before

ena-submission/src/ena_deposition/upload_external_metadata_to_loculus.py
```diff
@@ -26,101 +26,202 @@
...
+                        table_name=TableName.SUBMISSION_TABLE,
+                    )
+                except Exception as e:
+                    logger.exception(
+                        f"Failed to update status_all for {accession_version} to "
+                        f"{StatusAll.SENT_TO_LOCULUS}: {e}"
+                    )
 
 
 def upload_handle_errors(
```
anna-parker, line 227: your current rewrite never puts the database into this state so the error state so we never use this function and we never get a notification on slack about the issue

ena-submission/scripts/test_ena_submission_integration.py
```diff
@@ -445,15 +446,82 @@ def simple_submission(
...
+    args = mock_submit_external_metadata.call_args_list
+
+    assert len(args) == 1
+    payload = args[0][0][0]  # first positional argument of first call
+    assert payload["accession"] == "LOC_0001TLY"
+    assert payload["version"] == 1
+    assert set(payload["externalMetadata"]) == {"bioprojectAccession"}
+    assert payload["externalMetadata"]["bioprojectAccession"].startswith("PRJEB")
+    assert mock_submit_external_metadata.call_args[0][0]["accession"] == "LOC_0001TLY"
+    assert mock_submit_external_metadata.call_args[0][0]["version"] == 1
```
anna-parker, line 468: isn't this a repeat of:
assert payload["accession"] == "LOC_0001TLY"
assert payload["version"] == 1
```suggestion
```

anna-parker (changes requested):
Thanks for adding tests thats very cool!

This refactor introduces some issues that need to be resolved - if you would like we can add the tests and rename the functions of the last PR - that PR doesn't have these issues (and it also raises the same errors as here with one exception - I did not add a try catch for it `result` is a column in the db. However, this case shouldn't happen and should immediately cause the system to fail as it would mean the database is in an incorrect state. So I think we can merge that PR as bit faster)

ena-submission/src/ena_deposition/upload_external_metadata_to_loculus.py
anna-parker: Lets add the same TODO here that we have in another spot in the deposition pod:
```
# TODO: Also pass the full segment list from config so we can handle someone submitting
# a multi-segmented virus that has a main segment
```

ena-submission/scripts/test_ena_submission_integration.py
```diff
@@ -297,20 +322,22 @@ def check_project_submission_has_errors(
...
-    db_config: SimpleConnectionPool, sequences_to_upload: dict[str, Any]
+    db_config: SimpleConnectionPool, sequences_to_upload: dict[str, Any], single_segment: bool
 ) -> None:
     """Sets erz-accession to known public accessions"""
     for full_accession, data in sequences_to_upload.items():
         accession, version = full_accession.split(".")
         organism = data["organism"]
         if organism == "cchf":
+            segment_order = ["L"] if single_segment else ["L", "M"]
+            erz_accession = "ERZ24985816" if single_segment else "ERZ24784470"
```
anna-parker, line 333: Should we use `ERZ24908522` here as well as?

anna-parker: I'm confused - this doesn't return a GCA? maybe we should add a comment here

ena-submission/src/ena_deposition/upload_external_metadata_to_loculus.py
```diff
@@ -26,101 +26,202 @@
...
+                            "status_all": StatusAll.SENT_TO_LOCULUS,
+                            "finished_at": datetime.now(tz=pytz.utc),
+                        },
+                        table_name=TableName.SUBMISSION_TABLE,
+                    )
+                except Exception as e:
+                    logger.exception(
+                        f"Failed to update status_all for {accession_version} to "
+                        f"{StatusAll.SENT_TO_LOCULUS}: {e}"
+                    )
```
anna-parker, line 224:
```suggestion
                    )
                    update_with_retry(
                        db_config,
                        conditions=seq_key,
                        update_values={
                            "status_all": StatusAll.HAS_ERRORS_EXT_METADATA_UPLOAD,
                            "finished_at": datetime.now(tz=pytz.utc),
                        },
                        table_name=TableName.SUBMISSION_TABLE,
                    )
```

kubernetes/loculus/values.yaml
```diff
@@ -1756,6 +1756,7 @@ enaDeposition:
   enaDbName: Loculus
   enaUniqueSuffix: Loculus
   enaIsBroker: False
+  enaGithubTestUrl: "https://raw.githubusercontent.com/pathoplexus/ena-submission/test_branch/test/approved_ena_submission_list.json"
```
anna-parker, line 1759: Revert before merge:
```suggestion
```

ena-submission/scripts/test_ena_submission_integration.py
corneliusroemer: Yeah I found only one single-segment submission we've ever done (on the non-broker account) and that one doesn't have a GCA!

ena-submission/scripts/test_ena_submission_integration.py
corneliusroemer: indeed

ena-submission/src/ena_deposition/upload_external_metadata_to_loculus.py
corneliusroemer: Meh, I'm not sure we will ever do that? ENA is just for us and we just won't name it main?

ena-submission/src/ena_deposition/upload_external_metadata_to_loculus.py
corneliusroemer: I'm not sure how important this is as we'll notice otherwise but since we've already got the code :p

anna-parker: How will we notice?

ena-submission/scripts/test_ena_submission_integration.py
anna-parker: You can use `ERZ24908522` (it also goes back to your testing and should return one segment) - but as this tests the WAITING case its probably good to do this - maybe just add a comment :-)

corneliusroemer: Wait, ERZ24908522 never had an assembly submitted? I can't see the ERZ at all in processed.

anna-parker: oh it appears this is an example we submitted on the broker account... which is weird as the tests are definitely on the test account

anna-parker (approved):
Amazing thanks!
image

ena-submission/scripts/test_ena_submission_integration.py
anna-parker: omg... the tests are only checking submission of cchf... well this is great that we figured this out!!! @corneliusroemer thanks so much for flagging this!!!!

Issue #4607, opened 2025-07-04
Keycloak internal server error instead of login page


corneliusroemer:
Possibly related to vite upgrade to v7

Worrying that no integration tests noticed this

<img width="1043" height="419" alt="Image" src="https://github.com/user-attachments/assets/6ceb27af-7cda-45f3-b7b9-f0196b472075" />

corneliusroemer:
Recovered on its own - false alarm?

<img width="840" height="599" alt="Image" src="https://github.com/user-attachments/assets/1d6a7852-db14-42ee-9a3d-1498f6f13229" />

anna-parker:
From discussion with @theosanderson this is likely due to https://github.com/loculus-project/loculus/pull/4326 - we should figure out how to restart on previews

PR #4720, opened 2025-07-21
fix(deployment): Add LOCULUS_VERSION to keycloak to restart it on deployment

theosanderson:
resolves https://github.com/loculus-project/loculus/issues/4607
Adds something that will restart keycloak when loculus is bumped in the case that we are running development databases, for Loculus preview deployments.

🚀 Preview: https://restart-kc.loculus.org


chaoran-chen (approved):
Thanks! (I can't judge whether this solves the problem but we will see whether it occurs again.)

Issue #3683, opened 2025-02-13
Clearer error message when sequences submitted without segment identifiers for segmented virus


theosanderson:
Apparently one gets an error message like this:

![Image](https://github.com/user-attachments/assets/d0b6e873-fdc0-47be-bef6-ece9af9a9778)

One should get a message like. `Error: some sequence headers do not contain segment identifiers. Sequence headers should end with a segment identifier, e.g. "mysample_S", "mysample_L", etc., whereas you entered: `

https://loculus.slack.com/archives/C05G172HL6L/p1739441563468129

fhennig:
Just spotted this as 'Ready to work' but I'm not super sure how I'd do this, @anna-parker maybe you can write down some pointers or even take it over onto your list instead? 😅

fhennig:
Ok, so, I have some updates here.

> Clearer error message when sequences submitted **without segment identifiers** for segmented virus

_Actually_, the submission IDs in the screenshot _do_ contain a segment identifier, because whatever comes after the last underscore becomes the segment identifier. so `test_NIH19` becomes isolate `test` with segment `NIH19`.

I thought "well, let's make sure that the segment is actually one of the configured segments, and I did that in #3769 and that works! _But_ now Chaoran told me that it is actually by design that we do not enforce the segment names, because the pipeline is supposed to handle this. Analogous to how we allow submitting arbitrary metadata fields and let the pipeline handle it.

So, actually, the error message is correct and "Error: some sequence headers do not contain segment identifiers" wouldn't be right, all the submission IDs do in fact contain a segment name.

I don't see an obvious way to fix this, so I thought I'd ask for input again!

@theosanderson @chaoran-chen @anna-parker

theosanderson:
Hmm thanks!

I don't have a problem in principle with the idea of the pipeline handling it. It's just that here the backend does do logic that assumes that these are formatted as id_segment things - and here there is a bug because they are not, and it would be nice for us to give a more useful error.

The question would be what the use-case that is prompting us to give the pipeline flexibility to do this.

One idea would be that we would want to support bacterial contigs (https://github.com/loculus-project/loculus/issues/3539) and they might be: `sample1_1`, `sample1_2` to any arbitrary number. But atm we don't support those, and once we do we would have a specific organism configuration for them which would allow us to turn off this check. I don't think I atm see a compelling reason not to move this logic into the backend.

chaoran-chen:
[...]

PR #4721, opened 2025-07-21
feat(backend): remove validation of segments in fasta headers to allow submission of multi pathogen sequence files

### LINKED ISSUE #4708: Multi pathogen support: backend submission
The submission flow for single pathogen organisms should remain the same.

For multi pathogen:

Metadata: as usual, nothing special here.

Sequences: We don't want the user to be forced to put the segment name here.

```
>key1
ACTG
```

For multiple segments:

```
>key1_segment1
ACTG
>key1_segment2
GTCA
```

fengelniederhammer:
resolves #4708

### PR Checklist
  - IMO no extra manual testing is required, it's already covered by the automated backend tests and multi pathogen organisms don't work e2e yet.


backend/src/main/kotlin/org/loculus/backend/utils/ParseFastaHeader.kt
fengelniederhammer, line 1: There is probably docs on the endpoint that need be changed.

Are there any other docs that mention how to submit multi segmented?

backend/src/test/kotlin/org/loculus/backend/controller/submission/SubmitEndpointTest.kt
```diff
@@ -346,6 +319,123 @@ class SubmitEndpointTest(
...
+            .andExpect(status().isOk)
+            .andExpect(content().contentType(APPLICATION_JSON))
+            .andExpect(jsonPath("\$.length()").value(NUMBER_OF_SEQUENCES))
+            .andExpect(jsonPath("\$[0].submissionId").value("custom0"))
+            .andExpect(jsonPath("\$[0].accession", containsString(backendConfig.accessionPrefix)))
+            .andExpect(jsonPath("\$[0].version").value(1))
+    }
+
+    @Test
+    fun `GIVEN submission for multi pathogen organism with segment id THEN is rejected`() {
```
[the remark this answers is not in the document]
fengelniederhammer: I added a test that shows that multi segmented multi pathogen works.

backend/src/main/kotlin/org/loculus/backend/utils/ParseFastaHeader.kt
chaoran-chen: You could add a paragraph [here](https://github.com/loculus-project/loculus/blob/d2f8f66889d02ce612cf6166ef0f55c408de01c5/preprocessing/specification.md#pulling-unpreprocessed-data) to explain that segments are not validated but that the pipeline gets everything that the user submitted (on purpose).

chaoran-chen, line 1: This is bringing back #3683. A coulple months ago, due to that issue, we decided to validate the segments, and now that we're removing it again, the issue is back. This can be reproduced by submitting to CCHF the following files:

Metadata:

```
id	geoLocCountry	sampleCollectionDate
seq_001	Germany	2025-02-02
```

FASTA:

```
>seq_001
AATTGGCC
```

With this change, the error message is again:

image

On main, we would get this message:

image

anna-parker: I do believe its a requirement unless the backend starts handling multi-pathogens differently than multi-segmented organisms, in which case the FASTA parser could be changed accordingly. 

I think another alternative is to add a better error message in `validateSubmissionIdSetsForConsensusSequences` - but Im always happy to wait until a feature has been shown to work until it is merged

fengelniederhammer: > the backend starts handling multi-pathogens differently than multi-segmented organisms

Well, that's what this PR is about. That's what I implemented in this PR and in #4692. Or did I understand wrong what should be done for the multi-path feature? (That's why I started #4641 to get a common understanding how things should work as I wasn't involved in all the preliminary work)

fengelniederhammer: > in validation

Just to be sure: you mean the validation of the preprocessing results in the backend?

anna-parker: umm no I meant the fasta header validation/parser and validateSubmissionIdSetsForConsensusSequences function

fengelniederhammer: As discussed with @anna-parker and @chaoran-chen:
* We'll revert this so that we still include the validation of the fasta headers in the backend (so that we don't run into #3683 again)
* We can relax this somewhen later (it would be better to first adapt the preprocessing pipeline to deal with arbitrary key in the sequences (rather than the segment name). Then the preprocessing pipeline can also take over the task of segment assignment. Then in a next step, we can remove the validation here in the backend. (Those changes are rather unrelated to the multi pathogen feature)
* Anya mentioned that even for single segmented organisms, people might want to upload multiple segments in case someone was infected with multiple suborganisms. We'll have to keep that in mind and probably revisit this anyway once we want to support that.

fengelniederhammer: I think this is not relevant anymore. Feel free to reopen this if you think otherwise.

backend/src/test/kotlin/org/loculus/backend/controller/submission/ExtractUnprocessedDataEndpointTest.kt
```diff
@@ -279,3 +283,122 @@ class ExtractUnprocessedDataEndpointTest(
...
+
+        assertThat(responseBody.size, `is`(1))
+        assertThat(
+            responseBody[0].data.unalignedNucleotideSequences,
+            `is`(mapOf("main" to "A")),
+        )
+    }
+
+    @Test
+    fun `GIVEN multi segmented multi pathogen submission THEN the segment names are the keys of the sequences`() {
```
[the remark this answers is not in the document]
fengelniederhammer: Does that case exist?

What should we do in that case here? Say that if `main` is a known segment, then accept it without a segment name?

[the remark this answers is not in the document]
anna-parker: The rev-seq co-infection example includes influenza, rsv and enteroviruses - so some are segmented and some are not. We could say that this is out of scope for now - but did want to bring this up as it is an immediate use case for the multi-path feature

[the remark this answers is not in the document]
fengelniederhammer: Ok, should I create a subtask of #3984 to revisit this later?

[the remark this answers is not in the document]
anna-parker: > What should we do in that case here? Say that if main is a known segment, then accept it without a segment name?

One way to solve it - we could also move the check. We can also tackle this later - but we should create an issue for this specific case

PR #4723, opened 2025-07-21
test(deposition): add single-segment organism west-nile to tests

anna-parker:
resolves issue found in https://github.com/loculus-project/loculus/pull/4719#discussion_r2218374556 - west-nile was accidentally removed from the ENA workflow tests

### PR Checklist


corneliusroemer:
> Uggh this is failing because the single-segmented ERZ on the dev account (ERZ24985816) doesn't have a GCA

Yeah I noticed this when I looked into that accession. Shall we make a real submission of a single segment with our test account (using dummy content to make it clear it's just for test purposes) so we have one that we can use for tests?

This can't really be merged as is no?

PR #4747, opened 2025-07-22
Garbage collect outdated prepro entries

corneliusroemer:
- **draft**
- **Add test**
- **expand test a bit**
- **test fix**
- **make things persistent**
- **Adapt test to take organism into account**
- **Should work now**
- **add some comments to the test**
- **Add v3 for ebola sudan (for testing)**
- **undo testing code**
- **Batch delete, add log, use exposed**

resolves #

### PR Checklist


backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -1272,6 +1275,94 @@ class SubmissionDatabaseService(
...
+     * Garbage collects outdated preprocessed entries for all organisms.
+     * This is a convenience method that iterates over all distinct organisms in the database
+     * and calls [cleanUpOutdatedPreprocessingData] for each organism.
+     */
+    fun cleanUpOutdatedPreprocessingDataForAllOrganisms() {
+        CurrentProcessingPipelineTable.selectAll()
+            .forEach { row ->
+                cleanUpOutdatedPreprocessingData(
+                    organism = row[CurrentProcessingPipelineTable.organismColumn],
+                    earliestVersionToKeep = row[CurrentProcessingPipelineTable.versionColumn],
```
anna-parker, line 1358: In https://github.com/loculus-project/loculus/pull/4584 we only deleted versions that were less than the currentVersion -1, here we now delete any version less than the currentVersion - could this ever be an issue?

If not lets rename the variable `earliestVersionToKeep` to `currentPipelineVersion` to be more explicit.

kubernetes/loculus/templates/loculus-database-standin.yaml
```diff
@@ -5,7 +5,7 @@ kind: Deployment
 metadata:
   name: loculus-database
   annotations:
-    argocd.argoproj.io/sync-options: Replace=true
+    argocd.argoproj.io/sync-options: Replace=false
```
anna-parker, line 8: here as well 
```suggestion
    argocd.argoproj.io/sync-options: Replace=true
```

kubernetes/loculus/values.yaml
```diff
@@ -1752,7 +1759,7 @@ secrets:
 runDevelopmentKeycloakDatabase: true
 runDevelopmentMainDatabase: true
 runDevelopmentS3: true
-developmentDatabasePersistence: false
+developmentDatabasePersistence: true
```
anna-parker, line 1762:
```suggestion
developmentDatabasePersistence: false
```

kubernetes/loculus/values.yaml
```diff
@@ -1256,6 +1256,13 @@ defaultOrganisms:
           genes: [NP, VP35, VP40, GP, sGP, ssGP, VP30, VP24, L]
           nextclade_dataset_name: nextstrain/ebola/sudan
           nextclade_dataset_server: https://raw.githubusercontent.com/nextstrain/nextclade_data/ebola/data_output
+      - <<: *preprocessing
+        version: 3
+        configFile:
+          <<: *preprocessingConfigFile
+          genes: [NP, VP35, VP40, GP, sGP, ssGP, VP30, VP24, L]
+          nextclade_dataset_name: nextstrain/ebola/sudan
+          nextclade_dataset_server: https://raw.githubusercontent.com/nextstrain/nextclade_data/ebola/data_output
```
anna-parker, line 1265:
```suggestion
```

backend/src/test/kotlin/org/loculus/backend/service/submission/UseNewerProcessingPipelineVersionTaskTest.kt
```diff
@@ -106,4 +108,58 @@ class UseNewerProcessingPipelineVersionTaskTest(
...
+        convenienceClient.submitProcessedData(processedData, pipelineVersion = 1)
+        useNewerProcessingPipelineVersionTask.task()
+
+        convenienceClient.extractUnprocessedData(pipelineVersion = 2)
+        convenienceClient.submitProcessedData(processedData, pipelineVersion = 2)
+        useNewerProcessingPipelineVersionTask.task()
+
+        transaction {
+            assertThat(getExistingPipelineVersions(DEFAULT_ORGANISM), `is`(listOf(2L)))
+            assertThat(getExistingPipelineVersions(OTHER_ORGANISM), `is`(listOf(1L)))
```
anna-parker, line 135: sorry I don't quite understand why  OTHER_ORGANISM is not being updated to a higher version with the `useNewerProcessingPipelineVersionTask.task()` statement

corneliusroemer: Good question, this test comes from Felix.

My assumption is that the convenience client submits for the DEFAULT_ORGANISM, not other organism, hence the version is always still 1 for other organism. It won't be bumped as there's nothing to be bumped for other organism? Makes sense?

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
corneliusroemer: we don't keep `-1` around anymore in this version, see PR description for reasoning:
> Clean up all outdated prepro versions (deciding not to keep the latest unused one around as it creates a lot of db load to have twice the number of versions around)

backend/src/test/kotlin/org/loculus/backend/service/submission/UseNewerProcessingPipelineVersionTaskTest.kt
anna-parker: True the default organism is DEFAULT_ORGANISM - maybe we could add a comment?

anna-parker (approved):
Looks good to me - did you check in the db that the deletion worked?

PR #4748, opened 2025-07-22
refactor!(website): enable extra sequence-related nav items in Pathoplexus

chaoran-chen:
This refactoring is needed for https://github.com/pathoplexus/pathoplexus/pull/590 so that the top navigation for tools can be organism-specific.

### PR Checklist
- ~[ ] The implemented feature is covered by appropriate, automated tests.~
- ~[ ] Any manual testing that has been done is documented (i.e. what exactly was tested?)~

🚀 Preview: https://extra-sequence-related-na.loculus.org

### DIFF
--- a/website/src/routes/extraTopNavigationItems.ts
+++ b/website/src/routes/extraTopNavigationItems.ts
@@ -1,1 +1,8 @@
-export const extraTopNavigationItems = [];
+// The following code is in a separate file so that it can be overwritten by Pathoplexus.
+import type { TopNavigationItems } from './navigationItems.ts';
+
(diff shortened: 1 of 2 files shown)


theosanderson:
I guess this is mostly consistent with Browse and Submit - apart from the fact that when accessed from the front page those prompt for the specific organism. Possibly we can use both of our approaches.

chaoran-chen:
I'm also happy to go with your solution. The pathogen-specific tool pages mght be a little less findable but that would be fine in my opinion if we have a page that shows all tools.

PR #4570, opened 2025-07-01
EVs multi-pathogen

alejandra-gonzalezsanchez:
Adding EVs as segments for testing

🚀 Preview: Add `preview` label to enable

PR #4753, opened 2025-07-24
feat(kubernetes): merge multi pathogen reference genomes for LAPIS

### LINKED ISSUE #4710: Multi pathogen support: merge reference genomes for LAPIS config
Instead of taking only the first reference genome, the Helm chart should merge them, if there are multiple ones.

fengelniederhammer:
resolves #4710

### PR Checklist


website/public/images/organisms/enterovirus.jpg
corneliusroemer, line 1: Should be minified, 500kB is unnecessarily large

kubernetes/loculus/values.yaml
```diff
@@ -1663,6 +1663,233 @@ defaultOrganisms:
...
+          - name: 2C
+            sequence: "[[URL:https://raw.githubusercontent.com/alejandra-gonzalezsanchez/loculus-evs/master/enterovirus/evd68/2C-evd68.fasta]]"
+          - name: 3A
+            sequence: "[[URL:https://raw.githubusercontent.com/alejandra-gonzalezsanchez/loculus-evs/master/enterovirus/evd68/3A-evd68.fasta]]"
+          - name: 3B
+            sequence: "[[URL:https://raw.githubusercontent.com/alejandra-gonzalezsanchez/loculus-evs/master/enterovirus/evd68/3B-evd68.fasta]]"
+          - name: 3C
+            sequence: "[[URL:https://raw.githubusercontent.com/alejandra-gonzalezsanchez/loculus-evs/master/enterovirus/evd68/3C-evd68.fasta]]"
+          - name: 3D
+            sequence: "[[URL:https://raw.githubusercontent.com/alejandra-gonzalezsanchez/loculus-evs/master/enterovirus/evd68/3D-evd68.fasta]]"
```
corneliusroemer, line 1892: We shouldn't use raw.githubusercontent.com due to likelihood of getting rate limited -> port to Github pages

kubernetes/loculus/values.yaml
```diff
@@ -1663,6 +1663,233 @@ defaultOrganisms:
...
+          method: "minimizer"
+          minimizer_index: "https://raw.githubusercontent.com/alejandra-gonzalezsanchez/loculus-evs/master/evs_minimizer-index.json"
+          minimizer_parser:
+            - segment
+        #  method: "align"
+        #  nextclade_dataset_name: community/hodcroftlab/enterovirus/enterovirus/linked
+        #  nextclade_dataset_server: https://raw.githubusercontent.com/alejandra-gonzalezsanchez/nextclade_data/multi-pathogen-evs/data_output
+    enaDeposition:
+      configFile:
+        taxon_id: 12059 # All EVs A = 138948
```
corneliusroemer, line 1777: Doesn't work like this, needs big red warning

corneliusroemer (commented):
For completeness, might make sense to add a multi-segmented suborganism - otherwise the multi-segment path is entirely dead/untested for now and we might as well not have it

corneliusroemer:
> I was planning to give it both, the merged and the original version. It needs the merged version for that validation. So either we do that in the backend and duplicate the logic or we do it in the Helm chart which bloats the config file. I'd rather do it in the Helm chart since keeping the logic in one place seems more important to me.

I don't understand this. What do we need to validate and why? Helm is horrible, let's do as little in it as possible.

anna-parker:
Hey it doesn't look like enteroviruses are working on main, they are working on @alejandra-gonzalezsanchez's PR: https://github.com/loculus-project/loculus/pull/4570 (where the config was copied from from) - so it looks like sth is wrong with the config changes here - can we revert this PR until this is working?

theosanderson:
As I understood it it was expected that this wasn't working yet

anna-parker:
Hey @fengelniederhammer I now fixed the ingest error in https://github.com/loculus-project/loculus/pull/4767 but submission to the backend is throwing an odd error:
```
URL: https://backend-small-ingest-fixes.loculus.org/enteroviruses/submit
Method: POST
Status Code: 422
Response Content: {"type":"about:blank","title":"Unprocessable Entity","status":422,"detail":"Metadata file contains 9999 ids that are not present in the sequence file: AB037250.1.EV-A71, AB037251.1.EV-A71
```
is this expected? this looks odd to me

anna-parker:
@theosanderson and @fengelniederhammer I don't think the changes in https://github.com/loculus-project/loculus/pull/4721 will be able to solve this issue - ingest is submitting the enteroviruses as segmented but the validation will now enforce the submission of them as single-segmented.

There are multiple solutions to this:
1. have ingest now just submit all enteroviruses and not use nextclade sort - prepro can then assign the clades - this means a config change
2. remove the strict submissionId validation as I described in https://github.com/loculus-project/loculus/issues/4759#issuecomment-3132742765

theosanderson:
Given we want users to not have to specify what suborganism stuff is from I guess we want option 1?

fengelniederhammer:
> > Enteroviruses are single segmented.
> 
> Yes - but the config that you copied is telling ingest to submit them each as an individual segment - we did this to show what changes would still be required:

Ah I see. Then we should change that (As I said I didn't expect that this would work out of the box, but rather that we can continue to work on it until it works).  There is not much docs on how to configure the ingest so it's difficult to understand what that config actually means.

Issue #4745, opened 2025-07-22
Use sentence case, not title case


theosanderson:
It's a bit of a nitpick, but it's good to be consistent with using either Sentence case or Title Case.

IMO we should always use sentence case for all buttons, and IMO for all titles.

Personally I just think it looks better, and is more readable, but for a more objective view if you look at other sites like GitHub and Google they tend to do this (if someone has a notable exception do let me know!)

<img width="335" height="99" alt="Image" src="https://github.com/user-attachments/assets/82f314b2-d13c-4d14-8a52-b99f5555faea" />

<img width="1093" height="396" alt="Image" src="https://github.com/user-attachments/assets/5a22bfb7-033b-4f50-86ee-bfdeb7e20b22" />

chaoran-chen:
> if someone has a notable exception do let me know!

genspectrum.org :P (-> I have a small preference for sentence case but am also entirely fine to go with sentence case :))

https://www.apple.com/ has a bit of a mix. On buttons, they use sentence case (e.g., "Learn more") but in their navigations, theu use title case. 

<details><summary>apple screenshots</summary>

<img width="937" height="492" alt="Image" src="https://github.com/user-attachments/assets/2bc3b15b-3cb4-43c6-88a2-84d9c1a3a967" />

<img width="968" height="497" alt="Image" src="https://github.com/user-attachments/assets/2f050446-5518-4956-a397-7e12b2980d55" />
</details>

theosanderson:
Interesting point on Apple: I suspect it may partly be because they also have lots of product names and want to avoid the inconsistency that results. Once you get into something like https://developer.apple.com/documentation/Xcode/writing-code-with-intelligence-in-xcode they use sentence case for titles too

PR #4754, opened 2025-07-24
fix(website): Standardize towards sentence case 

theosanderson:
copy of https://github.com/loculus-project/loculus/pull/4746 with a branch name that can be previewed

Moves closer to consistency on sentence case re: https://github.com/loculus-project/loculus/issues/4745

- convert button texts and page titles to sentence case
- update UI headings in docs and dialogs
- adjust tests to match new sentence case labels

------
https://chatgpt.com/codex/tasks/task_e_687fabd313f883259f2aecd52dcc4c5c

🚀 Preview: https://codex-standardize-buttons.loculus.org

### DIFF
--- a/website/src/components/SearchPage/DownloadDialog/LinkOutMenu.tsx
+++ b/website/src/components/SearchPage/DownloadDialog/LinkOutMenu.tsx
@@ -189,7 +189,7 @@ function LinkOutMenuDataUseTermModal(props: {
                         className='px-4 py-2 bg-primary-600 text-white rounded-md font-medium hover:bg-primary-700 transition-colors'
                         onClick={props.onClick1}
                     >
-                        Include Restricted-Use
+                        Include restricted-use
                     </button>
                 </div>
             </div>
--- a/website/src/components/SequenceDetailsPage/DataUseTermsHistoryModal.tsx
+++ b/website/src/components/SequenceDetailsPage/DataUseTermsHistoryModal.tsx
@@ -45,14 +45,14 @@ const DataUseTermsHistoryDialog: FC<DataUseTermsHistoryContainerProps> = ({ data
                 <button className='btn btn-sm btn-circle btn-ghost absolute right-2 top-2'>✕</button>
             </form>
 
-            <h3 className='font-bold text-lg'>Data Use Terms History</h3>
+            <h3 className='font-bold text-lg'>Data use terms history</h3>
 
             <table className='table'>
                 <thead>
                     <tr>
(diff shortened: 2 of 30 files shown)


anna-parker (approved):
Think I still prefer title case but I see that sentence is more standard

website/src/components/SequenceDetailsPage/DataUseTermsHistoryModal.tsx
```diff
@@ -45,14 +45,14 @@ const DataUseTermsHistoryDialog: FC<DataUseTermsHistoryContainerProps> = ({ data
...
-            <h3 className='font-bold text-lg'>Data Use Terms History</h3>
+            <h3 className='font-bold text-lg'>Data use terms history</h3>
 
             <table className='table'>
                 <thead>
                     <tr>
                         <th>Changed</th>
                         <th>User</th>
-                        <th>Data Use Terms</th>
+                        <th>Data use terms</th>
```
chaoran-chen, line 55: Should this be changed or do we consider the DUTs as a special name? In the case above, this would mean "Data Use Terms history"

website/src/components/SearchPage/DownloadDialog/LinkOutMenu.tsx
```diff
@@ -189,7 +189,7 @@ function LinkOutMenuDataUseTermModal(props: {
                         className='px-4 py-2 bg-primary-600 text-white rounded-md font-medium hover:bg-primary-700 transition-colors'
                         onClick={props.onClick1}
                     >
-                        Include Restricted-Use
+                        Include restricted-use
```
chaoran-chen, line 192: In Pathoplexus DUT (https://pathoplexus.org/about/terms-of-use/data-use-terms), we write for example "governing how Restricted-Use Data can be used and shared onward, as well as how it must be attributed when used", i.e., specifically capitalizing Restricted-Use. Should we do it here as well?

website/src/components/SequenceDetailsPage/DataUseTermsHistoryModal.tsx
theosanderson: I don't see a reason to consider them a special name here - I think one can make more of a case for "Open Data Use Terms" and "Resticted Data Use Terms" being special things as they are specific instances of data use terms.

chaoran-chen: okay, sounds good to me :)

Issue #3394, opened 2024-12-06
Streaming/Parallelization error when updating the preprocessing version


anna-parker:
See details here: https://github.com/loculus-project/loculus/pull/3393#issuecomment-2523326844

I checked the grafana logs and there were indeed two pods running with v2 that were both receiving these errors, my branch was up-to-date with main so it seems that https://github.com/loculus-project/loculus/pull/3073 was not successful in preventing this race-condition. We should investigate more and figure out how best to test this.

Issue #1882, opened 2024-05-12
E2E tests put backend into state that errors with `ERROR: duplicate key value violates unique constraint "sequence_entries_preprocessed_data_pkey"``


corneliusroemer:
I've been investigating flaky E2E tests, improving the error messages to see more what's going on (see https://github.com/loculus-project/loculus/pull/1881/commits/0f1d4a2a91af1251847e3f0a42acc99361959328)

When I properly logged full backend errors, I noticed the following one (part of the "allow bulk deletion test"):

```
An unexpected error occurred while streaming, aborting the stream: java.sql.BatchUpdateException: Batch entry 0 INSERT INTO sequence_entries_preprocessed_data (accession, pipeline_version, processing_status, started_processing_at, "version") VALUES (('LOC_000Q6CP'), ('1'::int8), ('IN_PROCESSING'), ('2024-05-12 15:24:03.37018+00'::timestamp), ('2'::int8)) was aborted: ERROR: duplicate key value violates unique constraint "sequence_entries_preprocessed_data_pkey"
      Detail: Key (accession, version, pipeline_version)=(LOC_000Q6CP, 2, 1) already exists.  Call getNextException to see other errors in the batch.
```

I'm not sure if this is just a test error, or this shows that our backend can end up in a bad state that we haven't noticed yet in manual testing. Would be good to investigate at some point. As far as I'm aware, the E2E tests don't have any special privileges with regards to the backend, so such a state could also be entered by real users.

To reproduce:

1. Fire up an E2E cluster
2. Run `npx playwright test "pages/review/index.spec.ts" --repeat-each=10`
3. Observe

It seems enough to just run lots of review tests in parallel to trigger this state. I know that the test will fail due to race conditions between tests (they just look at total counts and are not isolated from each other). However that doesn't mean that simultaneous uploading should produce an illegal backend state.

<img width="2553" alt="image" src="https://github.com/loculus-project/loculus/assets/25161793/54a880e7-3e17-45a5-8d52-7ae342ddd089">

PR #4756, opened 2025-07-25
fix(backend): handle duplicate selects in extract-unprocessed gracefully, no longer stream results

corneliusroemer:
resolves https://github.com/loculus-project/loculus/issues/1882
resolves https://github.com/loculus-project/loculus/issues/3394

In extract-unprocessed, when multiple clients call at the same time, it can happen that they both select the same rows which then causes an insertion error due to duplicate primary keys.

The more parallel prepros we add the more often the errors happen. So this blocks scalability of preprocessing: we're stuck with slow preprocessing: right now mpox takes ~3hr. For SC2 it'd be unacceptably slow.

This PR:
1. Adds a failing test, provoking the error
1. Switches the endpoint to non-streaming to handle such errors gracefully by returning nothing instead of garbled ndjson.

### Screenshot

These errors should no longer happen with these changes:
image

### PR Checklist

### DIFF
--- a/backend/src/test/kotlin/org/loculus/backend/controller/submission/ExtractUnprocessedDataEndpointTest.kt
+++ b/backend/src/test/kotlin/org/loculus/backend/controller/submission/ExtractUnprocessedDataEndpointTest.kt
@@ -83,7 +83,7 @@ class ExtractUnprocessedDataEndpointTest(
                 startLatch.await() // Wait for signal to start
                 val response = client.extractUnprocessedData(
                     numberOfSequenceEntries = DefaultFiles.NUMBER_OF_SEQUENCES,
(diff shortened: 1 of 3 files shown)


backend/src/test/kotlin/org/loculus/backend/controller/submission/ExtractUnprocessedDataEndpointTest.kt
```diff
@@ -46,6 +46,9 @@ import java.net.URI
 import java.net.http.HttpClient
 import java.net.http.HttpRequest
 import java.net.http.HttpResponse
+import java.util.concurrent.ConcurrentHashMap
```
anna-parker, line 49: I think this isn't needed?
```suggestion
```

PR #4760, opened 2025-07-29
feat(backend): validate the preprocessing results against the merged reference genome

### LINKED ISSUE #4709: Multi pathogen support: backend validation of preprocessing results
To be sketched out properly. We'll have to adapt the validation logic that checks that all expected sequences are there.

fengelniederhammer:
resolves #4709

Suppose you have reference genomes that look something like
```json
{
    "firstSuborganism": {
        "nucleotideSequences": [
            {"name": "main", "sequence": "..."}
        ],
        "genes": [
            {"name": "someLongGene", "sequence": "..."},
            {"name": "someShortGene", "sequence": "..."}
        ]
    },
    "secondSuborganism": {
        "nucleotideSequences": [
            {"name": "main", "sequence": "..."}
        ],
        "genes": [
            {"name": "someLongGene", "sequence": "..."},
            {"name": "anotherShortGene", "sequence": "..."}
        ]
    }
}
```

then SILO will get the merged reference genome:
```json
{
    "nucleotideSequences": [
        {"name": "firstSuborganism", "sequence": "..."},
        {"name": "secondSuborganism", "sequence": "..."}
    ],
    "genes": [
        {"name": "firstSuborganism-someLongGene", "sequence": "..."},
        {"name": "firstSuborganism-someShortGene", "sequence": "..."},
        {"name": "secondSuborganism-someLongGene", "sequence": "..."},
        {"name": "secondSuborganism-anotherShortGene", "sequence": "..."}
    ]
}
```

The preprocessing result must match the input that SILO will get. Thus, the backend must make sure that the preprocessing results conforms to the merged reference genome.

Since we already generate the merged version in the Helm chart, I figured that it was easier to also add it to the backend config there. 

The alternative would have been to duplicate the logic how to merge them. The price that we pay is that it bloats the backend config.

The next step would be to check whether the submission flow works end-to-end in the backend.

### PR Checklist
  - is there anything that we need to update?

### DIFF
--- a/backend/src/test/kotlin/org/loculus/backend/service/submission/EmptyProcessedDataProviderTest.kt
+++ b/backend/src/test/kotlin/org/loculus/backend/service/submission/EmptyProcessedDataProviderTest.kt
@@ -49,7 +49,7 @@ class EmptyProcessedDataProviderTest {
                     referenceGenomes = mapOf(
                         SINGLE_REFERENCE_GENOME_KEY to referenceGenome,
                     ),
-                    mergedReferenceGenome = referenceGenome, // TODO ?
+                    mergedReferenceGenome = referenceGenome,
                 ),
             ),
             dataUseTerms = DataUseTerms(true, null),


backend/src/main/kotlin/org/loculus/backend/config/Config.kt
```diff
@@ -47,7 +47,16 @@ const val SINGLE_REFERENCE_GENOME_KEY = "singleReference"
...
-data class InstanceConfig(val schema: Schema, val referenceGenomes: Map<Suborganism, ReferenceGenome>)
+data class InstanceConfig(
+    val schema: Schema,
+    val referenceGenomes: Map<Suborganism, ReferenceGenome>,
+    /**
+     * In the single pathogen case, this will be equal to the single entry in `referenceGenomes`.
+     * In the multi-pathogen case, this will a flattened version of the `referenceGenomes` map,
+     * where the sequences of the suborganisms are merged into a list with names prefixed with the suborganism name.
+     */
+    val mergedReferenceGenome: ReferenceGenome,
```
anna-parker, line 58: So we wanted to just start by having the ebola viruses as a multi-pathogen and keeping the other viruses as is - how does this work now? What will the cchf sequences be in the `mergedReferenceGenome`? As it looks like we will be using this now for the processedData validation

backend/src/test/kotlin/org/loculus/backend/controller/submission/SubmissionConvenienceClient.kt
corneliusroemer, line 1: We don't need all the new cases and tests as for the backend multipathogen is the same as multisegment. It's exactly the same.

backend/src/test/kotlin/org/loculus/backend/controller/submission/SubmitProcessedDataEndpointTest.kt
```diff
@@ -758,6 +760,139 @@ class SubmitProcessedDataEndpointTest(
...
+        convenienceClient.getSequenceEntry(
+            accession = accessionVersion.accession,
+            version = accessionVersion.version,
+            organism = MULTI_PATHOGEN_ORGANISM,
+        )
+            .assertStatusIs(Status.PROCESSED)
+    }
+
+    @Test
+    fun `WHEN I submit data for multi pathogen organism with main segment THEN is rejected`() {
```
corneliusroemer, line 790: Unnecessary test, especially at this early stage in multi-pathogen feature work, it's already covered by normal tests that a segment needs to exist

corneliusroemer (commented):
Not sure why we need all those changes. The backend doesn't need to change, it doesn't care whether something is a suborganism or a segment. Existing segment functionality should work. Just give it the merged list and we're done. We can undo the previous backend changes, they aren't necessary then.

PR #4762, opened 2025-07-29
test(backend): test whole submission flow for multi pathogen organism

### LINKED ISSUE #4761: Multi pathogen support: write tests for whole backend submission flow
Make sure that everything work in the backend from end (file submission) to end (export for SILO).

fengelniederhammer:
resolves #4761

### PR Checklist
- [x ] The implemented feature is covered by appropriate, automated tests.


corneliusroemer (changes requested):
I've turned into draft PR as tests don't pass, formatting.

Even if, similar to #4760 I don't think this is the way to go. The backend doesn't need these changes. It should work as is using the existing segment features with at most small changes needed. No whole set of test cases, new business logic etc

The only backend change necessary is to remove the requirements on sequence ids having special underscore suffixes.

The bulk of the multi pathogen feature is:
- adapting preprocessing to assign segments/sub organism
- Adjusting the website to not be confusing when we start using the segment features for what aren't segments
- adjusting ena deposition to work with different taxa

PR #4767, opened 2025-07-29
feat(ingest): never remove segment field from metadata and filter out nextclade sort matches that are not supported

anna-parker:
resolves #

Main is failing with the error 
```
 File "/package/scripts/heuristic_group_segments.py", line 145, in main
    segment = values["segment"]
              ~~~~~~^^^^^^^^^^^
KeyError: 'segment'
```
`segment` is a keyword in the heuristic grouping script. However, in the enteroviruses config we renamed `segment` to genotype. This rename meant that `segment` was nolonger a field in the metadata - leading the heuristic grouping script to break. This PR should ensure that even if we rename `segment` the `segment` field should stay in the metadata as it is a required field for grouping.

This PR also covers the case where nextclade sort assigns a segment to a sequence which is not a valid segment - in this case we also drop these matches. This is important as the backend (for now) does not allow submission of unknown segments and will error - also ingest has been written with this state in mind and assumes we only assign known segments. 

This was fixed in @alejandra-gonzalezsanchez's PR: https://github.com/loculus-project/loculus/pull/4570

### PR Checklist

🚀 Preview: https://small-ingest-fixes.loculus.org

### DIFF
--- a/ingest/scripts/heuristic_group_segments.py
+++ b/ingest/scripts/heuristic_group_segments.py
@@ -211,11 +211,6 @@ def main(
                 if segment in group
             ]
         )
-        if not joint_key:
-            logger.warning(
-                f"Group {group} has no known segments, skipping."
-            )
-            continue
         for segment, accession in group.items():
             fasta_id_map[accession] = f"{joint_key}_{segment}"
 
--- a/ingest/scripts/heuristic_group_segments.py
+++ b/ingest/scripts/heuristic_group_segments.py
@@ -211,6 +211,11 @@ def main(
                 if segment in group
             ]
         )
+        if not joint_key:
+            logger.warning(
+                f"Group {group} has no known segments, skipping."
+            )
+            continue
         for segment, accession in group.items():
             fasta_id_map[accession] = f"{joint_key}_{segment}"
 
--- a/ingest/scripts/prepare_metadata.py
+++ b/ingest/scripts/prepare_metadata.py
@@ -114,6 +114,10 @@ def main(
             val = record.pop(from_key)
             record[to_key] = val
 
+        # Keep segment field if segmented and it is in rename
+        if config.segmented and "segment" in config.rename:
+            record["segment"] = record[config.rename["segment"]]
+
     keys_to_keep = set(config.rename.values()) | set(config.keep)
     if config.segmented:
         keys_to_keep.add("segment")
@@ -138,9 +142,8 @@ def main(
         filtered_record = {k: str(v) for k, v in record.items() if v is not None and str(v)}
 
         # rename "id" to "submissionId" for back-compatibility with old hashes
-        
         filtered_record["submissionId"] = filtered_record.pop("id")
-        
+        filtered_record["segment"] = record.get("segment", "")
 
         metadata_dump = json.dumps(filtered_record, sort_keys=True)
         prehash = metadata_dump + sequence_hash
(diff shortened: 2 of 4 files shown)


kubernetes/loculus/values.yaml
```diff
@@ -1665,7 +1665,7 @@ defaultOrganisms:
             sequence: "[[URL:https://corneliusroemer.github.io/seqs/artefacts/cchf/NP.fasta]]"
   enteroviruses:
     <<: *defaultOrganismConfig
-    enabled: false
+    enabled: true
```
anna-parker, line 1668: Revert before merging
```suggestion
    enabled: false
```

ingest/scripts/prepare_metadata.py
```diff
@@ -138,9 +142,8 @@ def main(
         filtered_record = {k: str(v) for k, v in record.items() if v is not None and str(v)}
 
         # rename "id" to "submissionId" for back-compatibility with old hashes
-        
         filtered_record["submissionId"] = filtered_record.pop("id")
-        
+        filtered_record["segment"] = record.get("segment", "")
```
anna-parker, line 146:
```suggestion
```

ingest/scripts/heuristic_group_segments.py
```diff
@@ -211,6 +211,11 @@ def main(
                 if segment in group
             ]
         )
+        if not joint_key:
```
anna-parker, line 214: this is only if nextclade sort finds a match which we do not have in our "known matches"

corneliusroemer, line 214: Let's fix this at the root, filter out non-matching much earlier

anna-parker: True this could be filtered out in `parse_nextclade_sort_output.py` - I will try to do that there

ingest/scripts/prepare_metadata.py
```diff
@@ -114,6 +114,10 @@ def main(
             val = record.pop(from_key)
             record[to_key] = val
 
+        # Keep segment field if segmented and it is in rename
+        if config.segmented and "segment" in config.rename:
+            record["segment"] = record[config.rename["segment"]]
```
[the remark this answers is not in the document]
corneliusroemer: Below we already keep segment if it's segmented? I'm confused.

corneliusroemer (commented):
The PR description doesn't really explain:
- why we need to filter out matches that aren't supported, theoretically we could just submit them anyways, I wouldn't expect their presence to break anything, just waste db storage
- why the change to prepare_metadata is necessary, as far as I understand the code there, we already keep the segment through keys-to_keep, see old line 117
- why we need to hack a rename in specifically for segmented ones

Maybe ingest should differentiate segmented and multi-path and use different config for it? Overall I still don't feel like I understand why we get an error at the moment.

kubernetes/loculus/values.yaml
```diff
@@ -1721,7 +1721,8 @@ defaultOrganisms:
           noInput: true
           generateIndex: true
           autocomplete: true
-          ingest: segment
+          preprocessing:
```
anna-parker, line 1724: oh this actually doesn't work - I wonder if its because `segment` is a custom field - definitely the previous fix worked

ingest/scripts/prepare_metadata.py
```diff
@@ -114,6 +114,11 @@ def main(
             val = record.pop(from_key)
             record[to_key] = val
 
+        # segment is a required field for the ingest grouping scripts
+        # Keep segment field if config.segmented and it is in rename
```
anna-parker, line 118:
```suggestion
        # Keep segment field if config.segmented even if it is specified to be renamed
```

corneliusroemer (approved):

PR #4770, opened 2025-07-30
feat(backend): only add the merged reference genomes to the backend config

fengelniederhammer:
Instead of the full non-merged reference genomes, only supply the merged one to the backend. That should give as a working preview with entero-viruses fastest.
The validation of the preprocessing results will simply work with that.
The initial submission will still be in the multi segmented style, but we can tackle that later.

Apparently requires https://github.com/loculus-project/loculus/pull/4767 to have working ingest for the EVs.

This is a partial revert of #4692:
- The values.yaml config changes stay (extra level of hierarchy for each suborganism)
- The rename in backend of `referenceGenomes` -> `referenceGenome` stays but the backend config goes back to what it was before, flat.
- Website changes stay

### PR Checklist
  - the organisms on the preview still seem to work as before.

🚀 Preview: https://onlymergedrefgenomeforbac.loculus.org

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/config/Config.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/config/Config.kt
@@ -43,11 +43,7 @@ enum class FileUrlType {
     override fun toString(): String = lowerCase(name)
 }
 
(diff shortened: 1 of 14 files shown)


anna-parker (approved):
LGTM - thanks!

PR #4779, opened 2025-07-31
refactor(backend): make code more readable

anna-parker:
resolves #

### Screenshot

Could someone explain to me why the config is using `outputFileUrlType: backend` by default?
Why do we have a website option actually I was just looking at [website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts](https://github.com/loculus-project/loculus/blob/main/website/src/pages/seq/%5BaccessionVersion%5D/%5BfileCategory%5D/%5BfileName%5D.ts) and there we seem to redirect to the backend

### PR Checklist

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
@@ -8,6 +8,7 @@ import com.fasterxml.jackson.databind.node.NullNode
 import com.fasterxml.jackson.databind.node.TextNode
 import mu.KotlinLogging
 import org.loculus.backend.api.DataUseTerms
+import org.loculus.backend.api.FileCategory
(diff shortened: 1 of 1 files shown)


backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
chaoran-chen, line 1: The title/commit message should be `refactor(backend)`

theosanderson:
I'd say it probably makes sense for the default to be the `website` as I think our typical usage for Loculus is with the website running and the `backend` option only exists so the website can be disabled. I generally think we should take stock of whether we should support Loculus with the website disabled (in my opinion no) because for example we now have the `loculus-info` endpoint which is required for the CLI and based in the website.

chaoran-chen (approved):
Looks good, thanks!

PR #4780, opened 2025-07-31
feat(prepro): trim leading/trailing whitespace in string/regex metadata fields 

### LINKED ISSUE #4704: prepro: Trim leading/trailing whitespace in all metadata fields
Convenience feature proposed by @theosanderson here: https://github.com/pathoplexus/pathoplexus/pull/586#discussion_r2213392875

I consider this low priority but wanted to create an issue nonetheless as I don't think it's harmful.

alejandra-gonzalezsanchez:
resolves #4704 

Changes in `def identity(...)` in script `preprocessing_functions.py`:
- Metadata fields with types `int`, `float`, or `boolean` are not affected. For example, if `input_datum = " 3"` and the field is expected to be an `int`, the function converts it with `output_datum = int(input_datum)`, resulting in `input_datum = 3`. In this case, stripping whitespace is unnecessary.
- Metadata fields that are `strings` are now cleaned using `.strip() `to remove leading/trailing spaces and newlines.

Changes in `def check_regex(...)` in script `preprocessing_functions.py`:
- The `regex_field` is handled separately in this function (discovered during testing). To avoid mismatches due to whitespace, `regex_field = regex_field.strip()` has been added.

> **⚠️ Note:** Consider adding more fields to testing, to revise all possible options. Currently added `authors` and `regex_field` in Case `strip_spaces_in_metadata`

### DIFF
--- a/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
@@ -176,32 +176,6 @@ class ProcessingFunctions:
                 ],
             )
         return result
-    
-    @staticmethod
-    def strip_whitespace_from_metadata(
-        input_data: InputMetadata,
-        output_field: str,
-        input_fields: list[str],
-        args: FunctionArgs, 
-    ) -> ProcessingResult:
-        """
-        Removes leading and trailing whitespace from all metadata fields in input_fields.
-        """
-        warnings: list[ProcessingAnnotation] = []
-        errors: list[ProcessingAnnotation] = []
-
-        output_field = {}
-
-        for field in input_fields:
-            value = input_data.get(field)
-            if value is None:
-                continue
-            if not isinstance(value, str):
-                continue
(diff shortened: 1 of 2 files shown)


preprocessing/nextclade/tests/test_metadata_processing_functions.py
```diff
@@ -485,6 +485,26 @@
         expected_errors=[],
         expected_warnings=[],
     ),
+    Case(
+        name="strip_spaces_in_metadata",
+        input_metadata={
+            "submissionId": "strip_spaces_in_metadata",
+            "name_required": "name",
+            "ncbi_required_collection_date": "2022-11-01",
+            "authors": " Smith, John II; Doe, A.B.C. ",
```
anna-parker, line 494: looks great! can you also add a test that newlines and tabs are also stripped?

alejandra-gonzalezsanchez: Done! :)

Issue #4814, opened 2025-08-05
Backend and multi-path: allow addition of suborganisms by not storing null fields in backend and imputing when streaming results


anna-parker:
Similar to in https://github.com/loculus-project/loculus/pull/3232 now that we have multi-pathogens we also want to facilitate the easy addition of new "segments" (i.e. suborganisms) and their genes. We can do this again in an efficient way where we do not store nulls in the backend but impute while streaming - this also means that we can drop the requirement on prepro to add nulls for all "missing" segments and genes - decreasing the amount of data we have to stream

Issue #4763, opened 2025-07-29
Preprocessing: the default return structures for processed data nucleotide insertions and aa insertions is different


anna-parker:
When a sequence has no insertions we return a `{"main": []}` for nucleotide insertions but `{}` for amino acid insertions in the ProcessedData - this is inconsistent

PR #4815, opened 2025-08-05
feat(backend): omit nulls in insertions and sequences and do not require prepro to upload null sequences

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/4814, resolves https://github.com/loculus-project/loculus/issues/4763

### Screenshot

Similar to in https://github.com/loculus-project/loculus/pull/3232 now that we have multi-pathogens we also want to facilitate the easy addition of new "segments" (i.e. suborganisms) and their genes. 

We can do this again in an efficient way where we do not store nulls in the backend but impute the `null` entries while streaming - this also means that we can drop the requirement on prepro to add nulls for all "missing" segments and genes - decreasing the amount of data we have to stream.

This PR omits `null` entries and empty lists for nucleotide and amino acid sequences as well as insertions.

### PR Checklist
- [x] Any manual testing that has been done is documented: confirmed that CCHF preview is same state as main as prepro no longer returns `null` when a segment is missing this is a way to confirm that LAPIS is still receiving the correct values.

🚀 Preview: https://sequence-hydration.loculus.org


chaoran-chen (approved):
Looks good to me, thank you!

Issue #4734, opened 2025-07-21
Is `_` a good separator for segments?


chaoran-chen:
Our current FASTA header format for submisisons of multi-segmented organisms is `<submission id>_<segment>`. At the same time, `_` is a valid character for the submisison ID, so given a header like `abc_def`, it is not clear whether `def` is part of the submission id or of the segment name.

We encountered this problem already a couple of months ago in #3683 where the backend gave a wrong/undesired error message because it the value behind the `_` wrongly as the segment. To address that, we changed the backend to only accept specified segments. Now, in #4721 for the multi-pathogen feature, we are reverting the validation, thus also bringing back the issue: https://github.com/loculus-project/loculus/pull/4721/files#r2220245994

This can be resolved if we can clearly determine what is supposed to be the segment name and what's the submission ID. In https://github.com/loculus-project/loculus/issues/4711#issuecomment-3089892970, the option was briefly mentioned to use `|` as the separator. We certainly don't want to forbid `_` to be used in a submisison ID but would we want to forbid `|`? I checked Pathoplexus and there are unfortunately a total of 10 WNV sequences with a `|`:

<details><summary>Sequences on Pathoplexus with | in submission ID</summary>

```
63722|2022-09-30
77216-2|2024-08-23
Candillargues-C9|2024-08-21
Echovert-C3-P1V|2024-08-29
Elysettes-C2|2024-09-03
Lattes-E12|2024-08-28
Lattes-MX|2024-08-28
Ollioules-BG3-Cx39|2024-08-02
Ollioules-BG3-MX|2024-08-02
Ollioules-BG3-MX|2024-08-23
```
</details>

<details><summary>Script</summary>

```
// Script to check for submission IDs containing "|" character across multiple pathoplexus endpoints
async function checkSubmissionIds() {
  const endpoints = [
    { name: "Crimean-Congo Hemorrhagic Fever Virus", url: "https://lapis.pathoplexus.org/cchf/sample/details?fields=submissionId" },
    { name: "Ebola Sudan", url: "https://lapis.pathoplexus.org/ebola-sudan/sample/details?fields=submissionId" },
    { name: "Ebola Zaire", url: "https://lapis.pathoplexus.org/ebola-zaire/sample/details?fields=submissionId" },
    { name: "HMPV", url: "https://lapis.pathoplexus.org/hmpv/sample/details?fields=submissionId" },
[...]

PR #4821, opened 2025-08-06
feat(backend): remove fasta header validation from backend and refactor how fasta and metadata are merged

anna-parker:
resolves #

### PR Checklist


backend/src/main/resources/db/migration/V1.16__rename_aux_table_columns.sql
```diff
@@ -0,0 +1,6 @@
+alter table sequence_upload_aux_table drop CONSTRAINT sequence_upload_aux_table_pkey;
+
+alter table sequence_upload_aux_table drop column segment_name;
```
anna-parker, line 3: we should improve the migration - maybe create a new field sequence_submission_id which will be submission_id + segment_name and submission_id will become metadata_submission_id for all entries in the aux table at the time of migration

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -247,6 +247,116 @@ def run_sort(
...
+            _, value = next(iter(input_unaligned_sequences.items()))
+            unaligned_nucleotide_sequences["main"] = value
+        return (
+            unaligned_nucleotide_sequences,
+            aligned_nucleotide_sequences,
+            errors,
+        )
+    for segment in config.nucleotideSequences:
+        aligned_nucleotide_sequences[segment] = None
+        unaligned_segment = [
```
anna-parker, line 295: update some tests that show that the dictionary can include the full submissionId and not just the segment

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -247,6 +247,116 @@ def run_sort(
     return alerts
 
 
+def assign_segment(
```
anna-parker, line 250: add some tests for the validation in prepro

anna-parker:
error for cchf: 
` ERROR: duplicate key value violates unique constraint "sequence_entries_preprocessed_data_pkey"`

preprocessing/nextclade/tests/test_nextclade_preprocessing.py
```diff
@@ -560,6 +558,196 @@ def invalid_sequence() -> str:
...
+        ),
+    ),
+]
+
+segment_validation_tests_multi_segments = [
+    Case(
+        name="accept any prefix for multi-segment",
+        input_metadata={},
+        input_sequence={
+            "prefix_ebola-sudan": sequence_with_mutation("ebola-sudan"),
```
anna-parker, line 630: in theory we could assert that the prefix is `submissionId`

integration-tests/tests/specs/features/search/lineage-field.spec.ts
```diff
@@ -7,7 +7,7 @@ import { v4 as uuidv4 } from 'uuid';
 const SEQUENCE = 'ATTGATCTCATCATTT';
 
 test('Lineage field lineage counts', async ({ page, pageWithGroup }) => {
-    test.setTimeout(95_000);
+    test.setTimeout(900_000);
```
anna-parker, line 10: this increase doesn't have an affect - weirdly I can do these steps without issues on the preview so I'm not sure why the intergration tests are failing

anna-parker:
> Is this something that's been discussed? Are we now requiring that all segments align to reference? (We didn't before right?) Mightn't that e.g. mean we reject some INSDC stuff we currently accept?

sorry this should say we will allow the option to assign segments using nextclade sort - this shouldn't change anything tho as we anyways only accept INSDC data that aligns and assign segments in ingest using either nextclade align or nextclade sort

backend/docs/db/schema.sql
chaoran-chen, line 1: I'm happy with removing the validation but I'm hesitant about the breaking change. Since the backend is anyways already doing the parsing of the FASTA header and has to identify what is the `submissionId`, whether a `segmentName` is provided and, if yes, what the `segmentName` is, I think that we should provide the pipeline with the parsed information. The pipeline already receives the `submisisonId` and putting it into the sequence dictionary keys is redundant.

If the reason is to avoid writing `main`, we can change it to use an empty string as key if no segment name is provided.

I also still think that `_` is not a good separator and that we should change it (#4734).

website/src/components/Edit/SequencesForm.tsx
```diff
@@ -52,14 +52,14 @@ export class EditableSequences {
      */
     static fromInitialData(initialData: SequenceEntryToEdit, segmentNames: string[]): EditableSequences {
         const emptyRows = this.emptyRows(segmentNames);
-        const existingDataRows = Object.entries(initialData.originalData.unalignedNucleotideSequences).map(
-            ([key, value]) => ({
+        const existingDataRows = Object.entries(initialData.processedData.unalignedNucleotideSequences)
```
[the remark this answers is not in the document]
anna-parker: @theosanderson - ok! I guess as the unaligned sequences could be processed it is good to return the actual original data. 

Then I guess we need to update this part of the form to let the users download their original data with the submissionId. 

I am actually wondering - as this does create a breaking change in the structure of the processedData response anyways - maybe we should add a map from the submissionId of the original data to the assigned segment/subtype. For example

```json
{
    accession,
    version,
    errors,
    warnings,
    data: {
        metadata,
        unalignedNucleotideSequences,
        alignedNucleotideSequences,
        nucleotideInsertions,
        alignedAminoAcidSequences,
        aminoAcidInsertions,
        files,
        submissionIdSegmentMap,
    }
}
```

PR #4836, opened 2025-08-11
feat(website): enable drag selection for checkboxes

theosanderson:
- allow selecting multiple table rows by dragging over checkboxes (works for unselecting too)
- cover drag selection with a unit test

chrome-capture-2025-08-11

------
https://chatgpt.com/codex/tasks/task_e_6894fe602954832589005851d83a0820

🚀 Preview: https://codex-add-drag-to-select.loculus.org


website/src/components/SearchPage/Table.spec.tsx
chaoran-chen, line 1: On my phone (Android, Chrome browser), the keyboard is now opening every time I click on a checkbox:

https://github.com/user-attachments/assets/a78d987f-261a-4284-bbe0-1053d5679700

theosanderson: Hmm interesting, I can't reproduce (also on Chrome on Android) but will have a think about what cause could be

theosanderson: What I can reproduce is that on desktop mode in android if I am focused on a text field then I click one of the checkboxes the focus stays on the text field – I guess it's the same sort of issue so should let me debug – thanks for spotting!

chaoran-chen: Works for me now, thanks and sorry for the delay!

PR #4853, opened 2025-08-13
feat(website): better organism-selector redirect page

theosanderson:
resolves #

### Screenshot
image


website/src/pages/organism-selector/[redirectTo].astro
Comment (chaoran-chen) [inline website/src/pages/organism-selector/[redirectTo].astro:1]: Very nice! I can try to review the code tomorrow or on Friday. One small comment while noticing it: with conventional commits, `style` usually refers to code style issues that don't have an effect on the behavior of the software (see e.g. the [Angular guidelines](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type)).

PR #4867, opened 2025-08-16
Feat use column order to order search UI

rneher:
We specify the order of table columns in the `values.yml`, but fields are alphabetical in other places. in particular in the search UI and in the sample details page, we often have a sub-optimal ordering. E.g. here where Country should come before division:

image

Here, we re-use the order from the columns to order the search UI

image

I have not managed to do the same for the sequence details page. 

### PR Checklist


theosanderson:
https://github.com/loculus-project/loculus/pull/4865 hopefully does the details page

rneher:
thanks, Theo! tbh, I was mostly trying to see how far I get. Maybe its useful, maybe not.

PR #4893, opened 2025-08-26
feat(website): allow filtering for blank metadata values

theosanderson:
Show `(blank)` option in search autocompletes to filter null metadata values

image

These are encoded in the URL as `_null_` 

Tested:
- That "any" still works: https://codex-add-filtering-for-b.loculus.org/cchf/search?visibility_versionStatus=true&versionStatus=
- That in general the `(blank)` works

------
https://chatgpt.com/codex/tasks/task_e_68ad876c45ac83259b096faad7c70025

🚀 Preview: https://codex-add-filtering-for-b.loculus.org

### DIFF
--- a/website/src/utils/search.spec.ts
+++ b/website/src/utils/search.spec.ts
@@ -3,7 +3,7 @@ import { describe, expect, it } from 'vitest';
 import { MetadataFilterSchema, NULL_QUERY_VALUE } from './search.ts';
 
 describe('MetadataFilterSchema', () => {
-    it('decodes <null> values from query state', () => {
+    it('decodes _null_ values from query state', () => {
         const schema = new MetadataFilterSchema([{ name: 'field1', type: 'string' }]);
         const result = schema.getFieldValuesFromQuery({ field1: NULL_QUERY_VALUE }, {});
         expect(result.field1).toBeNull();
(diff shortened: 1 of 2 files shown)


website/src/components/SearchPage/DownloadDialog/SequenceFilters.tsx
```diff
@@ -30,8 +30,9 @@ export interface SequenceFilter {
 
     /**
      * Return a map of keys to human-readable descriptions of the filters to apply.
+     * null values are maintained as null.
      */
-    toDisplayStrings(): Map<string, [string, string]>;
+    toDisplayStrings(): Map<string, [string, string | null]>;
```
anna-parker, line 35: naive question - can the values ever be `undefined`? should we also add that case here?

theosanderson:
> when I copy the URL it becomes: https://codex-add-filtering-for-b.loculus.org/ebola-sudan/search?geoLocCountry=%3Cnull%3E&lengthFrom=12, not that <null> becomes %3Cnull%3E - could this be an issue?

This is known and expected behaviour but yeah I did think that maybe it's not ideal.

Characters which would not be affected would be
```
Hyphen: -
Underscore: _
Period: .
Tilde: ~
```

Maybe `_null_` would be good?

anna-parker (approved):
tested this out and it looks good! One small comment about the URL encoding :-)

anna-parker:
> Maybe _null_ would be good?

That sounds good to me - maybe @corneliusroemer or @chaoran-chen have a suggestion?

PR #4913, opened 2025-09-01
fix(deposition): ensure we submit only once by fixing SQL query to return strictly changed rowcount

### LINKED ISSUE #4911: ENA submission acquires lock even when other pod has already got the lock because we don't check if column value was not "SUBMITTING" already
This is the root cause of #4895 

This code succeeds even if the state was already `SUBMITTING`

```py
        update_values: dict[str, Any] = {"status": Status.SUBMITTING}
            db_config,
            table_name=TableName.ASSEMBLY_TABLE,
            conditions=seq_key,
            update_values=update_values,
        )
        if number_rows_updated != 1:
            # state not correctly updated - do not start submission
            logger.warning(
                "assembly_table: Status update from READY to SUBMITTING failed - "
                "not starting submission."
            )
            continue
```

A row appears in rowcount even if it wasn't changed:

```py
        with con, con.cursor(cursor_factory=RealDictCursor) as cur:
            # Prevent sql-injection with table_name and column_name validation
            TableName.validate(table_name)
            for key in conditions:
                validate_column_name(table_name, key)

            query = f"UPDATE {table_name} SET "  # noqa: S608

            set_clause = ", ".join([f"{key}=%s" for key in update_values])
            query += set_clause

            where_clause = " AND ".join([f"{key}=%s" for key in conditions])
            query += f" WHERE {where_clause}"
            parameters = tuple(
                str(value) if (isinstance(value, (Status, StatusAll))) else value
                for value in update_values.values()
            ) + tuple(
                str(value) if (isinstance(value, (Status, StatusAll))) else value
                for value in conditions.values()
            )

            cur.execute(query, parameters)
            updated_row_count = cur.rowcount
            con.commit()
```

corneliusroemer:
resolves #4911

resolves #

### PR Checklist

### DIFF
--- a/ena-submission/src/ena_deposition/submission_db_helper.py
+++ b/ena-submission/src/ena_deposition/submission_db_helper.py
@@ -399,13 +399,23 @@ def update_db_where_conditions(
             query += set_clause
 
             where_clause = " AND ".join([f"{key}=%s" for key in conditions])
-            query += f" WHERE {where_clause}"
-            parameters = tuple(
-                str(value) if (isinstance(value, (Status, StatusAll))) else value
-                for value in update_values.values()
-            ) + tuple(
-                str(value) if (isinstance(value, (Status, StatusAll))) else value
-                for value in conditions.values()
+            # Avoid updating rows that would not change so the return value is actually the
+            # number of rows that were changed for real. See bug #4911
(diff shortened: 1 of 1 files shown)


anna-parker (approved):
I looked through all the places where the function is used and it looks like a safe change - the workflow tests also all succeeded so I think this is safe!

Issue #4852, opened 2025-08-13
`/get-original-metadata` locked if backend restarts during an upload


chaoran-chen:
The `/get-original-metadata` is locked if a submission is being uploaded (as in written to the database): [code](https://github.com/loculus-project/loculus/blob/d3ff84474c1ba367746ee04365e910ac9fe91afb/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt#L428).

And the responsible function `checkIfStillProcessingSubmittedData()` ([code](https://github.com/loculus-project/loculus/blob/d3ff84474c1ba367746ee04365e910ac9fe91afb/backend/src/main/kotlin/org/loculus/backend/model/SubmitModel.kt#L428)) checks whether the upload aux tables are empty.

A problem occurs if the backend restarts (e.g., due to a crash or version update) while an upload is onging. Then, the aux tables are left filled and `/get-original-metadata` remains locked.

We should (1) check whether the lock is actually still needed / there is a better way to implement it and (2) clear outdated entries from the aux tables.

Issue #1613, opened 2024-04-16
Can't upload more than 65k sequences


fengelniederhammer:
When I tried to upload 100,000 dummy sequences at once, I got this error:

```
Internal Server Error: org.postgresql.util.PSQLException: PreparedStatement can have at most 65,535 parameters. Please consider using arrays, or splitting the query in several ones, or using COPY. Given query has 100,000 parameters
```

![grafik](https://github.com/loculus-project/loculus/assets/92720311/f2462e47-5571-4f2f-a668-408611179291)

chaoran-chen:
I also just hit that. Is "performance" the right label? I'd like to collect issues related to scalability.

anna-parker:
Just as an update - I did this via the API - there is no warning for that submitter: the submission post request succeeds. But the sequences are stuck in upload_metadata_aux table without an accession and I see this same error message in the backend logs - it would be great to somehow prevent this type of submission via the API.

corneliusroemer:
The backend simply needs to batch when there are >65k rows, not so important as submitters can batch themselves but might be nice eventually.

Agree with Chaoran that this isn't really performance, just a limitation/bug at the moment that we can fix.

anna-parker:
I was looking into where this error comes from. It looks like we stream uploads to the auxTable: https://github.com/loculus-project/loculus/blob/main/backend/src/main/kotlin/org/loculus/backend/model/SubmitModel.kt#L167. So is this error occurring when we try to upload the data from the auxTable into the unprocessed data table or maybe when we try to update with new accessions? 

https://github.com/loculus-project/loculus/blob/main/backend/src/main/kotlin/org/loculus/backend/model/SubmitModel.kt#L155 - calls https://github.com/loculus-project/loculus/blob/main/backend/src/main/kotlin/org/loculus/backend/service/submission/UploadDatabaseService.kt#L296 which chatGPT warned exposed could optimize this to one query and that could hit the limits

https://github.com/loculus-project/loculus/blob/main/backend/src/main/kotlin/org/loculus/backend/service/submission/UploadDatabaseService.kt#L134 will also cause an error due to how we build the jsonB object.

We should probably batch at these locations

Issue #996, opened 2024-02-12
When uploading too many sequences (7k mpox) I get error SSL error "violation of protocol"


corneliusroemer:
When uploading all mpox sequences, the requests fails but not with one of the correct HTTP error codes, but with some SSL thing instead:

```
urllib3.exceptions.SSLError: EOF occurred in violation of protocol (_ssl.c:2406)

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/corneliusromer/micromamba/envs/py12/lib/python3.12/site-packages/requests/adapters.py", line 486, in send
    resp = conn.urlopen(
           ^^^^^^^^^^^^^
  File "/Users/corneliusromer/micromamba/envs/py12/lib/python3.12/site-packages/urllib3/connectionpool.py", line 847, in urlopen
    retries = retries.increment(
              ^^^^^^^^^^^^^^^^^^
  File "/Users/corneliusromer/micromamba/envs/py12/lib/python3.12/site-packages/urllib3/util/retry.py", line 515, in increment
    raise MaxRetryError(_pool, url, reason) from reason  # type: ignore[arg-type]
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
urllib3.exceptions.MaxRetryError: HTTPSConnectionPool(host='backend-940-query-metadata-for-re.loculus.org', port=443): Max retries exceeded with url: /mpox/submit?groupName=insdc_ingest_group&dataUseTermsType=OPEN (Caused by SSLError(SSLEOFError(8, 'EOF occurred in violation of protocol (_ssl.c:2406)')))
```

Repro:
1. Upload these sequences https://loculus.slack.com/files/U05FBBSP02F/F06K13HNGV6/mpox_7200.tar.zst

Instead of SSL error, I should get:
```
requests.exceptions.HTTPError: 413 Client Error: Request Entity Too Large for url: https://backend-940-query-metadata-for-re.loculus.org/mpox/submit?groupName=insdc_ingest_group&dataUseTermsType=OPEN
```
which is what I get when I upload 1k rather than 7k mpox sequences.

PR #4915, opened 2025-09-01
refactor(backend): simplify submission by getting rid of aux tables

corneliusroemer:
- **wip**
- **Finish first pass**
- **It compiles**

resolves #

### PR Checklist


backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
chaoran-chen, line 1: Hi @corneliusroemer! Would you like to outline your approach? Just curious because I also found the aux tables a bit complicated but didn't see an obvious solution without them.

chaoran-chen: @corneliusroemer I see that this PR removes streaming and restricts the payload size to ~500MB and I'm concerned this goes in the wrong direction. While the current approach does have issues (e.g. #4852), I think they can be addressed without removing the aux tables or reducing scalability. We only recently improved upload scalability to support submissions over 65k sequences (#1613), and I believe continuing to build towards more scalability is important.

theosanderson: I definitely agree that discussion would be the first step towards going ahead with something like this (though of course a draft PR can be part of the discussion)

anna-parker:
I also wanted to add that the aux table has some issues in its current form: https://github.com/loculus-project/loculus/issues/4852#issuecomment-3291206637

backend/src/main/kotlin/org/loculus/backend/config/BackendSpringConfig.kt
chaoran-chen, line 1: While, for the current Pathoplexus organisms, people tend to sequence and submit only small number of sequences at once, there are many other use cases and situations where uploading larger files would be desired. For example, if someone sets up a new (internal) instance, they might want to import all of their existing data. Also, for bacteria, where the genomes are much larger, a few hundred genomes may already reach GBs.

chaoran-chen, line 1: If we merge this PR, is there a good way forward to support large uploads again, or would merging this mean that we decide against supporting large uploads (at least in the mid term)? We've had several issues (created by ourselves) regarding failing uploads of large files (#1613, #996, #1226) and to cite you, @corneliusroemer, "it's clearly something we want to address but it's not necessary for pathoplexus to work", so I think it's important to have a plan how to proceed.

You wrote "reductions of memory usage without reintroducing extra tables are possible but can be done later if considered necessary" – what do you have in mind?

As far as I can see, any really scalable solution would require performing an external memory join. Hereby, using the database/aux tables seems like a good approach to me. Alternatively, we could write temporary files to disk but this seems even more complicated to me.

chaoran-chen, line 1: I'd be interested to hear why the current approach is considered as overly complex and complicating further development.

anna-parker: Keeping the aux table structure does make the switch to multi-pathogen upload where segment/subtype assignment is done in the preprocessing pipeline and sequences are just grouped by the backend more complex. 

It is still doable but quite ugly: https://github.com/loculus-project/loculus/pull/4821

I wonder though if we should actually first switch the grouping of sequences in the backend to https://github.com/loculus-project/loculus/issues/4734#issuecomment-3317572503 - this will still require aux table changes (and a change in the original-data structure as the segment will not be known in advance) but I do this will be slightly less complicated than my current PR proposal.

anna-parker: As Loculus cannot serve bacteria currently due to other limitations (i.e. our query engine is not built to support them) I'm not sure this is a very pressing argument. 

Potentially having a very well compressed file that when inflated is too large could happen - although if I'm honest I find it highly unlikely that we have a submitter that would submit so much data at once to us - cases like this are more raw read related where we have a method for handling such large files. 

So, personally, I'm not sure I really see the importance of keeping this functionality. However, as a compromise, if we change the submission format as discussed in https://github.com/loculus-project/loculus/issues/4734#issuecomment-3317572503 I think we can keep the aux table structure relatively simple and make the necessary changes.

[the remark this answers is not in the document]
corneliusroemer: > I think we should aim throughout Loculus to be able to have memory requirements which basically have the option to scale at most with the size of a single sequence entry 

That's the case here - memory is proportional to N_of_sequences_in_batch x length_of_sequence? I might be misunderstanding.

> Achieving that here requires some version of the auxTable. 

That's not true?

> Yes, if we merged this we could still revert it in the future, or we could change various other things in the future to improve scalability.

Aux tables complicate things unnecessarily - that hampers scalability

> I haven't seen the auxTables cause a lot of problems in prod

They are unnecessary complexity - they cause problems when developing/changing code

> I agree they add complexity to the codebase but IMO that is worthwhile given where I hope Loculus to get to in the coming years.

Why do you think we need aux tables? Which problem do they solve? Why are they worth the complexity? I think there are much better alternatives - not all sequences need to be in memory - that's just the simplest way of doing it.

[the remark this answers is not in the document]
corneliusroemer: In the current implementation yes but one can also stream the sequences without aux tables - if you really want to support monolithic 2GB sequence submissions there are simple ways to make this PR work with this without requiring 2 GB backend memory.

corneliusroemer: It's unnecessary complexity, there's a single 100 line SQL command here, that's not easy to maintain.

PR #4931, opened 2025-09-03
fix(deployment): Make the backend use the user given S3-endpoint for the file-sharing feature

fhennig:
I'm working on a prod deployment of Loculus, and I disabled the development S3 and configured my own bucket, and I found this setting still in my backend Pod:

```yaml
    - name: S3_BUCKET_ENDPOINT
      value: http://localhost:8084
```

The backend also generates pre-signed URLs with the `localhost:8084`, instead of the proper endpoint.

I changed it so that the backend uses the user-given endpoint.

I've tested this in our setup, with a Hetzner bucket, and it's working.

### Screenshot
n/A

### PR Checklist

### DIFF
--- a/kubernetes/loculus/templates/_urls.tpl
+++ b/kubernetes/loculus/templates/_urls.tpl
@@ -28,7 +28,7 @@
         {{- printf "http://%s:8084" $.Values.localHost -}}
     {{- end -}}
   {{- else -}}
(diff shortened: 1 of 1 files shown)


theosanderson:
(i agree the line seems wrong)

theosanderson (approved):

PR #4958, opened 2025-09-05
chore(backend): rename pipeline upgrade function to be explicit about caching and not overload

corneliusroemer:
Noticed when looking at https://github.com/loculus-project/loculus/pull/4862 that the function is overloaded which makes it harder to reason and doesn't make clear we cache so renamed the function and also added a doc string.

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
@@ -14,7 +14,6 @@ import org.jetbrains.exposed.sql.JoinType
 import org.jetbrains.exposed.sql.LongColumnType
 import org.jetbrains.exposed.sql.Op
 import org.jetbrains.exposed.sql.SortOrder
-import org.jetbrains.exposed.sql.SqlExpressionBuilder.eq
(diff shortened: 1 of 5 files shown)


backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -1276,7 +1273,15 @@ class SubmissionDatabaseService(
...
-    fun useNewerProcessingPipelineIfPossible(): Map<String, Long?> {
+
+    private var lastPreprocessedDataUpdate: String? = null
+
+    /** Check if there have been newly preprocessed data since the last update check.
+     * Returns empty map if no changes to the prepro table since last check.
+     * If there are changes, runs useNewerProcessingPipelineIfPossible for each organism
+     * and returns a map of organism name to newly set pipeline version (or null if no new version was set).
+     */
+    fun runPipelineUpgradeIfPreproTableChanged(): Map<String, Long?> {
```
theosanderson, line 1284: Thanks for the PR - I agree it's confusing as is and this is an improvement


```suggestion
    fun runPotentialPipelineUpgradeIfPreproTableChanged(): Map<String, Long?> {
```
I think it's kinds confusing as proposed that the top function implies a definite upgrade and the subfunction suggests its possible

PR #4963, opened 2025-09-07
feat(website): refactor restricted use banner to make it easy to modify 

theosanderson:
🚀 Preview: https://restricteduse.loculus.org

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "githubPullRequests.ignoredPullRequestBranches": [
+        "main"
+    ]
+}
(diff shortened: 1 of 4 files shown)


theosanderson:
@claude refactor 

<ErrorBox title='Restricted-Use sequence' level='warning'>
                    This sequence is only available under the <a href='https://pathoplexus.org/about/terms-of-use/restricted-data' className='underline'>Restricted Use Terms</a>. This prohibits publications that use this sequence as focal data during the restricted-use period, except with the consent of the sequence submitters.
                </ErrorBox>
                
   to be a component in its own separate file

corneliusroemer (commented):
Refactor makes sense but the PRs title should not be `feat(website)` but `refactor(website)`. There's not feature here really ;)

theosanderson:
Loculus did not previously make it possible to reconfigure the banner and now it does, that's the (undocumented) feature

PR #4989, opened 2025-09-09
fix(ena-submission): use random suffix in aliases for testing to avoid CI failures due to duplicate aliases

### LINKED ISSUE #4985: Use milliseconds in ena deposition assembly aliases for testing to avoid CI failures
We get CI failures if we run 2 ena deposition tests at the same time and they try to upload the same sequence within a second of each other. It might seem rare but it happened just now:

```
 In analysis, alias: "webin-genome-LOC_0001TLY.1_20250909_081842". The object being added already exists in the submission account with accession: "ERZ28436670". The 
```

Let's just add something random or milliseconds to the name to avoid this from happening again.

https://github.com/loculus-project/loculus/actions/runs/17576281593/job/49922184518

corneliusroemer:
resolves #4985

### PR Checklist

### DIFF
--- a/ena-submission/src/ena_deposition/ena_submission_helper.py
+++ b/ena-submission/src/ena_deposition/ena_submission_helper.py
@@ -4,8 +4,8 @@ import gzip
 import json
 import logging
 import os
+import random
 import re
-import secrets
 import string
 import subprocess  # noqa: S404
 import tempfile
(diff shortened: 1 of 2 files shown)


ena-submission/src/ena_deposition/ena_submission_helper.py
```diff
@@ -156,7 +158,9 @@ def get_alias(prefix: str, test=False, set_alias_suffix: str | None = None) -> X
     if set_alias_suffix:
         return XmlAttribute(f"{prefix}:{set_alias_suffix}")
     if test:
-        return XmlAttribute(f"{prefix}:{datetime.datetime.now(tz=pytz.utc)}")
+        entropy = "".join(secrets.choice(string.ascii_letters + string.digits) for _ in range(4))
```
anna-parker, line 161: I think might be cleaner
```
''.join(secrets.choices(string.ascii_letters + string.digits, k=4))
```

anna-parker (approved):

ena-submission/src/ena_deposition/ena_submission_helper.py
corneliusroemer: that would be nice but it errors?

```
>>> import string, secrets
>>> ''.join(secrets.choices(string.ascii_letters + string.digits, k=4))
Traceback (most recent call last):
  File "<python-input-3>", line 1, in <module>
    ''.join(secrets.choices(string.ascii_letters + string.digits, k=4))
            ^^^^^^^^^^^^^^^
```

image

PR #4992, opened 2025-09-09
docs: improve image schema documentation

corneliusroemer:
## Summary
- Clarify that the `tag` field in imageSpec can include a digest to absolutely pin the image (e.g., `v0.7.10.1@sha256:...`)
- Add proper enum constraints to `pullPolicy` field (Always, IfNotPresent, Never)
- Document default pullPolicy behavior based on tag value

### DIFF
--- a/kubernetes/loculus/values.schema.json
+++ b/kubernetes/loculus/values.schema.json
@@ -715,7 +715,7 @@
           "groups": ["image"],
           "type": "string",
           "enum": ["Always", "IfNotPresent", "Never"],
-          "description": "The pull policy to use. Always: always pull the image. IfNotPresent: pull if not present locally. Never: never pull, use local image only. If not specified, defaults to Always if tag is 'latest' or empty, otherwise IfNotPresent."
+          "description": "The pull policy to use. Always: always pull the image. IfNotPresent: pull if not present locally. Never: never pull, use local image only."
         }
       },
       "required": ["repository", "pullPolicy"],
--- a/kubernetes/loculus/values.schema.json
+++ b/kubernetes/loculus/values.schema.json
@@ -709,12 +709,13 @@
         "tag": {
           "groups": ["image"],
           "type": "string",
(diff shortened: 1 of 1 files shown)


kubernetes/loculus/values.schema.json
```diff
@@ -709,12 +709,13 @@
...
           "type": "string",
-          "description": "The tag to pull. Examples: latest, 0.5.7"
+          "description": "The tag to pull. Examples: latest, 0.5.7. Can include a digest to absolutely pin the image (tags can move): v0.7.10.1@sha256:96e431192d6493ed71ac10bc273710df21e7c520a77bebaa7d94ff859264d36c"
         },
         "pullPolicy": {
           "groups": ["image"],
           "type": "string",
-          "description": "The pull policy to use. Examples: IfNotPresent, Always."
+          "enum": ["Always", "IfNotPresent", "Never"],
+          "description": "The pull policy to use. Always: always pull the image. IfNotPresent: pull if not present locally. Never: never pull, use local image only. If not specified, defaults to Always if tag is 'latest' or empty, otherwise IfNotPresent."
```
anna-parker, line 718: this is a bit unclear: `The default value is IfNotPresent, setting the tag to 'latest' will default to Always`.

It is probably not possible to leave the field empty (`pullPolicy: `)?

corneliusroemer: hmm the schema requires it at the moment it seems
image

anna-parker: just to confirm there is no default now?

corneliusroemer:
I don't know what you mean

corneliusroemer:
Our schema currently requires this to be set if you look at the schema.

But if you don't do schema validation you might get away with setting it to
be blank, in which case we probably are fine and kubernetes standard
defaults apply which one can google now (rather than the defaults being
mentioned here in the schema)

Feels a bit like bikeshedding to discuss this little tweak here at this
depth

anna-parker (approved):

PR #4995, opened 2025-09-10
feat(website): multi pathogen: sequence details page: only show relevant sequences

### LINKED ISSUE #4772: Multi pathogen support: sequence details page: only show relevant sequences.
Hide the segments and genes that do not belong to the suborganism of the respective sequence entry.

Image

fengelniederhammer:
resolves #4772

### Screenshot

This makes the sequence viewer aware of multiple reference genomes. It only shows the segments and genes of the current suborganism. Display names are shortened accordingly (only showing the gene name and not `<suborganism>-<gene>`), while we need to query LAPIS with the composed version.

image

image

### PR Checklist
  - On the preview, single segmented (west-nile) and multi segmented (CCHF) details pages still show the sequence viewer as before (all segments and genes are there)
  - For Enteroviruses, the expected segments and genes are there. The genes only show `XXXX...`, but that seems to be an issue with the data and not with the website.

🚀 Preview: https://4772-actually.loculus.org

### DIFF
--- a/website/src/utils/sequenceTypeHelpers.ts
+++ b/website/src/utils/sequenceTypeHelpers.ts
@@ -1,26 +1,31 @@
 export type SequenceType =
-    | { type: 'nucleotide'; aligned: boolean; name: string }
-    | { type: 'aminoAcid'; aligned: true; name: string };
+    | { type: 'nucleotide'; aligned: boolean; name: SequenceName }
+    | { type: 'aminoAcid'; aligned: true; name: SequenceName };
 export type BaseType = SequenceType['type'];
 
-export const unalignedSequenceSegment = (segmentName: string): SequenceType => ({
+export type SequenceName = {
+    lapisName: string;
+    label: string;
+};
+
+export const unalignedSequenceSegment = (segmentName: SequenceName): SequenceType => ({
     type: 'nucleotide',
     aligned: false,
     name: segmentName,
 });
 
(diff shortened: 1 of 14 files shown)


website/vitest.setup.ts
```diff
@@ -21,20 +21,22 @@ import type { RuntimeConfig } from './src/types/runtimeConfig.ts';
 
 export const DEFAULT_GROUP_NAME = 'testGroup';
 
+export const testOrganism = 'testOrganism';
```
anna-parker, line 24: This refactor could already go on main

website/src/utils/sequenceTypeHelpers.ts
```diff
@@ -1,26 +1,31 @@
...
+export const geneSequence = (gene: SequenceName): SequenceType => ({
     type: 'aminoAcid',
     aligned: true,
     name: gene,
 });
 export const isUnalignedSequence = (type: SequenceType): boolean => type.type === 'nucleotide' && !type.aligned;
 export const isAlignedSequence = (type: SequenceType): boolean => type.type === 'nucleotide' && type.aligned;
-export const isGeneSequence = (gene: string, type: SequenceType): boolean =>
-    type.type === 'aminoAcid' && type.name === gene;
+export const isGeneSequence = (gene: SequenceName, type: SequenceType): boolean =>
```
anna-parker, line 30: why is this `type.name === gene` actually required?

PR #4996, opened 2025-09-10
feat!(config): use new cchf dataset

anna-parker:
resolves #

### Screenshot

This uses the newly merged CCHF dataset that now does QC and lineages! See details here: https://github.com/nextstrain/nextclade_data/pull/265

It also vastly improves alignment, especially of the M segment which has high variability at the start and end of the sequences - leading to many alignments previously failing the QC scores. 

Uses new lineage file: https://github.com/pathoplexus/silo-lineage-hierarchy-definitions/blob/main/definitions/cchf/S/2025-09-12/lineages.yaml

To release on PPX this requires a pipeline version update: https://github.com/pathoplexus/pathoplexus/pull/658 and the regroup-revoke cronjob will have to be run once.

### PR Checklist
- [ ] ~All necessary documentation has been adapted.~
- [ ] ~The implemented feature is covered by appropriate, automated tests.~
- [x] Any manual testing that has been done is documented (i.e. what exactly was tested?) -> tested the results on a persistent DB instance on loculus: https://github.com/loculus-project/loculus/pull/5003
Results: Exactly 1 sequence must be revoked as it can no longer be grouped : https://loculus.slack.com/archives/C07FP0BGHDZ/p1757518656855529

This makes sense as now there are 3 segments from the same institute submitted on the same day without further identifiers - making grouping infeasible: 
image

🚀 Preview: https://test-new-cchf.loculus.org

### DIFF
--- a/ingest/Snakefile
+++ b/ingest/Snakefile
@@ -60,7 +60,12 @@ if SEGMENTED:
         for segment in config["nucleotide_sequences"]:
             dataset_server_map[segment] = segment_identification.get(
                 "nextclade_dataset_server_map", {}
-            ).get(segment, segment_identification.get("nextclade_dataset_server", "https://data.clades.nextstrain.org/v3"))
+            ).get(
+                segment,
+                segment_identification.get(
+                    "nextclade_dataset_server", "https://data.clades.nextstrain.org/v3"
+                ),
+            )
             dataset_name_map[segment] = segment_identification.get(
                 "nextclade_dataset_name_map", {}
             ).get(
--- a/ingest/Snakefile
+++ b/ingest/Snakefile
@@ -62,7 +62,9 @@ if SEGMENTED:
                 "nextclade_dataset_server_map", {}
             ).get(
                 segment,
-                segment_identification.get("nextclade_dataset_server"),
+                segment_identification.get(
+                    "nextclade_dataset_server", config.get("nextclade_dataset_server")
+                ),
             )
             dataset_name_map[segment] = segment_identification.get(
                 "nextclade_dataset_name_map", {}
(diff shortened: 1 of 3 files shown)


ingest/Snakefile
```diff
@@ -60,7 +60,7 @@ if SEGMENTED:
         for segment in config["nucleotide_sequences"]:
             dataset_server_map[segment] = segment_identification.get(
                 "nextclade_dataset_server_map", {}
-            ).get(segment, segment_identification.get("nextclade_dataset_server"))
+            ).get(segment, segment_identification.get("nextclade_dataset_server", "https://data.clades.nextstrain.org/v3"))
```
corneliusroemer: I'm hesitant to setting defaults so deeply in code.

We could just set this as the default in the config - then we don't need to add it everywhere we use the server? Less chance for divergence/bugs?

anna-parker: ok! added to the defaults - its a bit ugly though as the server should only be set if segment_identification is set

ingest/Snakefile
```diff
@@ -60,7 +60,12 @@ if SEGMENTED:
...
             dataset_server_map[segment] = segment_identification.get(
                 "nextclade_dataset_server_map", {}
-            ).get(segment, segment_identification.get("nextclade_dataset_server"))
+            ).get(
+                segment,
+                segment_identification.get(
+                    "nextclade_dataset_server", config.get("nextclade_dataset_server")
+                ),
+            )
             dataset_name_map[segment] = segment_identification.get(
```
corneliusroemer, line 69: Not a fan of this code here with the many nested `.get`s - seems to work but this should probably be refactored at some point into a config.py script that is not Snakemake so we can use full Pylance on it.

anna-parker: lets maybe make an issue for this?

anna-parker: https://github.com/loculus-project/loculus/issues/5074

PR #5036, opened 2025-09-16
Feature branch for moving stuff from E2E to integration tests

theosanderson:


integration-tests/tests/specs/auth/logout.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+        await authPage.createAccount(testAccount);
+        await page.waitForLoadState('networkidle');
+
+        await authPage.logout();
+
+        await expect(page).toHaveURL(/\/logout$/);
+        await expect(page.getByText('You have been logged out')).toBeVisible();
+
+        const cookies = await page.context().cookies();
+        const authCookies = cookies.filter((cookie) => AUTH_COOKIE_NAMES.includes(cookie.name));
```
fengelniederhammer, line 25: Do we want to check whether those cookies are set before the logout? 

Testing for absence is always dangerous. You might end up with a test that does nothing but is still green.

fengelniederhammer (approved):
LGTM :+1:

integration-tests/tests/specs/auth/logout.spec.ts
theosanderson: Totally agreed but prefer to decouple this since I think it's a pre-existing issue and I don't have bandwidth this sec - made an issue!

PR #5047, opened 2025-09-17
feat(website): display configurable group creation guidelines

theosanderson:
## Summary
- add a `groupCreationGuidelines` option to the website configuration, including Helm template rendering and values schema validation
- pass the configured guidelines into the group creation flow and surface them above the form in the shared `GroupForm`
- cover the new display path with a unit test to ensure guidelines render when provided

## Testing
- npm run format
- npm run test
- npm run check-types

------
https://chatgpt.com/codex/tasks/task_e_68cad9b9fb188325a355fba467c43487

### DIFF
--- a/kubernetes/loculus/values.schema.json
+++ b/kubernetes/loculus/values.schema.json
@@ -798,6 +798,14 @@
       "type": ["string", "null"],
       "description": "A custom welcome message to be shown on the landing page"
     },
+    "groupCreationGuidelines": {
+      "groups": ["general"],
+      "type": "array",
+      "items": {
+        "type": "string"
+      },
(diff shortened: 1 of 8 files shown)


kubernetes/loculus/values.schema.json
```diff
@@ -798,6 +798,14 @@
       "type": ["string", "null"],
       "description": "A custom welcome message to be shown on the landing page"
     },
+    "groupCreationGuidelines": {
```
chaoran-chen, line 801: Would it make sense to have this again as an HTML as we do for a few other fields or do you think that it's better to have an array of non-HTML strings here?

PR #5055, opened 2025-09-19
refactor(website): don't duplicate the state in useUrlParamState

fengelniederhammer:
It's a derived state, let's make it a proper derived state. We don't need a separate, intermediate state that we need to be careful about syncing correctly.

### PR Checklist
~~- [ ] The implemented feature is covered by appropriate, automated tests.~~
  - `halfScreen` and `selectedSeq` still seem to work as before

🚀 Preview: https://useurlparamstate.loculus.org


theosanderson (approved):
Very satisfactory diff - thanks!

PR #5078, opened 2025-09-22
chore(website): ignore node modules during prettier runs

theosanderson:
## Summary
- ensure Prettier skips the local node_modules directory by adding it to the project's `.prettierignore`
- update the formatting scripts so that Prettier respects both the repository `.gitignore` and the project's `.prettierignore`, preventing third-party bundles from being parsed

## Testing
- npm run test
- npm run check-types
- npm run format

------
https://chatgpt.com/codex/tasks/task_e_68d1be70baec83258f878ddc335d2088


theosanderson:
This should already be achieved with the gitignore bit I think - thanks for clarifying above

PR #5085, opened 2025-09-23
feat(website): overhauled navigation and other design improvements

theosanderson:
This PR improves the design by:
- creating a clear hierarchy of Organisms -> [Browse / Submit]
- making navigation buttons have predictable behaviour (previously the Browse or Submit buttons did different things in different contexts)
- navigation now indicates what page you are currently on
- when first creating a group as part of an initial submission you are now redirected back to continue your submission

image

I won't merge until changes on the Pathoplexus side are ready to improve the front page to make up for the lack of a "Submit" button here

🚀 Preview: https://nav-exp5.loculus.org

### DIFF
--- a/website/src/layouts/BaseLayout.astro
+++ b/website/src/layouts/BaseLayout.astro
@@ -84,7 +84,7 @@ const lastTimeBannerWasClosed = Astro.cookies.get('lastTimeBannerWasClosed')?.va
                                 }}
                             />
                         </a>
-                        <a href='/' class='fancytitle leading-tight tracking-tight pb-1'>{websiteName}</a>
+                        <a href='/' class='fancytitle leading-tight tracking-tight pb-2'>{websiteName}</a>
                     </div>
                     <Navigation
                         implicitOrganism={implicitOrganism}


website/src/components/Navigation/NavigationTab.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+
+interface NavigationTabProps {
+    isActive?: boolean;
+    children: React.ReactNode;
+    as?: 'button' | 'a';
+    href?: string;
+    onClick?: () => void;
+    className?: string;
+}
```
chaoran-chen, line 10:
```suggestion
type NavigationTabProps = {
    isActive?: boolean;
    children: React.ReactNode;
    as?: 'a';
    href?: string;
    className?: string;
} |  {
    isActive?: boolean;
    children: React.ReactNode;
    as?: 'button';
    onClick?: () => void;
    className?: string;
}
```

Would that be a cleaner typing? This would allow removing the check `&& href` below which I found a bit strange as it led a `button` to be used if no `href` is provided even if the component user explicitly sets `as='a'`.

theosanderson: @claude try to implement this

website/src/components/Navigation/OrganismNavigation.tsx
```diff
@@ -0,0 +1,89 @@
...
+                <MenuItems className='absolute left-0 mt-1 w-56 bg-white rounded-lg shadow-lg border border-gray-200 z-50 focus:outline-none'>
+                    <div className='py-1'>
+                        {knownOrganisms.map((organism) => {
+                            const isActive = currentOrganism?.key === organism.key;
+
+                            return (
+                                <MenuItem key={organism.key}>
+                                    {({ focus }) => (
+                                        <a
+                                            href={`/${organism.key}/search`}
```
chaoran-chen, line 54: Can we reuse `routes.searchPage()`?

website/src/components/Navigation/OrganismNavigation.tsx
```diff
@@ -0,0 +1,89 @@
...
+                leaveTo='transform opacity-0 scale-95'
+            >
+                <MenuItems className='absolute left-0 mt-1 w-56 bg-white rounded-lg shadow-lg border border-gray-200 z-50 focus:outline-none'>
+                    <div className='py-1'>
+                        {knownOrganisms.map((organism) => {
+                            const isActive = currentOrganism?.key === organism.key;
+
+                            return (
+                                <MenuItem key={organism.key}>
+                                    {({ focus }) => (
```
chaoran-chen, line 52: This is quite deeply nested, would it make sense to extract this part into a separate component?

website/src/components/Navigation/SandwichMenu.tsx
```diff
@@ -8,50 +9,94 @@ import AccessionSearchBox from './AccessionSearchBox';
...
-                            {topNavigationItems.map(({ text, path }) => (
-                                <OffCanvasNavItem key={path} text={text} level={1} path={path} />
+                        <div className='flex-grow divide-y-2 divide-gray-300 border-y-2 border-gray-300'>
+                            <div className='py-3'>
+                                <h3 className='ml-4 mb-3 font-semibold text-gray-700'>Organisms</h3>
+                                <div className='ml-4 space-y-1'>
+                                    {knownOrganisms.map((organism) => (
+                                        <a
+                                            key={organism.key}
+                                            href={`/${organism.key}/search`}
```
chaoran-chen, line 61: Same as above: let's use `routes.searchPage()`?

website/src/components/Navigation/OrganismNavigation.tsx
```diff
@@ -0,0 +1,89 @@
...
+                                        <a
+                                            href={`/${organism.key}/search`}
+                                            className={`
+                                                flex items-center gap-3 px-4 py-2 text-sm transition-colors
+                                                ${
+                                                    isActive
+                                                        ? 'bg-primary-100 text-gray-900 font-semibold'
+                                                        : 'text-gray-700'
+                                                }
+                                                ${focus && !isActive ? 'bg-gray-50 text-gray-900' : ''}
```
chaoran-chen, line 62: At other locations in the PR, the classes were first stored in separate variables such as `baseClasses` and `stateClasses`. I found that quite readable, would it make sense to do it here as well?

website/src/components/User/GroupCreationForm.tsx
```diff
@@ -12,19 +13,35 @@ interface GroupManagerProps {
     accessToken: string;
 }
 
+const getContinueSubmissionFromLocation = (): ContinueSubmissionIntent | undefined => {
+    if (typeof window === 'undefined') {
+        return undefined;
```
chaoran-chen, line 18: Is this a branch that is expected or an error case?

website/src/components/User/GroupPage.tsx
```diff
@@ -63,12 +66,44 @@ const InnerGroupPage: FC<GroupPageProps> = ({
         queryFn: () => fetchSequenceCounts(groupId, clientConfig, organisms),
     });
 
+    const continueSubmissionCta = useMemo(() => {
```
chaoran-chen, line 69: What does `Cta` mean?

website/src/routes/navigationItems.ts
```diff
@@ -1,34 +1,47 @@
...
         return [browseItem];
     }
 
     const submitItem = {
-        text: 'Submit',
-        path:
-            organism !== undefined
-                ? routes.submissionPageWithoutGroup(organism)
-                : routes.organismSelectorPage('submission'),
+        text: 'Submit sequences',
```
chaoran-chen, line 42: "Browse data" and "Submit sequences" are quite long, in particular for small screens. I've a small preference for sticking with the shorter "Browse" and "Submit"

Image

chaoran-chen (commented):
Thanks very much! I've a few minor comments and questions, otherwise, it looks great!

website/src/routes/navigationItems.ts
theosanderson: I agree that we should improve the mobile experience here (I also think we should change Crimean Co.. to CCHF - I have a draft PR to add tooltips to organisms to help with this). I also agree that these may not be ideal final tiles.

I'd prefer to tackle this downstream and not block on it atm. (Personally I think something a bit longer than "Browse" and "Submit" might be good - esp. given that there were concerns about visibility).

website/src/components/User/GroupPage.tsx
theosanderson: "Call to action" - (this is what the LLM wrote - but IMO is reasonable to avoid overlong names and could become a standard pattern in the codebase)

website/src/components/User/GroupCreationForm.tsx
theosanderson: This will happen during SSR but not locally - we could improve improve this by passing the path through from astro so that we don't have to look at it on the client. Claude, let's try to do that.

theosanderson:
Thanks for the review! @claude aim to address chaoran's points except the one's I've suggested leaving til later

website/src/pages/user/createGroup.astro
```diff
@@ -7,8 +7,9 @@ import { getAccessToken } from '../../utils/getAccessToken';
 const accessToken = getAccessToken(Astro.locals.session)!;
 
 const clientConfig = getRuntimeConfig().public;
+const searchParams = Astro.url.searchParams.toString();
```
theosanderson, line 10: (though this could be done better IMO it's not worth another round of iteration)

website/src/components/Navigation/OrganismNavigation.tsx
theosanderson: So Claude failed to take this on and I think to me it makes sense to have the full headless UI hierarchy in this one file so I'll probably leave this unless you think it's a big deal

website/src/pages/user/createGroup.astro
theosanderson: (I did confirm that the functionality still works)

integration-tests/tests/pages/navigation.page.ts
theosanderson, line 1: Do not merge until PPX PR ready

corneliusroemer (commented):
Review in progress, not finished yet:

- I think the old top-right was better with text more centered. 
Old: image New: Google Chrome 2025-10-07 15 57 05
- Inconsistent underlining vs not underlining. I probably prefer not to underline as the background already changes: 

- Not sure about the fact that a selected tab doesn't get a grey background on hover - in contrast to unselected tabs:  
Not sure if it's better to not use grey background at all or use it for all.
- On dummy organism, there's insufficient padding at the end of the last filter at the bottom, other organisms have more padding there (which is better IMO):
  Google Chrome 2025-10-07 16 06 55 image
- Having the magnifying glass on the left (old) is IMO better than on the right as it causes less layout shift when clicking on it. This is new: 
 vs old: 
- Magnifying glass is now not vertically centered (visually), it was better previously. New: 
Google Chrome 2025-10-07 16 13 30, old: Google Chrome 2025-10-07 16 13 33. Also the color mismatches now, glass is still blue, but text for headings has switched to grey.

PR #5091, opened 2025-09-24
feat(website): multi pathogen support: show correct segments in download modal

### LINKED ISSUE #4997: Multi pathogen support: show correct segments in download modal
Image

What do we show here? On the search page, we don't have a "suborganism" yet.

We can disable the download modal as long as the search has not been narrowed down to a specific suborganism yet. Although that feels rather restrictive.

If not, then we probably need to show the full segment/gene names (`<suborganism>-<segment/gene>`) as download options?

We could also add the option to narrow it down first:
* Offer a dropdown which suborganism should be downloaded
* Only then let the user select which segment/gene should be downloaded
* Set the suborganism in the filter when sending the download request to LAPIS.

fengelniederhammer:
resolves #4997

Make sure that the correct segments and genes are displayed and downloaded. All downloads should now work in the multi pathogen case.

* When downloading raw sequences when no suborganism is selected yet, it will download all segments that are not null (#5092 is still open here).
* When downloading raw sequences when a suborganism is selected, it will download the corresponding segment only.

### Screenshot

image

### PR Checklist
  - sequence download works
    - for EV as expected
    - for west nile as before

### DIFF
--- a/website/src/components/SearchPage/DownloadDialog/DownloadForm.tsx
+++ b/website/src/components/SearchPage/DownloadDialog/DownloadForm.tsx
@@ -298,7 +298,7 @@ function getSequenceNames(
         return {
             nucleotideSequences: [],
             genes: [],
-            useMultiSegmentEndpoint: false, // LAPIS is multisegmented, since we're in multi pathogen case here. Use the "download all segments" endpoint which we get by pretending we're single segmented.
+            useMultiSegmentEndpoint: false, // When no suborganism is selected, use the "all segments" endpoint to download all available segments, even though LAPIS is multisegmented. That endpoint is available at the same route as the single segmented endpoint.
         };
     }
 
--- a/website/src/components/SearchPage/DownloadDialog/DownloadForm.tsx
+++ b/website/src/components/SearchPage/DownloadDialog/DownloadForm.tsx
@@ -9,13 +9,13 @@ import { routes } from '../../../routes/routes.ts';
 import { ACCESSION_VERSION_FIELD } from '../../../settings.ts';
 import type { Metadata, Schema } from '../../../types/config.ts';
 import { type ReferenceGenomesSequenceNames, SINGLE_REFERENCE } from '../../../types/referencesGenomes.ts';
-import type { SequenceName } from '../../../utils/sequenceTypeHelpers.ts';
 import {
     getMultiPathogenNucleotideSequenceNames,
     getMultiPathogenSequenceName,
     getSinglePathogenSequenceName,
     isMultiSegmented,
-} from '../../SequenceDetailsPage/SequencesDisplay/getSequenceNames.tsx';
+    type SequenceName,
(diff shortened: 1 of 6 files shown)


website/src/components/SequenceDetailsPage/SequencesDisplay/getSequenceNames.tsx
```diff
@@ -1,5 +1,11 @@
 import { type ReferenceGenomesSequenceNames, SINGLE_REFERENCE } from '../../../types/referencesGenomes.ts';
-import type { SequenceName } from '../../../utils/sequenceTypeHelpers.ts';
+import {
+    getMultiPathogenNucleotideSequenceNames,
+    getMultiPathogenSequenceName,
```
theosanderson, line 4: I'm wondering throughout this if `SequenceName` is the right name for this. I appreciate it was an existing name but maybe its more confusing now we have to do more with it. To me the sequence name would be e.g. the accession of a sequence entry. This is more segment name or contig name or something.

website/src/components/SearchPage/DownloadDialog/DownloadForm.tsx
```diff
@@ -269,3 +280,32 @@ export const DownloadForm: FC<DownloadFormProps> = ({
...
+            genes: genes.map(getSinglePathogenSequenceName),
+            useMultiSegmentEndpoint: isMultiSegmented(nucleotideSequences),
+        };
+    }
+
+    if (selectedSuborganism === null) {
+        return {
+            nucleotideSequences: [],
+            genes: [],
+            useMultiSegmentEndpoint: false, // LAPIS is multisegmented, since we're in multi pathogen case here. Use the "download all segments" endpoint which we get by pretending we're single segmented.
```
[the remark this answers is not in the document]
theosanderson: Will make a new comment with a suggestion but nbd

theosanderson, line 301:
```suggestion
            useMultiSegmentEndpoint: false, // LAPIS is multisegmented, since we're in multi pathogen case here, but we want to use the "download all segments" endpoint which we get by pretending we're single segmented.
```

I think linking the two sentences with a "but" helps my brain make sense of it, but nbd

website/src/components/SequenceDetailsPage/SequencesDisplay/getSequenceNames.tsx
theosanderson: We discussed and also want to rejig `SequenceName` to be something like `SegmentInfo | GeneInfo` or equivalent - we will tackle in a future PR

website/src/components/SearchPage/DownloadDialog/DownloadForm.tsx
fengelniederhammer: I already changed the comment to something else

PR #5093, opened 2025-09-24
feat(deployment): Add liveness probe to silo-deployment.yaml

theosanderson:
Uses https://github.com/GenSpectrum/LAPIS-SILO/pull/979 to ensure we restart SILO if it gets unhealthy (sorry for the delay I didn't notice this get added at the time)

🚀 Preview: https://silo-liveness.loculus.org


theosanderson:
I guess we need to bump silo

PR #5098, opened 2025-09-25
feat!(config): define lineage hierarchy in prepro where versions are defined for better DX

corneliusroemer:
resolves #

### PR Checklist


anna-parker:
A thought: while we are doing this should we stop silo preprocessing from crashing if there isn't a lineage defined for that pipeline version and have it fall back to the latest pipeline version lineage file?

theosanderson:
(i would separate those two things out)

PR #5100, opened 2025-09-25
feat(deployment): move SILO importer into python package and dedicated image

theosanderson:
This PR introduces a Python-based SILO importer to replace the existing shell script implementation. The new importer is packaged as a dedicated Python package with its own Docker image, improving maintainability and reliability of the data import process.

- Move SILO import logic from shell scripts to a structured Python package
- Create dedicated container image for the Python importer
- Refactor Kubernetes deployment to use separate containers for importing and preprocessing

As I mentioned a while back - anyone is welcome to improve this PR - I don't consider myself an owner/assignee on this feature and would prefer not to have to do much more on this personally.

🚀 Preview: https://pyimport.loculus.org


kubernetes/loculus/silo_import_wrapper.sh
```diff
@@ -1,54 +1,50 @@
...
+  echo "$(date '+%Y-%m-%dT%H:%M:%S%z') [silo-runner] $*"
+}
 
-input_data_dir="/preprocessing/input"
-current_etag_path="$input_data_dir/etag.txt"
-last_hard_refresh_time_path="$input_data_dir/last_hard_refresh_time.txt"
+INPUT_DIR="/preprocessing/input"
+RUN_FILE="$INPUT_DIR/run_silo"
+DONE_FILE="$INPUT_DIR/silo_done"
+SLEEP_SECONDS="${SILO_RUNNER_SLEEP_SECONDS:-1}"
```
anna-parker: lets maybe set this to 30seconds like it is now

kubernetes/loculus/templates/silo-deployment.yaml
```diff
@@ -65,15 +65,45 @@ spec:
             periodSeconds: 10
             failureThreshold: 3
             timeoutSeconds: 5
+        - name: silo-importer
```
anna-parker: Anya's thoughts: maybe we shouldn't split this up but have the silo-import job also run silo preprocessing - this way we don't need to make the bash script aware of the sentinel file structure and can do all that in one place.

kubernetes/silo-import/Dockerfile
anna-parker: I think the silo-import job should maybe be in its own folder and not under the kubernetes folder

kubernetes/silo-import/src/silo_import/config.py
```diff
@@ -0,0 +1,54 @@
...
+import os
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Dict, Optional
+
+
+@dataclass(frozen=True)
+class ImporterConfig:
+    backend_base_url: str
+    lineage_definitions: Optional[Dict[str, str]]
```
anna-parker: In my understanding this should be a dictionary from version to url 
```suggestion
    lineage_definitions: dict[int, str] | None
```

kubernetes/silo-import/src/silo_import/config.py
```diff
@@ -0,0 +1,54 @@
...
+
+    @classmethod
+    def from_env(cls) -> "ImporterConfig":
+        env = os.environ
+        backend_base_url = env.get("BACKEND_BASE_URL")
+        if not backend_base_url:
+            raise RuntimeError("BACKEND_BASE_URL environment variable is required")
+
+        lineage_definitions_raw = env.get("LINEAGE_DEFINITIONS")
+        lineage_definitions: Optional[Dict[str, str]] = None
```
anna-parker:
```suggestion
        lineage_definitions: dict[int, str] | None = None
```

kubernetes/silo-import/src/silo_import/paths.py
```diff
@@ -0,0 +1,34 @@
...
+    preprocessing_dir: Path
+    input_dir: Path
+    output_dir: Path
+    lineage_definition_file: Path
+    silo_input_data_path: Path
+    run_silo: Path
+    silo_done: Path
+
+    @classmethod
+    def from_root(cls, root: Path) -> "ImporterPaths":
```
anna-parker:
```suggestion
    def from_root(cls, root: Path) -> ImporterPaths:
```

kubernetes/silo-import/Dockerfile
anna-parker: started a PR with some formatting suggestions and this move: https://github.com/loculus-project/loculus/pull/5377

kubernetes/loculus/templates/silo-deployment.yaml
theosanderson: That's fine but then you'll need to compile Silo here

anna-parker (commented):
This is amazing - thanks for doing this @theosanderson!!

I made https://github.com/loculus-project/loculus/pull/5377 to address most of my nitpicks - I have resolved all the comments I made here that would be fixed by that PR. I left some suggestions - if you are ok with them we could possibly make those changes in a follow up PR - but this PR here includes all the changes I think would be good to do before we merge.

anna-parker (approved):
Approving - but would be good to still check that we are all on board with changing pipeline version to always be an int before we merge :-)

kubernetes/loculus/templates/silo-deployment.yaml
anna-parker: yeah sounds good to keep like this for now

PR #5102, opened 2025-09-25
fix(config): don't enforce schema on defaultOrganismConfig which serves as helper only, move non-shared stuff out of defaultOrganismConfig

corneliusroemer:
`defaultOrganismConfig` isn't used directly - it just serves as a place to put shared conifg, so it doesn't need to follow any particular schema

It's an error in the schema that defaultOrganismConfig must be a valid organism. There's no reason it should have to be valid as it's only present for inheritance purposes and DRYness of values.yaml. It doesn't make sense, for example, to enforce a `version` to be present - a default version doesn't make sense - neither does a default name etc. Only things that are meaningfully shared between more than one organism should live in defaultOrganism config.

Hence this PR relaxes what's allowed in defaultOrganismConfig. The schema still checks each individual organism config, so no truly invalid/problematic schema will fail to be noticed.

It also moves non-shared stuff out of defaultOrganismConfig

The overly strict schema prevented some clean-up on PPX: https://github.com/pathoplexus/pathoplexus/actions/runs/18014261615/job/51254960342?pr=678

### DIFF
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -56,12 +56,6 @@ lineageSystemDefinitions:
 defaultOrganismConfig: &defaultOrganismConfig
   schema: &schema
     loadSequencesAutomatically: true
-    organismName: "Ebola Sudan"
-    image: "/images/organisms/ebolasudan_small.jpg"
-    linkOuts:
-      - name: "Nextclade"
-        maxNumberOfRecommendedEntries: 5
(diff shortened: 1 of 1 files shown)


theosanderson (approved):
In some ways I think `defaultOrganismConfig` shouldn't really be part of the schema. We could create a new top level item `uncontrolled:` which can contain anything you want including that. But this seems fine

PR #5107, opened 2025-09-26
fix(deposition): don't single quote subprocess.run() args as we set shell=False

corneliusroemer:
I fixed the ena-webin-cli wrapper on bioconda by using execvp instead of shell invokation.

This way the wrapper can handle single quotes in center names.

However, the new wrapper reveals a bug in our current code: we were single quoting args in `subprocess.run()` calls. As subprocess doesn't go through the shell (we set `shell=False`) the quotes stay and cause issues.

Hence, this PR removes the single quoting which fixes things.

I also adapt our integration tests to ensure single quotes in center names now work - the original bug I was intending to fix with the wrapper edit.

### DIFF
--- a/ena-submission/environment.yml
+++ b/ena-submission/environment.yml
@@ -10,7 +10,7 @@ dependencies:
   # Extra dependencies
   - beautifulsoup4
   - click
-  - ena-webin-cli =9.0.1
+  - ena-webin-cli=9.0.1=hdfd78af_1
   - fastapi
   - pydantic
   - jsonlines


ena-submission/test/text_project_request.xml
```diff
@@ -1,8 +1,8 @@
 <PROJECT_SET>
-    <PROJECT center_name="Test institution" alias="2:Test organism:Test suffix">
-        <NAME>Test scientific name: Genome sequencing by Test group, Test institution</NAME>
-        <TITLE>Test scientific name: Genome sequencing by Test group, Test institution, Basel, Switzerland</TITLE>
-        <DESCRIPTION>Automated upload of Test scientific name sequences submitted by Test group, Test institution, Basel, Switzerland to Loculus</DESCRIPTION>
+    <PROJECT center_name="Test institution with funny characters \ / &apos;&quot;!@#$%^&amp;*()" alias="2:Test organism:Test suffix">
```
anna-parker, line 2: love the test but Im a bit surprised that `'` becomes `&apos;` in the xml? Is there a way to avoid this? I'm worried ENA will see center_name=`My team's project` in teh assembly and center_name=`My team&apos;s project` in the bioproject and not link them

anna-parker (approved):
approving as this is a clear improvement to current state - thanks @corneliusroemer!

ena-submission/test/text_project_request.xml
corneliusroemer: Yep - it's just escaping :)

PR #5112, opened 2025-09-28
chore(backend): implement Gradle dependency locking & let spring-boot manage flyway-database-postgresql version

corneliusroemer:
Add comprehensive dependency locking to prevent version conflicts like
those encountered in #5076 where Dependabot dependency updates caused
Flyway version mismatches and test failures.


backend/gradle.lockfile
```diff
@@ -0,0 +1,309 @@
+# This is a Gradle generated file for dependency locking.
```
anna-parker, line 1: @corneliusroemer can you explain to me how this file was generated? Just trying to fully understand this change

Issue #5135, opened 2025-09-29
Remove feature that locks `/get-original-metadata` when `auxTable` is non-empty


theosanderson:
https://github.com/loculus-project/loculus/blob/aeddcdfb6e83d195e3a00e5fd1c5df437dd497b9/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt#L428

makes `get-original-metadata` not work if the aux table is non-empty - it returns `423` instead. I'm uncertain that such a feature is needed.

This feature (IIUC) caused a problem @anna-parker saw on GenSpectrum Loculus where it stopped all submissions working because somehow stuff got stuck in the aux table, which of course would be a big problem

theosanderson:
OK, this was added in https://github.com/loculus-project/loculus/pull/2846 which provides more context on why

theosanderson:
I think I mostly understand why it happens. The whole processing of an upload to the main database is one long operation and if it is interrupted there is no way (at present) it can be restarted. And yeah, one could improve that but I don't think it will have especially bad effects.

chaoran-chen:
There is a very similar issue #4852 but as you've done more analysis of the issue, I'll close the old one :)

theosanderson:
> so it sounds like we think the aux table should be able to clear on its own

I don't think I've said this - as I understand it if for some reason the backend dies during a submission before it has released data into the normal tables and cleared the aux tables there is no mechanism for the aux tables to be cleared

theosanderson:
I asked an LLM and it confirmed that our existing kubernetes set up should entirely prevent concurrency of ingest. We just had that super confusing bug with the reloader and its interaction with old Cron jobs. I will make a PR to action this issue (which we can discuss!)

anna-parker:
>  if for some reason the backend dies during a submission before it has released data into the normal tables and cleared the aux tables there is no mechanism for the aux tables to be cleared

So I guess this was the problem on GS - I guess we need some way to clear the aux table. We could clear the aux table when the backend is restarted but if we ever ran two backend instances this could be an issue (is this actually ever possible? could we ever have two backends running?) otherwise we could clear entries from the aux table if they are in the aux table for too long?

theosanderson:
> could we ever have two backends running?

yes - the backend should be stateless - actually I think we already do

> I guess we need some way to clear the aux table

[...]

PR #5148, opened 2025-10-01
fix(backend): allow original metadata retrieval when aux table is non-empty

theosanderson:
resolves https://github.com/loculus-project/loculus/issues/5135

In the past we had a very confusing bug which led ingest to submit the same data multiple times. The true issue was that a kubernetes-reloader we were using triggered old remnant cron jobs to start immediately on roll-out of a new deployment and at the same time the ingest deployment also started. Then both would submit data and so we'd end up with two copies of the data. We battled this bug for a long time not understanding where the second data was coming from. I believe that as part of that we theorised that it might be coming from the time when data is in the aux tables before it moves to the main database table. So we merged https://github.com/loculus-project/loculus/pull/2846 which locks metadata retrieval when the aux table is non-empty. But now we have fixed the bug this should no longer required, and it would be good to remove this as it caused problems by blocking ingest when some stuff was left in the aux tables due to the backend restarting mid-submission.

- remove the HTTP 423 guard from `/get-original-metadata` so the endpoint continues to return data during uploads
- delete the auxiliary table polling helper that powered the lock check
- update the original metadata endpoint test to reflect the always-available behavior

------
https://chatgpt.com/codex/tasks/task_e_68dd274571248325969ba7b1d859c37a

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
@@ -406,10 +406,6 @@ open class SubmissionController(
             ),
         ],
     )
-    @ApiResponse(
-        responseCode = "423",
-        description = "Locked. New sequence entries are currently being uploaded.",
-    )
     @GetMapping("/get-original-metadata", produces = [MediaType.APPLICATION_JSON_VALUE])
     fun getOriginalMetadata(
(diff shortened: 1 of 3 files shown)


anna-parker (approved):
We should also clean up the data in the auxTable now if `uploadedAtColumn` is 2hours ago etc.

Started here with chatGPT: https://github.com/loculus-project/loculus/pull/5189

PR #5152, opened 2025-10-01
feat(website): allow cors on /loculus-info endpoint, add options route

theosanderson:
- add a reusable set of CORS headers for the `/loculus-info` API response, allowing access from any origin and explicitly listing supported request headers and methods
- attach the CORS headers to the JSON payload returned by the GET handler and implement an OPTIONS handler so that preflight requests receive the same permissive policy

------
https://chatgpt.com/codex/tasks/task_e_68dd73aa37fc8325b52713e1a1de0f7b

🚀 Preview: https://codex-add-cors-for-all-or.loculus.org

### DIFF
--- a/website/src/pages/loculus-info/index.ts
+++ b/website/src/pages/loculus-info/index.ts
@@ -3,6 +3,15 @@ import type { APIRoute } from 'astro';
 import { getRuntimeConfig, getWebsiteConfig } from '../../config';
 import { getAuthBaseUrl } from '../../utils/getAuthUrl';
 
+const corsHeaders = {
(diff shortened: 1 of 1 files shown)


fengelniederhammer (approved):

corneliusroemer (commented):
From the PR description I can't figure out the motivation for this - I suppose you want to access this from a tool under a different domain? There are probably also other endpoints we might want to add CORS for?

theosanderson:
>  I suppose you want to access this from a tool under a different domain

Yes, that's always a good guess for CORS :)

> There are probably also other endpoints we might want to add CORS for?

We have some already - and yeah I'm sure there are more we can add

PR #5164, opened 2025-10-04
chore(claude): create/make CLAUDE.md symlinks to AGENTS.md so claude auto-loads those into context

corneliusroemer:
Adds CLAUDE.md symlinks pointing to AGENTS.md in backend/, website/, and integration-tests/ and root directories for better compatibility with Claude Code.

Claude otherwise doesn't get the content added to context and it rarely explicitly reads those files. So current approach of top-level CLAUDE.md saying "look at agents.md" didn't work.


corneliusroemer:
> weird github doesn't show these diffs very well

Agreed. If you click on the file you can see it's a symbolic link - but that requires extra work.

Google Chrome 2025-10-05 12 43 23

Issue #5181, opened 2025-10-06
Author affliation search is not working


anna-parker:
Selecting an author affiliation makes the search page go blank: 

https://github.com/user-attachments/assets/f7c196fe-8332-47b3-a4d9-fb2979d588b0

anna-parker:
Error logs:
```
Uncaught TypeError: e.replace is not a function
    at aw (SearchFullUI.BMR98Z3Q.js:1:9726)
    at gu.toApiParams (SearchFullUI.BMR98Z3Q.js:1:8359)
    at SearchFullUI.BMR98Z3Q.js:67:25089
    at Object.vs [as useMemo] (index.BDobkqVR.js:22:21307)
    at r.useMemo (index.BOC3FD3Q.js:9:6116)
    at UO (SearchFullUI.BMR98Z3Q.js:67:25074)
    at Ru (index.BDobkqVR.js:22:16972)
    at Hu (index.BDobkqVR.js:24:3154)
    at sa (index.BDobkqVR.js:24:44850)
    at la (index.BDobkqVR.js:24:39783)

search?visibility_au…Affiliations=true:1 Uncaught (in promise) Error: A listener indicated an asynchronous response by returning true, but the message channel closed before a response was received

```

theosanderson:
@claude Any idea what's up?

theosanderson:
It's fine on Loculus main

anna-parker:
> It's fine on Loculus main

we don't use the same values in the config file - see details in https://github.com/loculus-project/loculus/pull/5182 -> reproduced issue there!

theosanderson:
👍 - I didn't know we were ever combining subString search and autocomplete, or that that had been possible. I'm not entirely sure how we'd do this (but will be great to see if your PR works). But for now we could easily just disable autocomplete on PPX.

theosanderson:
@claude Why did the typechecker not prevent this bug?

PR #5184, opened 2025-10-06
feat(config): prevent autocomplete and enableSubstringSearch from being set together in json schema

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/5181

Add a check to the json.schema to prevent autocomplete being used with enableSubString search

### PR Checklist
- [x] Any manual testing that has been done is documented (i.e. what exactly was tested?) -> test by adding both and ensuring that helm linter fails: https://github.com/loculus-project/loculus/actions/runs/18280002544/job/52040666765?pr=5184


corneliusroemer (approved):
Nice, while we're at it - there are possibly a few more that are incompatible? I'll have a think.

PR #5189, opened 2025-10-06
feat(backend): clean up auxTable entries that are older than 2 hours

anna-parker:
resolves #

### PR Checklist

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/service/submission/CleanUpAuxTableTask.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/service/submission/CleanUpAuxTableTask.kt
@@ -1,0 +1,30 @@
+package org.loculus.backend.service.maintenance
+
+import org.loculus.backend.log.AuditLogger
+import org.springframework.scheduling.annotation.Scheduled
+import org.springframework.stereotype.Component
+import java.time.Instant
+import java.time.temporal.ChronoUnit
+
+private val log = mu.KotlinLogging.logger {}
+
+@Component
+class CleanUpAuxTableTask(
+    private val auxTableService: MetadataUploadAuxTable,
+    private val auditLogger: AuditLogger,
+) {
+
+    /**
+     * Runs every hour and deletes auxTable entries older than 2 hours.
+     */
+    @Scheduled(fixedDelay = 1, timeUnit = java.util.concurrent.TimeUnit.HOURS)
+    fun task() {
+        val cutoff = Instant.now().minus(2, ChronoUnit.HOURS)
+        val deletedCount = auxTableService.deleteOlderThan(cutoff)
+
+        if (deletedCount > 0) {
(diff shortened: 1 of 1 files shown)


backend/src/main/kotlin/org/loculus/backend/service/submission/CleanUpAuxTableTask.kt
```diff
@@ -0,0 +1,40 @@
...
+    private val dateProvider: DateProvider,
+    private val auditLogger: AuditLogger,
+) {
+
+    /**
+     * Runs every hour and deletes auxTable entries older than 2 hours.
+     */
+    @Scheduled(fixedDelay = 1, timeUnit = java.util.concurrent.TimeUnit.HOURS)
+    fun task() {
+        val hourCutoff = 2L
```
theosanderson, line 26:
```suggestion
        val hourCutoff = 24L
```

While 2 hours is probably fine I can imagine some future situation where 2 hours might be a plausible time and this would be a really annoying cause-of-bug to have to chase down - hopefully 24 hours will never be a problem

theosanderson: I suspect we have timeouts so that atm this probably wouldn't work anyway which is why I said "future", but the idea of the aux tables is to stream into the DB from the upload (uncertain exactly to what extent that is the case atm). It is fairly plausible for someone on a slow connection (upload is typically slower than download) to be streaming their upload over a couple of hours.

backend/src/test/kotlin/org/loculus/backend/service/submission/CleanUpAuxTableTaskTest.kt
```diff
@@ -0,0 +1,81 @@
...
+        ).toLocalDateTime(DateProvider.timeZone)
+        uploadDatabaseService.batchInsertMetadataInAuxTable(
+            uploadId = uploadIdOld,
+            authenticatedUser = mockUser,
+            groupId = 1,
+            submittedOrganism = Organism("organism"),
+            uploadedMetadataBatch = listOf(MetadataEntry("submission id", mapOf("key" to "value"))),
+            uploadedAt = threeHoursOld,
+            null,
+        )
```
anna-parker, line 68:
```suggestion
        val oneDayOld = now.minus(
            25,
            DateTimeUnit.HOUR,
            DateProvider.timeZone,
        ).toLocalDateTime(DateProvider.timeZone)
        uploadDatabaseService.batchInsertMetadataInAuxTable(
            uploadId = uploadIdOld,
            authenticatedUser = mockUser,
            groupId = 1,
            submittedOrganism = Organism("organism"),
            uploadedMetadataBatch = listOf(MetadataEntry("submission id", mapOf("key" to "value"))),
            uploadedAt = oneDayOld,
            null,
        )
```

PR #5200, opened 2025-10-08
feat(website): multi pathogen support: edit page: handle preprocessing-assigned segments

### LINKED ISSUE #4999: Multi pathogen support: edit page
Image

What do we show?

fengelniederhammer:
resolves #4999

Change the sequence upload on the edit page and on the "Submit single sequence" page such that it doesn't assign segments anymore. This should happen in the preprocessing pipeline (#4847).
* This will only properly work once the backend is adapted accordingy. Until then, this PR will likely have failing integration tests.
* We could in principle extract the segment name / fasta header from the uploaded file. But that requires minor code changes in the website that I left for another PR (currently we only get the sequence of the fasta file, the header is discarded).

If we have original data on the edit page, then the segment name is taken from there. In all other cases, we simply enumerate them ("Segment 1", "Segment 2", etc.). That will also be part of the fasta header when submitting / the segment name when uploading edited data.

### Screenshot

CCHF "Submit single sequence" (`/cchf/submission/2/submit?inputMode=form`)

https://github.com/user-attachments/assets/6c092570-3fcf-4bf8-93f8-45c5cc598645

CCHF Edit page for a sequence entry that only has one segment in its original data:

https://github.com/user-attachments/assets/340471cb-b644-495b-b19e-26c5efb614fc

For EV this is rather boring. Either it has a sequence:
image

Or it doesn't:
image

### PR Checklist
  - "Submit single sequence"
    - works for CCHF - although we don't assign segments there anymore. We can only enumerate them.
    - works for EV
    - works for west nile as before
  - edit page
    - works for west nile as before
    - works for CCHF: segments that are present in the uploaded data remain present (but you can remove the files), it's possible to add additional segments until the maximum (3 in this case)
    - works for EV - the existing segment can be modified as for other single segmented viruses.

### DIFF
--- a/website/src/components/Edit/EditableSequences.spec.ts
+++ b/website/src/components/Edit/EditableSequences.spec.ts
@@ -59,7 +59,7 @@ describe('EditableSequences', () => {
         }
 
         expect(() => editableSequences.update('another key', 'GG')).toThrowError(
-            'Must not add more than 2 sequence file(s).',
+            'Maximum limit reached — you can add up to 2 sequence file(s) only.',
         );
     });
 
@@ -84,7 +84,7 @@ describe('EditableSequences', () => {
         expect(rows).deep.equals([{ label: 'Segment 1', value: 'ATCG', initialValue: null, key }]);
 
         expect(() => editableSequences.update('another key', 'GG')).toThrowError(
-            'Must not add more than 1 sequence file(s).',
+            'Maximum limit reached — you can add up to 1 sequence file(s) only.',
         );
     });
 


website/src/components/Edit/EditableSequences.ts
```diff
@@ -0,0 +1,127 @@
...
+    private readonly maxNumberOfRows: number;
+
+    public get rows(): Required<EditableSequenceFile>[] {
+        const rows = this.editableSequenceFiles.map((row, i) => ({
+            ...row,
+            label: row.label ?? `Segment ${i + 1}`,
+        }));
+        if (rows.length < this.maxNumberOfRows) {
+            rows.push({
+                label: `+ add new sequence`,
```
theosanderson, line 24: (I think we can style this thing quite a bit nicer - my mockup was intended more to illustrate the behaviour rather than a specific design -  but we don't need to block on it and I'll try to figure out something more specific)

website/src/components/Edit/EditableSequences.ts
```diff
@@ -0,0 +1,127 @@
...
+     * @param initialData The sequence entry to edit, from which the initial sequence data is taken.
+     * @param referenceGenomeLightweightSchema
+     */
+    static fromInitialData(
+        initialData: SequenceEntryToEdit,
+        referenceGenomeLightweightSchema: ReferenceGenomesLightweightSchema,
+    ): EditableSequences {
+        const existingDataRows = Object.entries(initialData.originalData.unalignedNucleotideSequences).map(
+            ([key, value]) => ({
+                label: `${key} segment`,
```
anna-parker: After I click edit on a multi-segment sequence and then submit again. Preprocessing fails - from the logs it looks like the unalignedSequences is sent to preprocessing with the keys `M segment`, `L segment` etc. instead of `M` and `S`

image

fengelniederhammer: That's because the code always showed "S segment" as the label there (when the segment name was "S"). But thinking about it, that probably doesn't make much sense anymore right? It would be better to show whatever the user used as FASTA header in the submission?

anna-parker:
I started reviewing but decided it was easiest to start https://github.com/loculus-project/loculus/pull/5382 with my suggestions :-)

PR #5204, opened 2025-10-09
refactor(backend): make a SQL view that assembles sequence entries from the aux tables

theosanderson:
- introduce a `aux_tables_as_sequence_entries ` view (via migration) that assembles metadata and sequence aux table contents into the JSON payload used during insertion - this simply moves the SQL logic from a big query in Kotlin `(mapAndCopy` below) into a view
- rename `UploadDatabaseService.mapAndCopy` to `insertFromAuxTablesIntoMainTable` so the method name reflects that it inserts staged data into `sequence_entries`, and update the submission flow to call the new method

We will need to change this view later due to the changes we want to make but IMO it's helpful to do so from this starting point.


chaoran-chen (approved):
Looks good to me, thanks!

PR #4904, opened 2025-08-28
wip - SILO 0.8

chaoran-chen:
resolves #

### PR Checklist
- [ ] All necessary documentation has been adapted.
- [ ] The implemented feature is covered by appropriate, automated tests.
- [ ] Any manual testing that has been done is documented (i.e. what exactly was tested?)

Issue #4639, opened 2025-07-08
Create a mock ENA and integration test deposition


theosanderson:
I feel like it shouldn't be too hard esp. with an LLM to create a mock ENA that provides enough functionality for us to test out deposition

PR #5225, opened 2025-10-14
feat(integration): add integration tests that run the ena cronjob

anna-parker:
resolves #

### Screenshot
Ensure that changes that would break the ena cronjob (e.g. https://github.com/loculus-project/loculus/pull/4904) will be caught by integration tests

### PR Checklist
- [x] Tested in https://github.com/loculus-project/loculus/pull/5221 that this test fails when the ena tests are broken e.g. due to a backend endpoint change

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -108,6 +108,18 @@ jobs:
           sed -n '/Running [0-9]\+ tests/,$p' output.txt >> $GITHUB_STEP_SUMMARY
           echo '```' >> $GITHUB_STEP_SUMMARY
           exit $EXIT_CODE
+      - name: Run get-ena-submission-list cronjob
+        run: |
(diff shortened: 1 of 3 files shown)


theosanderson (dismissed):
Copy pasting from Slack:

Basically I don't think we should keep adding more stuff to the test that looks like this, but it doesn't seem worth blocking for now so approving.

I think a lot more integration testing of deposition would be really nice some day: in the past I've wondered if we could create a basic [mock ENA](https://github.com/loculus-project/loculus/issues/4639), to facilitate that. Generally these sorts of tests will require e.g. submitting sequences, possibly revising them, which is stuff the integration test infra in playwright can do really well. So my suggestion would be to take this test and put it into a playwright test - there's a pattern to follow from the way we test the CLI. It would look something like [this](https://pastebin.com/0R8CtnLy) (llmed).

But it only takes 4 seconds so I guess I don't need to block on it for now

PR #5226, opened 2025-10-14
feat(deposition): add unit tests for the cronjob that uses the backend's get-released-data format

anna-parker:
This will help us to ensure we don't break anything if we ever refactor the cronjob.

The cronjob calls get-released-data and then uses that data to create a json with all data that we need to submit (this might change in future) - this means it partially transforms data. It also checks if data has already been submitted by us.

Previously we had no tests to cover this transformation. Or the database check. This unittest mocks slack calls, db checks and backend requests, however it checks that the JSON is formatted correctly and that we will send a slack notification with the correct text.

### PR Checklist
- ~Any manual testing that has been done is documented (i.e. what exactly was tested?)~ this does not touch any code but only adds tests

@maverbiest I know this is a lot to review - if you dont feel comfortable approving that is fine - but it would be cool if you could have a look at the code and ask me if you have any follow up questions :-)

### DIFF
--- a/ena-submission/scripts/test_get_ena_submission_list.py
+++ b/ena-submission/scripts/test_get_ena_submission_list.py
@@ -1,10 +1,11 @@
 # ruff: noqa: S101 (allow asserts in tests))
 import json
+import unittest
 from collections.abc import Iterator
-from types import SimpleNamespace
 from typing import Any
+from unittest.mock import ANY, Mock, patch
 
-import get_ena_submission_list as get_ena_submission_list_mod
+import get_ena_submission_list
 import orjsonl
 from deepdiff import DeepDiff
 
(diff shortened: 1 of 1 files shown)


corneliusroemer (commented):
I find the code hard to read - but not sure if that's a problem as long as you do and you trust that it tests what it's supposed to test.

Just to be super clear, this wouldn't fail if we changed the get-released-data endpoint as it uses example data that's committed to the repo.

So all this helps with is working on the code assuming get-released-data doesn't change (or adapt it if it does after updating the test data). 

The test here does _not_ fail if get-released-data suddenly changes its output. (That's kind of implied in the "unit test" I guess, but might be missed).

maverbiest (approved):
Looks good!

Issue #3881, opened 2025-03-21
Improve visibility of 'download all'


fhennig:
In https://github.com/loculus-project/loculus/pull/3809 we changed the download dialog to not be responsible for the "download all" option anymore, instead we want to make it easier for users to apply a "show all" filter in the search bar.

fhennig:
It's not entirely clear how this should be implemented yet

PR #5235, opened 2025-10-15
feat(website): add advanced search options modal

theosanderson:
resolves https://github.com/loculus-project/loculus/issues/3881

prompted by https://loculus.slack.com/archives/C06JCAZLG14/p1760527861693229

This is a quick hack to somewhat improve our surfacing of the fact that we restrict to `LATEST_VERSION` and exclude revocations and ability to change that.

In future I'd hope we could improve this with a more guided custom UI for these fields that explains things better.

image

image

For now the question is whether this is incrementally better than what we have atm

------
https://chatgpt.com/codex/tasks/task_e_68ef88390f14832583a509dc71a27b34

🚀 Preview: https://codex-add-advanced-option.loculus.org


anna-parker (approved):
I'm not very familiar with this part of the code base but this seems reasonable and the preview looks good!

PR #5255, opened 2025-10-17
chore(deps): add sha256 digest to lapis image

corneliusroemer:
Digest is from: https://github.com/genspectrum/LAPIS/pkgs/container/lapis/544882049?tag=v0.5.19


anna-parker (approved):
Ah! That also makes more sense

Issue #5185, opened 2025-10-06
Ideas to improve readability of auxtable code


theosanderson:
If we were to keep the aux tables we should try to improve the readability of the code.

I spent a bit of time looking at https://github.com/loculus-project/loculus/blob/main/backend/src/main/kotlin/org/loculus/backend/service/submission/UploadDatabaseService.kt to try to think about what the options would be for that. 

Perosnally I don't find the big SQL too horrendous to read but I acknowledge options will differ.

We should rename `mapAndCopy` to `insertFromAuxTablesIntoMainTable`  or something.

____

We could create a VIEW in a migration, something like:

```
CREATE VIEW sequence_entries_staging AS
SELECT
    m.accession,
    m.version,
    m.organism,
    m.submission_id,
    m.submitter,
    m.group_id,
    m.uploaded_at AS submitted_at,
    m.upload_id,
    jsonb_build_object(
        'metadata', m.metadata,
        'files', m.files,
        'unalignedNucleotideSequences', 
        COALESCE(
            jsonb_object_agg(
                s.segment_name,
                s.compressed_sequence_data::jsonb
            ) FILTER (WHERE s.segment_name IS NOT NULL),
            '{}'::jsonb
        )
    ) AS original_data
FROM
    metadata_upload_aux_table m
LEFT JOIN
    sequence_upload_aux_table s
    ON m.upload_id = s.upload_id 
    AND m.submission_id = s.submission_id
GROUP BY
    m.upload_id,
    m.accession,
    m.version,
    m.organism,
    m.submission_id,
    m.submitter,
    m.group_id,
    m.uploaded_at,
    m.metadata,
    m.files;
```

then the SQL in the code would just be 

```
INSERT INTO sequence_entries (
    accession,
    version,
    organism,
    submission_id,
    submitter,
    group_id,
    submitted_at,
    original_data
)
SELECT
    accession,
    version,
    organism,
    submission_id,
    submitter,
    group_id,
    submitted_at,
    original_data
FROM sequence_entry_staging
WHERE upload_id = ?
RETURNING accession, version, submission_id;
```

Issue #4850, opened 2025-08-12
Unreadable backend error if I accidentally remove sequence file when revising


anna-parker:

anna-parker:
@fengelniederhammer would you maybe have time to look into this?

anna-parker:
@fengelniederhammer it keeps coming up while we are modifying the edit page and I think it would be good to tackle now it shouldn't be too much work

anna-parker:
This error came up in @fengelniederhammer's PR: https://github.com/loculus-project/loculus/pull/5200 and in my PR: https://github.com/loculus-project/loculus/pull/5261 just after a segment was deleted. The odd thing is the logs show no data was sent (even metadata) - so this seems like an important error to investigate. Here are the logs for the ebola case on main:
```
Failed to submit revision for LOC_0008R2P.1 with error '{"message":"Request failed with status code 400","name":"AxiosError","stack":"It@https://main.loculus.org/_astro/createAuthorizationHeader.axCejEiN.js:3:1073\nw@https://main.loculus.org/_astro/createAuthorizationHeader.axCejEiN.js:3:5795\n","config":{"transitional":{"silentJSONParsing":true,"forcedJSONParsing":true,"clarifyTimeoutError":false},"adapter":["xhr","http","fetch"],"transformRequest":[null],"transformResponse":[null],"timeout":0,"xsrfCookieName":"XSRF-TOKEN","xsrfHeaderName":"X-XSRF-TOKEN","maxContentLength":-1,"maxBodyLength":-1,"env":{},"headers":{"Accept":"application/json, text/plain, */*","Content-Type":"application/x-www-form-urlencoded","Authorization":"Bearer ..."},"baseURL":"https://backend-main.loculus.org","method":"post","url":"/ebola-sudan/revise","data":{},"params":{},"allowAbsoluteUrls":true},"code":"ERR_BAD_REQUEST","status":400})}'
```

corneliusroemer:
A few more lines on how to reproduce would be good. IIUC:
1. Revise a sequence
2. Discard sequence file
3. Submit
4. Some error shown in modal

Can reproduce and get error message.

Backend `/revise` errors with 400 Bad Request:

```
{
    "type": "about:blank",
    "title": "Bad Request",
    "status": 400,
    "detail": "Submissions for organism ebola-sudan require a sequence file.",
    "instance": "/ebola-sudan/revise"
}
```

Error should be surfaced better.

PR #4783, opened 2025-07-31
feat!(prepro, config): assign segment with nextclade sort

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/4847

### Screenshot

Builds on https://github.com/loculus-project/loculus/pull/4821

You can use https://github.com/pathoplexus/dev_example_data/pull/2 for testing - see example submission in video:

https://github.com/user-attachments/assets/d6e2bc7f-b070-4230-a204-121d224ee241

Enable assignment of segments/subtypes using nextclade sort with the param:
```
classify_with_nextclade_sort: True
minimizer_index: <url_to_minimizer_index_used_by_nextclade_sort>
```
When this is set to true (per default) fasta headers must have the format <submissionId>_<segmentName> (as in current set up).

Additionally instead of having a dictionary for the nextclade datasets and servers we make `nucleotideSequences` a list of sequences: 
```
nextclade_dataset_name: 
    L: nextstrain/cchfv/linked/L
    M: nextstrain/cchfv/linked/M
    S: nextstrain/cchfv/linked/S
nextclade_dataset_server: https://raw.githubusercontent.com/nextstrain/nextclade_data/cornelius-cchfv/data_output
genes: [RdRp, GPC, NP]
```

```
nucleotideSequences:
  - name: L
    nextclade_dataset_name: nextstrain/cchfv/linked/L
    nextclade_dataset_tag: <optional - was previously incorrectly placed on an organism level> 
    nextclade_dataset_server: <optional overwrites nextclade_dataset_server for this seq>
    accepted_sort_matches: <optional, used for classify_with_nextclade_sort and require_nextclade_sort_match, if not given nextclade_dataset_name is used> 
    gene_prefix: <optional, prefix to add to genes produced by nextclade run, e.g. nextclade labels genes as `AV1` but we expect `EV1_AV1`, here `EV1` would be the prefix >
  - name: M
    nextclade_dataset_name: nextstrain/cchfv/linked/M
  - name: S
    nextclade_dataset_name: nextstrain/cchfv/linked/S
nextclade_dataset_server: https://raw.githubusercontent.com/nextstrain/nextclade_data/cornelius-cchfv/data_output
```

Note the templates now also generate the genes list from the merged config.

### PR Checklist
- [ ] All necessary documentation has been adapted.
- [ ] Create a minimizer for tests using: https://github.com/loculus-project/nextclade-sort-minimizer-creator
- [ ] Any manual testing that has been done is documented (i.e. what exactly was tested?)

[...]

PR #5261, opened 2025-10-17
feat!(backend): refactor multi-segment submission

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/4708, https://github.com/loculus-project/loculus/issues/4734

partially resolves https://github.com/loculus-project/loculus/issues/5392, https://github.com/loculus-project/loculus/issues/5185#issuecomment-3383043241

### Breaking Changes

When users submit to multi-segmented organisms and want to group multiple segments under one metadata entry they are required to add an additional `fastaId` column with a comma-separated list of the `fastaIds` (fasta header IDs) of the respective sequences. If no `fastaId` column is supplied the `submissionId` will be used instead and the backend will assume that (as in the single-segmented case) there is a one-to-one mapping of metadata `submissionId` to `fastaId`.

This requires:
- changes to the edit-page, started in https://github.com/loculus-project/loculus/pull/5382
- requires https://github.com/loculus-project/loculus/issues/4850 to be fixed -> this is leading to failing integration tests

### Future Steps

For simplicity in this PR prepro does not yet use a minimizer index to assign segments but parses the fasta header (as the backend did before). In future steps we should 
1. do a migration to change the originalData unalignedSequences keys for older processed versions
2. update the processedData return type to include a mapping from fasta header ID to segment
3. have prepro use a minimizer index to assign segments

### PR Checklist
- [x] Update ingest for multi-segmented organism (CCHF)
- [x] Use prepro updates from https://github.com/loculus-project/loculus/pull/4821 and https://github.com/loculus-project/loculus/pull/4783
- [ ] Update integration tests
- [x] Fix backend tests
- [x] Add backend test that multi-segmented can use id column instead of fastaId as well

🚀 Preview: https://multi-segment-submission.loculus.org

### DIFF
--- a/website/src/components/Edit/SequencesForm.tsx
+++ b/website/src/components/Edit/SequencesForm.tsx
@@ -6,7 +6,7 @@ import { FileUploadComponent } from '../Submission/FileUpload/FileUploadComponen
 import { PLAIN_SEGMENT_KIND, VirtualFile } from '../Submission/FileUpload/fileProcessing.ts';
 
 function generateAndDownloadFastaFile(fastaHeader: string, sequenceData: string) {
-    const trimmedHeader = fastaHeader.replaceAll(/[^a-zA-Z0-9_-]/g, '');
+    const trimmedHeader = fastaHeader.replace(/\s+/g, '');
     const fileContent = `>${trimmedHeader}\n${sequenceData}`;
 
     const blob = new Blob([fileContent], { type: 'text/plain' });
@@ -125,7 +125,7 @@ export class EditableSequences {
     getFastaIds(submissionId: string): string {
         const filledRows = this.rows.filter((row) => row.value !== null);
         return this.isMultiSegmented()
-            ? filledRows.map((sequence) => sequence.label.replaceAll(/[^a-zA-Z0-9_-]/g, '')).join(', ')
+            ? filledRows.map((sequence) => sequence.label.replace(/\s+/g, '')).join(', ')
             : submissionId;
     }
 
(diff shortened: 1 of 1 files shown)


backend/src/main/kotlin/org/loculus/backend/service/submission/UploadDatabaseService.kt
```diff
@@ -228,7 +236,11 @@ class UploadDatabaseService(
             LEFT JOIN
                 sequence_upload_aux_table
                 ON metadata_upload_aux_table.upload_id = sequence_upload_aux_table.upload_id 
-                AND metadata_upload_aux_table.submission_id = sequence_upload_aux_table.submission_id
+                AND EXISTS (
```
anna-parker: something is wrong with this function - when I submit multi-segmented sequences the unalignedNucleotideSequences object is empty?

anna-parker: oh no this is actually working fine! its just that issue with revising on the page and the website searching for exact segment names in the original data map

website/src/components/Edit/SequencesForm.tsx
```diff
@@ -6,24 +6,18 @@ import { FileUploadComponent } from '../Submission/FileUpload/FileUploadComponen
...
+    fastaHeader: string,
+    sequenceData: string
 ) {
-    const fileName =
-        isSingleSegment || !segmentKey ? `${accessionVersion}.fasta` : `${accessionVersion}_${segmentKey}.fasta`;
-
-    const header = isSingleSegment || !segmentKey ? accessionVersion : `${accessionVersion}_${segmentKey}`;
-
-    const fileContent = `>${header}\n${sequenceData}`;
+    const trimmed_header = fastaHeader.replaceAll(/[^a-zA-Z0-9]/g, '');
```
anna-parker:
```suggestion
    const trimmedHeader = fastaHeader.replaceAll(/[^a-zA-Z0-9]/g, '');
```

anna-parker:
I will close this in favor of https://github.com/loculus-project/loculus/pull/5398

PR #5264, opened 2025-10-20
4769 WIP - store compression dictionaries in database to make database self sufficient

### LINKED ISSUE #4769: Store compression dictionaries in database to make database self-sufficient
Sequences are currently zstd dictionary compressed using each sequence types reference as dictionary. Those references are hence required to make sense of the sequence data in the db.

We should store the dictionaries in the db to make the db self-sufficient in terms of information content.

We should then allow switching the exact dictionary used which will be helpful when we move sequence mapping from identifier based (backend) to sequence content based in preprocessing which is planned as part of multi pathogen.

Compressed data can say which dictionary was used so it's natural to in the future embed that dictionary information when we compress. Then all one needs to decompress is the compressed data which references a dictionary id which can be looked up in the db.

We would only once have to recompress sequences to embed the dictionary id in a migration then we'd be able to use a variety of dictionaries freely.

We should be able to write a migration to add dictionary information to the dicts: https://chatgpt.com/share/6889f227-988c-8005-a459-39245544a149

fengelniederhammer:
resolves #4769

https://github.com/loculus-project/loculus/issues/4769#issuecomment-3409473872

### PR Checklist

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/db/migration/V1_17_1__Migrate_Sequence_Entries_To_Use_Compression_Dict.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/db/migration/V1_17_1__Migrate_Sequence_Entries_To_Use_Compression_Dict.kt
@@ -35,6 +35,13 @@ class V1_17_1__Migrate_Sequence_Entries_To_Use_Compression_Dict(
         val db = Database.connect(context.configuration.dataSource)
 
         transaction(db) {
+            exec(
+                """
+                    LOCK TABLE sequence_entries IN EXCLUSIVE MODE;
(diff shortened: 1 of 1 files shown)


backend/docs/db/schema.sql
```diff
@@ -167,6 +167,41 @@ ALTER SEQUENCE public.audit_log_id_seq OWNER TO postgres;
 ALTER SEQUENCE public.audit_log_id_seq OWNED BY public.audit_log.id;
 
 
+--
+-- Name: compression_dictionaries_table; Type: TABLE; Schema: public; Owner: postgres
+--
+
+CREATE TABLE public.compression_dictionaries_table (
```
theosanderson: unexpected to put `table` in the name?

PR #5272, opened 2025-10-20
ci(backend): split test/lint into 2 jobs, don't fail codeql on backend test/lint failure - just assemble there

corneliusroemer:
- Previously, codeql would fail when backend tests/lint fail. We don't want to duplicate failures in CI, failure in one place shouldn't cause failure across the board. By running `./gradlew assemble` rather than `./gradlew build` codeql will no longer duplicate the existing tests.
- Split backend tests into 2 separate jobs: `lint` and `test` (test is everything except for lint). Right now, lint only runs once test succeeds. This delays lint and masks it when test fails.

### DIFF
--- a/.github/workflows/backend-tests.yml
+++ b/.github/workflows/backend-tests.yml
@@ -36,7 +36,7 @@ jobs:
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@v5
       - name: Run tests
-        run: ./gradlew test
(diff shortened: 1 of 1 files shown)


theosanderson (approved):
> Or do you mean codeql ran some things that we no longer do now?

yes, and it wasn't that I was saying we should definitely not make this change - just that I was waiting for someone else who knew better

Seems good now

PR #5299, opened 2025-10-22
feat(deposition): Distinguish between complete and incomplete (raw read) ENA metadata in notifications

corneliusroemer:
## Summary

This PR improves how we handle and notify about sequences with ENA-specific metadata fields.

Previously, all sequences with any ENA-specific metadata were triaged together into a single notification. This made it difficult to distinguish between:
1. Legitimate user submissions linking to raw reads (all fields present)
2. Potential user errors (only some fields present)

### Updated Notifications
- Two separate Slack notifications with clearer, more actionable messages
- Different output filenames to distinguish the two categories
- Complete metadata gets info-level logging, incomplete gets warning-level


anna-parker (changes requested):
see my comments :-)

PR #5314, opened 2025-10-23
feat(backend): migrate all sequence entries to use the compression dictionary stored in the data

### LINKED ISSUE #4769: Store compression dictionaries in database to make database self-sufficient

fengelniederhammer:
resolves #4769

Follow up to #5311 and the second part of https://github.com/loculus-project/loculus/issues/4769#issuecomment-3435703053

> The second migrates all existing data. It changes the backend to also only read data in the new format.

### PR Checklist

### DIFF
--- a/kubernetes/loculus/templates/loculus-backend.yaml
+++ b/kubernetes/loculus/templates/loculus-backend.yaml
@@ -34,7 +34,7 @@ spec:
               path: "/actuator/health/liveness"
               port: 8079
             periodSeconds: 5
-            failureThreshold: 36 # 3 minutes to start
+            failureThreshold: 180 # 15 minutes to start
           livenessProbe:
             httpGet:
               path: "/actuator/health/liveness"


backend/src/main/kotlin/org/loculus/backend/db/migration/V1_18_1__Migrate_Sequence_Entries_To_Use_Compression_Dict.kt
```diff
@@ -0,0 +1,229 @@
...
+                        metadata = originalData.metadata,
+                        files = originalData.files,
+                        unalignedNucleotideSequences = originalData.unalignedNucleotideSequences.mapValues {
+                            val value = it.value
+                            when {
+                                value == null -> null
+                                value.compressionDictId !== null -> value
+                                else -> CompressedSequence(
+                                    compressedSequence = value.compressedSequence,
+                                    compressionDictId = compressionDictService
```
anna-parker: @fengelniederhammer sorry if this is stupid but Im now rereading and I am confused - are't all the old unaligned sequences compressed on a segment level? Don't we need to get the dict from `getDictForSegmentOrGene`?

fengelniederhammer: Reading this now, I think you're right.

PR #5346, opened 2025-10-29
feat(prepro): provide request ID in backend requests to allow tracing requests

fhennig:
When a prepro Pod crashes with a timeout, we don't know what happened to the request in the backend, or we can only correlate it by looking at the time the Pod crashed.

The backend already reads the `x-request-id` header and uses the request ID in it, if given.

This PR sets a UUID as a request ID in the requests sent to the backend.

**Manual testing** - I deployed this to GenSpectrum staging and looked in loki and found a request successfully correlated.

(integration test failures seem unrelated?)

🚀 Preview: https://prepro-provide-request-id.loculus.org

### DIFF
--- a/preprocessing/nextclade/src/loculus_preprocessing/__main__.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/__main__.py
@@ -7,11 +7,7 @@ logger = logging.getLogger(__name__)
 
 
 def cli_entry() -> None:
-    logging.basicConfig(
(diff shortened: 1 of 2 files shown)


corneliusroemer (approved):
Thanks, this is good!

I wonder though whether there's a less tedious/verbose way of getting the request id into logs - maybe setting something on `logger` at the start of the function?

PR #5347, opened 2025-10-29
feat(backend): add request duration logging for streaming requests

fhennig:
In https://github.com/loculus-project/loculus/pull/5242 we added duration logging for requests, but for streamed requests the "200" response is sent immediately, but the actual data can take quite a bit longer.

This PR adds timing output for streamed requests. I'm also passing the name of the endpoint to the function, so it's easy to spot when among the long running requests there is one that usually shouldn't be there.

(I'm suspecting that `extract-unprocessed-data` is sometimes taking very long to respond, but the logs are not definitive yet)

**manual testing** - I tested this in GenSpectrum and it's running fine and I can see the numbers.

### PR Checklist

🚀 Preview: https://backend-time-streamed.loculus.org

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
@@ -38,6 +38,7 @@ import org.loculus.backend.auth.AuthenticatedUser
 import org.loculus.backend.auth.HiddenParam
 import org.loculus.backend.config.BackendConfig
 import org.loculus.backend.controller.LoculusCustomHeaders.X_TOTAL_RECORDS
+import org.loculus.backend.log.ORGANISM_MDC_KEY
 import org.loculus.backend.log.REQUEST_ID_MDC_KEY
 import org.loculus.backend.log.RequestIdContext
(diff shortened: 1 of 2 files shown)


corneliusroemer:
Rebasing mostly as an easy way to retrigger tests to see if they pass now (yesterday failed due to Azure down)

corneliusroemer (commented):
Thanks, good idea!

Why test on genspectrum though and not in preview?

I checked on preview:

Google Chrome 2025-10-30 08 50 25

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
corneliusroemer: I'd suggest we also log duration when the stream fails - that's also useful to know. Here: https://github.com/loculus-project/loculus/pull/5347/files#diff-a2395dbeb2c4c0b49554ea31019045f9586ec0d78868d6364d09b3183939e55aR520

corneliusroemer (changes requested):
This is great but can we also log duration when there's an error? Add it to line 518 the `log.error(e) "An unexpected error occurred while streamign"`

PR #5363, opened 2025-10-31
fix(website): manually configure retry of failed requests from LAPIS react query hooks as they don't mutate 

### LINKED ISSUE #5359: When first request to LAPIS from search UI fails, we don't retry
[trace-minimal.tar.gz](https://github.com/user-attachments/files/23255717/trace-minimal.tar.gz)

There's a flake (rare failure) in integration tests where the search UI ends up permanently in error state:

Image

I've traced this to the following chain of events:

1. When the search UI first hydrates on client side, it's by default disabled.
2. Search UI makes request to LAPIS (2 requests, one for count, one for details)
3. If one of those first requests post-hydration fails, e.g. due to `net::ERR_INCOMPLETE_CHUNKED_ENCODING`, search UI remains in error state.
4. Because we use a post request to LAPIS for those requests, react query hooks don't retry by default (per HTTP spec, POST can mean mutation, which isn't safe to retry - in our case we can safely retry as we only fetch data but we haven't set that flag apparently)

How to deflake:
- Enable retry for the safe-to-retry endpoints within react query hooks

There's no trace in kubernetes logs of the request failing. We can trace the request through LAPIS and SILO with the request id, and all looks good there. Website logs don't contain the request id - so we completely fail to log/notice the error.

We might want to try one or both of the following to improve debuggability of issues like that (and allow us also to see better how common such client errors are in general):
- Wrap detailsHook.mutate with onError to console.error in the web logs and to send to our telemetry. Include the X-Request-Id header so you can correlate with backend logs.
- Consider adding a custom frontend logger that POSTs failures to our (website) backend so we can aggregate “client saw ERR_INCOMPLETE_CHUNKED_ENCODING” events.

This is where I spotted the failure: https://github.com/loculus-project/loculus/actions/runs/18950984254/job/54115262458?pr=5356#step:22:84

And here are the traces/logs:
- Trace: [trace-minimal.tar.gz](https://github.com/user-attachments/files/23255761/trace-minimal.tar.gz)
- Logs: [kubernetes-logs-chromium (1).zip](https://github.com/user-attachments/files/23254245/kubernetes-logs-chromium.1.zip)

corneliusroemer:
resolves #5359 

When we switched to use post requests to query LAPIS, we forgot to enable retry upon failure. By default, POST isn't retried as in principle POSTs are allowed to mutate. We use POST to pass a large query string (e.g. accession list) and those _are_ safe to retry.

Lack of retrying causes occasional integration test flakes and might also cause occasional bad UX (requiring manual reload rather than resolving on its own).

I also got Copilot to improve DX of the hooks by removing boilerplate.

Code mostly from copilot session #5360 

### Future work

To cause such bugs faster, we might think about injecting failed requests on purpose. We could have a test that fails the first request deterministically for example. Don't think this is too important, but could be nice to have to improve coverage in future.

### Testing

I manually tested that when I block LAPIS in browser dev tools, then load the page, wait until LAPIS query has failed and then unblock, it automatically retries and succeeds.

### PR Checklist

### DIFF
--- a/website/AGENTS.md
+++ b/website/AGENTS.md
@@ -2,4 +2,5 @@
 
 Run `CI=1 npm run test`, `npm run check-types`, `npm run format`.
 
+Without `CI=1`, vitest thinks it’s in development mode and keeps watching for file changes and doesn't terminate. With `CI=1`, it knows it’s in automation mode, runs once, and exits cleanly.
 It's important to add `CI=1` envvar to `npm run test` because otherwise it keeps running in background.


website/AGENTS.md
```diff
@@ -1,3 +1,5 @@
 ## Checklist before committing code
 
 Run `CI=1 npm run test`, `npm run check-types`, `npm run format`.
+
```
anna-parker: this was still unclear for me, I think explaining `Without CI=1, your test runner thinks it’s in development mode and keeps watching for file changes.
With CI=1, it knows it’s in automation mode, runs once, and exits cleanly` might be clearer

PR #5382, opened 2025-11-06
feat!(website, prepro, backend, config, integration):multi pathogen - refactor multi segment submission in backend and edit page and have prepro assign segments

anna-parker:
resolves #4999 https://github.com/loculus-project/loculus/issues/4708, https://github.com/loculus-project/loculus/issues/4734, https://github.com/loculus-project/loculus/issues/5511

partially resolves https://github.com/loculus-project/loculus/issues/5392, https://github.com/loculus-project/loculus/issues/5185#issuecomment-3383043241

includes work done in https://github.com/loculus-project/loculus/pull/5398 and https://github.com/loculus-project/loculus/pull/5402

This PR additionally fixes submission, subtype assignment and search for EVs and other multi-path organisms.

### BREAKING CHANGES

When users submit to multi-segmented organisms and want to group multiple segments under one metadata entry they are now required to add an additional `fastaIds` column with a space -separated list of the `fastaId`s (fasta header IDs) of the respective sequences. If no `fastaIds` column is supplied the `submissionId` will be used instead and the backend will assume that (as in the single-segmented case) there is a one-to-one mapping of metadata `submissionId` to `fastaId`.

This new submission structure was voted for in microbioinfo: https://microbial-bioinfo.slack.com/archives/CB0HYT53M/p1760961465729399 and discussed in https://app.nuclino.com/Loculus/Development/2025-10-20-Weekly-6d5fe89f-8ded-4286-b892-d215e0a498f6 (and in other meetings)

Nextclade sort (uses a minimizer index for fast local alignment) or nextclade align (full sequence alignment to reference) will be used to assign segments/subtypes for all multi-segmented and multi-pathogen sequences (this is also done in ingest for grouping segments):
```
segment_classification_method: "minimizer" or "align"
minimizer_url: <url_to_minimizer_index_used_by_nextclade_sort>
```
For organisms without a nextclade dataset we still allow the fasta headers to be used to determine the segment/subtype - entries must have the format `<submissionId>_<segmentName>` (as in current set up).

As preprocessing now assigns segments it will return a map from the segment (or subtype) to the fastaId in the processedData, the map is called: `sequenceNameToFastaId`. This allows us to surface the segment assignment on the edit page.

### Nextclade Preprocessing pipeline config changes

Instead of having a dictionary for the nextclade datasets and servers we make `nucleotideSequences` a dictionary where each item includes all information required to run nextclade. I.e. we change from: 
```
nextclade_dataset_name:
    L: nextstrain/cchfv/linked/L
[...]

Issue #5095, opened 2025-09-25
Multi pathogen support: suborganism specific metadata search fields


fengelniederhammer:
On the search page: Think about whether we need metadata fields that are only available / chosen correctly once you selected a suborganism.

An example might be the length fields. It only makes sense to have the field when the corresponding suborganism was chosen.
<img width="720" height="645" alt="Image" src="https://github.com/user-attachments/assets/77b21ad8-bd2d-4777-8500-175a5a17bbc3" />

An alternative approach could be making those fields not "per segment", but merge the segment values into a single metadata field. (In the example: Have a single `length` field, then put the length of the suborganism sequence in there)

theosanderson:
I think ideally we'd only show stuff here if it applies to the selected suborganism, or possibly gray or out if not

PR #5467, opened 2025-11-19
feat(preprocessing): add "merged" metadata fields that contain the value of the only segment

fengelniederhammer:
another part of #5095

Add the boolean option `processing_spec[field].args.useFirstSegment` to the preprocessing config  to take the value of the first segment that has been found in the nextclade output, e.g.:
```yaml
        - name: totalUnknownNucs
          orderOnDetailsPage: 2800
          type: int
          header: "Alignment and QC metrics"
          noInput: true
          rangeSearch: true
          displayName: Total unknown nucs
          preprocessing:
            args:
              useFirstSegment: true
            inputs: {input: nextclade.totalMissing}
```

`useFirstSegment` is automatically set in the Helm chart for all `perSegment` metadata fields in the multi pathogen case. This won't work for multi-segmented multi pathogen organisms, but we decided to ignore that case for now anyway.

The field `organisms[organism].schema.nucleotideSequences` in the `values.yaml` is almost obsolete now. It's only still used in the preprocessing config template. Probably after #5382, the field can be removed altogether, since it's redundant anyway.

### Screenshot

grafik

### PR Checklist
  - the QC values (other than length) are not per segment anymore for EV.
  - They are still per segment for CCHF

🚀 Preview: https://5095-2-groupinprepro.loculus.org

### DIFF
--- a/preprocessing/nextclade/tests/test_nextclade_preprocessing.py
+++ b/preprocessing/nextclade/tests/test_nextclade_preprocessing.py
@@ -815,7 +815,7 @@ multi_pathogen_cases = [
                 processedFields=[
                     AnnotationSource(name="totalInsertions", type=AnnotationSourceType.METADATA)
                 ],
-                message="Expected exactly one valid segment, found multiple: ['ebola-sudan', 'ebola-zaire']",
+                message="Organism multi-ebola-test is configured to only accept one segment per submission, found multiple valid segments: ['ebola-sudan', 'ebola-zaire'].",
             ),
             ProcessingAnnotation(
                 unprocessedFields=[
@@ -829,7 +829,7 @@ multi_pathogen_cases = [
                 processedFields=[
                     AnnotationSource(name="totalSubstitutions", type=AnnotationSourceType.METADATA)
                 ],
-                message="Expected exactly one valid segment, found multiple: ['ebola-sudan', 'ebola-zaire']",
+                message="Organism multi-ebola-test is configured to only accept one segment per submission, found multiple valid segments: ['ebola-sudan', 'ebola-zaire'].",
             ),
         ],
         expected_warnings=[],


fengelniederhammer:
I just noticed that this breaks the length fields of EV. But the lengths of the other organisms are still working. IMO that's ok, since EV is on previews only anyway. What do you think? I would actually prefer to implement the lengths in a separate PR.

preprocessing/nextclade/tests/test_nextclade_preprocessing.py
```diff
@@ -383,6 +387,41 @@ def invalid_sequence() -> str:
             aminoAcidInsertions={},
         ),
     ),
+    Case(
```
anna-parker: IMO we don't need this test as it is the same as `name="with one succeeded and one not uploaded"` except that with ebola-sudanuploaded instead of ebola-zaire

theosanderson:
(The ideal behaviour for INSDC for single segment organisms is probably not to have suborganism name in it)

anna-parker:
Oh I had another thought! `generateBackendExternalMetadata` will also have to be changed to be single segmented for the EVs

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -475,6 +475,19 @@ def null_per_backend(x: Any) -> bool:
             return False
 
 
+def get_segment(spec: ProcessingSpec, unprocessed: UnprocessedAfterNextclade) -> str:
+    """Returns the segment to use based on spec args"""
+    if spec.args and spec.args.get("useFirstSegment", False) and unprocessed.nextcladeMetadata:
```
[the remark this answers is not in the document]
fengelniederhammer: If `valid_segments` is empty, what should we do? Also return a `failed to align` error?

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -69,12 +69,28 @@ def null_per_backend(x: Any) -> bool:
             return False
 
 
+def get_segment(spec: ProcessingSpec, unprocessed: UnprocessedAfterNextclade) -> str | None:
+    """Returns the segment to use based on spec args"""
+    if spec.args and spec.args.get("useFirstSegment", False) and unprocessed.nextcladeMetadata:
+        valid_segments = [key for key, value in unprocessed.nextcladeMetadata.items() if value]
+        if not valid_segments:
+            return None
+        if len(valid_segments) > 1:
```
anna-parker: I think this error would actually have to be returned in the processing result as an error (so that the user sees it and contacts the administrator). We have some open todos in the add_nextclade_metadata function for this - could you maybe add a todo here and I will draft a PR that returns proper errors in these cases?

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -629,6 +642,16 @@ def processed_entry_no_alignment(
     )
 
 
+def get_sequence_length(
+    unaligned_nucleotide_sequences: dict[SegmentName, NucleotideSequence | None],
+    segment: SegmentName | None,
+) -> int:
+    if segment is None:
+        return 0
+    sequence = unaligned_nucleotide_sequences.get(segment, None)
```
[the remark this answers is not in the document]
anna-parker: ah yes the default value for get is None in python so this can be removed but I think this is now outdated :-)

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
fengelniederhammer: Could you check whether the generated error annotation looks ok?

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -230,6 +268,26 @@ def get_output_metadata(
         spec.args = {} if spec.args is None else spec.args
         input_data: InputMetadata = {}
         input_fields: list[str] = []
+        if output_field == "length":
+            segment_name: SegmentName | None = "main"
```
anna-parker: actually we can reuse get_segment(spec, unprocessed) here to get the segment_name

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -70,12 +70,52 @@ def null_per_backend(x: Any) -> bool:
...
 def add_nextclade_metadata(
     spec: ProcessingSpec,
     unprocessed: UnprocessedAfterNextclade,
     nextclade_path: str,
 ) -> InputData:
-    segment = str(spec.args["segment"]) if spec.args and "segment" in spec.args else "main"
+    try:
+        segment = get_segment(spec, unprocessed)
+    except MultipleValidSegmentsError as e:
+        message = f"Expected exactly one valid segment, found multiple: {e.segments}"
```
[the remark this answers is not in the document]
fengelniederhammer: I agree with the first part - but IMO this is a user error (at least potentially), would you still add "please contact an administrator"?

PR #5469, opened 2025-11-19
refactor(integration tests): Refactor integration test fixtures pattern

### LINKED ISSUE #5468: integration tests fixtures refactor:`pageWithGroup`,`pageWithACreatedUser` is bad pattern
I think when I was implementing these I misunderstood how fixtures worked and thought they needed to return a page object. The pattern Cornelius uses in https://github.com/loculus-project/loculus/pull/5462 makes much more sense.

theosanderson:
…a-returning fixtures

The old pattern of fixtures returning Page objects (pageWithGroup, pageWithACreatedUser, pageWithReleasedSequence) was confusing and not idiomatic for Playwright fixtures.

This refactor follows the better pattern demonstrated in PR #5462:
- Fixtures now return data instead of page objects
- Tests use the regular 'page' fixture along with data fixtures
- Dependencies are explicit with void statements

Changes:
- auth.fixture.ts: pageWithACreatedUser → authenticatedUser (returns TestAccount)
- group.fixture.ts: pageWithGroup → removed, keeping groupId (returns number)
- sequence.fixture.ts: pageWithReleasedSequence → releasedSequence (returns string)
- Updated all 13 test files to use new pattern
- Updated README.md to document the new fixture pattern

This makes the fixtures easier to understand and compose, and follows Playwright best practices.

Fixes #5468

### DIFF
--- a/integration-tests/tests/specs/features/file-sharing.spec.ts
+++ b/integration-tests/tests/specs/features/file-sharing.spec.ts
@@ -18,23 +18,17 @@ test('submit single seq w/ 2 files thru single seq submission form', async ({
 }) => {
     test.setTimeout(180_000);
     void groupId;
-
     const submissionPage = new SingleSequenceSubmissionPage(page);
-
     await submissionPage.navigateToSubmissionPage(ORGANISM_NAME);
     await submissionPage.fillSubmissionFormDummyOrganism({
         submissionId: ID_1,
         country: COUNTRY_1,
         date: '2023-10-15',
     });
-
     await submissionPage.uploadExternalFiles(RAW_READS, FILES_DOUBLE, tmpDir);
-
     const reviewPage = await submissionPage.submitAndWaitForProcessingDone();
-
     await reviewPage.checkFilesInReviewDialog(FILES_DOUBLE);
     const searchPage = await reviewPage.releaseAndGoToReleasedSequences();
-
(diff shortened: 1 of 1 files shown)


integration-tests/tests/fixtures/sequence.fixture.ts
corneliusroemer: Ideally we'd return the accession here - we could make submissionId its own fixture

integration-tests/tests/specs/features/authenticated.spec.ts
```diff
@@ -2,8 +2,12 @@ import { expect } from '@playwright/test';
...
-    test('authenticated fixture runs and lands on user page', async ({ pageWithACreatedUser }) => {
-        await expect(pageWithACreatedUser).toHaveURL('/');
-        await expect(pageWithACreatedUser.getByRole('link', { name: 'My account' })).toBeVisible();
+    test('authenticated fixture runs and lands on user page', async ({
+        page,
+        authenticatedUser,
+    }) => {
+        void authenticatedUser;
+        await expect(page).toHaveURL('/');
+        await expect(page.getByRole('link', { name: 'My account' })).toBeVisible();
```
corneliusroemer: We can beef this up now by asserting on the authenticatedUser information - taking full advantage of the new pattern

PR #5477, opened 2025-11-19
refactor(backend): don't create empty concatenation, slight rewording of variables for clarity

corneliusroemer:
I noticed we concatenate nothing to yield an empty dict - that's unnecessary

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/service/submission/CompressionDictService.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/service/submission/CompressionDictService.kt
@@ -50,9 +50,9 @@ class CompressionDictService(private val backendConfig: BackendConfig, private v
         transaction {
             backendConfig.organisms.forEach { (organismString, instanceConfig) ->
                 val organism = Organism(organismString)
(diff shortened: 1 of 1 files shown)


backend/src/main/kotlin/org/loculus/backend/service/submission/CompressionDictService.kt
```diff
@@ -64,16 +64,22 @@ class CompressionDictService(private val backendConfig: BackendConfig, private v
...
-                    .sortedBy { it }
-                    .joinToString("")
-                val dictId = getDictIdOrInsertNewEntry(references, "${organism.name} - unaligned")
-                val unalignedDict = references.toByteArray()
-                val dictEntry = DictEntry(dictId, unalignedDict)
-
-                unalignedDictCache[organism] = dictEntry
-                cacheById[dictId] = unalignedDict
+                // No need to concatenate nothing, it'll never be used
+                if (nucleotideSequences.isNotEmpty()) {
```
fengelniederhammer: Nitpick: I'd prefer an early return instead.

PR #5495, opened 2025-11-20
create a separate interface for files with a fasta header

fengelniederhammer:
targetting https://github.com/loculus-project/loculus/pull/5382#discussion_r2511216041


anna-parker (approved):
oh awesome! thanks!!

PR #5502, opened 2025-11-21
fix(backend): validate fileId exists in submit edited endpoint, do validation in same place as file existence checks

corneliusroemer:
Fixes bug where we did not check that non-existing fileIds weren't uploaded in the submit edited endpoint. Two other endpoints (processed and submission of new data) did this check in two different places - rather than inside the logical place `validateFileExists` - so it was forgotten in the 3rd case.

Fixes bug where we didn't validate multipart upload in the submit-edited endpoint.

I discovered this as part of a refactor of the fileId existence validation into the existing file validation.

Thanks @fhennig for helping me understand the purpose/logic of the preexisting code.

### Screenshot / Manual testing

The new test fails on main - i.e. this PR really fixes the bug:

iTerm2 2025-11-21 14 22 36

Multipart upload validation is untested, see TODOs and #5503

### PR Checklist

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/service/submission/FileMappingPreconditionValidator.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/service/submission/FileMappingPreconditionValidator.kt
@@ -99,8 +99,8 @@ class FileMappingPreconditionValidator(
     }
 
     /**
-     1. Validate that the fileIds exist (have been requested for upload)
-     2. Check that a file has been uploaded for each fileId by checking S3 for its size
+     * 1. Validate that the fileIds exist (have been requested for upload)
+     * 2. Check that a file has been uploaded for each fileId by checking S3 for its size
      */
     fun validateFilesExist(fileIds: Set<FileId>): FileMappingPreconditionValidator {
         validateFileIdsExist(fileIds)


backend/src/main/kotlin/org/loculus/backend/service/submission/FileMappingPreconditionValidator.kt
```diff
@@ -70,13 +71,40 @@ class FileMappingPreconditionValidator(
...
+        val nonExistentFileIds = filesDatabaseService.getNonExistentFileIds(fileIds)
+        if (nonExistentFileIds.isNotEmpty()) {
+            throw UnprocessableEntityException(
+                "The following file IDs do not exist: " + nonExistentFileIds.joinToString(),
+            )
+        }
+        return this
+    }
+
+    fun validateFilesUploadedNotJustRequested(fileIds: Set<FileId>): FileMappingPreconditionValidator {
```
theosanderson:
```suggestion
    fun validateFilesUploaded(fileIds: Set<FileId>): FileMappingPreconditionValidator {
```

IMO this is over-long and the extra words don't really add clarity
(not blocking)

Issue #5511, opened 2025-11-22
Have EVs ingest data as single (not multi-segmented)


anna-parker:
Flagged in https://github.com/loculus-project/loculus/pull/5467#pullrequestreview-3492982460

Not all segment-specific fields are created in preprocessing, some already are set in ingest, for example INSDC accession. We have been ingesting EVs as a virus with segments and adding the `genotype` flag in ingest. 

When https://github.com/loculus-project/loculus/pull/5382 is merged ingest can submit EVs as single segmented and prepro will assign the segment. However, the `genotype` field will now be missing and will have to be added by preprocessing.

A PR on https://github.com/loculus-project/loculus/pull/5467/files requires:
1. Have ingest receive the `extractUniqueRawNucleotideSequenceNames` -> will ingest EVs as a single segment
2. Rebase on https://github.com/loculus-project/loculus/pull/5382 to allow submission to the backend
3. Add a new preprocessing.input flag to have preprocessing assign the genotype based on the assigned segment - I suggest sth like:
```
- name: genotype
  displayName: Genotype
  header: "Genotype"
  noInput: true
  generateIndex: true
  autocomplete: true
  preprocessing:
    inputs: {input: ASSIGNED_SEGMENT}
```

anna-parker:
completed in https://github.com/loculus-project/loculus/pull/5382

PR #5551, opened 2025-11-26
Add genotype

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/5511

### PR Checklist
- [x] Add a test
- [x] deduplicate code
- [x] make genotype a required field?

🚀 Preview: https://add-genotype.loculus.org


fengelniederhammer:
> * make genotype a required field?

Yes, I think that would be good. I wonder whether we can infer that automatically in the Helm chart? Since the website uses it as `suborganismIdentifierField`, it should not allow null values. But I'm also not sure whether that complicates things more than necessary.

PR #5553, opened 2025-11-27
chore(preprocessing): fix spelling in embl file error message

fengelniederhammer:
It looked weird before:
`embl_upload: Failed to create or upload EMBL file please contact your administrator.`

grafik

### DIFF
--- a/preprocessing/nextclade/src/loculus_preprocessing/prepro.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/prepro.py
@@ -570,7 +570,8 @@ def upload_flatfiles(processed: Sequence[SubmissionData], config: Config) -> Non
                     processedFields=[
                         AnnotationSource(name="embl_upload", type=AnnotationSourceType.METADATA)
                     ],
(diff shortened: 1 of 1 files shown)


preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -570,8 +570,7 @@ def upload_flatfiles(processed: Sequence[SubmissionData], config: Config) -> Non
                     processedFields=[
                         AnnotationSource(name="embl_upload", type=AnnotationSourceType.METADATA)
                     ],
-                    message="Failed to create or upload EMBL file "
-                    "please contact your administrator.",
+                    message="Failed to create or upload EMBL file. Please contact your administrator.",
```
anna-parker: would be nice to keep the lines shorter

fengelniederhammer: Ah I see `line-length = 100` (too low for my taste but ok). And yes, the CI should enforce this otherwise there will be more and more violations.

Issue #5608, opened 2025-12-02
Keycloak: README unclear, local installation is failing


anna-parker:
Hi! 

I tried to run `yarn storybook` to check if the changes in https://github.com/loculus-project/loculus/pull/5604 break keycloak but I cannot run this on main or on that preview. 

Node version:
```
$ node --version    
v24.11.1
```

I had trouble running the install docs on the quick start quide https://github.com/loculus-project/loculus/blob/main/keycloak/keycloakify/README.md#quick-start: 
```
aparker-adm@eve-501 keycloakify % nvm use             
corepack install
yarn install --immutable
zsh: command not found: nvm
Debugger attached.
Adding yarn@4.5.3 to the cache...
Waiting for the debugger to disconnect...
Debugger attached.
error This project's package.json defines "packageManager": "yarn@4.5.3". However the current global version of Yarn is 1.22.22.

Presence of the "packageManager" field indicates that the project is meant to be used with Corepack, a tool included by default with all official Node.js distributions starting from 16.9 and 14.19.
Corepack must currently be enabled by running corepack enable in your terminal. For more information, check out https://yarnpkg.com/corepack.
Waiting for the debugger to disconnect...
```

So I then went on to https://github.com/loculus-project/loculus/blob/main/keycloak/keycloakify/README.md#building-the-theme and was able to run the code but it is failing:

```
$ npm run storybook 
Debugger attached.

> loculus-keycloak-theme@0.0.1 storybook
> storybook dev -p 6006

Debugger attached.
storybook v9.1.16

(node:82117) [DEP0190] DeprecationWarning: Passing args to a child process with shell option true can lead to security vulnerabilities, as the arguments are not escaped, only concatenated.
(Use `node --trace-deprecation ...` to show where the warning was created)
info => Serving static files from ././public at /
info => Starting manager..
No story files found for the specified pattern: src/**/*.mdx
info => Starting preview..
info Using tsconfig paths for react-docgen
╭──────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                          │
│   Storybook 9.1.16 for react-vite started                                                │
│   107 ms for manager and 2.31 s for preview                                              │
│                                                                                          │
│    Local:            http://localhost:6006/                                              │
│    On your network:  http://10.221.193.87:6006/                                          │
│                                                                                          │
│   A new version (10.1.2) is available!                                                   │
│                                                                                          │
│   Upgrade now: npx storybook@latest upgrade                                              │
[...]

PR #5621, opened 2025-12-03
docs: Update README.md for keycloakify to explicitly contain nvm install

maverbiest:
instructions

resolves https://github.com/loculus-project/loculus/issues/5608

### PR Checklist
- [ ] ~The implemented feature is covered by appropriate, automated tests.~

Manually tested that the commands I included in the README work on my machine. Storybook output looked good as far as I could tell.

### DIFF
--- a/keycloak/keycloakify/README.md
+++ b/keycloak/keycloakify/README.md
@@ -15,7 +15,7 @@ Based on upstream commit: https://github.com/keycloakify/keycloakify-starter/com
 
 First, ensure [nvm](https://github.com/nvm-sh/nvm) is installed and that corepack is enabled (`corepack enable`).
 If you don't have the version of npm specified in `.nvmrc` installed, do so by running `nvm install` in this directory.
-Then, run the following commands (all together, not as individual commands):
+Then, run the following commands:
 ```bash
 nvm use
 corepack install


keycloak/keycloakify/README.md
```diff
@@ -13,6 +13,9 @@ Based on upstream commit: https://github.com/keycloakify/keycloakify-starter/com
 
 ## Quick start
 
+First, ensure [nvm](https://github.com/nvm-sh/nvm) is installed and that corepack is enabled (`corepack enable`).
+If you don't have the version of npm specified in `.nvmrc` installed, do so by running `nvm install` in this directory.
+Then, run the following commands (all together, not as individual commands):
```
fhennig: > Then, run the following commands (all together, not as individual commands)

What do you mean by "all together"? Like, pasted all together into the command line? Does that make any difference to how they are ran in the shell?

I just executed them one-by-one, like you'd usually do seeing these command line instructions, and it worked fine for me.

(not blocking, just surprised)

maverbiest: Ah yeah I indeed mean pasted all together into the command line so all three run at once.

I included this because @corneliusroemer mentioned this on the Slack thread yesterday (at least, that's how I interpreted it)

PR #5624, opened 2025-12-03
chore(ci): Revert `helm repo add` retry code and log ip instead

corneliusroemer:
Retrying (#5590) didn't help.

Getting IP will help Mittwald debug the problem if it comes back, see https://github.com/mittwald/kubernetes-secret-generator/issues/115#issuecomment-3607716359


anna-parker (approved):
~why are you logging ip? just curious~
read the issue - very cool!

PR #5625, opened 2025-12-03
feat(website): use segment 1 for uploads without fasta headers

anna-parker:
🚀 Preview: https://edit-page-fix.loculus.org/


theosanderson (commented):
Hey so this is the issue I was raising about duplicates:

image

(I deleted "segment 1" after uploading segment 2, then tried to add another segment)

corneliusroemer:
Would be good to add this in an integration test - we don't need to go all the way to approval to not take too much time, just the flow of adding, removing, readding.

PR #5639, opened 2025-12-05
refactor(website): edit page - remove side effect from rows getter

### LINKED ISSUE #5637: Codesmell: edit page function has side effect
See comment in https://github.com/loculus-project/loculus/pull/5631#discussion_r2589603028

`get rows()` has a side effect (it calls `getNextKey()`) this leads to the key of each newly submitted sequence increasing by an undefined amount as React is calling that getter multiple times per “one” user action. 

The main requirement for the key is uniqueness so this isnt a large issue but it should probably be improved on.

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/5637, https://github.com/loculus-project/loculus/issues/7377

get `rows()` had a side effect (it calls `getNextKey()`) this leads to the key of each newly submitted sequence increasing by an undefined amount as React is calling that getter multiple times per “one” user action.

Additionally, the previous code had an actual bug due to this side effect. When a re-edited file with an unchanged fastaHeader was dragged onto the same field, `update()` was called with that row's existing key. As the header matches itsself a false-positive "already exists" error is thrown, even though nothing is actually duplicated.

  On top of that, the old collision branch discarded whatever the caller passed and returned a filtered copy of the unmodified prior state. So the stored value/sequence never updates BUT **the FileUploadComponent (in SequencesForm.tsx) shows the newly dropped filename from its own local UI state, independent of whether the parent state actually accepted the change**.

### Screenshot
One of the changes here is that if you drop the sequence you edited back it the fileName will show and the sequence will be updated.

### PR Checklist

### DIFF
--- a/website/src/components/Edit/EditableSequences.ts
+++ b/website/src/components/Edit/EditableSequences.ts
@@ -17,7 +17,6 @@ type EditableSequenceFile = {
     value: string | null;
     initialValue: string | null;
 };
-
 export class EditableSequences {
     private static nextKey = 0;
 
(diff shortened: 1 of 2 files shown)


website/src/components/Edit/SequencesForm.tsx
```diff
@@ -37,31 +37,48 @@ export class EditableSequences {
...
-            ...row,
-            label: row.label ?? `Segment ${i + 1}`,
-        }));
-        if (rows.length < this.maxNumberOfRows) {
-            rows.push({
-                label: `Add a segment`,
-                fastaHeader: null,
-                value: null,
-                initialValue: null,
-                key: EditableSequences.getNextKey(),
```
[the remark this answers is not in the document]
fengelniederhammer: Why does that actually matter? AFAICS we don't show those keys anywhere, do we?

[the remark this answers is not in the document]
anna-parker: we dont but I think its good to ensure react code doesnt have unexpected side effects if we end up changing this code in future. I also hope splitting out the modification from the getter makes the `EditableSequences` class a bit more understandable

anna-parker:
@claude review

website/src/components/Edit/EditableSequences.ts
tombch, line 1: Ran this locally and encountered an issue for multi segmented case - adding a segment then re-adding the same segment results in an error toast, but the segment with the error is still visible:

Screenshot 2026-09-22 at 13-43-52 Submit sequences Loculus

anna-parker: hmm maybe there is some odd interaction with the folderUploadComponent that it stores the name - will investigate later! (if you find it out before me feel free to edit this PR)

tombch: Yeah claude has proposed to me a fix, tbh I don't really understand it so going to have more of a look

Issue #5615, opened 2025-12-02
Group Page: expose group information (without the emails) for all users even when not logged in


anna-parker:
When I click on a group page e.g. https://main.loculus.org/group/1 I want to see all information (except email) even when I am not logged in. 

The group page information on the website is here: /Users/aparker-adm/Documents/Github/loculus/website/src/pages/group/[groupId] - it calls the backend here: /Users/aparker-adm/Documents/Github/loculus/website/src/services/groupManagementClient.ts

This calls the backend endpoint: 
```
@Operation(description = "Get details of a group.")
    @ResponseStatus(HttpStatus.OK)
    @GetMapping("/groups/{groupId}", produces = [MediaType.APPLICATION_JSON_VALUE])
    fun getUsersOfGroup(
        @Parameter(
            description = "The id of the group to get details of.",
        ) @PathVariable groupId: Int,
    ): GroupDetails = groupManagementDatabaseService.getDetailsOfGroup(groupId)
```
/Users/aparker-adm/Documents/Github/loculus/backend/src/main/kotlin/org/loculus/backend/controller/GroupManagementController.kt

Steps:

First PR
1. Make the endpoint return different results if the user is logged in or not. 
2. Update the swaggerUI with this information.
Second PR
1. Use this new endpoint in the website

PR #5647, opened 2025-12-08
feat(backend): Expose a subset of group info to users that are not logged in

maverbiest:
resolves https://github.com/loculus-project/loculus/issues/5615 (first part)

### Background
We want to expose some group information when requests to /groups/{groupID} are made to the backend without authentication:
- Group details with the emailContact and user fields set to null should be returned when a GET request is made to the backend at groups/{groupId}.
- Full set of information should be returned by a GET request to the same endpoint by an authenticated user.

### Overview of changes:
- "/groups/*" added to "getEndpointsThatArePublic"
- "getDetailsOfGroup()" checks if user is logged in and decides what to return
- "GroupDetails" data class 'users' field is now optional
- "Group" data class 'emailContact' field is now optional
- To create a "Group" with redacted email, use "GroupEntity.toGroup(redactEmail = true)"

### What I tested manually
`curl` GET requests to the endpoint behave as expected: contactEmail is included when using an authentication token, and set to N/A when there is no token provided. I also checked the swagger-ui to verify that the documentation there is updated and reflects the new behaviour.

### Open issues & questions:
- Do we want to do something similar for the "/groups" endpoint? It could return a list of all Groups with the contact info redacted.

### PR Checklist

🚀 Preview: https://expose-group-info.loculus.org

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/service/groupmanagement/GroupManagementDatabaseService.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/service/groupmanagement/GroupManagementDatabaseService.kt
@@ -33,7 +33,6 @@ class GroupManagementDatabaseService(
         return when (user is AuthenticatedUser) {
             true -> {
                 val users = UserGroupEntity.find { UserGroupsTable.groupIdColumn eq groupId }
-                println("$users\n\n\n\n")
                 GroupDetails(
                     group = groupEntity.toGroup(),
                     users = users.map { User(it.userName) },


chaoran-chen (commented):
Looks great, thanks! I only have the same comments like Cornelius.

backend/src/main/kotlin/org/loculus/backend/api/GroupManagementTypes.kt
```diff
@@ -43,4 +43,20 @@ data class Group(
 
 data class User(val name: String)
 
-data class GroupDetails(val group: Group, val users: List<User>)
+data class ApiGroup(
```
anna-parker: instead of creating a new ApiGroup object I would probably just switch to updating the Group object so that the contactEmail is optional (might be easiest to then also use this in the other groups endpoints immediately in this PR then)

maverbiest: You're right, that's probably the most straightforward. As you suggested, I made the emailContact field on Group optional and got rid of ApiGroup.

This means that all group related endpoints are using the same Group type now, since they already were 😅 

Or did you mean that I can make the other endpoints available for unauthenticated users as well in this PR?

backend/src/main/kotlin/org/loculus/backend/service/groupmanagement/GroupManagementDatabaseService.kt
```diff
@@ -27,14 +27,25 @@ class GroupManagementDatabaseService(
...
-        val users = UserGroupEntity.find { UserGroupsTable.groupIdColumn eq groupId }
 
-        return GroupDetails(
-            group = groupEntity.toGroup(),
-            users = users.map { User(it.userName) },
-        )
+        return when (user is AuthenticatedUser) {
+            true -> {
+                val users = UserGroupEntity.find { UserGroupsTable.groupIdColumn eq groupId }
+                println("$users\n\n\n\n")
```
anna-parker: it looks like this was a line added for testing, lets remove it :-)
```suggestion
```

Issue #5646, opened 2025-12-08
deposition: improve typing


anna-parker:
Issues such as https://github.com/loculus-project/loculus/issues/5279 have occured due to poor typing of the database entries in the ena-deposition code. 

[pydantic](https://docs.pydantic.dev/latest/), [TypedDicts](https://typing.python.org/en/latest/spec/typeddict.html) or the ORM SQLALchemy might improve this.

### typedDicts
- do not error on bad type while code is running, but we can configure a linter to error with poor type, see example refactor PR: https://github.com/loculus-project/loculus/pull/5312
- more verbose code required than if we use pydantic, but can create a typeddict of a subset of fields for updating (can pydantic do this?)
### pydantic
- will error during runtime and thus causes some performance overhead
- we already use pydantic for validating the group info response from Loculus so using this for getting and setting database entries would make sense
### SQLAlchemy
- when passing query values as a dictionary SQLAlchemy validates that keys in a query exist in a database (does not validate their type)
- return a list of entries of entry type. This might actually satisfy our needs and clean up the db code.

PR #5652, opened 2025-12-08
refactor(deposition): add AccessionVersion as a type, add a primaryKey property to db table entry classes

anna-parker:
part of https://github.com/loculus-project/loculus/issues/5646

- add AccessionVersion and ProjectId as a type
- add a primaryKey property to db table entry classes
- use primaryKey object in ColumnCheckConfig
- refine functions like `last_version`, `is_revision` that expected a dictionary AccessionVersion object to take an AccessionVersion object
- convert StatusAll and Status into StrEnum - removing the requirement for a custom `def __str__(self)` function

### PR Checklist
- ~All necessary documentation has been adapted.~
- ~Any manual testing that has been done is documented (i.e. what exactly was tested?)~ -> integration tests pass this should be sufficient

🚀 Preview: https://ena-accessionversion.loculus.org

### DIFF
--- a/ena-submission/scripts/test_ena_submission_integration.py
+++ b/ena-submission/scripts/test_ena_submission_integration.py
@@ -702,7 +702,7 @@ class TestFirstPublicUpdate(TestSubmission):
         # Hence the 2 branches below
         if add_function == add_to_project_table:
             # Single key (like project_id)
-            conditions = asdict(entry.primary_key)
+            conditions = {"project_id": entity_id}
         else:
             conditions = asdict(entry.primary_key)
 


ena-submission/src/ena_deposition/create_assembly.py
```diff
@@ -586,7 +590,7 @@ def assembly_table_create(db_config: SimpleConnectionPool, config: Config, test:
         number_rows_updated = update_db_where_conditions(
```
corneliusroemer: Rather than converting at each call site - why not do the conversion in the callee?

corneliusroemer: Whenever you find yourself changing a lot of callers - consider modifying the callee instead

ena-submission/src/ena_deposition/create_assembly.py
```diff
@@ -348,9 +350,10 @@ def can_be_revised(config: Config, db_config: SimpleConnectionPool, entry: dict[
     3. metadata fields in manifest haven't changed since previous version, otherwise
        requires manual revision
     """
-    if not is_revision(db_config, entry):
+    seq_key = AccessionVersion(accession=entry["accession"], version=entry["version"])
```
anna-parker: it is being created here - not sure version is actually an int in this case

Issue #5447, opened 2025-11-17
Flaky integration test due to "revise" clicked before hydration


corneliusroemer:
Pretty sure this failure is because "Revise" was clicked before it was hydrated. The click happened but the page didn't change.

https://github.com/loculus-project/loculus/actions/runs/19392254322/job/55487397031?pr=5439#step:22:84

```

  1) [firefox-without-dep] › tests/specs/features/search/override-hidden-fields.spec.ts:8:5 › Override hidden fields 

    Test timeout of 200000ms exceeded.

    Error: locator.fill: Test timeout of 200000ms exceeded.
    Call log:
      - waiting for getByLabel('Collection date')

```

<img width="1803" height="854" alt="Image" src="https://github.com/user-attachments/assets/2499a758-769b-4481-a5a0-dfb4addcba72" />

corneliusroemer:
This is what playwright trace viewer shows:

<img width="1880" height="756" alt="Image" src="https://github.com/user-attachments/assets/8eedc21c-d8b6-4a67-a9f0-b965c2d58ff2" />

The href is there indeed. Not sure why there's the red triangle on the click target.

I see these on other successful clicks as well though so another red herring possibly.

PR #5666, opened 2025-12-09
chore(e2e): try deflaking revise button click by waiting for sequence to be visible first

corneliusroemer:
Attempt to mitigate #5447

Also centralize clicking of revise button click on page for better maintainability.

Testing: CI passes, so at least not breaking anything. Whether or not it resolves the issue we'll see over time.


theosanderson (approved):
Sounds fine to me - at some point in the future we may need to amend it as for some organisms the sequence viewer won't load by default,  but this is fine for now

theosanderson:
[Re PR title:] Until now we have not called these tests `e2e` (because that name was used for the other tests). I guess we can now if we want, no strong feelings.

PR #5696, opened 2025-12-12
refactor(deposition): use pydantic for config value validation

anna-parker:
part of https://github.com/loculus-project/loculus/issues/5646

- use pydantic to define config types/classes and subtypes/subclasses and validate config input is in expected format
- use new type objects in code (less dictionaries!!!)
- use proper types in tests (taxon_id has to be an int actually)
- remove one level of indirection in config (move values in enaDeposition map to standard level) to improve readability. 
- topology and molecule_type are validated during config creation (less functions required) 

This should not change any logic or functionality

### Screenshot
@maverbiest this is also a large change - would be nice if you had a look for understanding but I also don't expect you to approve :-) 

### PR Checklist
~All necessary documentation has been adapted.~
confirmed cronjob works: https://loculus.slack.com/archives/C07FP0BGHDZ/p1765614040247289
confirmed pods start on preview with the new config validation: 
image

### DIFF
--- a/ena-submission/scripts/test_ena_submission.py
+++ b/ena-submission/scripts/test_ena_submission.py
@@ -32,6 +32,8 @@ from ena_deposition.ena_submission_helper import (
     reformat_authors_from_loculus_to_embl_style,
 )
 from ena_deposition.ena_types import (
+    MoleculeType,
+    Topology,
     default_project_set,
     default_sample_set_type,
 )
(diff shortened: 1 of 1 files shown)


ena-submission/scripts/get_ena_submission_list.py
```diff
@@ -213,7 +213,7 @@ def get_ena_submission_list(config_file) -> None:
     all_entries_to_submit: dict[AccessionVersion, dict[str, Any]] = {}
     for organism in config.organisms:
         ena_specific_metadata_fields = [
-            value["name"] for value in config.organisms[organism]["externalMetadata"]
+            value.name for value in config.organisms[organism].externalMetadata
```
corneliusroemer: In general might be nice to have a more specific name for what the value is. This predates of course, just a thought reading the code

PR #5703, opened 2025-12-15
feat!(k8s): make silo-import refresh and poll intervals configurable

corneliusroemer:
Makes the silo-import refresh and polling intervals configurable through `values.yaml`, consolidating timing configuration under a new `siloImport` section.

For e2e/dev deployments reduce poll interval to 5s to speed up tests.

**Breaking change**: The `siloImportLimitSeconds` top-level field has been moved to `siloImport.siloTimeoutSeconds`.

### Before
```yaml
siloImportLimitSeconds: 3600  # Top-level field
# hardRefreshInterval: hardcoded to 3600 in Python
# pollInterval: hardcoded to 30 in Python
```

### After
```yaml
siloImport:
  siloTimeoutSeconds: 3600           # Timeout for SILO preprocessing
  hardRefreshIntervalSeconds: 3600   # Now configurable! Forces re-download regardless of ETag
  pollIntervalSeconds: 30            # Now configurable! Check interval for new data
```

## Motivation

Previously, the hard refresh interval (how often to force a full re-download ignoring ETags) and poll interval (how often to check for new data) were hardcoded in the Python code with defaults of 3600s and 30s respectively. This made it impossible to tune these values for different deployment scenarios without modifying code.

For example:
- Preview environments might want more frequent hard refreshes to catch issues
- Production environments might want longer poll intervals to reduce load
- Testing environments might want very short intervals for faster iteration

## Changes

1. **values.yaml**: Restructured to nest all silo-import timing configs under `siloImport`:
   - `siloImport.siloTimeoutSeconds` (renamed from `siloImportLimitSeconds`)
   - `siloImport.hardRefreshIntervalSeconds` (newly configurable)
   - `siloImport.pollIntervalSeconds` (newly configurable)

2. **values.schema.json**: Added validation schema for the new `siloImport` object with:
   - Proper descriptions for each field
   - Minimum value validation (`minimum: 1`)
   - Default values matching current behavior

3. **silo-deployment.yaml**: Updated to pass all three configuration values as environment variables to the silo-importer container

## Migration Guide

If you have customized `siloImportLimitSeconds` in your helm overrides, update:

```yaml
# Old
siloImportLimitSeconds: 7200

# New
siloImport:
  siloTimeoutSeconds: 7200
```

Default values remain unchanged, so deployments without overrides will continue to work as before.

### DIFF
--- a/kubernetes/loculus/templates/silo-deployment.yaml
+++ b/kubernetes/loculus/templates/silo-deployment.yaml
@@ -81,7 +81,11 @@ spec:
               value: {{ index $.Values.lineageSystemDefinitions $lineageSystem | toJson | quote }}
             {{- end }}
             - name: SILO_RUN_TIMEOUT_SECONDS
-              value: {{ $.Values.siloImportLimitSeconds | quote }}
+              value: {{ $.Values.siloImport.siloTimeoutSeconds | quote }}
+            - name: HARD_REFRESH_INTERVAL
+              value: {{ $.Values.siloImport.hardRefreshIntervalSeconds | quote }}
+            - name: SILO_IMPORT_POLL_INTERVAL_SECONDS
+              value: {{ $.Values.siloImport.pollIntervalSeconds | quote }}
           volumeMounts:
             - name: lapis-silo-database-config-processed
               mountPath: /preprocessing/input/reference_genomes.json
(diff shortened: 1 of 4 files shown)


corneliusroemer:
Plan is to a) reduce poll interval for integration tests to speed them up, b) increase hard refresh in general on AWS deployments (we trust etag now), c) potentially increase poll for staging

corneliusroemer:
Reducing poll interval to 5 seconds saves about 10% of E2E test execution 🎉 cheap win!

PR #5712, opened 2025-12-15
perf(backend): stream released data efficiently via optimized query (function)

corneliusroemer:
Performance investigations on prod db showed that external metadata joining was the culprit behind our very long running get-released-data queries.

Dropping the EM join reduced time from 200s to 2s for RSV-A. There was lots of looping for each EM row. Somehow the way our views were structured made postgres execute this query really inefficiently.

This PR adds a new postgres function that is optimized for streaming released data. It's a function not a view because views can't be parametrized (with organism in this case).

Further efficiency increase is due to not using the expensive sequence entries view for count and getting latest versions - this can be gleamed easily from just the sequence entries table alone.

## Summary
- add a typed Exposed binding for get_released_submissions so released streams stay schema-safe
- switch streamReleasedSubmissions to the function binding and keep count/latest queries on released-only base tables
- include the SQL migration defining get_released_submissions with early filters and pre-aggregated metadata/data use terms

## Testing

- [x] Tests pass
- [x] Previews look good
- [ ] Put on PPX staging to verify perf improvements


backend/docs/db/schema.sql
theosanderson: Some discussion in https://loculus.slack.com/archives/C05G172HL6L/p1765813389287539 about other options to potentially also solve the core issue that Cornelius has identified

anna-parker:
closing as we did this with a different function - thanks again @corneliusroemer !

PR #5726, opened 2025-12-16
ci: Disable claude review for dependabot PRs

theosanderson:
Disable the Claude code review GitHub Action for dependabot PRs since we don't have a dependabot secret set up.


corneliusroemer:
The copilot reviews were quite nice on dependabot PRs as no one had looked at them yet. So might be good to add a dependabot secret instead of disabling?

PR #5727, opened 2025-12-16
feat(db): improve perf with external_metadata  changes

theosanderson:
@corneliusroemer identified that

> Performance investigations on prod db showed that external metadata joining was the culprit behind our very long running get-released-data queries.

This is a way of fixing the issues he identified, inspired by his work. It avoids the previous situation where a GROUP BY was nested deep in a series of views.

🚀 Preview: https://claude-learn-database-vie.loculus.org


corneliusroemer:
One needs to rename migration file

theosanderson:
OK, thanks will clean up. Happy to have a big chat about architecture sometime! I haven't tested this at all really so I'll be relying heavily on your testing - with thanks!

corneliusroemer:
I'll keep playing with performance and might rewrite things more generally but being a bigger change it's much better to put this one out for now. I'll test with our usual pipeline - this should reliably show problems if there are any left.

PR #5737, opened 2025-12-17
feat(deposition): use unittest instead of pytest to test the cronjob that uses the backend's get-released-data format

maverbiest:
resolves #

This PR only switches out which testing framework is used for testing `scripts/get_ena_submission_list.py`. Instead of using pytest (as in https://github.com/loculus-project/loculus/pull/5226), we now use unittest. This is more convenient since unittest Mock objects keep track of how often and with what arguments they are called, whereas this needed to be implemented manually under pytest.

### PR Checklist
- [ ] ~Any manual testing that has been done is documented (i.e. what exactly was tested?)~ This PR only adds tests, not functionality. The test runs fine using both `pytest ./path/to/test.py` or `python ./path/to/test.py`

### DIFF
--- a/ena-submission/scripts/test_get_ena_submission_list.py
+++ b/ena-submission/scripts/test_get_ena_submission_list.py
@@ -56,7 +56,7 @@ class GetSubmissionListTests(unittest.TestCase):
 
         mock_upload_file_with_comment.return_value = {"ok": True}
         mock_fetch_released_entries.side_effect = fake_fetch_released_entries
-        mock_fetch_suppressed_accessions.return_value = "LOC01.1"
+        mock_fetch_suppressed_accessions.return_value = {"LOC01.1"}
 
         get_ena_submission_list.get_ena_submission_list.callback(config_file=str(CONFIG_FILE))  # type: ignore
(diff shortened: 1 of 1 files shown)


ena-submission/scripts/test_get_ena_submission_list.py
```diff
@@ -33,67 +35,51 @@ def fake_fetch_released_entries(config, organism) -> Iterator[dict[str, Any]]:
...
+    @patch("get_ena_submission_list.fetch_released_entries")
+    @patch("get_ena_submission_list.upload_file_with_comment")
+    @patch("get_ena_submission_list.slack_conn_init")
+    @patch("get_ena_submission_list.highest_version_in_submission_table")
+    @patch("get_ena_submission_list.db_init")
+    def test_happy_path_single_upload_and_file_content(
+        self,
+        mock_db_init: Mock,
+        mock_highest_version_in_submission_table: Mock,
+        mock_slack_conn_init: Mock,
```
anna-parker: I just realized - we anyways use a test config without the slack hook details - so we can actually get away with not mocking the `slack_conn_init`

Issue #4980, opened 2025-09-09
Suggest existing groups when creating duplicates


corneliusroemer:
**Problem:**
People keep creating groups with identical names (like having multiple "Lab ABC" groups). This creates confusion and splits communities.

**Solution - 3 stages:**

**Stage 1 - Basic warning (easy)**
When someone types "Research Lab" and "Research Lab" already exists, show:
```
⚠️ "Research Lab" already exists (5 members)
□ Create anyway   [Create] [Cancel]  
```

**Stage 2 - Fuzzy matching (medium)** 
Also catch similar names:
```
⚠️ Similar groups found:
📁 Research Lab (exact match, 5 members)
📁 Research Labs (3 members) 
📁 Lab Research Group (8 members)
```

**Stage 3 - Let people connect (harder)**
Show contact info so they can join existing groups instead:
```
📁 Research Lab (5 members)
   Contact: admin@example.com
   [Message Admin] [Request to Join]
```

**Why staged?** 
Stage 1 catches exact duplicates with zero risk. Stages 2-3 add fancier features once we know Stage 1 works.

**Acceptance criteria:**
- [ ] Warn when exact group name already exists
- [ ] User can override and create anyway
- [ ] No performance impact on group creation

PR #5766, opened 2025-12-19
feat(backend, website): alert users during group creation when groupname is in use

maverbiest:
resolves #

This PR will introduce changes to alert users during the submitting group creation process when they are trying to create a group with a name that is already in use. See Stage 1 of https://github.com/loculus-project/loculus/issues/4980 for details.

### New features
- Updated /groups endpoint. This endpoint still returns all groups when a simple GET request is made, but it now also accepts an optional request parameter `name` (i.e., /groups?name=myGroupName). If this parameter is provided, the list of groups is filtered to return only groups where the name is a case-insensitive perfect match to the request parameter.
- Website uses the updated endpoint to check for existing groups with the same name as the group being created. This happens in both the group creation and the group edit form.
- If one or more existing groups exist with the name: open a modal to alert the user and let them know that they can:
    1. Choose another group name
    2. Contact one of the existing groups using contact information shown in the modal
    3. Go ahead with group creation anyway

### What I tested manually
- Group creation/edit works as before when the new group's name does not already exist
- If the name does exist, an alert comes up containing contact info of the existing groups
- Clicking the 'create additional group' button disables the button, so spam clicking it does not result in multiple duplicate groups being created (was the case initially)

### Screenshot
The alert (table of existing groups is scrollable if there are enough existing groups):
Screenshot 2026-01-14 at 13 51 29

### PR Checklist

🚀 Preview: https://similar-groups-alert.loculus.org

### DIFF
--- a/website/src/components/Group/GroupForm.tsx
+++ b/website/src/components/Group/GroupForm.tsx
@@ -45,6 +45,11 @@ interface GroupFormProps {
      */
     defaultGroupData?: NewGroup;
     /**
+     * Provide this when editing existing group to prevent 'existing group alert'
+     * for the group that's currently being edited
+     */
+    editingGroupId?: number;
+    /**
      * A handler to call when the button is clicked (i.e. create or update a group).
      * @param group The new group information entered into the form.
      * @returns A submit success or error.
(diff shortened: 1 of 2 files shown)


website/src/components/Group/GroupForm.tsx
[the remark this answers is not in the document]
maverbiest: Hmm I'm a bit stumped: when I run the backend & website locally, I don't get this error. Editing a group to an existing name works without issue.

When I try the same on the preview, however, I'm also getting the internal server error you're describing

website/src/components/Group/ExistingGroupsModal.tsx
```diff
@@ -0,0 +1,75 @@
...
+                    </tbody>
+                </table>
+            </div>
+
+            <div className='mt-4 flex justify-center'>
+                <Button
+                    type='button'
+                    className='btn loculusColor text-white'
+                    onClick={() => void submitFromModal(newGroup)}
+                    disabled={isSubmitting}
```
anna-parker: I guess the `isSubmitting` field is only used in this modal - so maybe we could just have it defined in this modal and not also in the groupForm (for example we could set it before submitModal and the unset it)

maverbiest: It's also used in `submitGroup()` in the `GroupForm` component itself, though. I kind of like having it there so it also protects against spam submissions when submitting group creations & edits from outside the modal.

PR #5767, opened 2025-12-19
feat(backend, deposition): add a deposition status to the backend and use to filter get-released-data prior to streaming

anna-parker:
resolves #

### Summary

Add a new field to the sequences_entries_view called `ena_deposition_status` which is an enum of type:
```
enum class EnaDepositionStatus {
    INGESTED,
    READY,
    DEPOSITED,
}
```

This column is used as a coarse-grained filter option when calling the `get-released-data` endpoint. 

By adding the param `filterForEnaDeposition` the endpoints stops streaming sequences that were ingested from the INSDC or have been deposited by us in the INSDC with external metadata updated. Edge cases, such as revised versions, restricted data etc. are still handled by the ena cronjob.

The cronjob now takes 1min and 30seconds to run instead of the 15-20min it currently takes. 
```
14:40:08     INFO (           config.py: 135) - Submitting to ENA dev environment
...
14:41:31     INFO (     call_loculus.py: 160) - Fetching released data from https://backend-staging.pathoplexus.org/west-nile/get-released-data with request id 71af1dab-c6b1-4254-b7ae-5da099d29856
```
I also confirmed with spot checking that the new status was set correctly for these submissions. 

### PR Checklist


backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -694,63 +695,70 @@ class SubmissionDatabaseService(
...
-            SequenceEntriesView.jointDataColumn,
-            SequenceEntriesView.submitterColumn,
-            SequenceEntriesView.groupIdColumn,
-            SequenceEntriesView.submittedAtTimestampColumn,
-            SequenceEntriesView.releasedAtTimestampColumn,
-            SequenceEntriesView.submissionIdColumn,
-            SequenceEntriesView.pipelineVersionColumn,
-            DataUseTermsTable.dataUseTermsTypeColumn,
-            DataUseTermsTable.restrictedUntilColumn,
+    fun streamReleasedSubmissions(organism: Organism, enaDeposition: Boolean = false): Sequence<RawProcessedData> =
```
corneliusroemer: Why is the diff so large? Add more explanation to pr description as well

theosanderson:
@codex review pls

backend/docs/db/schema.sql
theosanderson: I guess my bigger picture questions would be whether some day we'll want to break up `DEPOSITED` into lots of substatuses - but it's not a problem for me to do this for now

Issue #5663, opened 2025-12-09
Multi Path: Multi Segment - issue with ASSIGNED_SEGMENT


anna-parker:
The website requires prepro to set a `suborganismIdentifierField` - I added ASSIGNED_SEGMENT to fix this. However, it returns the name of the "segment" that was identified by preprocessing. For the single segmented organisms this is the subtype - however for the multi-segmented organisms this will be `<subtype>_<segment>` - I believe we need to have prepro only return the subtype. This might require a prepro config update. Additionally, I believe `useFirstSegment` will also break in the same way for the multi-segmented multi-path case

anna-parker:
My proposed solution would be to add a segment field, but potentially the config could be improved on instead:
```
nextclade_sequence_and_datasets:
    - name: CV-A16 # name of the segment in proccessedData.alignedSegments results dict
      nextclade_dataset_name: enpen/enterovirus/cv-a16
      accepted_sort_matches: ["community/hodcroftlab/enterovirus/cva16", "community/hodcroftlab/enterovirus/enterovirus/linked/CV-A16"]
      gene_prefix: "CV-A16-"
      genes: ["VP4", "VP2", "VP3", "VP1", "2A", "2B", "2C", "3A", "3B", "3C", "3D"]
      segment: main # name of the segment returned in ASSIGNED_SEGMENT and useFirstSegment
```

anna-parker:
In the example above I actually dont know how to handle the case where the CCHF_A and CCHF_B subtype both have the same reference for that segment - which name should we return `CCHF_A-M` or `CCHF_B-M`? this kinda leads me back to https://github.com/loculus-project/loculus/issues/5664

Issue #5682, opened 2025-12-10
EVs: search page - should clades of suborganisms be selectable before a subtype is selected?


anna-parker:
I think that the clades should only be visible after selecting a genotype, as selecting a clade of a genotype is the same thing as selecting the genotype:

<img width="626" height="1386" alt="Image" src="https://github.com/user-attachments/assets/c7138b7f-3dce-49b4-a093-52b1e81fa400" />

Issue #5664, opened 2025-12-09
Multi path: Multi Segmented Organisms with reassortment


anna-parker:
When thinking about extending multi path to multi segmented organisms yesterday I realized the config structure is potentially flawed.

Currently the config structure for a multi-segmented multi-path would be:

```
referenceGenomes:
  CCHF_A:
    nucleotideSequences:
      - name: S
        sequence: ...
      - name: M
        sequence: ...
      - name: L
        sequence: ...
    genes:
      - name: gene1
        ...
  CCHF_B:
    nucleotideSequences:
      - name: S
        sequence: ...
      - name: M
        sequence: ...
      - name: L
        sequence: ...
    genes:
      - name: gene1
        ...
```
But multi-segmented viruses can reassort, meaning we could have a CCHF_A-S with a CCHF_B-M segment. 
Prepro can return `CCHF_A-S` and `CCHF_B-M` aligned sequences - but it will not know which subtype to set using ASSIGNED_SEGMENT. Additionally, the frontend now needs to know explicitly which subtype we are looking at i.e. `CCHF_A` or `CCHF_B` (LAPIS, the backend and prepro actually don't need to know this).

I think it makes sense to make `suborganismIdentifierField` a segment-level field. But I wonder if this fits in with the config structure. I wonder if the following structure would make more sense.

```
referenceGenomes:
  nucleotideSequences:
    - name: S
      references:
       CCHF_A: ...
       CCHF_B: ...
    - name: M
      references:
       sequence: ... # if the M segment only has one reference
    - name: L
      references:
       CCHF_A: ...
       CCHF_B: ...
  genes:
    - name: gene1
      references:
       CCHF_A: ...
       CCHF_B: ...
    ...
```

anna-parker:
In summary this is a switch from 
```
defaultOrganisms:
  organism:
    referenceGenomes:
      subtype1:
        nucleotideSequences:
          - segment1
          - segment2
      subtype2:
        nucleotideSequences:
          - segment1
          - segment2
```
to 
```
defaultOrganisms:
  organism:
    referenceGenomes:
      nucleotideSequences:
        - segment1
          references:
          - reference1: ...
          - reference2: ...
        - segment2
          references:
          - reference1: ...
          - reference2: ...
```

Note co-infection will not be possible with this new structure we would need to add in another level:

```
defaultOrganisms:
  sample:
    organism:
      referenceGenomes:
        nucleotideSequences:
          - segment1
            references:
[...]

PR #5799, opened 2026-01-05
feat!(config, website): flip config and website to segment-reference structure

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/5663, https://github.com/loculus-project/loculus/issues/5664 and https://github.com/loculus-project/loculus/issues/5682

This adds an example of a multi-reference multi-segment CCHF organism (only the segments S and M have multiple references)

image

### Architecture
As LAPIS does not have a specific way to define references of a segment each reference of a segment is stored as a unique segment in LAPIS - we use a namespaced naming convention to ensure there are no key clashes:

#### LAPIS Nucleotide sequence naming 
```
    match (multi_reference, multi_segment):
        case (False, _):
            return segmentName
        case (True, True):
            return f"{segmentName}-{referenceName}"
        case (True, False):
            return referenceName
```

#### LAPIS Gene sequence naming 
```
    match (multi_reference, multi_segment):
        case (False, _):
            return GeneName
        case (True, _):
            return f"{GeneName}-{referenceName}"
```

### Code changes

This is a large refactor to use the new structure in the website code. The code also deduplicates the calculation of the lapisName and displayName which was previously performed in multiple locations of the code.

It also fixes a bug in the way `referenceOnly` metadata fields are displayed - when references are not selected they are greyed out but still selectable, when a different reference than the one they are linked to it selected they are disabled (with a note) and when the reference they are linked to is activated they are selectable. 

When no reference is selected `referenceOnly` metadata are still downloadable - but are not shown on the search page.

### Config Changes
```
referenceGenomes:
      singleReference:
        nucleotideSequences:
          - name: "main"
            sequence: "[[URL:https://corneliusroemer.github.io/seqs/artefacts/ebola-sudan/reference.fasta]]"
            insdcAccessionFull: NC_002549.1
        genes:
          - name: NP
            sequence: "[[URL:https://corneliusroemer.github.io/seqs/artefacts/ebola-sudan/NP.fasta]]"
          ....
```
becomes:
```
referenceGenomes:
      - name: main
        references:
          - name: singleReference
            sequence: "[[URL:https://corneliusroemer.github.io/seqs/artefacts/ebola-sudan/reference.fasta]]"
            insdcAccessionFull: NC_002549.1
            genes:
              - name: NP
                sequence: "[[URL:https://corneliusroemer.github.io/seqs/artefacts/ebola-sudan/NP.fasta]]"
              - name: VP35
                sequence: "[[URL:https://corneliusroemer.github.io/seqs/artefacts/ebola-sudan/VP35.fasta]]"
```

Additionally, in prepro the `reference_name` field is renamed as `name` for consistency.

### PR Checklist
- [x] Add the lapisName vs displayName logic to the config structure at the start so that this doesnt have to be recalculated again and again in the code
- [x] Get the existing website tests to work
- [ ] Add integration tests for the multi-segment, multi-reference case
- [ ] Later steps: ensure ENA deposition works

🚀 Preview: https://restructure-anya.loculus.org

### DIFF
--- a/integration-tests/tests/specs/features/multiseg-multiref-submission-flow.spec.ts
+++ b/integration-tests/tests/specs/features/multiseg-multiref-submission-flow.spec.ts
@@ -1,0 +1,140 @@
+import { test } from '../../fixtures/group.fixture';
+import { BulkSubmissionPage, SingleSequenceSubmissionPage } from '../../pages/submission.page';
+import { expect } from '@playwright/test';
+import {
+    CCHF_L_SEGMENT_FULL_SEQUENCE,
+    CCHF_M_SEGMENT_FULL_SEQUENCE,
+    CCHF_S_SEGMENT_FULL_SEQUENCE,
+} from '../../test-helpers/test-data';
+
+test.describe('Multi-segment multi-reference submission flow', () => {
+    test('submit single sequence, edit and release', async ({ page, groupId }) => {
+        test.setTimeout(120_000);
+
+        void groupId;
+        const submissionPage = new SingleSequenceSubmissionPage(page);
+
+        await submissionPage.navigateToSubmissionPage('CCHF (Multi-Ref)');
+        await submissionPage.fillSubmissionForm({
+            submissionId: 'MSMR-TEST-001',
+            collectionCountry: 'Laos',
+            collectionDate: '2023-10-15',
+            authorAffiliations: 'Research Lab, University',
+        });
+        await submissionPage.fillSequenceData({ mySequence: 'should not align' });
+        await submissionPage.acceptTerms();
+        const reviewPage = await submissionPage.submitSequence();
+
+        await reviewPage.waitForAllProcessed();
+        const editPage = await reviewPage.editFirstSequence();
+
+        await editPage.discardSequenceFile();
+        await editPage.addSequenceFile(`>key\n${CCHF_S_SEGMENT_FULL_SEQUENCE}`);
+        await editPage.addSequenceFile(`>key2\n${CCHF_L_SEGMENT_FULL_SEQUENCE}`);
+        await editPage.fillField('Authors', 'Integration, Test');
+        await editPage.submitChanges();
+
+        await reviewPage.waitForAllProcessed();
+        const releasedPage = await reviewPage.releaseAndGoToReleasedSequences();
+
+        const accessionVersions = await releasedPage.waitForSequencesInSearch(1);
+        const firstAccessionVersion = accessionVersions[0];
+        await releasedPage.openPreviewOfAccessionVersion(`${firstAccessionVersion.accession}.1`);
+        const expectedDisplayName = new RegExp(
+            `^Display Name: Laos/${firstAccessionVersion.accession}\\.1`,
+        );
+        await expect(page.getByText(expectedDisplayName)).toBeVisible();
+        await expect(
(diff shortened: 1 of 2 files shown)


theosanderson:
@codex can you review?

website/src/utils/getSegmentAndGeneInfo.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+    for (const [segmentName, segmentData] of Object.entries(referenceGenomes)) {
+        const isSingleReference = Object.keys(segmentData).length === 1;
+        const selectedRef = selectedReferences[segmentName] ?? null;
+
+        const refForNaming = isSingleReference ? null : selectedRef;
+
+        nucleotideSegmentInfos.push(getSegmentInfoWithReference(segmentName, refForNaming, !isMultiSegmented));
+        if (!isSingleReference) {
+            isMultiSegmented = true;
```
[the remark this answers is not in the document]
anna-parker: I think this is accurate - the multiSegmented boolean is currently used for different things in different files (if the data is actually multi-segmented or if LAPIS has multi-segments) - its sth I want to improve on with a better data structure that uses a different name for these two cases

website/src/utils/sequenceTypeHelpers.ts
```diff
@@ -62,3 +52,151 @@ export const isUnalignedSequence = (type: SequenceType): boolean => type.type ==
...
+}
+
+export function getGeneLapisName(geneName: string, referenceName: string, isMultiReferenced: boolean): string {
+    if (!isMultiReferenced) {
+        return geneName;
+    }
+    return `${geneName}-${referenceName}`;
+}
+
+export function toReferenceGenomes(values: ReferenceGenomesSchema): ReferenceGenomes {
```
theosanderson: The reason the kind of object this produced used to be called `ReferenceGenomeLightweightSchema` (which I agree I don't love as a name) is that it doesn't actually contain the reference genome sequence itself. IMO in this PR it's confusing that there are two types:

`ReferenceGenomesSchema` - contains sequences, I think?

`ReferenceGenomes` - doesn't contain sequences

If anything I'd probably expect those to be named in reverse?

theosanderson: So I had called the without-sequence version `lightweightSchema` for that reason (`ReferenceGenomesLightweightSchema`). I think I probably do prefer that to `ReferenceGenomes`, but I'm very open to other options too! @claude maybe you could try suggesting 20 options

theosanderson: I think my pref. of those is `ReferenceGenomesInfo` - structure seems a bit too specifically about e.g. topology and stuff, when this encompasses a bit more

anna-parker: Also quite consistent with the SegmentInfo and GeneInfo types :-)

theosanderson: Thanks!!

backend/docs/organismWithSuborganisms.md
```diff
@@ -76,34 +76,32 @@ defaultOrganisms:
           segments:
             - name: main
               references:
-              - reference_name: CV-A16
+              - referenceName: CV-A16
```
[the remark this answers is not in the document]
anna-parker: changed as discussed in the codex issue :-)

PR #5805, opened 2026-01-07
fix(deployment): hotfix bug in ingest config

theosanderson:
We were missing the right context for `.referenceGenomes`

Ingest now works:

🚀 Preview: https://fixbug3.loculus.org


anna-parker (approved):
thanks! preview looks healthy!

PR #5830, opened 2026-01-11
refactor(integration-tests): remove unused methods

chaoran-chen:
This removes unused code in the integration tests.

(This is using the branch from #5827 as base to avoid conflicts.)


theosanderson:
At first glance I'm not sure about this for most of these. These are all `.page.ts` files. 

I think it's fine for them to have functions that aren't used atm. IMO in their ideal state these files would capture every single functionality of the page in question.

Practical reasons to keep them:
- we might want to use them later
- (more relevant for me) they describe how to interact with these pages using playwright to AI agents

theosanderson:
But maybe we've ended up with a lot of duplication and maybe this broad brush approach is the easiest way to fix it, idk

chaoran-chen:
The problem is that we don't know whether they are actually correct and they can easily get out-of-date. I stumbled over this [here](https://github.com/loculus-project/loculus/pull/5827/files#diff-db886a04f25e74a7ff128cca180aacb5e10fded8c9dbe04b6551e7c8f43864b6L124) where Claude called the `submitRevision` function but it's actually wrong (there is no confirm button) and it took me a while to realize that the function had been unused (what made it more difficult was that the function itself was not directly unused but the functions that used it were).

theosanderson:
> The problem is that we don't know whether they are actually correct and they can easily get out-of-date.

I guess IMO this is true of features that we haven't yet added tests for too - but my solution there would be to add tests rather than remove them (and same here). [It's not a huge deal either way though, and I definitely agree with a good proportion of the removals]

chaoran-chen:
Closed in favor of #5852.

PR #5839, opened 2026-01-15
feat(website): Cumulative graph of submission on groups page

theosanderson:
## feat(website): Add cumulative submissions chart to group page

Adds a line chart showing cumulative submissions over time on the group page, with each organism as its own series.

https://dates-graph.loculus.org/group/1

image


emmahodcroft:
I found this on slack again and just didn't want it to be forgotten 😆

theosanderson:
@codex review

kubernetes/loculus/values.schema.json
```diff
@@ -1060,6 +1060,13 @@
       "default": "LOC_",
       "description": "Prefix used for accession numbers"
     },
+    "metadataItemForCumulativeGroupGraph": {
```
anna-parker: maybe we could be more specific (in the name) that this is a date field

website/src/components/User/CumulativeSubmissionsChart.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+export type TimeSeriesDataPoint = {
+    date: string;
+    count: number;
+};
+
+export type TimeSeriesData = Record<string, TimeSeriesDataPoint[]>;
+
+// Note: Colors will repeat if there are more than 8 organisms
+const ORGANISM_COLORS = ['#54858c', '#e6a756', '#7b68a6', '#5aa469', '#d4776b', '#4a90a4', '#9b8b6e', '#c97b84'];
```
anna-parker, line 26: hmm... do we want to use the same colors we already use for the mutation badges for consistency?

website/src/components/User/GroupPage.tsx
```diff
@@ -294,14 +322,53 @@ async function fetchSequenceCounts(groupId: number, clientConfig: ClientConfig,
...
+            } catch (error) {
+                void logger.error(`Failed to fetch sequence count for ${key}: ${JSON.stringify(error)}`);
                 counts[key] = 0;
             }
         }),
     );
     return counts;
 }
 
+async function fetchTimeSeriesData(
```
anna-parker, line 334: In theory LAPIS has a `/component/queriesOverTime` endpoint that I think could be used for this, but this seems quite logical :-)

website/src/components/User/GroupPage.tsx
```diff
@@ -294,14 +322,53 @@ async function fetchSequenceCounts(groupId: number, clientConfig: ClientConfig,
...
+            const url = clientConfig.lapisUrls[key];
+            if (!url) {
+                data[key] = [];
+                return;
+            }
+            try {
+                const response = await axios.post(`${url}/sample/aggregated`, {
+                    [GROUP_ID_FIELD]: groupId,
+                    [VERSION_STATUS_FIELD]: versionStatuses.latestVersion,
+                    [IS_REVOCATION_FIELD]: 'false',
```
tombch, line 352: Should revoked versions be excluded from the graph, as they were still a submission at one point?

anna-parker: I do think its good to remove them as they could inflate the numbers

anna-parker (approved):
Looks good! As this code is very contained Im happy to merge - I have more layout related suggestions (like using the same colors as the badges) and maybe also we could place this graph next to the cumulative counts instead of below - but can be done in a follow up PR :-)

website/src/components/User/GroupPage.tsx
theosanderson: Yeah it's debatable but I think it's more conservative to exclude them as people may then resubmit fixed versions as new entries

kubernetes/loculus/values.schema.json
theosanderson: @theosanderson-agent rename to dateFieldForGroupGraph

website/src/components/User/CumulativeSubmissionsChart.tsx
theosanderson: I think they are optimised for slightly different purposes - the badges are designed to have black text readable on them, these should have decent contrast on a grey bg. Some of the badge colors I can imagine being an issue for that so will leave this for now (though I'm not saying these colours are perfect). Happy to revisit.

kubernetes/loculus/values.schema.json
theosanderson-agent: Done in 5dffaf931 — renamed to `dateFieldForGroupGraph` across the schema, values, templates, and website code.

Issue #5821, opened 2026-01-09
Link to existing files when submitting on website


chaoran-chen:
It is possible (via the API) to link to the same file from multiple sequence entries of the same group. This may e.g. be useful for having one fastq file that contains reads from different pathogens. It would be great if this can also be done when submitting through the website.

Does anyone have thoughts regarding how this could look like?

theosanderson:
I would possibly say that this could rely on https://github.com/loculus-project/loculus/issues/5823 (i.e. notionally the uploader re-uploads the data). It's just occurred to me that we could actually run the MD5 locally and skip the upload in such cases.

PR #5856, opened 2026-01-19
feat(backend, website): download of original data for revisions

### LINKED ISSUE #3167: Revision in bulk
Because the processed data that you can download easily from the website and the original data that you need to revise sequences are different, it is currently difficult to get the data for revision if you don't have the orginal data around. We should find a solution to facilitate bulk revision.

### LINKED ISSUE #5296: Make it easier for users to do batch revisions
We don't have an easy way for users to do batch revisions. This means they need to have their old original metadata which they need to add loculus accessions to (and revise) or they download the metadata from us with the loculus accessions. The issue with the second option (although it is probably easiest for submitters) users download the processed data and not the original data. For example we reformat authors lists from:
`Tiangay P.M.S Kallom` to `Tiangay P. M. S. Kallom`. 

No change is seen on the page but the ena deposition pod flagged this as a change in author's lists and wanted us to do a manual revision.

An easy solution is to change the ena deposition pod to compare the final authors output and not the original data in this case but in general this is an issue that will probably present itself in other ways.

https://loculus.slack.com/archives/C05G172HL6L/p1761121808948159

chaoran-chen:
resolves #3167, resolves https://github.com/loculus-project/loculus/issues/5296

This PR adds the feature to download original submission data (metadata and sequences) for released sequences, making it easier to prepare revisions in bulk. Previously, users could only download processed data from the website, but revisions require the original submitted data format.

A new backend endpoint `POST /{organism}/get-original-data` returns a zip file containing metadata.tsv and sequences.fasta in the same format expected for submissions. The download is limited to 500 entries as it has not been optimized for memory efficiency (no streaming or out-of-memory storage). On the website, a new button has been added to the released sequences page for downloading the file.

The design of providing a zip download would be compatible with @theosanderson's proposal in https://github.com/loculus-project/loculus/issues/5821#issuecomment-3731045121 to "provide a zip download which has the whole directory structure to upload in order to put in the existing sequences -- except that instead of each file actually containing the contents it just contains the text `md5:da23ef` or whatever, and that's used as a pointer to the existing file". 

### Screenshot

image

### PR Checklist
  - Workflow of downloading original data (test organism and CCHF), changing something, and using the files for revision

🚀 Preview: https://download-original-data.loculus.org

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
@@ -287,7 +287,7 @@ data class EditedSequenceEntryData(
 data class UnprocessedData(
     @Schema(example = "LOC_000S01D") override val accession: Accession,
     @Schema(example = "1") override val version: Version,
-    val data: OriginalDataWithFileUrls<GeneticSequence>,
+    val data: UnprocessedDataContentWithFileUrls<GeneticSequence>,
     @Schema(description = "The submission id that was used in the upload to link metadata and sequences")
     val submissionId: String,
     @Schema(description = "The username of the submitter")
@@ -298,7 +298,7 @@ data class UnprocessedData(
     val submittedAt: Long,
 ) : AccessionVersionInterface
 
-data class OriginalDataInternal<SequenceType, FilesType>(
+data class UnprocessedDataContentInternal<SequenceType, FilesType>(
     @Schema(
         example = "{\"date\": \"2020-01-01\", \"country\": \"Germany\"}",
         description = "Key value pairs of metadata, as submitted in the metadata file",
@@ -316,9 +316,10 @@ data class OriginalDataInternal<SequenceType, FilesType>(
     val files: Map<String, List<FilesType>>? = null,
 )
 
-typealias OriginalData<SequenceType> = OriginalDataInternal<SequenceType, FileIdAndName>
-typealias OriginalDataWithFileUrls<SequenceType> =
-    OriginalDataInternal<SequenceType, FileIdAndNameAndReadUrl>
+typealias UnprocessedDataContent<SequenceType> = UnprocessedDataContentInternal<SequenceType, FileIdAndName>
+typealias OriginalData<SequenceType> = UnprocessedDataContent<SequenceType>
+typealias UnprocessedDataContentWithFileUrls<SequenceType> =
+    UnprocessedDataContentInternal<SequenceType, FileIdAndNameAndReadUrl>
 
 data class AccessionVersionUnprocessedMetadata(
     override val accession: Accession,
@@ -326,6 +327,24 @@ data class AccessionVersionUnprocessedMetadata(
     val submitter: String,
     val isRevocation: Boolean,
     val unprocessedMetadata: Map<String, String?>?,
+) : AccessionVersionInterface
+
+data class GetOriginalDataRequest(
+    @Schema(
+        description = "The group ID to download data for.",
+    )
+    val groupId: Int,
+    @Schema(
+        description = "Filter by specific accessions. If not provided, all accessions for the group are returned.",
+    )
+    val accessionsFilter: List<Accession>? = null,
+)
+
+data class UnprocessedDataDownloadEntry(
+    override val accession: Accession,
+    override val version: Version,
+    val submissionId: String,
+    val unprocessedData: UnprocessedDataContent<GeneticSequence>,
 ) : AccessionVersionInterface
 
 enum class Status {
--- a/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
@@ -24,6 +24,7 @@ import org.loculus.backend.api.DataUseTerms
 import org.loculus.backend.api.DataUseTermsType
 import org.loculus.backend.api.EditedSequenceEntryData
 import org.loculus.backend.api.ExternalSubmittedData
+import org.loculus.backend.api.GetOriginalDataRequest
 import org.loculus.backend.api.GetSequenceResponse
 import org.loculus.backend.api.Organism
 import org.loculus.backend.api.ProcessedData
@@ -41,6 +42,10 @@ import org.loculus.backend.controller.LoculusCustomHeaders.X_TOTAL_RECORDS
 import org.loculus.backend.log.ORGANISM_MDC_KEY
 import org.loculus.backend.log.REQUEST_ID_MDC_KEY
 import org.loculus.backend.log.RequestIdContext
+import org.loculus.backend.model.ACCESSION_HEADER
+import org.loculus.backend.model.FASTA_IDS_HEADER
+import org.loculus.backend.model.FASTA_IDS_SEPARATOR
+import org.loculus.backend.model.METADATA_ID_HEADER
 import org.loculus.backend.model.RELEASED_DATA_RELATED_TABLES
 import org.loculus.backend.model.ReleasedDataModel
 import org.loculus.backend.model.SubmissionParams
@@ -49,7 +54,12 @@ import org.loculus.backend.service.datauseterms.DataUseTermsPreconditionValidato
 import org.loculus.backend.service.groupmanagement.GroupManagementPreconditionValidator
 import org.loculus.backend.service.submission.SubmissionDatabaseService
 import org.loculus.backend.utils.Accession
+import org.loculus.backend.utils.FastaEntry
+import org.loculus.backend.utils.FastaWriter
+import org.loculus.backend.utils.GetOriginalDataHelpers
 import org.loculus.backend.utils.IteratorStreamer
+import org.loculus.backend.utils.TsvWriter
+import org.loculus.backend.utils.makeUniqueIds
 import org.slf4j.MDC
 import org.springframework.http.HttpHeaders
 import org.springframework.http.HttpStatus
@@ -69,10 +79,16 @@ import org.springframework.web.bind.annotation.ResponseStatus
 import org.springframework.web.bind.annotation.RestController
 import org.springframework.web.multipart.MultipartFile
 import org.springframework.web.servlet.mvc.method.annotation.StreamingResponseBody
+import java.time.OffsetDateTime
+import java.time.ZoneOffset
+import java.time.format.DateTimeFormatter
 import java.util.UUID
 import io.swagger.v3.oas.annotations.parameters.RequestBody as SwaggerRequestBody
 
 private val log = KotlinLogging.logger { }
+private val ORIGINAL_DATA_DOWNLOAD_TIMESTAMP_FORMAT = DateTimeFormatter.ofPattern("yyyyMMdd'T'HHmmss'Z'")
+
+const val MAX_ORIGINAL_DATA_DOWNLOAD_ENTRIES = 500
 
 @RestController
 @RequestMapping("/{organism}")
@@ -456,6 +472,112 @@ open class SubmissionController(
         }
 
         return ResponseEntity(streamBody, headers, HttpStatus.OK)
+    }
+
+    @Operation(
+        description = "Download original data (metadata and sequences) as a zip file, suitable for revisions. " +
+            "It is limited to $MAX_ORIGINAL_DATA_DOWNLOAD_ENTRIES entries.",
+    )
+    @ResponseStatus(HttpStatus.OK)
+    @PostMapping(
+        "/get-original-data",
+        consumes = [MediaType.APPLICATION_JSON_VALUE],
+        produces = ["application/zip"],
+    )
+    fun getOriginalData(
+        @PathVariable @Valid organism: Organism,
+        @HiddenParam authenticatedUser: AuthenticatedUser,
+        @RequestBody body: GetOriginalDataRequest,
+    ): ResponseEntity<StreamingResponseBody> {
+        groupManagementPreconditionValidator.validateUserIsAllowedToModifyGroup(body.groupId, authenticatedUser)
+
+        val entryCount = transaction {
+            submissionDatabaseService.countOriginalDataDownloadEntries(
+                organism,
+                body.groupId,
+                body.accessionsFilter,
+            )
+        }
+
+        if (entryCount > MAX_ORIGINAL_DATA_DOWNLOAD_ENTRIES) {
+            throw UnprocessableEntityException(
+                "Download is limited to $MAX_ORIGINAL_DATA_DOWNLOAD_ENTRIES entries. " +
+                    "Requested download would include $entryCount entries. " +
+                    "Please filter fewer sequences.",
+            )
+        }
+
+        val headers = HttpHeaders()
+        headers.contentType = MediaType.parseMediaType("application/zip")
+        headers.set(
+            HttpHeaders.CONTENT_DISPOSITION,
+            "attachment; filename=\"${originalDataDownloadFilename(organism)}\"",
+        )
+
+        val instanceConfig = backendConfig.getInstanceConfig(organism)
+        val hasConsensusSequences = instanceConfig.schema.submissionDataTypes.consensusSequences
+        val isMultiSegmented = instanceConfig.referenceGenome.nucleotideSequences.size > 1
+
+        val streamBody = StreamingResponseBody { responseBodyStream ->
+            val startTime = System.currentTimeMillis()
+            MDC.put(REQUEST_ID_MDC_KEY, requestIdContext.requestId)
+            MDC.put(ORGANISM_MDC_KEY, organism.name)
+
+            try {
+                java.util.zip.ZipOutputStream(responseBodyStream).use { zipOut ->
+                    transaction {
+                        val data = submissionDatabaseService.streamUnprocessedDataForOriginalDataDownload(
+                            organism,
+                            body.groupId,
+                            body.accessionsFilter,
+                        ).toList()
+
+                        val metadataIds = makeUniqueIds(data.map { it.submissionId })
+                        val uniqueFastaIdsByEntry =
+                            GetOriginalDataHelpers.uniqueFastaIdsByEntry(data, isMultiSegmented)
+
+                        zipOut.putNextEntry(java.util.zip.ZipEntry("metadata.tsv"))
+                        GetOriginalDataHelpers.writeMetadataTsv(
+                            data,
+                            metadataIds,
+                            uniqueFastaIdsByEntry,
+                            zipOut,
+                            isMultiSegmented,
+                        )
+                        zipOut.closeEntry()
+
+                        if (hasConsensusSequences) {
+                            zipOut.putNextEntry(java.util.zip.ZipEntry("sequences.fasta"))
+                            GetOriginalDataHelpers.writeSequencesFasta(
+                                data,
+                                metadataIds,
+                                uniqueFastaIdsByEntry,
+                                zipOut,
+                                isMultiSegmented,
+                            )
+                            zipOut.closeEntry()
+                        }
+                    }
+                }
+            } catch (e: Exception) {
+                val duration = System.currentTimeMillis() - startTime
+                log.error(e) { "[get-original-data] Error after ${duration}ms: $e" }
+                throw e
+            } finally {
+                MDC.remove(REQUEST_ID_MDC_KEY)
+                MDC.remove(ORGANISM_MDC_KEY)
+            }
+
+            val duration = System.currentTimeMillis() - startTime
+            log.info { "[get-original-data] Completed in ${duration}ms" }
+        }
+
+        return ResponseEntity(streamBody, headers, HttpStatus.OK)
+    }
+
+    private fun originalDataDownloadFilename(organism: Organism): String {
+        val timestamp = OffsetDateTime.now(ZoneOffset.UTC).format(ORIGINAL_DATA_DOWNLOAD_TIMESTAMP_FORMAT)
+        return "${organism.name}_original_data_$timestamp.zip"
     }
 
     @Operation(description = APPROVE_PROCESSED_DATA_DESCRIPTION)
--- a/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
@@ -532,6 +532,9 @@ open class SubmissionController(
                             body.accessionsFilter,
                         ).toList()
 
+                        // metadataIds: the unique metadata ids in the same order as the original submission ids.
+                        // uniqueFastaIdsByEntry: per entry (in the same order), a map from the original FASTA id to
+                        // the unique FASTA id used in the download.
                         val metadataIds = makeUniqueIds(data.map { it.submissionId })
                         val uniqueFastaIdsByEntry =
                             GetOriginalDataHelpers.uniqueFastaIdsByEntry(data, isMultiSegmented)
--- a/backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
@@ -14,6 +14,8 @@ import org.jetbrains.exposed.sql.LongColumnType
 import org.jetbrains.exposed.sql.Op
 import org.jetbrains.exposed.sql.QueryParameter
 import org.jetbrains.exposed.sql.SortOrder
+import org.jetbrains.exposed.sql.SqlExpressionBuilder.eq
+import org.jetbrains.exposed.sql.SqlExpressionBuilder.inList
 import org.jetbrains.exposed.sql.SqlExpressionBuilder.less
 import org.jetbrains.exposed.sql.SqlExpressionBuilder.plus
 import org.jetbrains.exposed.sql.Transaction
@@ -54,7 +56,6 @@ import org.loculus.backend.api.GeneticSequence
 import org.loculus.backend.api.GetSequenceResponse
 import org.loculus.backend.api.Organism
 import org.loculus.backend.api.OriginalData
-import org.loculus.backend.api.OriginalDataWithFileUrls
 import org.loculus.backend.api.PreprocessingStatus.IN_PROCESSING
 import org.loculus.backend.api.PreprocessingStatus.PROCESSED
 import org.loculus.backend.api.ProcessedData
@@ -69,6 +70,8 @@ import org.loculus.backend.api.Status.APPROVED_FOR_RELEASE
 import org.loculus.backend.api.SubmissionIdMapping
 import org.loculus.backend.api.SubmittedProcessedData
 import org.loculus.backend.api.UnprocessedData
+import org.loculus.backend.api.UnprocessedDataContentWithFileUrls
+import org.loculus.backend.api.UnprocessedDataDownloadEntry
 import org.loculus.backend.api.fileIds
 import org.loculus.backend.api.getFileId
 import org.loculus.backend.auth.AuthenticatedUser
@@ -194,13 +197,13 @@ class SubmissionDatabaseService(
             .chunked(streamBatchSize)
             .map { chunk ->
                 val chunkOfUnprocessedData = chunk.map {
-                    val originalData = compressionService.decompressSequencesInOriginalData(
+                    val unprocessedData = compressionService.decompressSequencesInOriginalData(
                         it[table.unprocessedDataColumn]!!,
                     )
-                    val originalDataWithFileUrls = OriginalDataWithFileUrls(
-                        originalData.metadata,
-                        originalData.unalignedNucleotideSequences,
-                        originalData.files?.let {
+                    val unprocessedDataWithFileUrls = UnprocessedDataContentWithFileUrls(
+                        unprocessedData.metadata,
+                        unprocessedData.unalignedNucleotideSequences,
+                        unprocessedData.files?.let {
                             it.mapValues {
                                 it.value.map { f ->
                                     val presignedUrl = s3Service.createUrlToReadPrivateFile(f.fileId)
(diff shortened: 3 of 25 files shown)


backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
```diff
@@ -451,12 +459,145 @@ open class SubmissionController(
...
+        "/get-original-data",
+        consumes = [MediaType.APPLICATION_JSON_VALUE],
+        produces = ["application/zip"],
+    )
+    fun getOriginalData(
+        @PathVariable @Valid organism: Organism,
+        @HiddenParam authenticatedUser: AuthenticatedUser,
+        @RequestBody body: GetOriginalDataRequest,
+    ): ResponseEntity<StreamingResponseBody> {
+        val entryCount = transaction {
```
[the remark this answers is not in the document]
anna-parker: ah sorry I thought we included entryCount in the header - but that doesnt seem to the case so its fine to have this in a separate transaction

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -1218,6 +1230,7 @@ class SubmissionDatabaseService(
         groupIdsFilter: List<Int>?,
         statusesFilter: List<Status>?,
         fields: List<String>?,
+        accessionVersionsFilter: List<AccessionVersion>?,
```
anna-parker: again this change can probably be reverted as streamOriginalMetadata is not used by the new functionality but streamOriginalData... however, these functions could potentially be merged to limit the amount of duplciation

website/src/services/lapisClientSideApi.ts
```diff
@@ -0,0 +1,19 @@
+import axios from 'axios';
+
+import type { DetailsResponse } from '../types/lapis.ts';
+
+/**
+ * Client-side LAPIS API functions.
+ * Use these for imperative LAPIS calls in React components (e.g., button click handlers).
+ * For reactive data fetching, use the hooks from serviceHooks.ts instead.
+ */
+export async function fetchDetailsFromLapis(
```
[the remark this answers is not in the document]
theosanderson: (sorry that I didn't reply to this but `lapidClientHooks` != `lapisClient.ts` I think?)

backend/src/main/kotlin/org/loculus/backend/utils/TsvWriter.kt
```diff
@@ -0,0 +1,21 @@
...
+        .setHeader(*headers.toTypedArray())
+        .get()
+        .print(writer)
+
+    fun writeRow(values: List<String?>) {
+        printer.printRecord(values)
+    }
+
+    override fun close() {
+        printer.flush()
```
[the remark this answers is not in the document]
chaoran-chen: Leaving this as flush-only because this writer is used inside ZipOutputStream entries; closing here would close the whole zip before later entries. The caller owns the stream. [This was unintentionally written by Codex which was (not anymore!) apparently connected to my GitHub account and decided to comment when asked to address the issues..]

backend/src/main/kotlin/org/loculus/backend/utils/FastaWriter.kt
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    fun writeAll(entries: Iterable<FastaEntry>) {
+        for (entry in entries) {
+            write(entry)
+        }
+    }
+
+    override fun close() {
+        writer.flush()
```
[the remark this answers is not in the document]
chaoran-chen: Same here: closing the wrapped writer would close the ZipOutputStream while the controller still needs to add entries. Flushing is intentional because the caller owns the stream. [This was unintentionally written by Codex which was (not anymore!) apparently connected to my GitHub account and decided to comment when asked to address the issues..]

website/src/services/lapisClientSideApi.ts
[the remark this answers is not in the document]
chaoran-chen: `lapisClient.ts` pulls in server-side config/logging and is not browser-safe. This helper is for an imperative client-side click handler; the hooks still cover reactive fetches. [This was unintentionally written by Codex which was (not anymore!) apparently connected to my GitHub account and decided to comment when asked to address the issues..]

backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
```diff
@@ -328,6 +328,24 @@ data class AccessionVersionOriginalMetadata(
     val originalMetadata: Map<String, String?>?,
 ) : AccessionVersionInterface
 
+data class GetOriginalDataRequest(
+    @Schema(
+        description = "The group ID to download data for.",
+        required = true,
+    )
+    val groupId: Int,
```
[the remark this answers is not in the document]
chaoran-chen: This has been removed.

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
```diff
@@ -451,12 +459,145 @@ open class SubmissionController(
...
+
+            try {
+                java.util.zip.ZipOutputStream(responseBodyStream).use { zipOut ->
+                    transaction {
+                        val data = submissionDatabaseService.streamOriginalData(
+                            authenticatedUser,
+                            organism,
+                            body.groupId,
+                            body.accessionsFilter,
+                        ).toList()
```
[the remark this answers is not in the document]
chaoran-chen: That's enough for now, streaming shall be implemented separately in the future.

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -1258,6 +1272,67 @@ class SubmissionDatabaseService(
...
+            accessionsCondition
+    }
+
+    fun countOriginalData(
+        authenticatedUser: AuthenticatedUser,
+        organism: Organism,
+        groupId: Int,
+        accessionsFilter: List<String>?,
+    ): Long {
+        groupManagementPreconditionValidator.validateUserIsAllowedToModifyGroup(groupId, authenticatedUser)
```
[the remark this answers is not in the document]
chaoran-chen: Done - it's now an explicit function to validate it.

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
```diff
@@ -451,12 +459,145 @@ open class SubmissionController(
...
+        val metadataKeys = data.flatMap { it.originalData.metadata.keys }.toSet().sorted()
+        val headers = if (isMultiSegmented) {
+            listOf("id", "accession", "fastaIds") + metadataKeys
+        } else {
+            listOf("id", "accession") + metadataKeys
+        }
+
+        TsvWriter(outputStream, headers).use { writer ->
+            for (entry in data) {
+                val id = "${entry.accession}.${entry.version}"
```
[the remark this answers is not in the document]
chaoran-chen: This has been resolved: it is now using the original id. In cases of duplications, it will add a suffix `_{i}` which `i` being an increasing integer starting with 1 that is not in the dataset.

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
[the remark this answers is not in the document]
chaoran-chen: Follow-up issue:  #6544

website/src/components/SearchPage/DownloadDialog/DownloadOriginalDataButton.tsx
```diff
@@ -0,0 +1,128 @@
...
+            const timestamp = new Date().toISOString().replace(/[:.]/g, '-').slice(0, 19);
+            const filename = `${organism}_original_data_${timestamp}.zip`;
+            downloadBlob(result.blob, filename);
+        } catch (err) {
+            const message = err instanceof Error ? err.message : 'Download failed';
+            setError(message);
+        } finally {
+            setIsDownloading(false);
+        }
+    }, [sequenceFilter, backendUrl, accessToken, organism, groupId, sequenceCount, fetchAccessions]);
```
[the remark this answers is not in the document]
chaoran-chen: Done - the block is now wrapped in a try-catch.

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
chaoran-chen: These changes have been reverted.

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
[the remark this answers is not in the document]
chaoran-chen: These changes have been reverted.

website/src/components/SearchPage/DownloadDialog/DownloadOriginalDataButton.tsx
```diff
@@ -0,0 +1,128 @@
...
+            if (sequenceCount !== undefined) {
+                const apiParams = sequenceFilter.toApiParams();
+                const selectedVersions = apiParams.accessionVersion;
+                if (!Array.isArray(selectedVersions)) {
+                    throw new Error('No accessions selected');
+                }
+                accessionVersions = selectedVersions;
+            } else {
+                accessionVersions = await fetchAccessions();
+                if (accessionVersions.length > MAX_DOWNLOAD_ENTRIES) {
```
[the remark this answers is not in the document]
chaoran-chen: This is now checked consistently.

backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
```diff
@@ -328,6 +328,24 @@ data class AccessionVersionUnprocessedMetadata(
...
+        description = "Filter by specific accessions. If not provided, all accessions for the group are returned.",
+    )
+    val accessionsFilter: List<Accession>? = null,
+)
+
+data class OriginalDataResponse(
+    override val accession: Accession,
+    override val version: Version,
+    val submissionId: String,
+    val originalData: OriginalData<GeneticSequence>,
```
anna-parker: This PR was probably initially made before we merged https://github.com/loculus-project/loculus/pull/6254/changes. Users need the unprocessed not the original data (if we had to perform a db surgery on the data e.g. to fix author format after an entry was submitted so that the submission didnt error we dont want users to submit the original data before the db surgery as it will now error)

anna-parker: I guess `OriginalData` and `UnprocessedData` use the same type, but maybe we could still make this clearer - sorry this weirdness was more introduced by me than by you!

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -1304,6 +1307,73 @@ class SubmissionDatabaseService(
...
+                    (SequenceEntriesView.versionColumn eq SequenceEntriesTable.versionColumn)
+            },
+        )
+
+        return originalDataQuery
+            .select(
+                SequenceEntriesView.accessionColumn,
+                SequenceEntriesView.versionColumn,
+                SequenceEntriesView.submissionIdColumn,
+                SequenceEntriesTable.originalDataColumn,
```
anna-parker: here we need to select the unprocessedDataColumn (see my comment above)

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
```diff
@@ -457,6 +470,107 @@ open class SubmissionController(
...
+        "/get-original-data",
+        consumes = [MediaType.APPLICATION_JSON_VALUE],
+        produces = ["application/zip"],
+    )
+    fun getOriginalData(
+        @PathVariable @Valid organism: Organism,
+        @HiddenParam authenticatedUser: AuthenticatedUser,
+        @RequestBody body: GetOriginalDataRequest,
+    ): ResponseEntity<StreamingResponseBody> {
+        submissionDatabaseService.validateOriginalDataDownloadAllowed(authenticatedUser, body.groupId)
```
anna-parker: as this function is just the following validation I would perhaps just inline the function:
```suggestion
        groupManagementPreconditionValidator.validateUserIsAllowedToModifyGroup(groupId, authenticatedUser)
```

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
```diff
@@ -457,6 +470,107 @@ open class SubmissionController(
...
+                    "Requested download would include $entryCount entries. " +
+                    "Please filter fewer sequences.",
+            )
+        }
+
+        val headers = HttpHeaders()
+        headers.contentType = MediaType.parseMediaType("application/zip")
+        headers.set(
+            HttpHeaders.CONTENT_DISPOSITION,
+            "attachment; filename=\"${organism.name}_original_data.zip\"",
```
anna-parker: maybe we could add a timestamp to the filename - Ive found it really helpful to have with our other downloads

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
```diff
@@ -457,6 +470,107 @@ open class SubmissionController(
...
+        )
+
+        val instanceConfig = backendConfig.getInstanceConfig(organism)
+        val hasConsensusSequences = instanceConfig.schema.submissionDataTypes.consensusSequences
+        val isMultiSegmented = instanceConfig.referenceGenome.nucleotideSequences.size > 1
+
+        val streamBody = StreamingResponseBody { responseBodyStream ->
+            val startTime = System.currentTimeMillis()
+            MDC.put(REQUEST_ID_MDC_KEY, requestIdContext.requestId)
```
anna-parker, line 523: nice!

backend/src/main/kotlin/org/loculus/backend/utils/UniqueIds.kt
```diff
@@ -0,0 +1,25 @@
+package org.loculus.backend.utils
+
+fun makeUniqueIds(ids: List<String>): List<String> {
```
anna-parker, line 3: Not blocking: but as usedIds are mutated as a side effect I asked chatGPT if there is a more standard way to do this and it suggested:
```
fun makeUniqueIds(ids: List<String>): List<String> {
    val originalIds = ids.toSet()
    val usedIds = mutableSetOf<String>()

    fun nextUniqueId(id: String): String =
        generateSequence(1) { it + 1 }
            .map { suffix -> "${id}_$suffix" }
            .first { it !in originalIds && it !in usedIds }

    return ids.map { id ->
        val uniqueId = if (id in usedIds) nextUniqueId(id) else id
        usedIds += uniqueId
        uniqueId
    }
}
```

anna-parker: a thought is also if it would be nicer to not add a suffix if the submissionId is duplicated but use the accession as a submissionId then

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
```diff
@@ -457,6 +470,107 @@ open class SubmissionController(
...
+            try {
+                java.util.zip.ZipOutputStream(responseBodyStream).use { zipOut ->
+                    transaction {
+                        val data = submissionDatabaseService.streamOriginalData(
+                            organism,
+                            body.groupId,
+                            body.accessionsFilter,
+                        ).toList()
+
+                        val metadataIds = makeUniqueIds(data.map { it.submissionId })
```
anna-parker, line 538: reading through this I wonder if we should just output the accession.version and not try to return the submissionIds as they are not unique. Creating the unique submissionIds for the metadata and the fastaIds starts getting really tedious.

anna-parker: another option proposed by @corneliusroemer is to throw if the submissionId is not unique

backend/src/main/kotlin/org/loculus/backend/utils/GetOriginalDataHelpers.kt
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    fun writeMetadataTsv(
+        data: List<OriginalDataResponse>,
+        metadataIds: List<String>,
+        fastaIdsByEntry: List<UniqueFastaIdsForEntry>,
+        outputStream: java.io.OutputStream,
+        isMultiSegmented: Boolean,
+    ) {
+        val metadataKeys = data.flatMapTo(mutableSetOf()) { it.originalData.metadata.keys }.sorted()
```
anna-parker: the fileIds arent output into the metadata file I believe? and we dont export them in the zip folder separately either

website/src/components/SearchPage/SearchFullUI.tsx
```diff
@@ -205,6 +214,15 @@ export const InnerSearchFullUI = ({
     const totalSequences = aggregatedHook.data?.data[0].count ?? undefined;
     const linkOutSequenceCount = downloadFilter.sequenceCount() ?? totalSequences;
 
+    const fetchAccessions = useCallback(async (): Promise<string[]> => {
```
anna-parker, line 218: I think we can add this to the hooks like we do for the `aggregatedHook` above

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
chaoran-chen: I switched from accession version to submission ids just yesterday due to the request of @theosanderson in https://github.com/loculus-project/loculus/pull/5856#discussion_r2792849671 😅 Throwing seems problematic to me: no where did we define that submission ids are supposed to be unique across submissions, so making the feature unusable for people who reuse submission ids doesn't make sense to me.

backend/src/main/kotlin/org/loculus/backend/utils/GetOriginalDataHelpers.kt
chaoran-chen: Indeed, this is missing and I can imagine implementing it together with https://github.com/loculus-project/loculus/issues/5821#issuecomment-3731045121

> we provide a zip download which has the whole directory structure to upload in order to put in the existing sequences -- except that instead of each file actually containing the contents it just contains the text md5:da23ef or whatever, and that's used as a pointer to the existing file.

(We don't want to create a zip file with all the real potentially large files, but otherwise, there is currently no way to revise files from the website, so there's not much what we can do now without implementing #5821 first)

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
anna-parker: I find it odd that the submissionIds can be changed (i.e. have a suffix added) - to me it feels like we should return consistent output, I think @theosanderson suggested to throw an error actually if the submissionIds are not unique

backend/src/main/kotlin/org/loculus/backend/utils/GetOriginalDataHelpers.kt
anna-parker: ah apologies I misread the description!

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
chaoran-chen: I think that it is desirable that users can easily download all their submitted data (currently potentially limited due to missing streaming but that's not a fundamental restriction) and if we throw, having one duplication would prevent it entirely. If we really think that duplicated submission IDs are undesireable, it would be better to actually prevent it at submission. For this feature, if we think that changing submission IDs is particularly problematic, I could imagine showing a warning or including a file in the zip that shows a map of renamings (but rather in a future PR than now).

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
anna-parker: @chaoran-chen this is important to change

chaoran-chen: Thanks for pointing this out, this is indeed very important! I fixed it in 4db619b0dcb7f95145e8b51c4328e80d872cb044

backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
chaoran-chen: Thanks for pointing this out! It's fixed in 4db619b0dcb7f95145e8b51c4328e80d872cb044

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
chaoran-chen: Good idea! It's added in 3bdecec3bd988321d407a5555670edf703b43630

backend/src/main/kotlin/org/loculus/backend/utils/UniqueIds.kt
chaoran-chen: The suggested code is a bit shorter but I personally find it less easy to read but rather prefer a pain while loop, so I'd prefer keeping it as is.

chaoran-chen: For the suffix idea, I created #6624.

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
chaoran-chen: Done: e4db73e52485d5aa8bff6f4a516b2374d5775b36

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
chaoran-chen: Thanks for the discussions on Slack. I have created the following tickets based on it: #6623, #6624, #6625 and #6626 and suggest addressing them in future PRs.

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
anna-parker: thanks!

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
anna-parker: sounds good! thanks! I will resolve my comment :-)

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
```diff
@@ -69,10 +79,16 @@ import org.springframework.web.bind.annotation.ResponseStatus
...
 import org.springframework.web.multipart.MultipartFile
 import org.springframework.web.servlet.mvc.method.annotation.StreamingResponseBody
+import java.time.OffsetDateTime
+import java.time.ZoneOffset
+import java.time.format.DateTimeFormatter
 import java.util.UUID
 import io.swagger.v3.oas.annotations.parameters.RequestBody as SwaggerRequestBody
 
 private val log = KotlinLogging.logger { }
+private val ORIGINAL_DATA_DOWNLOAD_TIMESTAMP_FORMAT = DateTimeFormatter.ofPattern("yyyyMMdd'T'HHmmss'Z'")
```
anna-parker, line 89: thank you!

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
```diff
@@ -69,10 +79,16 @@ import org.springframework.web.bind.annotation.ResponseStatus
...
+
+const val MAX_ORIGINAL_DATA_DOWNLOAD_ENTRIES = 500
```
anna-parker, line 91: does this max also hold for mpox or do we need to allow it to be configurable per organism?

backend/src/main/kotlin/org/loculus/backend/utils/GetOriginalDataHelpers.kt
```diff
@@ -0,0 +1,94 @@
...
+import org.loculus.backend.model.FASTA_IDS_SEPARATOR
+import org.loculus.backend.model.FastaId
+import org.loculus.backend.model.METADATA_ID_HEADER
+import kotlin.collections.component1
+import kotlin.collections.component2
+import kotlin.collections.iterator
+import kotlin.collections.orEmpty
+
+/**
+ * In the Map, each entry is a segment.
```
anna-parker:
```suggestion
 * Map from original FastaId to unique FastaId for each segment in an entry.
```

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
```diff
@@ -457,6 +473,112 @@ open class SubmissionController(
...
+            try {
+                java.util.zip.ZipOutputStream(responseBodyStream).use { zipOut ->
+                    transaction {
+                        val data = submissionDatabaseService.streamUnprocessedDataForOriginalDataDownload(
+                            organism,
+                            body.groupId,
+                            body.accessionsFilter,
+                        ).toList()
+
+                        val metadataIds = makeUniqueIds(data.map { it.submissionId })
```
anna-parker, line 538: I would maybe add a comment here that this is a list of unique metadataIds in the same order as the original metadataIds, and a list of maps from old fastaIds to new unique fastaIds in the same order as the metadataIds.

backend/src/test/kotlin/org/loculus/backend/controller/submission/GetOriginalDataEndpointTest.kt
```diff
@@ -0,0 +1,519 @@
...
+        assertThat(header[0], `is`("id"))
+        assertThat(header[1], `is`("accession"))
+
+        val firstDataRow = lines[1].split("\t")
+        assertThat(firstDataRow[0], `is`(SubmitFiles.DefaultFiles.submissionIds[0]))
+        assertThat(firstDataRow[1], `is`(accessionVersions[0].accession))
+    }
+
+    @Test
+    fun `GIVEN data exists THEN FASTA headers match metadata ids`() {
```
anna-parker: this is only true for single segmented organisms

backend/src/test/kotlin/org/loculus/backend/controller/submission/GetOriginalDataEndpointTest.kt
```diff
@@ -0,0 +1,519 @@
...
+
+        val fastaHeaderIds = sequencesFasta.lines()
+            .filter { it.startsWith(">") }
+            .map { it.removePrefix(">") }
+            .toSet()
+
+        assertThat(fastaHeaderIds, `is`(allFastaIdsFromMetadata))
+    }
+
+    private fun extractZipContents(zipContent: ByteArray): Pair<String, String> {
```
anna-parker: this function appears to be duplicated in backend/src/test/kotlin/org/loculus/backend/controller/submission/OriginalDataRevisionWorkflowTest.kt

integration-tests/tests/specs/features/download-original-data.spec.ts
```diff
@@ -0,0 +1,177 @@
...
+            expect(fs.existsSync(sequencesPath)).toBe(true);
+
+            const metadataContent = fs.readFileSync(metadataPath, 'utf8');
+            const metadataLines = metadataContent
+                .split('\n')
+                .filter((line) => line.trim().length > 0);
+
+            expect(metadataLines.length).toBe(3);
+
+            const header = metadataLines[0].split('\t');
```
anna-parker, line 79: maybe we want to confirm the metadata is there but not blocking

integration-tests/tests/specs/features/download-original-data.spec.ts
```diff
@@ -0,0 +1,177 @@
...
+        const searchPage = new SearchPage(page);
+        await searchPage.goToReleasedSequences(TEST_ORGANISM, groupId);
+        const accessionVersions = await searchPage.waitForSequencesInSearch(
+            3,
+            SEARCH_INDEXING_TIMEOUT,
+        );
+
+        expect(accessionVersions).toHaveLength(3);
+
+        const rows = searchPage.getSequenceRows();
```
anna-parker, line 143: Im confused why we dont select the sequence created above with the submissionId `select-download-${timestamp}-${i}`

anna-parker (approved):
Looks good! I tested out the full submission, download of original data and revision procedure on this preview and it works well! I also confirmed it works when the submissionId and fastaIds are duplicated.

I just wonder if it would make sense to merge https://github.com/loculus-project/loculus/pull/6629 into this before merging this into main - but we could also do one after another.

kubernetes/loculus/values.yaml
chaoran-chen, line 1: Revert!

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
chaoran-chen: I added mpox now to the preview here, let's see!

backend/src/test/kotlin/org/loculus/backend/controller/submission/GetOriginalDataEndpointTest.kt
chaoran-chen: Added a new helper function that's called by both tests

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
chaoran-chen: I just downloaded >400 mpox sequences. It's slow but worked.

integration-tests/tests/specs/features/download-original-data.spec.ts
chaoran-chen: As far as I can see, the submisison ID is not displayed by default on the Released sequences page.

anna-parker: ah... I guess I just wondered if this test ends up actually selecting the submitted sequences or older ones, but I guess doesnt matter too much! I will resolve :-)

chaoran-chen:
Thansk very much for the review, @anna-parker! I'm merging this now. I think it's good to merge #6629 separately so that the commits keep the feature changes apart from the refactorings.

PR #5860, opened 2026-01-21
fix(website): fix selection bug in Table when vertical shift

### LINKED ISSUE #5849: Selecting sequences on search/released sequences page is buggy if there is a vertical shift
If there is a vertical shift when selecting sequences (which happens in certain combinations of screen width and number of sequences), the drag&select-multipe-sequences feature is causing an unintended behavior that leads none or more than one sequence being selected.

https://github.com/user-attachments/assets/3cc5755a-d37d-4505-840c-dcae55b294c0

(cc @theosanderson as you probably find this interesting ;))

chaoran-chen:
resolves #5849

This fixes the selection problem when there is a vertical shift in the Table by only activating the drag-selection if the cursor is actually being moved vertically.

### Screenshot

https://github.com/user-attachments/assets/1c902b3e-7005-4325-bc96-a62c79ab0705

### PR Checklist
- ~[ ] The implemented feature is covered by appropriate, automated tests.~
    - See video

### DIFF
--- a/website/src/components/SearchPage/Table.tsx
+++ b/website/src/components/SearchPage/Table.tsx
@@ -125,7 +125,6 @@ export const Table: FC<TableProps> = ({
 
     const mouseDownSelection = useRef('');
     const dragSelecting = useRef({ active: false, value: false, startY: 0 });
-    const checkboxMouseDownFired = useRef(false);
 
     useEffect(() => {
         const handleMouseUp = () => {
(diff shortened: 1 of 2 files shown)


website/src/components/SearchPage/Table.tsx
theosanderson: Thanks a lot for identifying and fixing the bug. What would you think about a slight simplification in https://github.com/loculus-project/loculus/pull/5865 - just handling everything with the td. My initial implementation always seemed a bit random in allowing dragging over the td, but not clicking and IMO a bit more real estate to click for selection would be good (and the code is a bit simpler). But it's nbd either way, so I'll approve this now.

Issue #3027, opened 2024-10-21
Re-order panels on sequence-details page 


emmahodcroft:
At the moment our panels ("Submission details"; "Data use terms") are not in a bad order, but perhaps they could do with some re-ordering based on what is most valuable/important to those looking at the page. For example, for many the "Sample details" (sample date, location, some other info) is probably more immediately of interest than the "Alignment state and QC metrics"

This is a discussion issue as a new order is not decided, but it would be nice to talk about this, then we can re-order.

Example of current:
![image](https://github.com/user-attachments/assets/13bd6b16-3bbe-413f-8b8e-b6eb4dcc5a30)

theosanderson:
IMO it would be nice to get an INSDC accession / links into probably the top left section (alongside other stuff)

emmahodcroft:
We should also re-think 'submission related' fields - perhaps exchanging some of these for things like 'accession' number place higher

PR #5967, opened 2026-02-12
feat!(website, config): add option to order headers in sequence details page

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/3027

### Screenshot
Adds a way to specify the order of headers on the sequence details and improves the ordering of fields in the search bar by using the orderOnDetails page if available.

I added this option to the private-virus preview but thought I would split it out for easier review :-)

### PR Checklist
Confirmed on preview new header order is applied and that this is also true for CCHF
image

🚀 Preview: https://improve-detailspage.loculus.org

### DIFF
--- a/website/src/components/SequenceDetailsPage/getDataTableData.spec.ts
+++ b/website/src/components/SequenceDetailsPage/getDataTableData.spec.ts
@@ -29,16 +29,8 @@ describe('getDataTableData', () => {
         ).toBeUndefined();
     });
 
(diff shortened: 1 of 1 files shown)


kubernetes/loculus/templates/_common-metadata.tpl
theosanderson: My recollection is that I already implemented this such that the order is determined by the order of the items themselves (e.g. the average order of all items in a group) - I think we already have the ability to set the order, and I'm not sure we want extra bloat in the config..

PR #5968, opened 2026-02-12
feat(website): use standard autocomplete field for reference selector

theosanderson:
This brings in the counts of each reference

🚀 Preview: https://autocomplete.loculus.org

### DIFF
--- a/website/src/components/SearchPage/SearchForm.spec.tsx
+++ b/website/src/components/SearchPage/SearchForm.spec.tsx
@@ -167,7 +167,6 @@ describe('SearchForm', () => {
                     referenceSelection={{
                         referenceIdentifierField: 'My genotype',
                         selectedReferences: {},
-                        setSelectedReferences: vi.fn(),
                     }}
                 />
             </QueryClientProvider>,
@@ -233,7 +232,6 @@ describe('SearchForm', () => {
                     referenceIdentifierField: 'My genotype',
 
                     selectedReferences: { main: 'suborganism1' },
-                    setSelectedReferences: vi.fn(),
                 },
             });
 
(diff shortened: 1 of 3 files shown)


website/src/components/SearchPage/SearchForm.tsx
```diff
@@ -191,8 +191,10 @@ export const SearchForm = ({
                                 filterSchema={filterSchema}
                                 referenceGenomesInfo={referenceGenomesInfo}
                                 referenceIdentifierField={referenceSelection.referenceIdentifierField}
-                                selectedReferences={referenceSelection.selectedReferences}
-                                setSelectedReferences={referenceSelection.setSelectedReferences}
```
anna-parker: ah I see! what you had in mind... but I think we need to either restructure the referenceSelection or have setSomeFieldValues also set the new referenceSelection for this to have the desired effect

website/src/components/SearchPage/ReferenceSelector.tsx
```diff
@@ -30,85 +28,61 @@ export const ReferenceSelector: FC<ReferenceSelectorProps> = ({
...
+                                fieldName,
+                            }}
+                            setSomeFieldValues={setSomeFieldValues}
+                            fieldValue={(fieldValues[fieldName] as string | undefined) ?? ''}
+                        />
 
                         <p className='text-xs text-gray-600 mt-2'>
-                            Select a {formatLabel(labelsBySegment[segment] ?? '')} to enable mutation search and
-                            download of aligned sequences
+                            Select a {formatLabel(label)} to enable mutation search and download of aligned sequences
```
anna-parker: I was going to suggest we can hide this after selection but without the selectedReferences param it isnt that easy I guess we can do that in a follow up PR :-)

Issue #4275, opened 2025-06-02
Make use of new SILO preprocessing "append" mode (incremental updates)


corneliusroemer:
SILO recently added incremental preprocessing update functionality: https://github.com/GenSpectrum/LAPIS-SILO/pull/720

We should see whether we can take advantage of this feature to speed up preprocessing, so that when a new mpox sequence is released, it takes less than 10min for it to show up (current time on PPX).

Would be cool if @Taepper could guide us here a little :)

PR #5969, opened 2026-02-12
feat(deployment, backend, silo-importer): use incremental update option for silo imports

theosanderson:
resolves https://github.com/loculus-project/loculus/issues/4275

Incremental Ingest Docs: https://github.com/GenSpectrum/LAPIS-SILO/commit/6875b5ed7b6de8565c15239fd0d7dfde1b82f1c2#diff-e2a8d31441dd77b633a6e25a93f6e89136521a6b6b73a64578c92c2e68740b10R14

### PR Checklist
- [x] Any manual testing that has been done is documented (i.e. what exactly was tested?) -> confirmed preview comes up and incremental upload is working according to the logs

### DIFF
--- a/loculus-silo/src/silo_import/runner.py
+++ b/loculus-silo/src/silo_import/runner.py
@@ -1,9 +1,9 @@
 from __future__ import annotations
 
 import logging
+import re
 import shutil
 import time
-from datetime import UTC, datetime
 
 from .config import ImporterConfig
 from .constants import SPECIAL_ETAG_NONE
@@ -53,9 +53,17 @@ class ImporterRunner:
             return True
         return time.time() - self.last_hard_refresh >= self.config.hard_refresh_interval
 
-    def _current_timestamp_iso(self) -> str:
-        """Return the current UTC time as an ISO-8601 string."""
-        return datetime.now(tz=UTC).strftime("%Y-%m-%dT%H:%M:%S")
+    @staticmethod
+    def _parse_etag_timestamp(etag: str) -> str:
+        """Extract the ISO-8601 timestamp from a backend ETag.
+
+        The backend ETag format is ``"<timestamp>"`` (with double quotes),
+        where ``<timestamp>`` is a UTC ISO-8601 string such as
+        ``2024-01-15T10:30:00.123456``.  This method strips the surrounding
+        quotes and weak-validator prefix (W/) so the value can be passed
+        directly as a ``releasedSince`` parameter.
+        """
+        return re.sub(r'^(?:W/)?"(.*)"$', r"\1", etag)
 
     def _update_lineage_if_needed(
         self, pipeline_version: int | None, hard_refresh: bool = False
@@ -144,7 +152,7 @@ class ImporterRunner:
         # Mark success and update state
         self.current_etag = download.etag
         self.has_existing_silo_db = True
-        self.last_successful_import_time = self._current_timestamp_iso()
+        self.last_successful_import_time = self._parse_etag_timestamp(download.etag)
 
         if hard_refresh:
             self.last_hard_refresh = time.time()
@@ -202,7 +210,7 @@ class ImporterRunner:
 
         # Mark success and update state
         self.current_etag = download.etag
-        self.last_successful_import_time = self._current_timestamp_iso()
+        self.last_successful_import_time = self._parse_etag_timestamp(download.etag)
 
         logger.info(
             "Incremental append complete (%d records); waiting %s seconds",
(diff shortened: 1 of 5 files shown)


theosanderson:
@theosanderson-agent investigate backend test failures

theosanderson-agent:
**Backend test failure investigation:**

The backend test failures were a **CI flake**, not caused by the code changes in this PR.

**Root cause:** All 377 test failures stemmed from a single root cause: `Failed to determine suitable jdbc url`. The Spring Boot test context failed to load because the Testcontainers PostgreSQL container did not start properly in the CI environment. Once the first context load failed, all subsequent tests failed with `ApplicationContext failure threshold (1) exceeded`.

**Evidence:**
- The code changes (adding `releasedSince` parameter to `countReleasedSubmissions` and `streamReleasedSubmissions`) compile cleanly and are type-safe
- Tests pass when run locally
- Main branch backend tests have been passing consistently (no failures around the same timeframe)
- I re-ran the failed CI job and it **passed successfully**

No code changes needed.

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
```diff
@@ -312,6 +312,10 @@ open class SubmissionController(
         @Parameter(
             description = "(Optional) Only retrieve all released data if Etag has changed.",
         ) @RequestHeader(value = HttpHeaders.IF_NONE_MATCH, required = false) ifNoneMatch: String?,
+        @Parameter(
+            description = "(Optional) Only retrieve data released after this " +
```
anna-parker: lets add strictly after to make this more clear

theosanderson-agent:
Fixed the CI failures. The 4 new incremental tests (`test_runner_incremental_append_after_initial_full`, `test_runner_append_fallback_to_full_on_failure`, `test_runner_hard_refresh_forces_full_preprocessing`, `test_runner_incremental_skips_when_not_modified`) were passing `lineage_definitions={1: "http://lineage"}` (`dict[int, str]`) instead of the correct `{"test": {1: "http://lineage"}}` (`dict[str, dict[int, str]]`). This caused both the mypy type errors and the runtime `AttributeError: 'str' object has no attribute 'get'` in `update_lineage_definitions()`. Fixed in 75fc53f.

loculus-silo/src/silo_import/runner.py
```diff
@@ -100,11 +127,65 @@ def run_once(self) -> None:
...
+        try:
+            download = self.download_manager.download_release(
+                self.config,
+                self.paths,
+                last_etag,
+                released_since=self.last_successful_import_time,
+            )
+        except (NotModifiedError, HashUnchangedError) as skip:
+            logger.info("Skipping incremental append: %s", skip)
+            if skip.new_etag is not None:
```
anna-parker: wait... should we update an etag is the download fails? I dont think so - that means we need to retry with that etag I think

anna-parker:
@theosanderson, @corneliusroemer and @chaoran-chen does this sound reasonable?

For incremental updates to work we need to be able to compute the additional sequences since the last time we downloaded all sequences. In order to do this we need to ask the backend for all new sequences since a specific time. Defining this time is hard.

We have a couple contraints.
1. The backend streams get-released-data, this doesnt lock the database so if sequences are added to the database while streaming they will be added to the stream. To partially resolve this issue we add a record count to the header that is calculated at the start of the stream and delete requests where there are more or less records than specified in the header.
2. SILO cannot handle uploads where a sequence already exists so we cannot resend data to SILO.

In more detail:

silo-import sends request (1) > backend starts stream (2) > backend finished stream (3) > silo-importer finishes request (4)

- If the silo-importer stores the time (1), we can run into the issue that additional sequences were added between time 1 and 4 and when we rerequest sequences with after time 1 we will end up with duplicates and SILO will fail/die
- If the silo-importer stores the time (4), we run into the issue that if the backend received any sequences between time 3 and 4 these sequences will be ignored until the next hard refresh.


Therefore, I think the most clean approach would be to have the backend send time2 in the header along with the record count. As we ignore requests where the record count in the header doesnt align with the record count at the start, this will be the true time since the last change in the backend.

loculus-silo/src/silo_import/runner.py
```diff
@@ -56,7 +109,7 @@ def run_once(self) -> None:
             )
         else:
             logger.info(
-                f"Soft refresh: time_since_last={time.time() - self.last_hard_refresh:.1f}s, "
+                f"Full preprocessing: time_since_last={time.time() - self.last_hard_refresh:.1f}s, "
```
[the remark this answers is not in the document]
theosanderson: and when would we use each of those?

PR #5983, opened 2026-02-14
[POC] website: structure sequence details page, suppress 0 values in metadata

rneher:
Very naive PR that structures the sequence details page and unclutters  the alignment metrics

current [main](https://main.loculus.org/cchf-multi-ref/search?selectedSeq=LOC_001H47E.1):
image

this PR:
image

### DIFF
--- a/website/src/components/SequenceDetailsPage/DataTable.tsx
+++ b/website/src/components/SequenceDetailsPage/DataTable.tsx
@@ -58,7 +58,7 @@ const DataTableComponent: React.FC<Props> = ({
     const mutationSections = dataTableData.table.filter(({ header }) =>
         header.toLowerCase().includes('mutation')
     );
-
+    console.log("DataTable", dataTableData)
     return (
         <div>
             {dataTableData.topmatter.sequenceDisplayName !== undefined && (
--- a/website/src/components/SequenceDetailsPage/getDataTableData.ts
+++ b/website/src/components/SequenceDetailsPage/getDataTableData.ts
@@ -115,6 +115,10 @@ export function getDataTableData(listTableDataEntries: TableDataEntry[]): DataTa
         let combinedRows = combineAlignmentLengthAndCompleteness(rows);
         // Combine host fields
         combinedRows = combineHostFields(combinedRows);
+        // Suppress collection date bounds if equal
+        combinedRows = suppressEqualCollectionDateBounds(combinedRows);
+        // Combine geo-location fields
+        combinedRows = combineGeoLocationFields(combinedRows);
 
         const definedOrders = combinedRows.map((r) => r.orderOnDetailsPage).filter((o): o is number => o !== undefined);
         const meanOrder =
@@ -187,7 +191,7 @@ function combineHostFields(rows: TableDataEntry[]): TableDataEntry[] {
 
         if (isHostTaxonId) {
             const hostTaxonId = currentRow.value.toString();
-            
+
             // Look for hostNameScientific and hostNameCommon entries
             const hostNameScientificIndex = rows.findIndex(
                 (row, idx) =>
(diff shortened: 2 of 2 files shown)


website/src/components/SequenceDetailsPage/DataTable.tsx
```diff
@@ -45,43 +45,111 @@ const DataTableComponent: React.FC<Props> = ({
...
     return (
         <div>
             {dataTableData.topmatter.sequenceDisplayName !== undefined && (
-                <div className='pr-6 mb-4 italic'>Display Name: {dataTableData.topmatter.sequenceDisplayName}</div>
+                <div className='pr-6 mb-4 italic'>Isolate: {dataTableData.topmatter.sequenceDisplayName}</div>
             )}
             {dataTableData.topmatter.authors !== undefined && dataTableData.topmatter.authors.length > 0 && (
                 <div className='pr-6 mb-4'>
                     <AuthorList authors={dataTableData.topmatter.authors} />
+                    {authorSection.flatMap(({ rows }) => rows).map((entry: TableDataEntry, index: number) => (
```
anna-parker: I would probably still show the "key" or name of these fields, i.e. make it clear that this grey field is author affiliations

website/src/components/SequenceDetailsPage/DataTable.tsx
```diff
@@ -45,43 +45,111 @@ const DataTableComponent: React.FC<Props> = ({
...
+            )}
+
+            {/* Divider before Alignment section */}
+            {alignmentSections.length > 0 && <hr className='my-8 border-t-2 border-gray-200' />}
+
+            {/* Alignment Section Header */}
+            {alignmentSections.length > 0 && (
+                <div>
+                <h2 className='text-xl font-bold mb-2'>Alignment and QC
+                {hasReferenceAccession && (
```
anna-parker: (This is a comment on existing code) I think we can possibly just delete this ReferenceSequenceButton and use the same link as below:
```
hasReferenceAccession && (
                                <h2 className='pt-2 text-xs text-gray-500'>
                                    <AkarInfo className='inline-block h-4 w-4 mr-1 -mt-0.5' />
                                    Mutations called relative to the <ReferenceDisplay reference={references} /> reference
                                    {references.length > 1 ? 's' : ''}
                                </h2>
                            )}
```

website/src/components/SequenceDetailsPage/getDataTableData.ts
```diff
@@ -110,16 +110,246 @@ export function getDataTableData(listTableDataEntries: TableDataEntry[]): DataTa
...
+            result.push(currentRow);
+        }
+
+        processedIndices.add(i);
+    }
+
+    return result;
+}
+
+function suppressEqualCollectionDateBounds(rows: TableDataEntry[]): TableDataEntry[] {
```
anna-parker: we can actually hide the collection date bounds on the sequence details page by adding `hideOnSequenceDetailsPage: true` to the `sampleCollectionDateRangeLower` and `sampleCollectionDateRangeUpper` - which means less custom code

website/src/components/SequenceDetailsPage/DataTableEntry.tsx
```diff
@@ -11,10 +11,9 @@ interface Props {
 
 const DataTableComponent: React.FC<Props> = ({ data, dataUseTermsHistory }) => {
     const { label, type } = data;
-
     return (
         <>
-            {type.kind === 'metadata' && (
+            {type.kind === 'metadata' && data.value !== undefined && data.value !== 0 && (
```
anna-parker: I wonder if we should only filter out 0s in the alignment and QC details - just wondering if we might accidentally filter out some important metadata field with this filter

website/src/components/SequenceDetailsPage/getDataTableData.ts
```diff
@@ -110,16 +110,246 @@ export function getDataTableData(listTableDataEntries: TableDataEntry[]): DataTa
             (a, b) =>
                 (a.orderOnDetailsPage ?? Number.POSITIVE_INFINITY) - (b.orderOnDetailsPage ?? Number.POSITIVE_INFINITY),
         );
-        const definedOrders = rows.map((r) => r.orderOnDetailsPage).filter((o): o is number => o !== undefined);
+
+        // Combine length and completeness entries
+        let combinedRows = combineAlignmentLengthAndCompleteness(rows);
```
anna-parker: These three functions depend on the metadata field names and would probably be the biggest change from the current structure - @theosanderson what are your thoughts?

theosanderson:
Thank you - this became https://github.com/loculus-project/loculus/pull/5989

PR #5985, opened 2026-02-16
feat(website): use orderOnDetailsPage in search, metadata download, column selector and search field selector

anna-parker:
resolves #

### PR Checklist

### DIFF
--- a/website/src/components/SequenceDetailsPage/getDataTableData.ts
+++ b/website/src/components/SequenceDetailsPage/getDataTableData.ts
@@ -66,9 +66,7 @@ export type HeaderGroup<T> = {
     meanOrder: number;
 };
 
-export function buildHeaderGroups<T extends HasOrder>(
-    source: Map<string, T[]>,
-): HeaderGroup<T>[] {
+export function buildHeaderGroups<T extends HasOrder>(source: Map<string, T[]>): HeaderGroup<T>[] {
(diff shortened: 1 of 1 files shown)


kubernetes/loculus/values.schema.json
theosanderson: Hey so the order in the downloads on PPX (at least for the default feeds) is already in an intended order that we specifically chose, which IMO is better than the one in this preview.

A) I'm not sure these things should always all be in the same order
B) I think to consider any proposals on this we should probably look at a Pathoplexus preview

theosanderson: > The header groups themselves are ordered alphabetically in the selector. Whereas, on the detailsPage we order the headers based on the average order of the fields in each header. I am not a fan of this alphabetical ordering as I think e.g. Sample details should be at the top of the selector.

Yep, this suggestion sounds fine for me - I agree the selector isn't great atm in various ways!

> Selected search fields are currently in the order they arise in the config which I find quite ugly. For example admin regions are before country:

I think this order is something we probably need to be able to specify specifically for the search fields. Atm we can do that (by controlling the order). It may not be perfect now - I agree Country should be above, but that's just because we've put it in the wrong order. I'd be very open to us adding a `orderInSearchDisplay`. I'd also be open to that inheriting from `order` if unset. I don't disagree that the order will often want to be similar, but I think we want to maintain the ability to determine it for search.

Issue #5270, opened 2025-10-20
Seqset organism listed as "N/A" for unversioned accessions


corneliusroemer:
1. Add unversioned accession to seqset
2. Observe that the seqset page shows organism as N/A

<img width="1037" height="165" alt="Image" src="https://github.com/user-attachments/assets/0b7f6229-0508-4fc0-b750-db32a022d900" />

PR #6003, opened 2026-02-18
fix(website): Support bare accessions in SeqSet metadata queries

theosanderson:
resolves https://github.com/loculus-project/loculus/issues/5270

This PR enhances the SeqSet metadata fetching logic to handle both versioned accessions (e.g., `NC_045512.2`) and bare accessions (e.g., `NC_045512`) by querying them separately against the LAPIS API.

**Key Changes:**
- Split accessions into two groups: versioned (containing `.`) and bare (without `.`)
- Query versioned accessions using the `accessionVersion` parameter
- Query bare accessions using the `accession` parameter with `versionStatus: latestVersion` filter to retrieve the latest version
- Refactored result handling to use a `MetadataEntry` type that tracks both the record data and the appropriate lookup key
- Updated metadata map creation to use the correct accession form as the key (versioned or bare)

This allows the component to correctly resolve metadata for both types of accession formats that may be present in SeqSet records.

image

🚀 Preview: https://claude-fix-seqset-bare-ac.loculus.org

### DIFF
--- a/.github/workflows/format-website-on-dispatch.yml
+++ b/.github/workflows/format-website-on-dispatch.yml
@@ -36,7 +36,6 @@ jobs:
         run: |
           git config --global user.name 'Loculus bot'
           git config --global user.email 'bot@loculus.org'
-          git checkout -- website/package-lock.json
           git add -A
           git commit -m "Automated code formatting" || echo "No changes to commit"
           git push origin HEAD:${{ github.head_ref }}


website/src/components/SeqSetCitations/SeqSetRecordsTableWithMetadata.tsx
```diff
@@ -45,47 +68,51 @@ const fetchRecordsMetadata = async (
...
+            );
+        }
 
-        metadataMap.set(String(recordData.accessionVersion), metadata);
+        const results = await Promise.all(queries);
+        for (const { data, keyField } of results) {
+            for (const record of data) {
+                const key = String(record[keyField]);
+                const metadata: RecordMetadata = {
+                    accession: key,
```
anna-parker: hmm... the keyField is sometimes `accession` and sometimes `accessionVersion` - wont this create a dictionary where the key can be an accession or an accessionVersion? do we want this?

PR #6007, opened 2026-02-19
fix: isolate keyring file per test to fix CLI test flake #5424

### LINKED ISSUE #5424: Flaky CLI integration test: apparently logged out despite having logged in shortly before
Need more error output from CLI to be able to debug better. Also looking at keycloak logs could help. Most likely something like timeout.

```
  1) [cli-tests] › tests/specs/cli/release.spec.ts:5:12 › CLI Release Command › should handle sequence release workflow 

    Error: Failed to submit sequence 1/2

    Exit code: 1

    STDOUT:
    ✗ Not logged in

    STDERR:
    Error: Please log in first using 'loculus auth login'
```

https://github.com/loculus-project/loculus/actions/runs/19300499266/job/55196619378?pr=5409#step:22:41

theosanderson-agent:
## Summary

Fixes #5424 — the flaky CLI integration test where tests intermittently fail with "Not authenticated. Please run 'loculus auth login' first."

### Root cause

The `PlaintextKeyring` backend (used in CI via `keyrings.alt`) stores **all** keyring data in a single shared file: `~/.local/share/python_keyring/keyring_pass.cfg`. Even though each test instance uses a unique keyring **service name** (`loculus-cli-test-<uuid>`), these are just different INI sections in the **same physical file**.

The `PlaintextKeyring._write_config_value()` method performs a non-atomic **read → modify → write** cycle with no file locking:

1. Reads entire file into memory
2. Modifies in-memory config
3. Truncates and rewrites entire file

When parallel Playwright workers (4 by default with `fullyParallel: true`) run CLI tests simultaneously, concurrent writes to this file cause a classic race condition — one test's write clobbers another test's token entry. The affected test then calls `keyring.get_password()`, finds its token missing, and fails with "Not authenticated."

The ~5% failure rate matches the narrow timing window required for the race to occur.

### Fix

Give each test instance its own `XDG_DATA_HOME` directory (UUID-based), so the `PlaintextKeyring` writes to a completely isolated file per test. This eliminates the shared-file contention.

Changes to `integration-tests/tests/pages/CliPage.ts`:
- Add a per-test `dataHome` temp directory
- Pass `XDG_DATA_HOME` in the CLI command environment
- Clean up the directory after each test

### Validation

**11 consecutive CI runs on the fix, all green, zero flakes.**

| Run | Result |
|-----|--------|
| 22196044432 | ✅ |
| 22196045471 | ✅ |
| 22196046483 | ✅ |
| 22196047671 | ✅ |
| 22196048715 | ✅ |
| 22196049777 | ✅ |
| 22196051086 | ✅ |
| 22196052130 | ✅ |
| 22196053084 | ✅ |
| 22196054063 | ✅ |
| 22195510170 | ✅ |

## Test plan
- [x] 10+ consecutive CI runs with no flakes

via [Happy](https://happy.engineering)

### DIFF
--- a/integration-tests/tests/pages/CliPage.ts
+++ b/integration-tests/tests/pages/CliPage.ts
@@ -1,6 +1,6 @@
 import { exec } from 'child_process';
 import { promisify, stripVTControlCharacters } from 'util';
-import { writeFile, unlink } from 'fs/promises';
+import { writeFile, unlink, rm } from 'fs/promises';
 import { join } from 'path';
 import { tmpdir } from 'os';
 import { test } from '@playwright/test';
@@ -21,6 +21,7 @@ export class CliPage {
     private baseUrl: string;
     private keyringService: string;
     private configFile: string;
+    private dataHome: string;
 
     constructor() {
         const uuid = randomUUID();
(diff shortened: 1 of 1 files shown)


anna-parker (approved):
not an expert but this seems to make sense - thanks!

PR #6025, opened 2026-02-25
fix(website): convert null field values to empty strings in GET URL p…

theosanderson:
…arams

When building GET request URLs (e.g. LAPIS download URLs), null field values were being converted to the literal string "null" via String(null) instead of empty strings. LAPIS interprets empty string params as "field is null" for GET requests, whereas the string "null" would search for the literal text "null".

The fix converts null values to empty strings in
FieldFilterSet.toUrlSearchParams() and updates DownloadUrlGenerator to allow empty string values through (previously filtered by truthiness check).

https://claude.ai/code/session_017kbxfHYZZzyhUvVgEfyEeG

resolves #

### PR Checklist

🚀 Preview: https://claude-fix-null-get-param.loculus.org

### DIFF
--- a/website/src/components/SearchPage/DownloadDialog/SequenceFilters.tsx
+++ b/website/src/components/SearchPage/DownloadDialog/SequenceFilters.tsx
@@ -172,8 +172,10 @@ export class FieldFilterSet implements SequenceFilter {
 
             if (Array.isArray(value)) {
                 if (value.length > 0) {
-                    result.push([key, value]);
+                    result.push([key, value.map((v: any) => (v === null ? '' : v))]);
                 }
+            } else if (value === null) {
+                result.push([key, '']);
             } else {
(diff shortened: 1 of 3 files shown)


website/src/components/SearchPage/DownloadDialog/SequenceFilters.tsx
```diff
@@ -172,8 +172,10 @@ export class FieldFilterSet implements SequenceFilter {
 
             if (Array.isArray(value)) {
                 if (value.length > 0) {
-                    result.push([key, value]);
+                    result.push([key, value.map((v: any) => (v === null ? '' : v))]);
```
anna-parker: just to check in this case we drop empty strings" "" and convert null to "". This works for the current field is `(blank)` in the search filter, but should we stil keep the empty strings?

PR #6068, opened 2026-03-02
docs: add funding section

chaoran-chen:
This adds a funding section to the landing page. Please feel free to directly add additional funding sources!

### Screenshot

image

### DIFF
--- a/docs/src/content/docs/index.mdx
+++ b/docs/src/content/docs/index.mdx
@@ -62,6 +62,20 @@ import members from '../../team.json';
   small={true}
 />
 
+## Funding
+
+This project has received funding from NIH<span class="align-super text-xs">1</span> through the [Pathogen Data Network](https://pathogendatanetwork.org/) and [swissuniversities](https://www.swissuniversities.ch/) through the Programme Open Science.
+
+<p>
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/index.mdx
```diff
@@ -62,6 +62,20 @@ import members from '../../team.json';
   small={true}
 />
 
+## Funding
+
+This project has received funding from NIH<span class="align-super text-xs">\*)</span> through the [Pathogen Data Network](https://pathogendatanetwork.org/) and [swissuniversities](https://www.swissuniversities.ch/) through the Programme Open Science.
```
anna-parker: not blocking could we use `*1` instead of `*)`?

anna-parker: true `*` is nice - but Im also ok with `1`

emmahodcroft:
So is plan to have 'layman' wording at top and 'full official paragraph of the funder' in a footnote?

chaoran-chen:
@emmahodcroft, so far, it's just based on what SIB suggested and it seems like a good start. When we get a longer list of funders, we can reorganize.

PR #6075, opened 2026-03-04
feat(actions): added YAML linting action using `yamlfmt`, and linted existing YAML files

tombch:
This PR adds an action `yamlfmt-lint` which checks for correctly formatted YAML, for all YAML files excluding the helm templates and Github actions. 

The linting is done using [yamlfmt](https://github.com/google/yamlfmt), and a couple of settings have been added in the `.yamlfmt.yml`:

- `trim_trailing_whitespace: true`: trims trailing whitespace, and was required to preserve multiline script/run tags in GitHub actions (mentioned here: https://github.com/google/yamlfmt/issues/185).
- `retain_line_breaks_single: true`: any instance of multiple line breaks is squashed down into a single line break.

Any suggestions for this are very welcome! Some caveats I have noticed with the formatting is that it does not preserve any indenting/aligning within comments (mentioned here: https://github.com/google/yamlfmt/issues/290) 

~For the Github actions, it forces any run/script tags to use `|-` rather than `|`. Possibly the workflows shouldn't be included in linting?~ (Github actions are now excluded from linting)

### PR Checklist

### DIFF
--- a/.github/workflows/yamlfmt-lint.yml
+++ b/.github/workflows/yamlfmt-lint.yml
@@ -10,7 +10,7 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     container:
-      image: ghcr.io/google/yamlfmt:latest
+      image: ghcr.io/google/yamlfmt:0.21.0
     steps:
       - name: Checkout repository
         uses: actions/checkout@v6


anna-parker:
Hi @tombch this is amazing! if the github actions are hard to do you can start with excluding them and we can add them in a later PR - thanks for finding this and working on it!

PR #6102, opened 2026-03-09
feat(prepro): Improving warning message when displayName regex fails

maverbiest:
Adding a more informative warning message when parsing an identifier field during displayName creation for directly submitted sequences doesn't work.

Old message:
```python
"identifier string could not be parsed using provided regex_pattern"
```

New message:
```python
f"identifier string '{identifier}' could not be parsed, using ACCESSION_VERSION in displayName instead"
```

### PR Checklist

### DIFF
--- a/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
@@ -1201,7 +1201,6 @@ class ProcessingFunctions:
         collector_id = input_data.get("specimenCollectorSampleId", None)
         submission_id = input_data.get("submissionId", None)
         warnings: list[ProcessingAnnotation] = []
-        errors: list[ProcessingAnnotation] = []
         if submission_id is None:
             return ProcessingResult(
                 datum=None,
@@ -1306,7 +1305,6 @@ class ProcessingFunctions:
                             ),
                         )
                     )
-                    errors = errors + extract_result.errors
                 identifier = extract_result.datum
 
         if identifier is None:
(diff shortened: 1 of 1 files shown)


preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -1293,19 +1294,20 @@ def replace_identifier(values, replacement):
...
+                if extract_result.datum is None:
                     # regex extraction of ID field failed, fall back to ACCESSION_VERSION
                     warnings.append(
                         ProcessingAnnotation.from_fields(
                             input_fields,
                             [output_field],
                             AnnotationSourceType.METADATA,
                             message=(
-                                "identifier string could not be parsed using provided regex_pattern"
+                                f"identifier string '{identifier}' could not be parsed, using ACCESSION_VERSION in displayName instead"
```
anna-parker: this looks good!

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -1293,19 +1293,19 @@ def replace_identifier(values, replacement):
                     input_fields=[],
                     args={"pattern": regex_pattern, "capture_group": "identifier"},
                 )
-                identifier = extract_result.datum
-                if identifier is None:
+                if extract_result.datum is None:
```
anna-parker: Note on a pre-existing issue in the code: its a bit confusing that we update the `identifier` variable, so that it corresponds to different things at different parts of the code. Ideally, we would call it input_identifer or preparsing_identifier and afterwards only identifier

Issue #6100, opened 2026-03-05
Preprocessing: support metadata produced from generated metadata


anna-parker:
At the moment preprocessing generates each metadata field independently from the input metadata submitted by the user or obtained from the nextclade run. However, it is not configured to use generated metadata to create new metadata fields. 

For example the `concatenate` takes country, date and accession as input and joins them to create a displayName field. The processed country field is  processed by `process_options` to validate it is a valid country. We do not perform this processing in `concatenate` and `concatenate` thus accepts invalid countries. The only way to add this validation would be to create a new function with the necessary input of both the `concatenate` and `process_options`, this is what we had to do with the `build_display_name` and my new `assign_custom_lineage` function. 

Should we alternatively restructure the pipeline to allow metadata fields to be created from processed metadata fields? i.e. allow `concatenate` to take the processed country field as input instead of the non-processed input field. This raises a question of how many layers of fields can/shoudl we allow. Given input metadata field I1 and I2, we now support output field A produced from I1 and I2. If we allow B produced from A and I2 we might still want to create C from B and I1...

maverbiest:
Thanks for opening the discussion! This is something I was also thinking about while working on prepro, however it's not something I have experience with. 

I guess a naive approach would be to have some sort of priority system where functions run in sequential groups in an order configured in the values.yaml. But I don't have a clear idea of how to not make this super error-prone when configuring it.

As far as I'm aware, a way similar problems are solved by software like Nextflow, Snakemake etc. is to create a [DAG](https://en.wikipedia.org/wiki/Directed_acyclic_graph) based on dependencies between different entities and use this to determine the order in which things need to run. For our usecase, each metadata field would declare which other metadata fields it depends on, which we'd use to build a graph. I can imagine there are libraries with existing implementations for this but it's not something I'm too familiar with. Would be happy to investigate, though :)

PR #6137, opened 2026-03-14
feat(prepro): Ordered processing of metadata fields

maverbiest:
resolves #https://github.com/loculus-project/loculus/issues/6100

Currently, metadata fields are processed in the order they are declared in `values.yaml`. This change adds support for determining a processing order at runtime that enables generating metadata fields from other generated metadata fields.

As described in https://github.com/loculus-project/loculus/issues/6100, this enables functions such as `concatenate` to use values that have already been validated by another processing function.

Another practical example is for the upcoming hostname validation functionality. With the new implementation, we can validate `hostNameScientific` and map it to a `taxonId`. Then, we can use that `taxonId` to derive additional fields such as `hostNameCommon`. In the existing state, we would need separate functions that each repeat the same hostname validation before producing different outputs.

## Summary:
- Processing order is determined by looking for metadata `inputs` that are prefixed with a `METADATA_DEPENDENCY_PREFIX` (currently set to `processed.`). If a metadata field has such inputs, this field is processed only after all its dependencies have been processed.
- Dependencies have to be explicitly declared in the config via this prefix: existing `values.yaml` files will be processed exactly as they were before and do not need to be updated
- I configured a dependency in the `values.yaml` for west-nile (in `build_display_name`), so that we can observe the behaviour on previews.

## Manual testing
As noted above, I configured a dependency for west-nile. On the preview of this PR, the same number of west-nile sequences (as well as other organisms) come up as on main's preview. I also did some spot checks on west-nile sequences in the preview to check that the processed field is incorportated in the displayName, and not the unprocessed one. 

### PR Checklist

🚀 Preview: https://ordered-prepro.loculus.org

### DIFF
--- a/preprocessing/nextclade/tests/test_metadata_processing_functions.py
+++ b/preprocessing/nextclade/tests/test_metadata_processing_functions.py
@@ -761,6 +761,13 @@ def test_preprocessing_metadata_dependencies(test_case_def: Case, config_depende
     processed_entry = process_single_entry(test_case, config_dependency)
     verify_processed_entry(processed_entry, test_case.expected_output, test_case.name)
 
+    wrong_order = tuple(
+        ["depends_on_A"] + [i for i in config_dependency.processing_order if i != "depends_on_A"]
+    )
+    config_dependency.processing_order = wrong_order
+    processed_entry = process_single_entry(test_case, config_dependency)
+    assert processed_entry.data.metadata != test_case.expected_output.data.metadata
+
 
 def test_preprocessing_without_consensus_sequences(config: Config) -> None:
     sequence_name = "entry without sequences"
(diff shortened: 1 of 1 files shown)


preprocessing/nextclade/src/loculus_preprocessing/config.py
```diff
@@ -225,6 +227,23 @@ def generate_argparse_from_model(config_cls: type[BaseModel]) -> argparse.Argume
...
+    """Return a valid order for processing metadata fields based on their dependencies.
+
+    Dependencies are derived from input fields in `config.processing_spec`. A DAG is
+    constructed and topologically sorted to ensure each field is processed after the
+    fields it depends on.
+    """
+    dag: dict[str, set[str]] = {k: set() for k in config.processing_spec.keys()}
+    for node, spec in config.processing_spec.items():
+        for dependency in spec.inputs.values():
+            if dependency == node or dependency not in dag:
```
anna-parker: this makes an assumption that an output and input field only have the same name if we use the identity function, which Im not sure is always true - we should probably check this

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -349,10 +350,19 @@ def get_output_metadata(
...
             if isinstance(unprocessed, UnprocessedAfterNextclade):
-                input_metadata = add_input_metadata(spec, unprocessed, input_path, config=config)
-                input_data[arg_name] = input_metadata.datum
-                errors.extend(input_metadata.errors)
-                warnings.extend(input_metadata.warnings)
+                # Try to get required intput field from processed metadata values
+                # If it does not exists there, get it through add_input_metadata
+                processed = output_metadata.get(input_path)
+                if processed is not None:
+                    input_data[args_name] = processed  # type: ignore
```
anna-parker:
```suggestion
                    input_data[arg_name] = processed  # type: ignore
```

preprocessing/nextclade/src/loculus_preprocessing/config.py
anna-parker: I made a comment about this for Monday - think it should be fine if we just document this behavior but I want to double check - but overall really cool - think this will be a great help!

Issue #6140, opened 2026-03-16
Bug: joinToString misuse in SubmissionDatabaseService discards accession versions in error messages


theosanderson-agent:
## Bug Description

In `SubmissionDatabaseService.kt`, the `andThatSequenceEntriesHaveNoErrors()` method uses Kotlin's `joinToString` with trailing lambda syntax, which passes the lambda as a **transform** parameter instead of as the separator. This silently replaces every accession version string with the literal `", "`, making the error message useless to users.

## Location

`backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt` — in the `andThatSequenceEntriesHaveNoErrors()` method:

```kotlin
sequenceEntriesWithErrors.map {
    "${it[SequenceEntriesView.accessionColumn]}.${it[SequenceEntriesView.versionColumn]}"
}.joinToString { ", " },
```

## Expected Behavior

Error message should read:
> Accession versions have errors: LOC_00001.1, LOC_00002.1

## Actual Behavior

Error message reads:
> Accession versions have errors: , , , 

Because `joinToString { ", " }` passes `{ ", " }` as the `transform` lambda (each element is replaced with the literal string `", "`), not as the separator argument.

## Fix

Change `joinToString { ", " }` to `joinToString(", ")` (parentheses, not curly braces), or simply `joinToString()` since `", "` is already the default separator.

```kotlin
}.joinToString(", "),
```

PR #6144, opened 2026-03-16
fix: correct joinToString usage in error messages

saschabuehrle:
## Bug
[Issue #6140](https://github.com/loculus-project/loculus/issues/6140) — joinToString misuse in SubmissionDatabaseService discards accession versions in error messages

## Fix
Changed `joinToString { ", " }` to `joinToString(", ")` in AccessionPreconditionValidator. The curly brace syntax passes the lambda as a transform parameter, replacing each element with the literal string ", ", instead of using it as the separator.

## Testing
Error messages will now correctly display accession versions like "LOC_00001.1, LOC_00002.1" instead of just ", , , ".

Happy to address any feedback.

Greetings, saschabuehrle


theosanderson:
Thanks but we will use https://github.com/loculus-project/loculus/pull/6143 due to added tests

Issue #3135, opened 2024-10-31
Refactor and test version comments


theosanderson:
Currently versionComments are implemented in a pretty confusing way for developers:
- versionComments can either be stored as a metadata item called `versionComment` or as a column called `versionComment`
- For revisions the first approach is used
- For revocations the second approach is used

Instead we should just use one approach :)

anna-parker:
Just for some background I did this as a workaround because revocations do not have metadata.

anna-parker:
@theosanderson-agent can you draft a PR where we attempt to solve this difference in implementations with the last suggestion?

Add the revocation versionComment to the originalDataColumn and update the processedDataColumn with the contents of the originalDataColumn automatically for revocations. -> I think this is the best approach I can think of

You should then also write a migration to copy over the contents of the existing versionComment column to these new metadata jsonb columns and delete that column in the db

PR #6198, opened 2026-03-26
refactor(backend): move revocation versionComment from dedicated column to metadata

theosanderson-agent:
## Summary

This PR unifies how `versionComment` is stored for revisions and revocations, addressing #3135.

**Problem:** Previously, `versionComment` was stored differently depending on the type of version entry:
- For **revisions**, it was stored as a metadata field in `originalData`, processed through the preprocessing pipeline, and ended up in `processedData.metadata`
- For **revocations**, it was stored in a dedicated `version_comment` column on the `sequence_entries` table, then manually merged into released data via special-case code in `ReleasedDataModel`

This inconsistency made the codebase harder to understand and maintain.

**Solution:** Following the approach suggested in #3135, this PR:

1. **Stores revocation `versionComment` in `originalData.metadata`** — When creating a revocation with a version comment, the comment is stored in the `originalData` JSONB column as a metadata entry, compressed using the standard compression service

2. **Automatically constructs `processedData` for revocations in the database view** — The `sequence_entries_view` is updated to build `joint_metadata` for revocations directly from `originalData.metadata`. This approach is pipeline-version-independent, meaning the versionComment survives processing pipeline version changes without needing to reprocess revocations

3. **Removes the dedicated `version_comment` column** — A migration (V1.26) copies existing `version_comment` values into `originalData.metadata` for all affected revocations, then drops the column

4. **Removes special-case code** — The `versionComment` field is removed from `RawProcessedData`, and the conditional metadata merging in `ReleasedDataModel` is removed. The versionComment now flows through the same path for both revisions and revocations

### Files changed

- **Migration `V1.26`**: Migrates existing data and recreates the view
- **`SubmissionDatabaseService.kt`**: `revoke()` now stores versionComment in `originalData`; `streamReleasedSubmissions()` no longer reads `versionCommentColumn`; `RawProcessedData` no longer has `versionComment` field
- **`SequenceEntriesTable.kt` / `SequenceEntriesView.kt`**: Remove `versionCommentColumn`
- **`ReleasedDataModel.kt`**: Remove conditional versionComment addition for revocations
- **`EarliestReleaseDateFinderTest.kt`**: Remove `versionComment` field from test data

### Key design decision

Rather than inserting into `sequence_entries_preprocessed_data` for revocations (which would be tied to a specific pipeline version and lost when the pipeline version changes), the view constructs `joint_metadata` directly from `originalData` for revocations. This means the versionComment is always available regardless of pipeline version changes, without needing to reprocess revocations.

## Test plan

- [x] Verify compilation passes (confirmed locally)
- [x] Verify lint passes (confirmed locally)
- [x] Run full backend test suite (requires PostgreSQL infrastructure)
- [x] Verify revocation with versionComment shows correctly in released data on preview
image

- [x] Verify revocation without versionComment still works
image

(currently on main we also see None if this is empty - I think this is probably an issue with the website I assume it submits revocations with an empty string and not None, this isnt too important atm)
- [x] Verify revision versionComment is unaffected
image

- [x] Verify migration correctly moves existing version_comment values

🚀 Preview: https://refactor-version-comment.loculus.org

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
@@ -52,6 +52,7 @@ import org.loculus.backend.api.FileIdAndNameAndReadUrl
 import org.loculus.backend.api.GeneticSequence
 import org.loculus.backend.api.GetSequenceResponse
 import org.loculus.backend.api.Organism
+import org.loculus.backend.api.OriginalData
 import org.loculus.backend.api.OriginalDataWithFileUrls
 import org.loculus.backend.api.PreprocessingStatus.IN_PROCESSING
 import org.loculus.backend.api.PreprocessingStatus.PROCESSED
@@ -743,7 +744,6 @@ class SubmissionDatabaseService(
             SequenceEntriesView.accessionColumn,
             SequenceEntriesView.versionColumn,
             SequenceEntriesView.isRevocationColumn,
-            SequenceEntriesView.versionCommentColumn,
             SequenceEntriesView.jointDataColumn,
             SequenceEntriesView.submitterColumn,
             SequenceEntriesView.groupIdColumn,
@@ -786,7 +786,6 @@ class SubmissionDatabaseService(
                     DataUseTermsType.fromString(it[DataUseTermsTable.dataUseTermsTypeColumn]),
                     it[DataUseTermsTable.restrictedUntilColumn],
                 ),
-                versionComment = it[SequenceEntriesView.versionCommentColumn],
                 dataUseTermsChangeDate = it[DataUseTermsTable.changeDateColumn],
             )
         }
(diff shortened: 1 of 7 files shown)


anna-parker:
Created a new preview off main, set DB to persistent and revoked two sequences with and without versionComments: LOC_00004V5.2 with comment LOC_00002X3.2 without, confirmed that migration suceeded and versionComments are still visible: 
image
https://github.com/loculus-project/loculus/pull/6266

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -992,6 +988,28 @@ class SubmissionDatabaseService(
             ),
         )
 
+        if (versionComment != null) {
```
theosanderson: @theosanderson-agent  - would there be away to avoid this branch with so much duplication?

theosanderson: @theosanderson-agent

theosanderson-agent: Done in 8f6392661. Combined the post-insert UPDATE into the `insertSelect` itself by passing the compressed `originalData` as a `QueryParameter` (used for both `original_data` and `unprocessed_data`), so the if-branch and the second SQL statement are gone. When no version comment is supplied, the metadata is `{}` — consistent with V1.27 step 1b, which already ensures every revocation row has non-null `unprocessed_data`. Verified `RevokeEndpointTest`, `GetReleasedDataEndpointTest`, `GetReleasedDataFileSharingEndpointTest`, `GetUnprocessedMetadataEndpointTest`, `GetSequencesEndpointTest`, and `ApproveProcessedDataEndpointTest` all still pass.

theosanderson: erghh, sorry for pushing to your branch - happy to fix it back to how it was. but what would you think about this?

anna-parker: oh absolutely no worries! this is anyways a theo bot branch :-)

anna-parker: but also otherwise Im normally not too bothered with people pushing to my branches unless Im actively working on it and even then its no biggie :-)

theosanderson: Ah yeah, it probably pushed because it was its own branch :)

anna-parker (approved):
much nicer!

theosanderson:
@codex review

theosanderson (approved):
lgtm - haven't tried it out!

anna-parker:
I quickly tested again just to be sure!

PR #6209, opened 2026-03-30
feat(website): add tooltips for metadata field descriptions to search ui 

tombch:
Adds tooltips for metadata field definitions in search, column and download field selector modals. This could be quite a convenient place (imo) to provide additional information on processed metadata fields but open to alternate suggestions!

### Screenshot

image

### PR Checklist


anna-parker (commented):
heyhey - think this is overall a great improvement, but I think we dont need a new `rangeDefinitionField` for this (I might be wrong though)

Update: ah! it would also be cool to test this on a PPX preview to see if we need to add description to some internal metadata fields still :-)

Issue #3262, opened 2024-11-20
Create new fiel


theosanderson:

corneliusroemer:
This would work well with a set of migration scripts that can be applied to original data to deterministically reshape it to whatever we want the preprocessing pipeline to see. That would be a less hacky way: rather than doing one-off db surgery, apply scripts to original data:

- Script 1 for changing author format: Overwrite these author fields with those values
- Hypothetical script 2 for renaming a field: rename all field keys `geoLocCity` -> `city`
- ...

We could contain complexity as opposed to have it infect the processing pipeline which could theoretically accept all previous versions based on timestamps of submission etc, but would become unwieldy.

PR #6254, opened 2026-04-09
feat!(backend): add column in backend for raw original data that should never be modified

anna-parker:
https://github.com/loculus-project/loculus/issues/3262

Adds a new field to the database called `unprocessed_data`. It is for now populated with a copy of the `original_data` jsonb column. `unprocessed_data` will be used instead of `original_data` for calls to preprocessing/requests for original data. `original_data` on the other hand will not be used and serves only as a backup/storage of the initial metadata we recieved. This will simplify the reprocessing process as we can migrate `unprocessed_data` in a db migration which is not stored in the code but still keep a record of the data originally submitted to us. 

The biggest code change is the subsequent field renaming that I did to make the distrinction between original and unprocessed data clearer in the code.

## Breaking change
The `get-original-data` endpoint is now called `get-unprocessed-data` endpoint.

### PR Checklist
Preview comes up, revision (which uses affected endpoint) via the webpage is WAI: LOC_00001V8.2

🚀 Preview: https://add-new-column.loculus.org


backend/src/main/kotlin/org/loculus/backend/service/submission/UploadDatabaseService.kt
```diff
@@ -218,6 +219,12 @@ class UploadDatabaseService(
                 m.submitter,
                 m.group_id,
                 m.uploaded_at,
+                jsonb_build_object(
```
anna-parker: hmm I guess it is a bit ugly to build this twice

maverbiest (approved):
Looks good!

Connected to the DB via port forwarding. Checked that this query returns all rows:
```sql
SELECT * FROM public.sequence_entries
WHERE original_data = unprocessed_data;
```

And that this query returns no rows:
```sql
SELECT * FROM public.sequence_entries
WHERE original_data != unprocessed_data;
```

I also did spot checks on the preview web page and data seems to be coming up fine, with all the data as expected in the sequence details view I checked.

corneliusroemer:
I just came across this when analyzing the database size, I was surprised we were storing the same data twice. Then found this PR via git blame.

Do we use the original data at all? Is it just there for the historical record to be viewed by administrators?

Might be useful to add some numbers on the impact of duplicating: Ingesting all organisms on main (no revisions, so prod is larger) `original_data` is 326MB, out of a total of 842MB for `sequence_entries` table and total db size 1731MB. So this PR increased `sequence_entries` by ~63% and overall db by ~23%.

theosanderson:
> Is it just there for the historical record to be viewed by administrators

Indeed – it lowers the stakes of db surgery / migrations

I don't think DB size is too much of an issue – and in terms of performance these will be toasts anyway

anna-parker:
@corneliusroemer db size is basically negligable atm we pay like 1dollar a month for it and 2k for db compute

PR #6255, opened 2026-04-10
feat!(backend, website): add a new backend endpoint to get organism for an accession

### LINKED ISSUE #6265: Accession search and details pages must call each LAPIS instance to figure out which organism an accession belongs to
I thought we had an issue for this but couldnt find it! This was noticed recently by @tombch when creating https://github.com/loculus-project/loculus/issues/4840

Calling each LAPIS instance to determine which organism an accession belongs to doesnt scale well the more organisms we add. Additionally, this information is easily available from the backend as the accessionVersion is a primary key.

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/6265?reload=1

Generated with claude CLI

### Backend: New endpoint GET /get-details

Given a list of accession and/or accessionVersions return details such as isRevocation and organism. This allows us to e.g. know which organism an accession belongs to without querying every LAPIS instance. Currently LAPIS is configured on an organism level and cannot perform such queries, if LAPIS becomes able to do this we can always remove this endpoint.
                                                                                                                     
  New files:                                                                                                            
  - SearchController.kt — top-level controller (no organism prefix) with the new endpoint                            
                                                                                                                        
  Modified files:
  - SubmissionTypes.kt — added AccessionVersionWithOrganism data class (accession, version, organism, isRevocation,     
  submittedAt as unix timestamp)                                                                                        
  - SubmissionDatabaseService.kt — added streamSequenceEntryVersionsForAccessions() that queries sequence_entries_view
  for released entries matching the given accessions/accession-versions                                                 
  - SequenceEntriesView.kt — added accessionIsOneOf() helper method                                                     
   
  Endpoint behavior:                                                                                                    
  - ?accessionsOrAccessionVersions=LOC_000S01D → returns all released versions of that accession
  - ?accessionsOrAccessionVersions=LOC_000S01D.1 → returns that specific released version                                           
  - Only returns APPROVED_FOR_RELEASE entries (no auth required)             
  - Returns NDJSON, sorted by accession then version                                                                    
                                                                                                                        ### Website: Use backend instead of all-LAPIS queries                                                                     
                                                                                                                        
  findOrganismAndData.ts (details page): Instead of querying all LAPIS instances in parallel, now calls backend once to 
  get organism, then calls only the correct LAPIS instance. The "no version" redirect case is handled directly from the 
  backend result (finds max version).                                                                                   
                  
  getVersionsData.ts (versions page): Instead of querying all LAPIS instances, calls backend once to get organism, then 
  queries only that organism's LAPIS for version history.
                                                                                                                        
  SeqSetRecordsTableWithMetadata.tsx (seqsets): Instead of querying all LAPIS instances in parallel for each record,    
  calls backend once to get organism for all accessions, then queries only the relevant LAPIS instances (one per
  distinct organism in the seqset).                                                                                     
                  
  backendClient.ts — added getSequenceEntryVersions() method that parses the NDJSON response.                           
   
  backend.ts — added accessionVersionWithOrganism Zod schema and AccessionVersionWithOrganism type. 

### PR Checklist


backend/src/test/kotlin/org/loculus/backend/controller/submission/GetDetailsEndpointTest.kt
```diff
@@ -0,0 +1,204 @@
...
+            .expectNdjsonAndGetContent<AccessionVersionWithOrganism>()
+
+        assertThat(result, `is`(empty()))
+    }
+
+    @Test
+    fun `GIVEN unreleased sequence entry THEN does not return it`() {
+        convenienceClient.prepareDefaultSequenceEntriesToInProcessing()
+
+        val result = submissionControllerClient.getDetails(accessions = listOf("LOC_000S01D"))
```
anna-parker: hmmm.... how do we know that `LOC_000S01D` is the accession of an unreleased sequence entry?

backend/src/test/kotlin/org/loculus/backend/controller/submission/GetDetailsEndpointTest.kt
```diff
@@ -0,0 +1,204 @@
...
+
+    @Test
+    fun `GIVEN released revocation entry THEN it is included in results`() {
+        convenienceClient.prepareRevokedSequenceEntries()
+        val allReleased = convenienceClient.prepareDefaultSequenceEntriesToApprovedForRelease()
+        val accession = allReleased.first().accession
+
+        // Get all accessions including the revoked one
+        val revokedAccessionVersions = submissionControllerClient.getDetails(accessions = listOf(accession))
+            .expectNdjsonAndGetContent<AccessionVersionWithOrganism>()
```
anna-parker: add a check that revocation is True

backend/src/test/kotlin/org/loculus/backend/controller/submission/GetDetailsEndpointTest.kt
```diff
@@ -0,0 +1,204 @@
...
+
+        assertThat(revokedAccessionVersions.size, greaterThan(0))
+    }
+
+    @Test
+    fun `GIVEN released data THEN endpoint is publicly accessible without authentication`() {
```
anna-parker: all tests were called without jwt so this is a pointless test

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -886,6 +887,47 @@ class SubmissionDatabaseService(
         )
     }
 
+    fun streamDetailsForAccessions(
```
anna-parker, line 890: I wonder if I should actually move this to a new file called SearchService.kt?

PR #6259, opened 2026-04-10
feat(website): Metadata documentation for generated fields

tombch:
This PR does the following:

- Adds definitions for common metadata fields in [`‎kubernetes/loculus/templates/_common-metadata.tpl`](https://github.com/loculus-project/loculus/pull/6259/changes#diff-7e3a901948919db7b4aba7f0baeca99179831c5f217f9595531a6d673497eb6b), as well as missing definitions in the default Loculus `values.yaml`.
- Adds the basic files/config for navigating to and viewing metadata documentation in Loculus (without needing e.g. a Pathoplexus overlay).
- Updates the documentation structure with tabs for input and generated fields, ability to link to a specific field on the page in the URL, and a search bar.
- Adds support to the `FormattedText` component for optional rendering of `http`/`https` URLs as links, and this is used within the metadata docs.

resolves #

### PR Checklist

🚀 Preview: https://generated-field-docs.loculus.org


anna-parker (commented):
Already looks great here and on the PPX as well - it might be good to already add the descriptions for the INSDC fields here as well

website/src/routes/extraTopNavigationItems.ts
```diff
@@ -1,7 +1,12 @@
 // The following code is in a separate file so that it can be overwritten by Pathoplexus.
 import type { TopNavigationItems } from './navigationItems.ts';
 
-export const extraStaticTopNavigationItems = [];
+export const extraStaticTopNavigationItems: TopNavigationItems = [
```
anna-parker: I guess as this will now lead all instances running loculus to have a docs link with `Fill this in with documentation for your loculus instance.` maybe we dont want to add a link to the docs by default to the top? @theosanderson do you have thoughts?

website/src/components/OrganismMetadataTable/OrganismMetadataTableSelector.tsx
```diff
@@ -12,50 +13,40 @@ export type OrganismMetadata = {
...
-    const [selectedOrganismKey, setSelectedOrganismKey] = useState('');
-    const selectedOrganism = organisms.find((o) => o.key === selectedOrganismKey) ?? null;
-
-    const handleOrganismSelect = (event: { target: { value: string } }) => {
-        setSelectedOrganismKey(event.target.value);
-    };
+}> = ({ organisms }) => {
+    const [selectedOrganismKey, setSelectedOrganismKey] = useState<string>('');
 
+    // Set the initial selected organism based on the URL parameter
```
anna-parker, line 21: not a react expert but my understanding is we added popstate to get ensure that the state of the page and the URL are kept in sync when a user would navigate forwards and backwards, could this now get out of sync?

website/src/routes/extraTopNavigationItems.ts
tombch: Personally I feel like it would be quite useufl to have the docs link as it provides automatic metadata documentation for people, although I could remove the 'fill this in with documentation...' stuff (as well as removing the similar one for seqsets perhaps?)

website/src/components/OrganismMetadataTable/OrganismMetadataTableSelector.tsx
tombch: (We discussed this and determined removing the `popstate` function doesn't cause any issues, as the url is still kept in sync by the `useEffect` on page load and the `handleOrganismSelect` function when a user selects a new organism)

kubernetes/loculus/templates/_common-metadata.tpl
```diff
@@ -43,6 +48,7 @@ fields:
     hideOnSequenceDetailsPage: true
     header: Submission details
     orderOnDetailsPage: 5010
+    definition: {{ printf "Name of the %s user which submitted the sequence record." $.Values.name | quote }}
```
anna-parker: who submitted?

kubernetes/loculus/values.yaml
```diff
@@ -121,6 +121,9 @@ defaultOrganismConfig: &defaultOrganismConfig
     metadata:
       - name: sampleCollectionDate
         displayName: Collection date
+        definition: "The date on which the sample was collected."
+        guidance: "Please format `YYYY-MM-DD`. If day is unknown, use `YYYY-MM` (e.g., `2020-03`). If both day and month are unknown, use `YYYY` (e.g., `2020`). Please provide at least a year."
```
```suggestion
        guidance: "Please format the date as `YYYY-MM-DD`. If day is unknown, use `YYYY-MM` (e.g., `2020-03`). If both day and month are unknown, use `YYYY` (e.g., `2020`). Please provide at least a year."
```

kubernetes/loculus/values.yaml
```diff
@@ -187,6 +191,7 @@ defaultOrganismConfig: &defaultOrganismConfig
         noInput: true
       - name: ncbiReleaseDate
         displayName: NCBI release date
+        definition: Date the viral nucleotide accession was first released in NCBI Virus.
```
```suggestion
        definition: Date the viral nucleotide accession was first released on the INSDC.
```

kubernetes/loculus/values.yaml
```diff
@@ -198,6 +203,7 @@ defaultOrganismConfig: &defaultOrganismConfig
         columnWidth: 100
       - name: earliestReleaseDate
         displayName: Earliest release date
+        definition: The earliest release date for the accession, across all versions in both Loculus and NCBI Virus.
```
anna-parker: in general I think we should replaceNCBI Virus with either Genbank or INSDC

kubernetes/loculus/values.yaml
```diff
@@ -207,6 +213,7 @@ defaultOrganismConfig: &defaultOrganismConfig
       - name: ncbiUpdateDate
         type: date
         displayName: NCBI update date
+        definition: Date the viral nucleotide accession was last updated in NCBI Virus.
```
anna-parker: also here

kubernetes/loculus/values.yaml
```diff
@@ -594,13 +614,18 @@ defaultOrganismConfig: &defaultOrganismConfig
         orderOnDetailsPage: 520
       - name: specimenCollectorSampleId
         displayName: Isolate name
+        definition: "A de-identified ID for the sample the sequence was obtained from."
```
anna-parker, line 618: I know this is copied from PPX but I do find 'de-identfied ID' odd

kubernetes/loculus/values.yaml
```diff
@@ -2155,6 +2240,7 @@ defaultOrganisms:
           header: "Clade & Lineage"
           isSequenceFilter: true
           displayName: Closest reference
+          definition: "The reference genome this sequence was most closely assigned to by Nextclade."
```
```suggestion
          definition: "The reference genome this sequence was most closely assigned to according to Nextclade."
```

anna-parker (approved):
Not blocking (can be done in a follow up PR), for the multi segmented fields potentially we should modify the description (potentially by pre-appending sth like Segment L: <description>)

Image

Additionally the files description is empty (I fear this might be handled differently than the other metadata fields so feel free to move this to a follow up PR)

Image

kubernetes/loculus/values.yaml
tombch: The only similar things I can think of is either 'masked ID' or 'pseudonymised ID', but I'm not sure if these are equivalent to de-identified data?

kubernetes/loculus/values.yaml
tombch: Updated to use INSDC

kubernetes/loculus/values.yaml
tombch: Updated to use INSDC

kubernetes/loculus/values.yaml
tombch: Updated to use INSDC

kubernetes/loculus/values.yaml
tombch: Closing this as resolved as it was discussed to stick with de-identification: https://loculus.slack.com/archives/C06JCAZLG14/p1777022595852749?thread_ts=1777014456.371549&cid=C06JCAZLG14

PR #6264, opened 2026-04-10
feat(deposition): refactor to use SQLAlchemy

### LINKED ISSUE #5293: Use SQLAlchemy in ENA deposition
I use psycopg2 to manage pools and query the db, however I would like to switch to sqlalchemy as it would give me proper type-safety.

The https://github.com/loculus-project/loculus/blob/main/ena-submission/src/ena_deposition/submission_db_helper.py would need to be rewritten for this. Instead of using custom search functions such as `find_conditions_in_db` write the queries in the sqlAlchemy. 

For example:
```
find_conditions_in_db(
                    db_config,
                    TableName.PROJECT_TABLE,
                    conditions={"group_id": group_id, "organism": organism, "status": "READY"},
                )
```
should be converted to one easy query in the ORM and the response should be typed ProjectTableEntries:
```
stmt = select(ProjectTable).where(
    ProjectTable.group_id == group_id,
    ProjectTable.organism == organism,
    ProjectTable.status == "READY"
)
result = session.execute(stmt)
entries = result.scalars().all() 
```

Side Notes:
- sql alchemy now needs to manage its own thread pool

### LINKED ISSUE #5646: deposition: improve typing

anna-parker:
Claude helped me do this!

resolves https://github.com/loculus-project/loculus/issues/5293
partially resolves https://github.com/loculus-project/loculus/issues/5646

[SQLAlchemy](https://docs.sqlalchemy.org/en/20/tutorial/index.html) is a popular python SQL database toolkit, it is comprised of 2 components: the core which uses a database driver (we continue to use psycopg2) to establish and manage database connections and an Object Relational Mapper (ORM) which maps SQL structures and commands into python objects.

Currently the ENA deposition code uses plain `psycopg2` and I manage all connections in the code, writing raw SQL transactions myself. This works quite well but `psycopg2` returns database transactions as dictionaries where the keys are of type str and correspond to column names and the values are of type Any. psycopg2 do not check column names exist or that values are used with correct types in the code. 

### Changes

1. Defines table objects via declarative mapping classes: https://docs.sqlalchemy.org/en/20/tutorial/metadata.html#declaring-mapped-classes and dataclass wrappers: https://docs.sqlalchemy.org/en/21/orm/dataclasses.html#declarative-dataclass-mapping
2. SQLAlchemy now manages the connections to the database via an "engine" (psycopg2 is still the database driver), the engine manages the connection pool (removing the requirement for custom connection pool code that I added before)
3. Database operations are now performed using SQLAlchemy's ORM layer within [sessions](https://docs.sqlalchemy.org/en/20/tutorial/orm_data_manipulation.html) (removing the requirement to open DB connections ourselves and excecute raw SQL statements). (ORM) Objects are added to sessions and the session ensures they are passed to the DB as transactions in an optimized automated manner (can be forced using "flush"). Objects persist in a Session using a "unit of work pattern". Note that although the session manages DB connections, commit operations still need to be added explicitly. SQLAlchemy performs JSONB dumps and validates inserted values to prevent SQL injections. After changes are commited, the objects still persist in the session and are linked to their location in memory via an "identity map", this can be used to retrieve e.g. (generated) primary keys. Note objects stay attached to a session until the session is closed. 
4. The largest improvement over the current state is that return values/rows are mapped into typed dataclasses instead of dictionary objects. This alone woudl have prevent most of the typing error encountered in the codebase.

### Limitations

Note the `update_db_where_conditions` and `find_conditions_in_db` still pass dictionaries. I did this primarily to restrict the changes in this PR to refactoring changes and not logic changes. The dictionaries that are passed are not typed.

To improve the typing of these functions we could move to using typedDicts as suggested in the issue or write out typed ORM queries each time instead of using a dictionary of table columns/values to filter on/ update. 

### PR Checklist
Tested on staging - see notes in https://loculus.slack.com/archives/C05G172HL6L/p1777376294911849

### DIFF
--- a/ena-submission/scripts/get_ena_submission_list.py
+++ b/ena-submission/scripts/get_ena_submission_list.py
@@ -113,7 +113,7 @@ def assign_ena_organism(
 
 def filter_for_submission(
     config: Config,
-    db_pool: Engine,
+    db_engine: Engine,
     entries_iterator: Iterator[dict[str, Any]],
     ena_organisms: list[ENAOrganism],
 ) -> SubmissionResults:
@@ -134,7 +134,7 @@ def filter_for_submission(
     entries_with_external_metadata: set[Accession] = set()
     revoked_entries: set[Accession] = set()
     logger.debug("Querying ENA db for latest version of submissions")
-    highest_submitted_version = highest_version_in_submission_table(db_pool)
+    highest_submitted_version = highest_version_in_submission_table(db_engine)
     logger.debug("Starting processing of data from Loculus backend")
     suppressed_accessions = fetch_suppressed_accessions(config)
     for idx, entry in enumerate(entries_iterator):
@@ -257,7 +257,7 @@ def get_ena_submission_list(config_file) -> None:
 
     output_file_suffix = "ena_submission_list.json"
 
-    db_pool = db_init(
+    db_engine = db_init(
         db_password_default=config.db_password,
         db_username_default=config.db_username,
         db_url_default=config.db_url,
@@ -276,7 +276,7 @@ def get_ena_submission_list(config_file) -> None:
         released_entries = fetch_released_entries(config, loculus_organism)
         logger.info("Starting to stream released entries. Filtering for submission...")
         submission_results = filter_for_submission(
-            config, db_pool, released_entries, input_map[loculus_organism]
+            config, db_engine, released_entries, input_map[loculus_organism]
         )
         if submission_results.entries_to_submit:
             logger.info(
--- a/ena-submission/scripts/get_ena_submission_list.py
+++ b/ena-submission/scripts/get_ena_submission_list.py
@@ -22,7 +22,7 @@ from ena_deposition.submission_db_helper import (
     db_init,
     highest_version_in_submission_table,
 )
-from psycopg2.pool import SimpleConnectionPool
+from sqlalchemy import Engine
 
 logger = logging.getLogger(__name__)
 logging.basicConfig(
@@ -113,7 +113,7 @@ def assign_ena_organism(
 
 def filter_for_submission(
     config: Config,
-    db_pool: SimpleConnectionPool,
+    db_pool: Engine,
     entries_iterator: Iterator[dict[str, Any]],
     ena_organisms: list[ENAOrganism],
 ) -> SubmissionResults:
@@ -134,7 +134,7 @@ def filter_for_submission(
     entries_with_external_metadata: set[Accession] = set()
     revoked_entries: set[Accession] = set()
     logger.debug("Querying ENA db for latest version of submissions")
-    highest_submitted_version = highest_version_in_submission_table(db_conn_pool=db_pool)
+    highest_submitted_version = highest_version_in_submission_table(db_pool)
     logger.debug("Starting processing of data from Loculus backend")
     suppressed_accessions = fetch_suppressed_accessions(config)
     for idx, entry in enumerate(entries_iterator):
--- a/ena-submission/src/ena_deposition/api.py
+++ b/ena-submission/src/ena_deposition/api.py
@@ -1,5 +1,6 @@
 import logging
 import threading
+from typing import cast
 
 import uvicorn
 from fastapi import FastAPI, HTTPException
@@ -25,7 +26,9 @@ def get_bio_sample_accessions(engine: Engine) -> dict[str, str]:
     with Session(engine) as session:
         stmt = select(SampleTableEntry).where(SampleTableEntry.status == str(Status.SUBMITTED))
         results = list(session.scalars(stmt).all())
-    return {row.accession: row.result["biosample_accession"] for row in results if row.result}
+    return {
+        row.accession: cast(str, row.result["biosample_accession"]) for row in results if row.result
+    }
 
 
 def get_insdc_accessions(engine: Engine) -> dict[str, list[str]]:
@@ -36,7 +39,9 @@ def get_insdc_accessions(engine: Engine) -> dict[str, list[str]]:
         results = list(session.scalars(stmt).all())
     return {
         row.accession: [
-            row.result[key] for key in row.result if key.startswith("insdc_accession_full")
+            cast(str, row.result[key])
+            for key in row.result
+            if key.startswith("insdc_accession_full")
         ]
         for row in results
         if row.result
--- a/ena-submission/src/ena_deposition/api.py
+++ b/ena-submission/src/ena_deposition/api.py
@@ -50,6 +50,7 @@ def get_insdc_accessions(engine: Engine) -> dict[str, list[str]]:
 
 def init_app(config: Config):
     app.state.config = config
+    app.state.engine = db_init(config.db_password, config.db_username, config.db_url)
 
 
 @app.get("/")
@@ -59,8 +60,7 @@ def read_root():
 
 @app.get("/submitted", response_model=SubmittedAccessionsResponse)
 def submitted_insdc_accessions():
-    config = app.state.config
-    engine = db_init(config.db_password, config.db_username, config.db_url)
+    engine = app.state.engine
     try:
         insdc_accessions = get_insdc_accessions(engine)
         all_insdc_accessions = [item for sublist in insdc_accessions.values() for item in sublist]
--- a/ena-submission/src/ena_deposition/api.py
+++ b/ena-submission/src/ena_deposition/api.py
@@ -3,12 +3,12 @@ import threading
 
 import uvicorn
 from fastapi import FastAPI, HTTPException
-from psycopg2.extras import RealDictCursor
-from psycopg2.pool import SimpleConnectionPool
 from pydantic import BaseModel
+from sqlalchemy import Engine, select
+from sqlalchemy.orm import Session
 
 from .config import Config
-from .submission_db_helper import db_init
+from .submission_db_helper import AssemblyTableEntry, SampleTableEntry, Status, db_init
 
 logger = logging.getLogger(__name__)
 
@@ -21,39 +21,33 @@ class SubmittedAccessionsResponse(BaseModel):
     biosampleAccessions: list[str]  # noqa: N815
 
 
-def get_bio_sample_accessions(db_conn_pool: SimpleConnectionPool) -> dict[str, str]:
-    con = db_conn_pool.getconn()
-    try:
-        with con, con.cursor(cursor_factory=RealDictCursor) as cur:
-            query = "SELECT accession, result FROM sample_table WHERE STATUS = 'SUBMITTED'"
-            cur.execute(query)
-            results = cur.fetchall()
-    finally:
-        db_conn_pool.putconn(con)
-
-    return {result["accession"]: result["result"]["biosample_accession"] for result in results}
-
-
-def get_insdc_accessions(db_conn_pool: SimpleConnectionPool) -> dict[str, list[str]]:
-    con = db_conn_pool.getconn()
-    try:
-        with con, con.cursor(cursor_factory=RealDictCursor) as cur:
-            query = (
-                "SELECT accession, result FROM assembly_table "
-                "WHERE STATUS IN ('SUBMITTED', 'WAITING')"
-            )
-            cur.execute(query)
-            results = cur.fetchall()
-    finally:
-        db_conn_pool.putconn(con)
-
+def get_bio_sample_accessions(engine: Engine) -> dict[str, str]:
+    with Session(engine) as session:
+        stmt = select(SampleTableEntry).where(
+            SampleTableEntry.status == str(Status.SUBMITTED)
+        )
+        results = list(session.scalars(stmt).all())
     return {
-        result["accession"]: [
-            result["result"][key]
-            for key in result["result"]
+        row.accession: row.result["biosample_accession"]
+        for row in results
+        if row.result
+    }
+
+
+def get_insdc_accessions(engine: Engine) -> dict[str, list[str]]:
+    with Session(engine) as session:
+        stmt = select(AssemblyTableEntry).where(
+            AssemblyTableEntry.status.in_([str(Status.SUBMITTED), str(Status.WAITING)])
+        )
+        results = list(session.scalars(stmt).all())
+    return {
+        row.accession: [
+            row.result[key]
+            for key in row.result
             if key.startswith("insdc_accession_full")
         ]
-        for result in results
+        for row in results
+        if row.result
     }
 
 
@@ -69,11 +63,11 @@ def read_root():
 @app.get("/submitted", response_model=SubmittedAccessionsResponse)
 def submitted_insdc_accessions():
     config = app.state.config
-    db_conn_pool = db_init(config.db_password, config.db_username, config.db_url)
+    engine = db_init(config.db_password, config.db_username, config.db_url)
     try:
-        insdc_accessions = get_insdc_accessions(db_conn_pool)
+        insdc_accessions = get_insdc_accessions(engine)
         all_insdc_accessions = [item for sublist in insdc_accessions.values() for item in sublist]
-        bio_samples = list(get_bio_sample_accessions(db_conn_pool).values())
+        bio_samples = list(get_bio_sample_accessions(engine).values())
         return {
             "status": "ok",
             "insdcAccessions": all_insdc_accessions,
--- a/ena-submission/src/ena_deposition/api.py
+++ b/ena-submission/src/ena_deposition/api.py
@@ -23,15 +23,9 @@ class SubmittedAccessionsResponse(BaseModel):
 
 def get_bio_sample_accessions(engine: Engine) -> dict[str, str]:
     with Session(engine) as session:
-        stmt = select(SampleTableEntry).where(
-            SampleTableEntry.status == str(Status.SUBMITTED)
-        )
+        stmt = select(SampleTableEntry).where(SampleTableEntry.status == str(Status.SUBMITTED))
         results = list(session.scalars(stmt).all())
-    return {
-        row.accession: row.result["biosample_accession"]
-        for row in results
-        if row.result
-    }
+    return {row.accession: row.result["biosample_accession"] for row in results if row.result}
 
 
 def get_insdc_accessions(engine: Engine) -> dict[str, list[str]]:
@@ -42,9 +36,7 @@ def get_insdc_accessions(engine: Engine) -> dict[str, list[str]]:
         results = list(session.scalars(stmt).all())
     return {
         row.accession: [
-            row.result[key]
-            for key in row.result
-            if key.startswith("insdc_accession_full")
+            row.result[key] for key in row.result if key.startswith("insdc_accession_full")
         ]
         for row in results
         if row.result
--- a/ena-submission/src/ena_deposition/api.py
+++ b/ena-submission/src/ena_deposition/api.py
@@ -24,7 +24,7 @@ class SubmittedAccessionsResponse(BaseModel):
 
 def get_bio_sample_accessions(engine: Engine) -> dict[str, str]:
     with Session(engine) as session:
-        stmt = select(SampleTableEntry).where(SampleTableEntry.status == str(Status.SUBMITTED))
+        stmt = select(SampleTableEntry).where(SampleTableEntry.status == Status.SUBMITTED)
         results = list(session.scalars(stmt).all())
     return {
         row.accession: cast(str, row.result["biosample_accession"]) for row in results if row.result
@@ -34,7 +34,7 @@ def get_bio_sample_accessions(engine: Engine) -> dict[str, str]:
 def get_insdc_accessions(engine: Engine) -> dict[str, list[str]]:
     with Session(engine) as session:
         stmt = select(AssemblyTableEntry).where(
-            AssemblyTableEntry.status.in_([str(Status.SUBMITTED), str(Status.WAITING)])
+            AssemblyTableEntry.status.in_([Status.SUBMITTED, Status.WAITING])
         )
         results = list(session.scalars(stmt).all())
     return {
--- a/ena-submission/src/ena_deposition/check_external_visibility.py
+++ b/ena-submission/src/ena_deposition/check_external_visibility.py
@@ -268,7 +268,7 @@ def check_and_update_visibility_for_column(
     )
 
     for entity in entities_needing_check:
-        entity_id = asdict(entity.primary_key)
+        entity_id = asdict(entity.pkey)
         accessions = get_accessions_to_check(entity, column_config)
 
         if not accessions:
--- a/ena-submission/src/ena_deposition/check_external_visibility.py
+++ b/ena-submission/src/ena_deposition/check_external_visibility.py
@@ -18,7 +18,7 @@ from http import HTTPStatus
 
 import pytz
 import requests
-from psycopg2.pool import SimpleConnectionPool
+from sqlalchemy import Engine
 
 from ena_deposition.config import Config
 from ena_deposition.submission_db_helper import (
@@ -26,7 +26,6 @@ from ena_deposition.submission_db_helper import (
     ProjectTableEntry,
     SampleTableEntry,
     Status,
-    TableName,
     db_init,
     find_conditions_in_db,
     update_db_where_conditions,
@@ -45,7 +44,6 @@ class EntityType(Enum):
 class ColumnCheckConfig:
     """Configuration for checking a specific table column"""
 
-    table_name: TableName
     entry_class: type[ProjectTableEntry | SampleTableEntry | AssemblyTableEntry]
     visibility_column: str
     accession_field_name_prefix: str  # Field prefix in result dict (e.g. "insdc_accession_full")
@@ -152,28 +150,24 @@ class NCBIVisibilityChecker(VisibilityChecker):
 # Configuration mapping: (EntityType, column_name) -> ColumnCheckConfig
 COLUMN_CONFIGS = {
     (EntityType.PROJECT, "ena_first_publicly_visible"): ColumnCheckConfig(
-        table_name=TableName.PROJECT_TABLE,
         entry_class=ProjectTableEntry,
         visibility_column="ena_first_publicly_visible",
         accession_field_name_prefix="bioproject_accession",
         checker_class=ENAVisibilityChecker,
     ),
     (EntityType.PROJECT, "ncbi_first_publicly_visible"): ColumnCheckConfig(
-        table_name=TableName.PROJECT_TABLE,
         entry_class=ProjectTableEntry,
         visibility_column="ncbi_first_publicly_visible",
         accession_field_name_prefix="bioproject_accession",
         checker_class=NCBIVisibilityChecker,
     ),
     (EntityType.SAMPLE, "ena_first_publicly_visible"): ColumnCheckConfig(
-        table_name=TableName.SAMPLE_TABLE,
         entry_class=SampleTableEntry,
         visibility_column="ena_first_publicly_visible",
         accession_field_name_prefix="biosample_accession",
         checker_class=ENAVisibilityChecker,
     ),
     (EntityType.SAMPLE, "ncbi_first_publicly_visible"): ColumnCheckConfig(
-        table_name=TableName.SAMPLE_TABLE,
         entry_class=SampleTableEntry,
         visibility_column="ncbi_first_publicly_visible",
         accession_field_name_prefix="biosample_accession",
@@ -181,14 +175,12 @@ COLUMN_CONFIGS = {
     ),
     # Assemblies - ENA nucleotide accessions
     (EntityType.ASSEMBLY, "ena_nucleotide_first_publicly_visible"): ColumnCheckConfig(
-        table_name=TableName.ASSEMBLY_TABLE,
         entry_class=AssemblyTableEntry,
         visibility_column="ena_nucleotide_first_publicly_visible",
         accession_field_name_prefix="insdc_accession_full",  # Prefix for multi-segment accessions
         checker_class=ENAVisibilityChecker,
     ),
     (EntityType.ASSEMBLY, "ncbi_nucleotide_first_publicly_visible"): ColumnCheckConfig(
-        table_name=TableName.ASSEMBLY_TABLE,
         entry_class=AssemblyTableEntry,
         visibility_column="ncbi_nucleotide_first_publicly_visible",
         accession_field_name_prefix="insdc_accession_full",  # Prefix for multi-segment accessions
@@ -196,14 +188,12 @@ COLUMN_CONFIGS = {
     ),
     # Assemblies - ENA GCA accessions
     (EntityType.ASSEMBLY, "ena_gca_first_publicly_visible"): ColumnCheckConfig(
-        table_name=TableName.ASSEMBLY_TABLE,
         entry_class=AssemblyTableEntry,
         visibility_column="ena_gca_first_publicly_visible",
         accession_field_name_prefix="gca_accession",
         checker_class=ENAVisibilityChecker,
     ),
     (EntityType.ASSEMBLY, "ncbi_gca_first_publicly_visible"): ColumnCheckConfig(
-        table_name=TableName.ASSEMBLY_TABLE,
         entry_class=AssemblyTableEntry,
         visibility_column="ncbi_gca_first_publicly_visible",
         accession_field_name_prefix="gca_accession",
@@ -213,21 +203,17 @@ COLUMN_CONFIGS = {
 
 
 def get_entities_needing_column_check(
-    pool: SimpleConnectionPool, column_config: ColumnCheckConfig
+    engine: Engine, column_config: ColumnCheckConfig
 ) -> list[SampleTableEntry | ProjectTableEntry | AssemblyTableEntry]:
-    """Get entities (db rows for Project/Sample/Assembly) that don't have a timestamp
-    for a specific visibility column"""
-
-    data_in_submission_table: list[dict] = find_conditions_in_db(
-        pool,
-        table_name=column_config.table_name,
+    """Get entities that don't have a timestamp for a specific visibility column"""
+    return find_conditions_in_db(
+        engine,
+        column_config.entry_class,
         conditions={
             column_config.visibility_column: None,
-            "status": Status.SUBMITTED,
+            "status": str(Status.SUBMITTED),
         },
     )
-
-    return [column_config.entry_class(**row) for row in data_in_submission_table]
 
 
 def get_accessions_to_check(
@@ -258,7 +244,7 @@ def get_accessions_to_check(
 
 def check_and_update_visibility_for_column(
     config: Config,
-    pool: SimpleConnectionPool,
+    engine: Engine,
     entity_type: EntityType,
     column_name: str,
 ):
@@ -276,7 +262,7 @@ def check_and_update_visibility_for_column(
         return
 
     logger.debug(f"Checking {entity_type.value}.{column_name} for visibility")
-    entities_needing_check = get_entities_needing_column_check(pool, column_config)
+    entities_needing_check = get_entities_needing_column_check(engine, column_config)
     logger.info(
         f"Found {len(entities_needing_check)} {entity_type.value}s needing {column_name} check"
     )
@@ -318,8 +304,8 @@ def check_and_update_visibility_for_column(
                 "publicly visible, updating database."
             )
             updated_count = update_db_where_conditions(
-                pool,
-                table_name=column_config.table_name,
+                engine,
+                model_class=column_config.entry_class,
                 conditions=entity_id,
                 update_values={column_config.visibility_column: first_visible_timestamp},
             )
@@ -342,19 +328,19 @@ def check_and_update_visibility_for_column(
             )
 
 
-def check_and_update_visibility_all_columns(config: Config, pool: SimpleConnectionPool):
+def check_and_update_visibility_all_columns(config: Config, engine: Engine):
     """Check and update visibility for all configured (entity_type, column) combinations"""
 
     for entity_type, column_name in COLUMN_CONFIGS:
         try:
-            check_and_update_visibility_for_column(config, pool, entity_type, column_name)
+            check_and_update_visibility_for_column(config, engine, entity_type, column_name)
         except Exception as e:
             logger.error(f"Error checking {entity_type.value}.{column_name}: {e}", exc_info=True)
 
 
 def check_and_update_visibility(config: Config, stop_event: threading.Event):
     """Main loop function"""
-    pool = db_init(config.db_password, config.db_username, config.db_url)
+    engine = db_init(config.db_password, config.db_username, config.db_url)
 
     while True:
         start_time = time.time()
@@ -362,7 +348,7 @@ def check_and_update_visibility(config: Config, stop_event: threading.Event):
             logger.info("check_and_update_visibility stopped due to exception in another task")
             return
 
-        check_and_update_visibility_all_columns(config, pool)
+        check_and_update_visibility_all_columns(config, engine)
         logger.debug("check_and_update_visibility finished, sleeping for a while")
 
         gca_cache.clear()
--- a/ena-submission/src/ena_deposition/check_external_visibility.py
+++ b/ena-submission/src/ena_deposition/check_external_visibility.py
@@ -257,9 +257,6 @@ def check_and_update_visibility_for_column(
 
     # Get the appropriate visibility checker
     visibility_checker: VisibilityChecker = column_config.checker_class()
-    if not visibility_checker:
-        logger.error(f"No checker found for {column_config.checker_class}")
-        return
 
     logger.debug(f"Checking {entity_type.value}.{column_name} for visibility")
     entities_needing_check = get_entities_needing_column_check(db_engine, column_config)
--- a/ena-submission/src/ena_deposition/check_external_visibility.py
+++ b/ena-submission/src/ena_deposition/check_external_visibility.py
@@ -203,11 +203,11 @@ COLUMN_CONFIGS = {
 
 
 def get_entities_needing_column_check(
-    engine: Engine, column_config: ColumnCheckConfig
+    db_engine: Engine, column_config: ColumnCheckConfig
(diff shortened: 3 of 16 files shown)


ena-submission/src/ena_deposition/check_external_visibility.py
```diff
@@ -213,22 +203,18 @@ def check_visibility(self, config: Config, accession: str) -> datetime | None:
...
-        pool,
-        table_name=column_config.table_name,
+    """Get entities that don't have a timestamp for a specific visibility column"""
+    return find_conditions_in_db(
+        engine,
+        column_config.entry_class,
         conditions={
             column_config.visibility_column: None,
-            "status": Status.SUBMITTED,
+            "status": str(Status.SUBMITTED),
```
anna-parker: Im not a fan of having to cast this to a string

ena-submission/src/ena_deposition/submission_db_helper.py
```diff
@@ -154,348 +126,315 @@ class ProjectId:
...
+
+class Base(MappedAsDataclass, DeclarativeBase):
+    pass
+
+
+class SubmissionTableEntry(Base):
+    """Maps to submission_table. Primary key: (accession, version)."""
+
+    __tablename__ = "submission_table"
+    __table_args__: typing.ClassVar[dict[str, Any]] = {"schema": "ena_deposition_schema"}
```
anna-parker, line 134: Im not quite sure why this is needed

ena-submission/src/ena_deposition/submission_db_helper.py
```diff
@@ -154,348 +126,315 @@ class ProjectId:
...
+    started_at: Mapped[datetime | None] = mapped_column(default=None)
+    finished_at: Mapped[datetime | None] = mapped_column(default=None)
+    unaligned_nucleotide_sequences: Mapped[dict[str, str | None]] = mapped_column(
+        JSONB, default_factory=dict
+    )
+    center_name: Mapped[str | None] = mapped_column(default=None)
+    external_metadata: Mapped[dict[str, str | Sequence[str]] | None] = mapped_column(
+        JSONB, default=None
+    )
+    project_id: Mapped[int | None] = mapped_column(default=None)
```
anna-parker, line 160: this is actually a foreign key

ena-submission/src/ena_deposition/submission_db_helper.py
```diff
@@ -154,348 +126,315 @@ class ProjectId:
...
+    the model class, so a typo raises AttributeError at import/call time
+    rather than a silent runtime KeyError.
+    """
+    with Session(engine) as session:
+        stmt = select(model_class)
+        for col_name, value in conditions.items():
+            col = getattr(model_class, col_name)
+            stmt = stmt.where(col.is_(None)) if value is None else stmt.where(col == value)
+        rows = list(session.scalars(stmt).all())
+        session.expunge_all()
```
anna-parker, line 288: this removes all objects from the session but Im not quite sure why its needed...

anna-parker: https://docs.sqlalchemy.org/en/20/tutorial/orm_data_manipulation.html#committing

ena-submission/src/ena_deposition/submission_db_helper.py
```diff
@@ -154,348 +126,315 @@ class ProjectId:
...
+    generate an equality check.  Column names are Python attribute names on
+    the model class, so a typo raises AttributeError at import/call time
+    rather than a silent runtime KeyError.
+    """
+    with Session(engine) as session:
+        stmt = select(model_class)
+        for col_name, value in conditions.items():
+            col = getattr(model_class, col_name)
+            stmt = stmt.where(col.is_(None)) if value is None else stmt.where(col == value)
+        rows = list(session.scalars(stmt).all())
```
anna-parker, line 287: session.scalars() returns ORM objects

ena-submission/src/ena_deposition/submission_db_helper.py
anna-parker: Ah it is required here because session.scalars() returns ORM objects and loads them into the session, but e.g. session.execute doesnt - hence no need to expunge_all

ena-submission/src/ena_deposition/submission_db_helper.py
```diff
@@ -154,348 +126,315 @@ class ProjectId:
...
+    logger.debug(f"Deleting records from '{model_class.__name__}' where {conditions}")
+    with Session(engine) as session:
+        stmt = delete(model_class)
+        for col_name, value in conditions.items():
+            col = getattr(model_class, col_name)
+            stmt = stmt.where(col.is_(None)) if value is None else stmt.where(col == value)
+        result = session.execute(stmt)
+        session.commit()
+        deleted_rows = result.rowcount
+    logger.debug(f"Deleted {deleted_rows} rows from '{model_class.__name__}' where {conditions}")
```
anna-parker, line 316: I need to check why the model_class can't access the __table_name__

ena-submission/src/ena_deposition/submission_db_helper.py
```diff
@@ -527,179 +469,90 @@ def _do_update():
...
-                    json.dumps(sample_table_entry.errors),
-                    json.dumps(sample_table_entry.warnings),
-                    str(sample_table_entry.status),
-                    sample_table_entry.started_at,
-                    sample_table_entry.finished_at,
-                    json.dumps(sample_table_entry.result),
-                ),
-            )
-            con.commit()
+        with Session(engine, expire_on_commit=False) as session:
```
anna-parker, line 495: I don't understand why `expire_on_commit=False` has been added here...

anna-parker: this means that I can still access object attributes after the transaction has completed and the session has been closed

ena-submission/src/ena_deposition/submission_db_helper.py
anna-parker: technically this isnt needed as the table schema is specified in the engine db conection args but it is apparently good practice to keep it in the table definitons

ena-submission/src/ena_deposition/submission_db_helper.py
```diff
@@ -154,348 +126,315 @@ class ProjectId:
...
+    status: Mapped[str] = mapped_column(default=str(Status.READY))
+    started_at: Mapped[datetime | None] = mapped_column(default=None)
+    finished_at: Mapped[datetime | None] = mapped_column(default=None)
+    center_name: Mapped[str | None] = mapped_column(default=None)
+    result: Mapped[dict[str, str | Sequence[str]] | None] = mapped_column(JSONB, default=None)
+    ena_first_publicly_visible: Mapped[datetime | None] = mapped_column(default=None)
+    ncbi_first_publicly_visible: Mapped[datetime | None] = mapped_column(default=None)
 
     @property
     def primary_key(self) -> ProjectId:
```
anna-parker: I should maybe rename this as sqlalchemy also has a primary_key object, it will not be overwritten but it gets confusing, I can maybe call it pkey or sth

ena-submission/src/ena_deposition/submission_db_helper.py
anna-parker: it actually does exist this a typing issue but I cant be bothered to correct it, model_class.__name__ will just return TableEntry instead of Table but this is sufficient

ena-submission/src/ena_deposition/check_external_visibility.py
tombch: You can use `Status.SUBMITTED.value` for enums!

anna-parker: ooh nice! Will changed in a bit!

anna-parker: SQLAlchemy actually supports enum types when parsing strings from db tables, which I believe is a better choice here to prevent ever having to cast an enum to a string :-) https://docs.sqlalchemy.org/en/21/core/type_basics.html#sqlalchemy.types.Enum

ena-submission/src/ena_deposition/submission_db_helper.py
```diff
@@ -154,348 +123,327 @@ class ProjectId:
...
-                    where_conditions.append(f"{key} = %s")
-                    params.append(type_conversion(value))
-            if where_conditions:
-                query += " WHERE " + " AND ".join(where_conditions)
 
-            cur.execute(query, params)
-            results = cur.fetchall()
-    finally:
-        db_conn_pool.putconn(con)
+T = TypeVar(
```
maverbiest: I don't think this TypeVar is ever used atm: all helper functions have the form `helper[T](..., model_class type[T],..) -> T`. AFAIK, this does not constrain the input type, it just links the input type to the output type.

If we'd rework the signatures to be `helper(..., model_class T,..) -> T`, typecheckers could actually check whether we pass in one of the four types of table entries

ena-submission/src/ena_deposition/submission_db_helper.py
```diff
@@ -527,179 +478,90 @@ def _do_update():
         return 0
 
 
-def add_to_project_table(
-    db_conn_pool: SimpleConnectionPool, project_table_entry: ProjectTableEntry
-) -> int | None:
-    con = db_conn_pool.getconn()
+def add_to_project_table(engine: Engine, entry: ProjectTableEntry) -> int | None:
```
maverbiest, line 476: I think the four `add_to_X_table` helpers could be consolidated into one `add_to_table` method with signature `def add_to_table(engine: Engine, entry: T) -> T | None`

The return type would still be truthy on success and falsy otherwise (so works in place of current bool return values), and since it returns the entry on success, you could still get the `project_id` from it.

ena-submission/src/ena_deposition/submission_db_helper.py
```diff
@@ -154,348 +123,327 @@ class ProjectId:
...
+        stmt = select(model_class)
+        for col_name, value in conditions.items():
+            col = getattr(model_class, col_name)
+            stmt = stmt.where(col.is_(None)) if value is None else stmt.where(col == value)
+        rows = list(session.scalars(stmt).all())
+        session.expunge_all()
+        return rows
+
+
+def delete_records_in_db[T](
```
maverbiest: This helper is never user in the production code, just used once in a test. Might make sense to inline it there and remove the helper

ena-submission/src/ena_deposition/submission_db_helper.py
```diff
@@ -154,348 +123,327 @@ class ProjectId:
...
+        stmt = select(model_class).where(
+            (status_col == Status.HAS_ERRORS)
+            | ((status_col == Status.SUBMITTING) & (started_at_col < min_start_time))
+        )
+        rows = list(session.scalars(stmt).all())
+        session.expunge_all()
+        return rows
 
 
 def find_stuck_in_submission_db(
```
maverbiest, line 339: This helper is only used once, so could be inlined at the call site (no preference either way, just flagging)

ena-submission/src/ena_deposition/submission_db_helper.py
```diff
@@ -154,348 +123,327 @@ class ProjectId:
...
+        stmt = select(SubmissionTableEntry).where(
+            SubmissionTableEntry.status_all == StatusAll.HAS_ERRORS_EXT_METADATA_UPLOAD,
+            SubmissionTableEntry.started_at < min_start_time,
+        )
+        rows = list(session.scalars(stmt).all())
+        session.expunge_all()
+        return rows
 
 
 def find_waiting_in_db(
```
maverbiest, line 355: This helper is only used once, so could be inlined at the call site (no preference either way, just flagging)

ena-submission/src/ena_deposition/submission_db_helper.py
```diff
@@ -154,348 +123,327 @@ class ProjectId:
...
+        rows = list(session.scalars(stmt).all())
+        session.expunge_all()
+        return rows
 
 
-def update_db_where_conditions(
-    db_conn_pool: SimpleConnectionPool,
-    table_name: TableName,
-    conditions: Mapping[str, str | int],
+def update_db_where_conditions[T](
```
maverbiest: For `update_db_where_conditions` and `update_with_retry` could consider adding thin wrappers for calling them with primary keys, e.g.:
```
  def update_by_pkey[T](
      engine: Engine,
      model_class: type[T],
      pkey: AccessionVersion | ProjectId,
      updates: dict[str, Any],
  ) -> int:
      return update_db_where_conditions(
          engine,
          model_class=model_class,
          conditions=asdict(pkey),
          update_values=updates,
      )
```

This would save us having to construct `seq_key = {"accession": row.accession, "version": row.version}` in many places.

ena-submission/src/ena_deposition/submission_db_helper.py
```diff
@@ -154,348 +126,315 @@ class ProjectId:
...
-            updated_row_count = cur.rowcount
-            con.commit()
-    except (Exception, psycopg2.DatabaseError) as e:
-        con.rollback()
+            stmt = stmt.values(**update_values)
+            result = session.execute(stmt)
+            session.commit()
+            updated_row_count = result.rowcount
+    except Exception as e:
         logger.warning(f"update_db_where_conditions errored with: {e}")
```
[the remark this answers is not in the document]
maverbiest: I think I'd prefer removing the try/except from this function and having the caller deal with any exceptions. Would also allow for cleaner retry logic e.g., `retry=retry_if_exception_type(OperationalError),` in `update_with_retry` to only retry on transient DB erorrs

ena-submission/src/ena_deposition/submission_db_helper.py
maverbiest: Yes it'd only make sense if these methods return the actualy entry, right?

ena-submission/src/ena_deposition/submission_db_helper.py
```diff
@@ -527,179 +478,90 @@ def _do_update():
...
         return False
-    finally:
-        db_conn_pool.putconn(con)
 
 
-def is_revision(db_config: SimpleConnectionPool, seq_key: AccessionVersion) -> bool:
-    """Check if the entry is a revision"""
-    version = seq_key.version
-    if version == 1:
+def is_revision(engine: Engine, seq_key: AccessionVersion) -> bool:
```
maverbiest: The name of this function is a bit confusing to me. It returns True _only_ if the seq_key provided is the highest version in a chain of versions. So for version 3 out of 3, it returns True, but for version 2 out of 3 it returns False (which is what the doc string says, so that's fine).

My confusion comes from the fact that, personally, I'd still call version 2 out of 3 a revision: it's a revised version of the original entry. Maybe we could we rename this to `is_highest_version`?

ena-submission/src/ena_deposition/submission_db_helper.py
```diff
@@ -527,179 +478,90 @@ def _do_update():
...
     )
-    all_versions = sorted([entry["version"] for entry in sample_data_in_submission_table])
-    return len(all_versions) > 1 and version == all_versions[-1]
+    all_versions = sorted(row.version for row in rows)
+    return len(all_versions) > 1 and seq_key.version == all_versions[-1]
 
 
-def last_version(db_config: SimpleConnectionPool, seq_key: AccessionVersion) -> int | None:
-    if not is_revision(db_config, seq_key):
+def last_version(engine: Engine, seq_key: AccessionVersion) -> int | None:
```
maverbiest: I was similarly confused by this function, but here the docstring also seems to actually contradict the behaviour (unless I'm mistaken): this function _always_ returns the second to last version of a submission if `seq_key` is the latest of multiple versions of an accession (as checked by `is_revision`). 

However, if the `seq_key` I'm providing is for version 2 out of 3, this function will return None. Based on the docstring, I'd expect it to give me back version 1 instead, as it's the previous version number.

ena-submission/src/ena_deposition/upload_external_metadata_to_loculus.py
```diff
@@ -25,39 +28,36 @@
 logger = logging.getLogger(__name__)
 
 
-def _get_results_of_single_db_record(
-    db_config: SimpleConnectionPool,
-    table_name: TableName,
+def _get_result_of_single_db_record(
+    db_config: Engine,
+    model_class: type,
```
maverbiest: We could constrain this to only take one of `(SubmissionTableEntry, ProjectTableEntry,SampleTableEntry, AssemblyTableEntry)` (potentially via a TypeVar as in `submission_db_helper.py`)

ena-submission/src/ena_deposition/upload_external_metadata_to_loculus.py
```diff
@@ -158,16 +156,16 @@ def get_external_metadata_and_send_to_loculus(
...
             db_config,
-            table_name=TableName.SUBMISSION_TABLE,
+            SubmissionTableEntry,
             conditions={"status_all": status},
         ):
-            accession = entry["accession"]
-            version = entry["version"]
+            accession = entry.accession
+            version = entry.version
             accession_version = f"{accession}.{version}"
```
maverbiest: We should use `entry.pkey` here, which we can then also use for the rest of this function

ena-submission/src/ena_deposition/api.py
```diff
@@ -21,44 +22,35 @@ class SubmittedAccessionsResponse(BaseModel):
...
-        db_conn_pool.putconn(con)
-
-    return {result["accession"]: result["result"]["biosample_accession"] for result in results}
-
+def get_bio_sample_accessions(engine: Engine) -> dict[str, str]:
+    with Session(engine) as session:
+        stmt = select(SampleTableEntry).where(SampleTableEntry.status == Status.SUBMITTED)
+        results = list(session.scalars(stmt).all())
+    return {
+        row.accession: cast(str, row.result["biosample_accession"]) for row in results if row.result
```
maverbiest, line 30: I'm curious why we need to `cast` the biosample_accession to a string now but we didn't before?

ena-submission/src/ena_deposition/submission_db_helper.py
```diff
@@ -93,31 +87,6 @@ class TableName(StrEnum):
     ASSEMBLY_TABLE = "assembly_table"
     SUBMISSION_TABLE = "submission_table"
```
maverbiest:
```suggestion

```

maverbiest: `TableName` is not used anymore

ena-submission/src/ena_deposition/check_external_visibility.py
maverbiest: I don't think `visibility_checker` can ever be `None`, right? Could we skip this check?

ena-submission/src/ena_deposition/check_external_visibility.py
```diff
@@ -276,13 +262,13 @@ def check_and_update_visibility_for_column(
...
     logger.debug(f"Checking {entity_type.value}.{column_name} for visibility")
-    entities_needing_check = get_entities_needing_column_check(pool, column_config)
+    entities_needing_check = get_entities_needing_column_check(engine, column_config)
     logger.info(
         f"Found {len(entities_needing_check)} {entity_type.value}s needing {column_name} check"
     )
 
     for entity in entities_needing_check:
-        entity_id = asdict(entity.primary_key)
+        entity_id = asdict(entity.pkey)
```
maverbiest, line 268: We could replace all instances of `seq_key = {"accession": row.accession ...}` with this approach

ena-submission/scripts/get_ena_submission_list.py
```diff
@@ -113,7 +113,7 @@ def assign_ena_organism(
 
 def filter_for_submission(
     config: Config,
-    db_pool: SimpleConnectionPool,
+    db_pool: Engine,
```
maverbiest:
```suggestion
    engine: Engine,
```

ena-submission/scripts/get_ena_submission_list.py
```diff
@@ -134,7 +134,7 @@ def filter_for_submission(
     entries_with_external_metadata: set[Accession] = set()
     revoked_entries: set[Accession] = set()
     logger.debug("Querying ENA db for latest version of submissions")
-    highest_submitted_version = highest_version_in_submission_table(db_conn_pool=db_pool)
+    highest_submitted_version = highest_version_in_submission_table(db_pool)
```
maverbiest:
```suggestion
    highest_submitted_version = highest_version_in_submission_table(engine)
```

ena-submission/scripts/get_ena_submission_list.py
maverbiest:
```suggestion
    engine = db_init(
```

maverbiest:
```suggestion
            config, engine, released_entries, input_map[loculus_organism]
```

ena-submission/src/ena_deposition/create_sample.py
```diff
@@ -157,22 +157,26 @@ def construct_sample_set_object(
...
+            ]
         ),
         sample_attributes=SampleAttributes(sample_attribute=sample_attributes),
     )
     return SampleSetType(sample=[sample_type])
 
 
-def set_sample_table_entry(db_config, row, seq_key, config: Config):
+def set_sample_table_entry(
+    db_config: Engine, row: SubmissionTableEntry, seq_key: dict, config: Config
```
maverbiest: Maybe rename `db_config` -> `engine`? (here and elsewhere)

ena-submission/src/ena_deposition/create_sample.py
```diff
@@ -288,14 +289,13 @@ def submission_table_update(db_config: SimpleConnectionPool):
             raise RuntimeError(error_msg)
 
 
-def is_old_version(db_config: SimpleConnectionPool, seq_key: AccessionVersion) -> bool:
+def is_old_version(db_config: Engine, seq_key: AccessionVersion) -> bool:
```
maverbiest: This is almost the same function as `submission_db_helper.is_revision` (although see my comment there). Maybe they could be combined

ena-submission/src/ena_deposition/create_sample.py
```diff
@@ -332,30 +332,28 @@ def sample_table_create(db_config: SimpleConnectionPool, config: Config, test: b
     """
     conditions = {"status": Status.READY}
     ready_to_submit_sample = find_conditions_in_db(
-        db_config, table_name=TableName.SAMPLE_TABLE, conditions=conditions
+        db_config, SampleTableEntry, conditions=conditions
     )
     logger.debug(f"Found {len(ready_to_submit_sample)} entries in sample_table in status READY")
     for row in ready_to_submit_sample:
-        seq_key = AccessionVersion(accession=row["accession"], version=row["version"])
+        seq_key = AccessionVersion(accession=row.accession, version=row.version)
```
maverbiest:
```suggestion
        seq_key = row.pkey
```

ena-submission/src/ena_deposition/create_assembly.py
```diff
@@ -243,37 +243,31 @@ def submission_table_start(db_config: SimpleConnectionPool, config: Config) -> N
...
-        db_config, table_name=TableName.SUBMISSION_TABLE, conditions=conditions
-    )
+    ready_to_submit = find_conditions_in_db(db_config, SubmissionTableEntry, conditions=conditions)
     if len(ready_to_submit) > 0:
         logger.debug(
             f"Found {len(ready_to_submit)} entries in submission_table in status SUBMITTED_SAMPLE"
         )
     for row in ready_to_submit:
-        seq_key = {"accession": row["accession"], "version": row["version"]}
+        seq_key = {"accession": row.accession, "version": row.version}
```
maverbiest: Could use `row.pkey` here as well (and elsewhere in the file)

ena-submission/src/ena_deposition/create_assembly.py
```diff
@@ -299,26 +293,26 @@ def submission_table_update(db_config: SimpleConnectionPool) -> None:
...
     for row in submitting_assembly:
-        seq_key = {"accession": row["accession"], "version": row["version"]}
+        seq_key = {"accession": row.accession, "version": row.version}
 
         corresponding_assembly = find_conditions_in_db(
-            db_config, table_name=TableName.ASSEMBLY_TABLE, conditions=seq_key
+            db_config, AssemblyTableEntry, conditions=seq_key
         )
-        if len(corresponding_assembly) == 1 and corresponding_assembly[0]["status"] == str(
+        if len(corresponding_assembly) == 1 and corresponding_assembly[0].status == str(
```
maverbiest, line 309:
```suggestion
        if len(corresponding_assembly) == 1 and corresponding_assembly[0].status == status.SUBMITTED.value
```

ena-submission/src/ena_deposition/create_assembly.py
```diff
@@ -361,20 +355,20 @@ def can_be_revised(config: Config, db_config: SimpleConnectionPool, entry: dict[
     3. metadata fields in manifest haven't changed since previous version, otherwise
        requires manual revision
     """
-    seq_key = AccessionVersion(accession=entry["accession"], version=entry["version"])
+    seq_key = AccessionVersion(accession=submission_row.accession, version=submission_row.version)
```
maverbiest:
```suggestion
    seq_key = submission_row.pkey
```

ena-submission/src/ena_deposition/create_assembly.py
```diff
@@ -383,33 +377,43 @@ def can_be_revised(config: Config, db_config: SimpleConnectionPool, entry: dict[
...
             error = (
                 "Assembly cannot be revised because biosampleAccession in new version: "
                 f"{new_sample_accession} differs from last version: {previous_sample_accession}"
             )
             logger.error(error)
-            update_assembly_error(db_config, [error], seq_key=entry, update_type="revision")
+            update_assembly_error(
+                db_config,
+                [error],
+                seq_key={"accession": submission_row.accession, "version": submission_row.version},
```
maverbiest:
```suggestion
                seq_key=asdict(seq_key),
```

ena-submission/src/ena_deposition/create_assembly.py
```diff
@@ -383,33 +377,43 @@ def can_be_revised(config: Config, db_config: SimpleConnectionPool, entry: dict[
...
             error = (
                 "Assembly cannot be revised because bioprojectAccession in new version: "
                 f"{new_project_accession} differs from last version: {previous_study_accession}"
             )
             logger.error(error)
-            update_assembly_error(db_config, [error], seq_key=entry, update_type="revision")
+            update_assembly_error(
+                db_config,
+                [error],
+                seq_key={"accession": submission_row.accession, "version": submission_row.version},
```
maverbiest:
```suggestion
                seq_key=asdict(seq_key),
```

ena-submission/src/ena_deposition/create_assembly.py
```diff
@@ -431,30 +435,34 @@ def can_be_revised(config: Config, db_config: SimpleConnectionPool, entry: dict[
             f"last version: {json.dumps(differing_fields)}"
         )
         logger.error(error)
-        update_assembly_error(db_config, [error], seq_key=entry, update_type="revision")
+        update_assembly_error(
+            db_config,
+            [error],
+            seq_key={"accession": submission_row.accession, "version": submission_row.version},
```
maverbiest:
```suggestion
                seq_key=asdict(seq_key),
```

ena-submission/src/ena_deposition/create_assembly.py
```diff
@@ -431,30 +435,34 @@ def can_be_revised(config: Config, db_config: SimpleConnectionPool, entry: dict[
...
     return True
 
 
-def is_flatfile_data_changed(db_config: SimpleConnectionPool, entry: dict[str, Any]) -> bool:
+def is_flatfile_data_changed(db_config: Engine, submission_row: SubmissionTableEntry) -> bool:
     """
     Check if change in sequence or flatfile metadata has occurred since last version.
     """
-    seq_key = AccessionVersion(accession=entry["accession"], version=entry["version"])
+    seq_key = AccessionVersion(accession=submission_row.accession, version=submission_row.version)
```
maverbiest:
```suggestion
    seq_key = submission_row.pkey
```

maverbiest: (and use throughout function instead of `{"accession": ...}`)

ena-submission/src/ena_deposition/create_assembly.py
```diff
@@ -553,31 +567,32 @@ def assembly_table_create(db_config: SimpleConnectionPool, config: Config):
...
-        center_name = sample_data_in_submission_table[0]["center_name"]
+        submission_row = submission_rows[0]
+        center_name = submission_row.center_name
 
         sample_accession, study_accession = get_project_and_sample_results(
-            db_config, sample_data_in_submission_table[0]
+            db_config, submission_row
         )
 
         if is_revision(db_config, seq_key):
```
maverbiest: `can_be_revised` already checks whether the entry `is_revision`, so we could directly do `if can_be_revised(...):`, although the current logic is probably easier to read so fine with leaving as-is

ena-submission/src/ena_deposition/create_assembly.py
```diff
@@ -553,31 +567,32 @@ def assembly_table_create(db_config: SimpleConnectionPool, config: Config):
...
     )
     if len(ready_to_submit_assembly) > 0:
         logger.debug(
             f"Found {len(ready_to_submit_assembly)} entries in assembly_table in status READY"
         )
     for row in ready_to_submit_assembly:
-        seq_key = AccessionVersion(accession=row["accession"], version=row["version"])
-        sample_data_in_submission_table = find_conditions_in_db(
-            db_config, table_name=TableName.SUBMISSION_TABLE, conditions=asdict(seq_key)
+        seq_key = AccessionVersion(accession=row.accession, version=row.version)
```
maverbiest:
```suggestion
        seq_key = row.pkey
```

ena-submission/src/ena_deposition/create_assembly.py
```diff
@@ -633,21 +644,21 @@ def assembly_table_create(db_config: SimpleConnectionPool, config: Config):
...
                 update_values=update_values,
-                table_name=TableName.ASSEMBLY_TABLE,
+                model_class=AssemblyTableEntry,
             )
         else:
             update_assembly_error(
                 db_config,
                 assembly_creation_results.errors,
-                seq_key=row,
+                seq_key={"accession": row.accession, "version": row.version},
```
maverbiest:
```suggestion
                seq_key=asdict(seq_key),
```

ena-submission/src/ena_deposition/create_assembly.py
```diff
@@ -656,31 +667,35 @@ def assembly_table_update(db_config: SimpleConnectionPool, config: Config, time_
...
-    time = datetime.now(tz=pytz.utc)
-    if not _last_ena_check or time - timedelta(minutes=time_threshold) > _last_ena_check:
+    now = datetime.now(tz=pytz.utc)
+    if not _last_ena_check or now - timedelta(minutes=time_threshold) > _last_ena_check:
         logger.debug("Checking state in ENA")
         for row in waiting:
-            seq_key = {"accession": row["accession"], "version": row["version"]}
-            sample_data_in_submission_table = find_conditions_in_db(
-                db_config, table_name=TableName.SUBMISSION_TABLE, conditions=seq_key
+            seq_key = {"accession": row.accession, "version": row.version}
```
maverbiest:
```suggestion
            seq_key = row.pkey
```

ena-submission/src/ena_deposition/create_assembly.py
```diff
@@ -656,31 +667,35 @@ def assembly_table_update(db_config: SimpleConnectionPool, config: Config, time_
...
+            organism = config.enaOrganisms[submission_rows[0].organism]
             # Previous means from the last time the entry was checked, from db
-            previous_result = row["result"]
-            segment_order = previous_result["segment_order"]
+            segment_order = row.result.get("segment_order") if row.result else None
+            erz_accession = row.result.get("erz_accession") if row.result else None
+            if not erz_accession or not segment_order:
+                logger.warning(
+                    f"Missing erz_accession or segment_order for {seq_key['accession']} version "
+                    f"{seq_key['version']} - cannot check ENA for accession yet."
```
maverbiest:
```suggestion
                    f"Missing erz_accession or segment_order for {seq_key.accession} version "
                    f"{seq_key.version} - cannot check ENA for accession yet."
```

maverbiest (commented):
Very cool! I think this move is definitely an improvement :)

I added a bunch of comments, but many are for the same thing in different places, just let me know if anything's unclear. For me, the most important things are:

1. Looking at the helper functions in `submission_db_helper.py`: many of these are only used once, so maybe don't warrant being their own function. On the other hand, I think adding `update_by_pkey()` and `update_by_pkey_with_rertry` that take primary keys may be nice (see point 2)
2. In a lot of places, there is now `seq_key: {"accession": row.accession, "version": row.version}`. IMO, these should be replaced by `row.pkey`, now that this property exists. These could then be cast using `asdict(seq_key)` as needed (this already happens sporadically)
3. It's currently a bit inconsistent that in some places, the `Engine` instance is referred to as `db_config`, and in other places as `engine`. Not a big deal but thought I'd raise it
4. I think the four `add_to_X_table` helper functions can be consolidated
5. It's not directly related to this PR, but I think the current implementations of `is_revision` and `last_version` are a bit unclear. Maybe they could be looked at since this is anyway a larger refactor

ena-submission/src/ena_deposition/api.py
anna-parker: ah because row.result is now properly typed as `result: Mapped[dict[str, str | Sequence[str]]` - so it could be a string or a list[str], hence we need to tell mypy we are correct to use it as a str here now

anna-parker: as a next step we should perhaps type the results dictionary better but I think this is ok for now :-)

ena-submission/src/ena_deposition/submission_db_helper.py
anna-parker: I think its fine to have it here for now :-)

ena-submission/src/ena_deposition/submission_db_helper.py
anna-parker: I would prefer to not inline as this is a larger function and I think it would be good to keep as one function, we could perhaps move it but I kind of like having all db modification code in this file

ena-submission/src/ena_deposition/submission_db_helper.py
anna-parker: same here

ena-submission/src/ena_deposition/submission_db_helper.py
anna-parker: ah yes that sounds like a good next step - but maybe we can do this in a follow up PR :-)

ena-submission/src/ena_deposition/create_project.py
```diff
@@ -105,7 +106,7 @@ def set_project_table_entry(db_engine: Engine, config: Config, row: SubmissionTa
     """Set bioprojectAccession for entry with custom bioprojectAccession"""
     logger.debug(f"Accession {row.accession} already has bioprojectAccession in metadata")
     group_key = {"group_id": row.group_id, "organism": row.organism}
-    seq_key = {"accession": row.accession, "version": row.version}
+    seq_key = asdict(row.pkey)
```
anna-parker: ah yes this is not the pkey of the project

ena-submission/src/ena_deposition/submission_db_helper.py
anna-parker: true! good catch!

ena-submission/src/ena_deposition/submission_db_helper.py
anna-parker: My plan is to rework these functions in a follow up PR to not take untyped dictionaries as input - so I think it is fine to leave this as is for now :-)

ena-submission/src/ena_deposition/submission_db_helper.py
anna-parker: although tbf project does need different logic than sample and assembly

ena-submission/src/ena_deposition/submission_db_helper.py
anna-parker: yeah this is probably not well documented, so the ena submission pod will/should never receive a version to 2 and 3 - it will only receive the latest not yet submitted version, this is why this is a safe check, but probably we should make this more explicit

anna-parker: I improved the naming now a bit and now also explicitly check if sth is a revision but cannot be submitted

ena-submission/src/ena_deposition/create_sample.py
anna-parker: oh excellent catch! I have modified always nice to remove some code!

anna-parker:
Hi @maverbiest! I have done the following:

1. Function structure has stayed the same (we can make changes in a follow up PR)
2. row.pkey is now used consistently across the code base
3. db_engine for Engine objects is now used consistently across the code base
4. I also chose to keep the individual add_to_X_table helper functions as is for now
5. Thanks for catching the fact that is_old_version is basically the same as is_latest_revision! I have modified the code to use this and improved the naming of the revise functions

~Update: I thought I fixed the integration tests but that is not the case - I will investigate by running them locally tomorrow!~ appears to have been flake!

ena-submission/src/ena_deposition/submission_db_helper.py
```diff
@@ -154,348 +116,326 @@ class ProjectId:
...
     )
     return updated_row_count
 
 
-def update_with_retry(
-    db_config: SimpleConnectionPool,
+def update_with_retry[
+    T: (SubmissionTableEntry, ProjectTableEntry, SampleTableEntry, AssemblyTableEntry)
+](
+    db_engine: Engine,
```
maverbiest, line 421:
```suggestion
    engine: Engine,
```

ena-submission/src/ena_deposition/submission_db_helper.py
```diff
@@ -154,348 +116,326 @@ class ProjectId:
...
-    def _do_update():
+    *conditions* and *update_values* map column names to values.
+    They are included in log messages to aid debugging.
+    """
+
+    def _do_update() -> int:
         number_rows_updated = update_db_where_conditions(
-            db_config,
-            table_name=table_name,
+            db_engine,
```
maverbiest, line 435: Maybe `engine` instead of `db_engine`, to be consistent with the other functions

ena-submission/src/ena_deposition/create_sample.py
```diff
@@ -332,30 +303,28 @@ def sample_table_create(db_config: SimpleConnectionPool, config: Config, test: b
...
     ready_to_submit_sample = find_conditions_in_db(
-        db_config, table_name=TableName.SAMPLE_TABLE, conditions=conditions
+        db_engine, SampleTableEntry, conditions=conditions
     )
     logger.debug(f"Found {len(ready_to_submit_sample)} entries in sample_table in status READY")
     for row in ready_to_submit_sample:
-        seq_key = AccessionVersion(accession=row["accession"], version=row["version"])
-        if is_old_version(db_config, seq_key):
+        seq_key = row.pkey
+        if is_revision(db_engine, seq_key) and not is_latest_revision(db_engine, seq_key):
```
maverbiest: This line now does two separate DB hits for the same information (which is requested again on line 340).

Maybe we could replace these two functions into one that tells you both if something is a revision and if it is the latest revision? You'd call it once at the top of each iteration and use the info throughout:
```
  def revision_status(engine: Engine, seq_key: AccessionVersion) -> tuple[bool, bool]:
      if seq_key.version == 1:
          return (False, False)
      rows = find_conditions_in_db(
          engine, SubmissionTableEntry, {"accession": seq_key.accession}
      )
      versions = sorted(row.version for row in rows)
      is_rev = len(versions) > 1
      is_latest = is_rev and seq_key.version == versions[-1]
      return (is_rev, is_latest)
```
(possibly adding a RevisionStatus dataclass for nicer return type)

But maybe it's beyond the scope of this PR

ena-submission/src/ena_deposition/submission_db_helper.py
anna-parker: I used db_engine everywhere now :-) - I think its easier to search for in the code like this and it is a db_engine so think this is good :-)

ena-submission/src/ena_deposition/create_sample.py
anna-parker: ah yes I should probably at least reuse the output of the check instead of repeating the db query again - that is quite a waste! I think having these two functions separate does improve readability however (sth Cornelius always comments in his reviews) so I might keep it as 2 separate functions for now :-)

maverbiest (commented):
Looks great, thank you!

I just noticed that the `[ena-submission-workflow-tests / Tests (pull_request)` action is failing now, is that because of a change or flake? If that gets cleared up I will approve :)

maverbiest (approved):

ena-submission/src/ena_deposition/submission_db_helper.py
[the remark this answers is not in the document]
anna-parker: lets also make these changes in a follow up PR :-) - I think we can actually remove some of the retry code now we are using sqlalchemy

PR #6273, opened 2026-04-13
feat(deposition): remove unnecessary states

anna-parker:
Small refactor, I realized while going over the code that I introduced a bunch of states in the main "submission_table" which tracks the global state of submissions that aren't used or aren't necessary. 

## Unused States

Can be immedietly deleted: 

    HAS_ERRORS_PROJECT = "HAS_ERRORS_PROJECT"
    HAS_ERRORS_ASSEMBLY = "HAS_ERRORS_ASSEMBLY"
    HAS_ERRORS_SAMPLE = "HAS_ERRORS_SAMPLE"

## Unnecessary states:

    SUBMITTING_PROJECT = "SUBMITTING_PROJECT"
    SUBMITTING_SAMPLE = "SUBMITTING_SAMPLE"

I set the `submission_table` state to submitting when starting a project or sample submission, but this is actually unnecessary as the project and sample table have full ownership of the project/sample submission cycle. Thus, this state was only used in the project and sample creation scripts when it was updated.

The state `READY_TO_SUBMIT` is sufficient to trigger the start of the project creation. The project creation will update this state to `SUBMITTED_PROJECT` when the submission is complete. The full state of the submission is tracked in the project table.

The state `SUBMITTED_PROJECT` is sufficient to trigger the start of the sample creation, again the sample_table tracks the state of the sample submission. The sample creation script updates the submission_table state to `SUBMITTED_SAMPLE` when the sample submission is complete.

Note project and sample submission should take seconds, if they fail the project creation and sample creation scripts will send us notifications.

### Migration

`SUBMITTING_PROJECT` and `SUBMITTING_SAMPLE` should only appear in the db table if the sequence submission is erroring. Administrators should check the failing sequences. If the errors need manual intervention they should be handled accordingly.

### PR Checklist


ena-submission/src/ena_deposition/create_sample.py
```diff
@@ -450,7 +405,6 @@ def create_sample(config: Config, stop_event: threading.Event):
             return
         logger.debug("Checking for samples to create")
         submission_table_start(db_config, config=config)
-        submission_table_update(db_config)
 
         sample_table_create(db_config, config, test=config.test)
```
anna-parker, line 409:
```suggestion
        sample_table_create(db_config, config, test=config.test)
        submission_table_start(db_config, config=config)  # update submission_table state after creation
```

ena-submission/src/ena_deposition/create_project.py
```diff
@@ -416,7 +366,6 @@ def create_project(config: Config, stop_event: threading.Event):
             return
         logger.debug("Checking for projects to create")
         submission_table_start(db_config, config)
-        submission_table_update(db_config)
 
         project_table_create(db_config, config, test=config.test)
```
anna-parker, line 370:
```suggestion
        project_table_create(db_config, config, test=config.test)
        submission_table_start(db_config, config)  # update submission_table state after creation
```

ena-submission/src/ena_deposition/create_project.py
```diff
@@ -416,9 +366,9 @@ def create_project(config: Config, stop_event: threading.Event):
...
+        submission_table_start(db_config, config)  # update submission_table state after creation
```
anna-parker: potentially this function could be renamed as sync_state_with_submission_table

maverbiest: This would be nice I think, yes!

maverbiest (commented):
This looks like a nice simplification of the deposition process!

I just wanted to check with you if I understand the flow before I approve since I'm still getting familiar with this code, hope that's okay:

Is it true that the first call to `submission_table_start()` creates an entry in the `sample_table` with the default state of `READY`. Then `sample_table_create()` runs, which sets it to `SUBMITTED` in the `sample_table`, and then the second call to `submission_table_start()` will update the status from `SUBMITTED_PROJECT` to `SUBMITTED_SAMPLE` in the `submission_table`?

(I think renaming `submission_table_start()` would be nice, as you suggested)

maverbiest (approved):
Thank you!

PR #6278, opened 2026-04-14
feat(deposition): add constants for mapping between database keys and Loculus field names for external metadata fields

### LINKED ISSUE #5646: deposition: improve typing

anna-parker:
partially resolves https://github.com/loculus-project/loculus/issues/5646

The ENA deposition pod uses JSONB fields (like the backend). This PR makes the field names CONSTANTS hopefully reducing the number of ways they could be misspelled leading to errors.

### PR Checklist
- [x] Any manual testing that has been done is documented - I ran the integration tests locally to see the state of the entries in the DB

### DIFF
--- a/ena-submission/src/ena_deposition/config.py
+++ b/ena-submission/src/ena_deposition/config.py
@@ -1,6 +1,7 @@
 import logging
 import os
 from dataclasses import field
+from typing import Final
 
 import dotenv
 import yaml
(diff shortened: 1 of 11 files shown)


ena-submission/src/ena_deposition/config.py
```diff
@@ -10,6 +11,18 @@
 
 logger = logging.getLogger(__name__)
 
+# Constants for mapping between database keys and Loculus field names for external metadata fields
```
maverbiest, line 15: I think a setup like
```python
class DB_KEY(StrEnum):
    BIOPROJECT_ACCESSION = auto()
    VERSIONED_NUCCORE_ACCESSION_PREFIX = "insdc_accession_full"
    ...

class LOCULUS_KEY(StrEnum):
    BIOPROJECT_ACCESSION = "bioprojectAccession"
    ...
````
or similar would be nice (and save a bunch of individual imports)

Comment (anna-parker) [reply ena-submission/src/ena_deposition/config.py:15]: I like that!

Comment (anna-parker) [reply ena-submission/src/ena_deposition/config.py:15]: done:-)

File: ena-submission/src/ena_deposition/create_project.py [hunk 2]
```diff
@@ -153,7 +153,7 @@ def set_project_table_entry(db_engine: Engine, config: Config, row: SubmissionTa
     project_table_entry = ProjectTableEntry(
         group_id=row.group_id,
         organism=row.organism,
-        result={"bioproject_accession": bioproject},
+        result={DBKeys.BIOPROJECT_ACCESSION: bioproject},
```
maverbiest, line 156: Here (and other places where we use the enum variant as a key), the `type` of the key in the dict changes from `str` to `DBKeys`, probably won't ever matter for us since you'd have to do a check on `type(key)` for it to make a difference, but just flagging

We could take `DBKeys.BIOPROJECT_ACCESSION.value` if we care

anna-parker: also dont think its a problem - if anything we might be able to use this to our advantage to prevent using the wrong keys :-)

anna-parker: ah but maybe this is actually a potential issue without value for the jsonb parsing... I will ssh into a preview and check if it is needed

anna-parker:
closing in favor of https://github.com/loculus-project/loculus/pull/7221

PR #6282, opened 2026-04-15
feat(website): new docs menu and ensure agreement checkbox uses underline for links

theosanderson:
Preview at https://preview-update-loculus-f47de2.pathoplexus.org

This PR contains two unrelated changes.

- redesigning the docs menu, see https://loculus.slack.com/archives/C061ZQQM3N1/p1776272384904849
- adding something that ensures that when we use custom agreement HTML text links are styled with underline

Apologies for randomly combining them

### DIFF
--- a/website/src/components/Navigation/DocsMenu.tsx
+++ b/website/src/components/Navigation/DocsMenu.tsx
@@ -56,18 +56,21 @@ interface MenuItemProps {
     currentPageUrl: string;
 }
 
-const MenuItem: FC<MenuItemProps> = ({ page, currentPageUrl }) => (
-    <li className='border-b border-gray-200 last:border-0'>
-        <a
-            href={page.url}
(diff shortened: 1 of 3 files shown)


maverbiest (approved):
Looks good! I think it's an improvement to the design, although I also agree with the slack discussion that the 'blueness' is something we could look at at some point.

It's unrelated to this PR, but to me the `How to` section seems a bit awkwardly named; it's inconsistent with the other section in the menu `Concepts`, as well as the lower-level links contained in the section (e.g., How to: Creating an account). Would `Guides` or similar be a better name for it? (might just be personal taste so can of course ignore this)

theosanderson:
Thanks – don't disagree on the menu but it would be outside of this PR anyway as it's on PPX (and I can't promise to personally get to it)

PR #6291, opened 2026-04-17
chore(taxonomy): Refactor taxonomy service

maverbiest:
In the taxonomy service, entries retrieved from the DB are currently represented as `dict[str, str | int | None]`, with data fields being accessed via keys (taxon["common_name"]). This is error prone and it means we don't have nice typing across the service for data fetched from the DB.

This PR adds a `Taxon(BaseModel)` class that represents a DB entry. It implements a `from_row()` method that constructs a Taxon from a DB row and validates the fields. This allows for better typing and nicer access of data fields (e.g., `taxon["common_name"]` -> `taxon.common_name`)

### PR Checklist

### DIFF
--- a/taxonomy/taxonomy_service/test/test_api.py
+++ b/taxonomy/taxonomy_service/test/test_api.py
@@ -133,10 +133,3 @@ class ApiTest(unittest.TestCase):
             response.json()["detail"]
             == f"Unable to find common name for taxon {cellular_organisms['tax_id']}"
         )
-
-    def test_assign_host_categories(self):
-        homo_sapiens = mock_taxa["Homo sapiens"]
-        response = client.get(f"/taxa/{homo_sapiens['tax_id']}/host-categories")
-
-        assert len(response.json()) == 2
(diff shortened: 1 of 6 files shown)


taxonomy/taxonomy_service/test/test_api.py
```diff
@@ -25,9 +25,16 @@
         "tax_id": 9605,
         "common_name": "humans",
         "scientific_name": "Homo",
-        "parent_id": 207598,
+        "parent_id": 9442,  # skipping some levels for testing purposes
         "depth": 30,
     },
+    "Primates": {
```
anna-parker: fine with adding this, but to check my understanding - this isnt being used by the tests right?

anna-parker (approved):

taxonomy/taxonomy_service/test/test_api.py
maverbiest: Ah yea you're right it's not used here. 

It's left over from when I was testing on-the-fly lineage creation for a taxonomy in this branch and wanted a bit of a deeper hierarchy. I can get rid of it since it's not related to this refactor

PR #6292, opened 2026-04-17
feat(deployment): add traefikVersion toggle for v2/v3

theosanderson-agent:
## Summary

Adds a `traefikVersion` Helm value (default `2`) that selects the Middleware CRD `apiVersion` group used by the chart:

- `traefikVersion: 2` → `traefik.containo.us/v1alpha1` (Traefik v2, unchanged default)
- `traefikVersion: 3` → `traefik.io/v1alpha1` (Traefik v3)

The templates compute a single `$traefikApiVersion` variable via a ternary on the value, and every `Middleware` resource in `templates/ingressroute.yaml` and `templates/lapis-ingress.yaml` (compression, redirect-to-https, basic-auth, noindex-robots-header, redirect-www, cors-all-origins, strip-*-prefix, redirect-slash) renders with that group. `values.schema.json` declares `traefikVersion` as an integer with `enum: [2, 3]`.

While in the area, this also changes the keycloak ingress path from `/{+}` to `/` with `pathType: Prefix`. The `/{+}` syntax causes 404s under Traefik v3; `/` achieves the same catch-all behavior using standard Ingress path semantics and works identically under v2.

This lets existing v2 deployments stay on their current defaults and lets operators migrate to Traefik v3 by flipping a single value — no chart fork or template patching required.

## Test plan

- [ ] `helm lint kubernetes/loculus -f kubernetes/loculus/values.yaml` passes
- [ ] `helm template kubernetes/loculus --set traefikVersion=2` emits only `traefik.containo.us/v1alpha1` for Middleware resources
- [ ] `helm template kubernetes/loculus --set traefikVersion=3` emits only `traefik.io/v1alpha1` for Middleware resources
- [ ] Deploy against a Traefik v2 cluster with defaults — website, backend, keycloak, lapis, minio ingresses all route correctly
- [ ] Deploy against a Traefik v3 cluster with `traefikVersion: 3` — same
- [ ] Verify keycloak login flow works under both versions (path `/` no longer 404s on v3)
- [ ] Schema rejects values other than `2` or `3` for `traefikVersion`

### DIFF
--- a/kubernetes/loculus/templates/ingressroute.yaml
+++ b/kubernetes/loculus/templates/ingressroute.yaml
@@ -1,11 +1,12 @@
-apiVersion: traefik.containo.us/v1alpha1
+{{- $traefikApiVersion := ternary "traefik.io/v1alpha1" "traefik.containo.us/v1alpha1" (eq (int $.Values.traefikVersion) 3) -}}
+apiVersion: {{ $traefikApiVersion }}
 kind: Middleware
 metadata:
   name: compression-middleware
 spec:
   compress: {}
 ---
-apiVersion: traefik.containo.us/v1alpha1
+apiVersion: {{ $traefikApiVersion }}
 kind: Middleware
 metadata:
   name: redirect-middleware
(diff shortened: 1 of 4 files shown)


anna-parker (approved):
lgtm! we can try switching to v3 when we migrate to the new aws account

PR #6293, opened 2026-04-17
feat(taxonomy): endpoint to get taxonomic SILO lineage

maverbiest:
This PR adds a new endpoint to the taxonomy service: `POST /silo-lineage`
The endpoint takes a list of taxon ids to keep (in the request body), and returns the NCBI taxonomy with as SILO-compatible YAML:
```sh
'1':
  aliases:
  - 'Taxon 1: root'
  parents: []
'131567':
  aliases:
  - 'Taxon 131567: cellular organisms'
  parents:
  - '1'
'2759':
  aliases:
  - 'Taxon 2759: Eukaryota; eukaryotes'
  parents:
  - '131567'
...
  ```

The enpoint has two query parameters:

```python
        prune: bool = False
                                If False, return a lineage that contains all provided tax_ids,
                                as well as all taxa on the path from the provided tax_ids to the root.
                                If True, return a lineage that only contains the provided tax_ids
                                (and root), but with the overall hierarchy preserved by connecting
                                the children of pruned taxa to the parent taxon.
        allow_large: bool = False
                                If False, return status 413 if the generated yaml file is
                                larger than `LARGE_FILE_THRESHOLD`
```

`LARGE_FILE_THRESHOLD` is set to 2.5MB. The reason we refuse to send back lineages larger than this size by default is that LAPIS/SILO uses Jackson to parse YAML files, which has a limit of 3MB. If we send a lineage file over the limit to SILO it will crash and take down the instance.

## Manual testing
I made POST requests agains this endpoint with curl on the command line using all 449 host organisms we currently have in PPX (across all organisms). With `prune=true`, the resulting yaml is 41KB in size, with `prune=false`, the resulting yaml contains 960 taxa and is 76KB in size.

I also have this branch https://github.com/loculus-project/loculus/pull/6302 where I'm implementing the `silo-importer` side of things. I made a preview off this branch, and SILO pods are healthy and I can use the taxonomic lineage to filter host organisms.

### PR Checklist

### DIFF
--- a/taxonomy/taxonomy_service/README.md
+++ b/taxonomy/taxonomy_service/README.md
@@ -16,7 +16,7 @@ The DB version is configured in `values.yaml` (`taxonomyService.dbVersion`).
 | `GET /` | Health check |
 | `GET /taxa?scientific_name=<string>` | Endpoint used to validate user input. If the input is valid, the details of all associated taxa are returned. Currently only supports validation of scientific names (case-insensitive) through the `scientific_name` query parameter. |
 | `GET /taxa/{tax_id}?find_common_name=<boolean>` |  Endpoint to use once a valid taxon ID is found. Looks up a taxon by NCBI taxon ID and returns it. If find_common_name=true, returns the nearest ancestor (including self) that has a common name. |
-| `POST /silo-lineage?prune=<boolean>&allow_large=<boolean>` |  Return a taxonomy based on the taxa provided in the request body. The taxonomy is returned as a SILO lineage yaml file. |
+| `POST /silo-lineage?prune=<boolean>&allow_large=<boolean>` |  Return a taxonomy based on the taxa provided in the request body (should be provided as {values: [<ids>]}). The taxonomy is returned as a SILO lineage yaml file. |
 
 ## Updating the NCBI database
 
--- a/taxonomy/taxonomy_service/README.md
+++ b/taxonomy/taxonomy_service/README.md
@@ -16,6 +16,7 @@ The DB version is configured in `values.yaml` (`taxonomyService.dbVersion`).
 | `GET /` | Health check |
 | `GET /taxa?scientific_name=<string>` | Endpoint used to validate user input. If the input is valid, the details of all associated taxa are returned. Currently only supports validation of scientific names (case-insensitive) through the `scientific_name` query parameter. |
 | `GET /taxa/{tax_id}?find_common_name=<boolean>` |  Endpoint to use once a valid taxon ID is found. Looks up a taxon by NCBI taxon ID and returns it. If find_common_name=true, returns the nearest ancestor (including self) that has a common name. |
+| `POST /silo-lineage?prune=<boolean>&allow_large=<boolean>` |  Return a taxonomy based on the taxa provided in the request body. The taxonomy is returned as a SILO lineage yaml file. |
 
 ## Updating the NCBI database
 
--- a/taxonomy/taxonomy_service/src/taxonomy_service/api.py
+++ b/taxonomy/taxonomy_service/src/taxonomy_service/api.py
@@ -271,7 +271,10 @@ def get_silo_lineage(
     if missing:
         logger.warning(f"one or more requested taxa don't exist: {sorted(missing)}")
         for m in missing:
-            lineage[str(m)] = {"aliases": [], "parents": []}
+            lineage[str(m)] = {
+                "aliases": [f"Taxon {m}"],
+                "parents": ["1"],  # attach these directly to root node
+            }
 
     return Response(
         content=yaml.dump(lineage, sort_keys=False), media_type="application/yaml"
--- a/taxonomy/taxonomy_service/src/taxonomy_service/api.py
+++ b/taxonomy/taxonomy_service/src/taxonomy_service/api.py
@@ -206,10 +206,13 @@ def get_silo_lineage(
     for child_list in children.values():
         child_list.sort()
 
-    missing = (keep_ids - {1}) - observed
-    logger.warning(f"WARNING: one or more requested taxa don't exist: {missing}")
-
     lineage = build_pruned_lineage(1, keep_ids, children, None)
+    missing = (keep_ids - {1}) - observed
+    if missing:
+        logger.warning(f"WARNING: one or more requested taxa don't exist: {missing}")
+        for m in missing:
+            lineage[m] = {"aliases": [], "parents": []}
+
     return Response(
         content=yaml.dump(lineage, sort_keys=False), media_type="application/yaml"
     )
--- a/taxonomy/taxonomy_service/src/taxonomy_service/api.py
+++ b/taxonomy/taxonomy_service/src/taxonomy_service/api.py
@@ -111,16 +111,16 @@ def fetch_common_name(db_conn: sqlite3.Connection, taxon: Taxon) -> Taxon | None
     return None
 
 
-def get_spanning_tree(db_conn: sqlite3.Connection, tax_ids: set[int]) -> list[Taxon]:
+def get_spanning_tree(
+    db_conn: sqlite3.Connection, tax_ids: set[int]
+) -> list[tuple[int, int]]:
     """Run a recursive CTE against the database to collect the path
     from each taxon in tax_ids to the root node.
 
     Uses UNION to prevent having to evaluate parts of paths shared by
     several input taxa multiple times.
     """
-    keep_ids = tax_ids | {1}  # always include root node
-    placeholders = ",".join("?" * len(keep_ids))
-
+    placeholders = ",".join("?" * len(tax_ids))
     rows = db_conn.execute(
         f"""
       WITH RECURSIVE ancestors AS (
@@ -133,23 +133,19 @@ def get_spanning_tree(db_conn: sqlite3.Connection, tax_ids: set[int]) -> list[Ta
       )
       SELECT * FROM ancestors
     """,
-        list(keep_ids),
+        list(tax_ids),
     ).fetchall()
 
-    return [Taxon.from_row(row) for row in rows]
+    return [(row["tax_id"], row["parent_id"]) for row in rows]
 
 
-def map_child_nodes(tree: list[Taxon]) -> dict[int, list[int]]:
-    """Create a dict where keys are taxon ids and values are lists off
-    all children associated with a taxon
-    """
+def map_child_nodes(tree: list[tuple[int, int]]) -> dict[int, list[int]]:
     children: dict[int, list[int]] = defaultdict(list)
-    for taxon in tree:
-        if taxon.tax_id != taxon.parent_id:
-            children[taxon.parent_id].append(taxon.tax_id)
+    for tax_id, parent_id in tree:
+        if tax_id != parent_id:
+            children[parent_id].append(tax_id)
     for child_list in children.values():
         child_list.sort()
-
     return children
 
 
@@ -226,7 +222,7 @@ def get_silo_lineage(
     payload: SubtreeRequestBody,
     db: DbConnection,
 ) -> Response:
-    taxon_ids = set()
+    taxon_ids = {1}  # always include the root node
     for i in payload.tax_ids:
         try:
             taxon_ids.add(int(i))
--- a/taxonomy/taxonomy_service/src/taxonomy_service/api.py
+++ b/taxonomy/taxonomy_service/src/taxonomy_service/api.py
@@ -165,7 +165,7 @@ def build_silo_lineage(
 
     if node_id in keep_ids:
         taxon = tree[node_id]
-        alias = f"{node_id}; {taxon.scientific_name}"
+        alias = f"Taxon {node_id}: {taxon.scientific_name}"
         if taxon.common_name is not None:
             alias += f"; {taxon.common_name}"
         result[str(node_id)] = {
--- a/taxonomy/taxonomy_service/src/taxonomy_service/api.py
+++ b/taxonomy/taxonomy_service/src/taxonomy_service/api.py
@@ -207,9 +207,12 @@ def get_silo_lineage(
         child_list.sort()
 
     lineage = build_pruned_lineage(1, keep_ids, children, None)
+
     missing = (keep_ids - {1}) - observed
     if missing:
-        logger.warning(f"WARNING: one or more requested taxa don't exist: {missing}")
+        logger.warning(
+            f"WARNING: one or more requested taxa don't exist: {sorted(missing)}"
+        )
         for m in missing:
             lineage[m] = {"aliases": [], "parents": []}
 
--- a/taxonomy/taxonomy_service/src/taxonomy_service/api.py
+++ b/taxonomy/taxonomy_service/src/taxonomy_service/api.py
@@ -111,18 +111,51 @@ def fetch_common_name(db_conn: sqlite3.Connection, taxon: Taxon) -> Taxon | None
     return None
 
 
-def build_pruned_lineage(
+def get_spanning_tree(db_conn: sqlite3.Connection, tax_ids: set[int]):
+    keep_ids = set(tax_ids) | {1}
+    placeholders = ",".join("?" * len(keep_ids))
+
+    rows = db_conn.execute(
+        f"""
+      WITH RECURSIVE ancestors AS (
+          SELECT tax_id, parent_id
+          FROM taxonomy WHERE tax_id IN ({placeholders})
+          UNION
+          SELECT t.tax_id, t.parent_id
+          FROM taxonomy t JOIN ancestors a ON t.tax_id = a.parent_id
+          WHERE t.tax_id != t.parent_id
+      )
+      SELECT * FROM ancestors
+    """,
+        list(keep_ids),
+    ).fetchall()
+
+    return rows
+
+
+def map_child_nodes(tree):
+    children: dict[int, list[int]] = defaultdict(list)
+    for node in tree:
+        if node["tax_id"] != node["parent_id"]:
+            children[node["parent_id"]].append(node["tax_id"])
+    for child_list in children.values():
+        child_list.sort()
+
+    return children
+
+
+def build_silo_lineage(
     node_id: int,
     keep_ids: set[int],
     children: dict[int, list[int]],
     parent_in_output: int | None,
-) -> dict[int, dict]:
-    result: dict[int, dict] = {}
+) -> dict[str, dict]:
+    result: dict[str, dict] = {}
 
     if node_id in keep_ids:
-        result[node_id] = {
+        result[str(node_id)] = {
             "aliases": [],
-            "parents": [parent_in_output] if parent_in_output is not None else [],
+            "parents": [str(parent_in_output)] if parent_in_output is not None else [],
         }
         next_parent = (
             node_id  # this node becomes the nearest-kept-parent for descendants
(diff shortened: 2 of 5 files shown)


taxonomy/taxonomy_service/README.md
```diff
@@ -16,6 +16,7 @@ The DB version is configured in `values.yaml` (`taxonomyService.dbVersion`).
 | `GET /` | Health check |
 | `GET /taxa?scientific_name=<string>` | Endpoint used to validate user input. If the input is valid, the details of all associated taxa are returned. Currently only supports validation of scientific names (case-insensitive) through the `scientific_name` query parameter. |
 | `GET /taxa/{tax_id}?find_common_name=<boolean>` |  Endpoint to use once a valid taxon ID is found. Looks up a taxon by NCBI taxon ID and returns it. If find_common_name=true, returns the nearest ancestor (including self) that has a common name. |
+| `POST /silo-lineage?prune=<boolean>&allow_large=<boolean>` |  Return a taxonomy based on the taxa provided in the request body. The taxonomy is returned as a SILO lineage yaml file. |
```
anna-parker: I assume this is too large to make this a GET request?

anna-parker: I might also add a quick description of the request format :-)

taxonomy/taxonomy_service/src/taxonomy_service/api.py
```diff
@@ -107,6 +114,106 @@ def fetch_common_name(db_conn: sqlite3.Connection, taxon: Taxon) -> Taxon | None
     return None
 
 
+def get_spanning_tree(db_conn: sqlite3.Connection, tax_ids: set[int]) -> list[Taxon]:
```
anna-parker: not too important but might be nice to move these to a new file

taxonomy/taxonomy_service/src/taxonomy_service/api.py
```diff
@@ -1,15 +1,22 @@
...
 
-from taxonomy_service.datatypes import Taxon
+from taxonomy_service.datatypes import SubtreeRequestBody, Taxon
 
 from .config import Config
 
+# LAPIS/SILO uses Jackson to parse YAML files, which by default sets a limit of 3MB
+# If we send a lineage file over the limit to SILO it will crash and take down the
+# instance, so for large files we send back status 413 instead.
+LARGE_FILE_THRESHOLD = 2.5 * 1024 * 1024
```
anna-parker: great catch!

maverbiest: I'm now debating whether it would be nicer to instead have a `?max_size=<float>` query parameter on the endpoint so clients can tell the server themselves what their limit on filesize is?

anna-parker: hmm yeah that might be better - but I think its ok to do it in a follow up PR if we would like to get this merged faster :-)

taxonomy/taxonomy_service/src/taxonomy_service/api.py
```diff
@@ -145,6 +252,57 @@ def get_taxon(
...
+        prune: bool     If False, return a lineage that contains all provided tax_ids,
+                        as well as all taxa on the path from the provided tax_ids to the root.
+                        If True, return a lineage that only contains the provided tax_ids
+                        (and root), but with the overall hierarchy preserved by connecting
+                        the children of pruned taxa to the parent taxon.
+        allow_large: bool
+                        If False, return status 413 if the generated yaml file is
+                        larger than `LARGE_FILE_THRESHOLD`
+    """
+    taxon_ids = {1} | set(payload.tax_ids)  # always include the root node
```
anna-parker: why always include the root node?

taxonomy/taxonomy_service/src/taxonomy_service/api.py
```diff
@@ -107,6 +114,106 @@ def fetch_common_name(db_conn: sqlite3.Connection, taxon: Taxon) -> Taxon | None
...
+    for child_list in children.values():
+        child_list.sort()
+
+    return children
+
+
+def prune_tree(
+    tree: list[Taxon],
+    keep_ids: set[int],
+    root_id: int,
```
anna-parker: instead of always using the root node you could just use the MRCA of the keep_ids

taxonomy/taxonomy_service/src/taxonomy_service/api.py
```diff
@@ -107,6 +114,106 @@ def fetch_common_name(db_conn: sqlite3.Connection, taxon: Taxon) -> Taxon | None
...
+          FROM taxonomy t JOIN ancestors a ON t.tax_id = a.parent_id
+          WHERE t.tax_id != t.parent_id
+      )
+      SELECT * FROM ancestors
+      ORDER BY depth, tax_id
+    """,
+        list(tax_ids),
+    ).fetchall()
+
+    return [Taxon.from_row(row) for row in rows]
```
anna-parker: ChatGPT came up with this cool suggestion to find the most recent common ancestor
```
WITH RECURSIVE ancestors AS (
    -- starting nodes
    SELECT
        tax_id AS original_tax_id,
        taxonomy.*
    FROM taxonomy
    WHERE tax_id IN ({placeholders})

    UNION

    -- walk upward
    SELECT
        a.original_tax_id,
        t.*
    FROM taxonomy t
    JOIN ancestors a
      ON t.tax_id = a.parent_id
    WHERE t.tax_id != t.parent_id
)

SELECT
    tax_id,
    parent_id,
    depth,
    COUNT(DISTINCT original_tax_id) AS matched_starting_nodes
FROM ancestors
GROUP BY tax_id, parent_id, depth
HAVING COUNT(DISTINCT original_tax_id) = {number_of_placeholders}
ORDER BY depth DESC
LIMIT 1;
```

taxonomy/taxonomy_service/README.md
maverbiest: Yea the client has to provide all taxon IDs that it wants the lineage to contain. The worry was that we could run into URL length limits if we did a GET endpoint with taxon IDs in a query param.

In a way POST is not super inappropriate since we are asking to create a new SILO lineage (the server just doesn't store it after creation)

taxonomy/taxonomy_service/src/taxonomy_service/api.py
maverbiest: Just for convenience of having a unified tree :) But yea MRCA might be nicer (if more of a hassle to figure out)

taxonomy/taxonomy_service/src/taxonomy_service/api.py
anna-parker: way to get the tree only up[ to the MRCA:
```
WITH RECURSIVE ancestors AS (
    SELECT
        tax_id AS original_tax_id,
        taxonomy.*
    FROM taxonomy
    WHERE tax_id IN ({placeholders})

    UNION

    SELECT
        a.original_tax_id,
        t.*
    FROM taxonomy t
    JOIN ancestors a
      ON t.tax_id = a.parent_id
    WHERE t.tax_id != t.parent_id
),

mrca AS (
    SELECT tax_id, depth
    FROM ancestors
    GROUP BY tax_id, depth
    HAVING COUNT(DISTINCT original_tax_id) = {number_of_placeholders}
    ORDER BY depth DESC
    LIMIT 1
)

SELECT DISTINCT a.*
FROM ancestors a
JOIN mrca m
  ON a.depth >= m.depth
ORDER BY a.depth, a.tax_id;
```

maverbiest: Ahhh nice, this might be the one!

maverbiest: Although there might be an issue with the last `SELECT. DISTINCT a.*` deduplication, as it doesn't correctly deduplicate across the original rows and the additional rows introduced in the recursive CTE. Claude suggests this?
```
WITH RECURSIVE ancestors AS (
      SELECT
          tax_id AS original_tax_id,
          taxonomy.*
      FROM taxonomy
      WHERE tax_id IN ({placeholders})

      UNION

      SELECT
          a.original_tax_id,
          t.*
      FROM taxonomy t
      JOIN ancestors a ON t.tax_id = a.parent_id
      WHERE t.tax_id != t.parent_id
  ),
  mrca AS (
      SELECT tax_id, depth
      FROM ancestors
      GROUP BY tax_id, depth
      HAVING COUNT(DISTINCT original_tax_id) = {n}
      ORDER BY depth DESC
      LIMIT 1
  )
  SELECT t.*
  FROM taxonomy t
  CROSS JOIN mrca m
  WHERE t.tax_id IN (SELECT tax_id FROM ancestors)
    AND t.depth >= m.depth
  ORDER BY t.depth, t.tax_id;
  ```

Comment (anna-parker) [reply taxonomy/taxonomy_service/src/taxonomy_service/api.py]: ah this is to avoid the extra recursive bookkeeping column, yeah sounds good!

File: taxonomy/taxonomy_service/src/taxonomy_service/api.py [hunk 7]
```diff
@@ -107,6 +114,106 @@ def fetch_common_name(db_conn: sqlite3.Connection, taxon: Taxon) -> Taxon | None
...
+    indexed_by_id: dict[int, Taxon] = {t.tax_id: t for t in tree}
+    children = map_child_nodes(tree)
+    return list(_prune(indexed_by_id, keep_ids, children, root_id, root_id).values())
+
+
+def _prune(
+    tree_by_id: dict[int, Taxon],
+    keep_ids: set[int],
+    children: dict[int, list[int]],
+    node_id: int,
```
anna-parker: ~I would still call this root_id~

Update: sorry only now understood the function, I guess node_id and pruned_ancestor_id would maybe be clearer

taxonomy/taxonomy_service/test/test_api.py
```diff
@@ -126,3 +128,112 @@ def test_get_common_name_not_found(self):
...
+
+    def test_returns_full_ancestry_for_single_taxon(self):
+        homo_sapiens = mock_taxa["Homo sapiens"]
+        response = self._post([homo_sapiens["tax_id"]])
+
+        assert response.status_code == codes.ok
+        lineage = yaml.safe_load(response.text)
+        assert set(lineage.keys()) == {"1", "131567", "9605", "9606"}
+
+    def test_root_always_included_with_no_parents(self):
```
anna-parker: i guess then this test will be obsolete

taxonomy/taxonomy_service/test/test_api.py
```diff
@@ -126,3 +128,112 @@ def test_get_common_name_not_found(self):
...
+        assert lineage["9605"]["aliases"] == ["Taxon 9605: Homo; humans"]
+
+    def test_alias_omits_common_name_when_absent(self):
+        homo_sapiens = mock_taxa["Homo sapiens"]
+        response = self._post([homo_sapiens["tax_id"]])
+
+        lineage = yaml.safe_load(response.text)
+        assert lineage["9606"]["aliases"] == ["Taxon 9606: Homo sapiens"]
+
+    def test_empty_tax_ids_returns_root_only(self):
```
anna-parker: and I guess here it will be an empty set - which I think does make sense

taxonomy/taxonomy_service/src/taxonomy_service/api.py
maverbiest: I added a docstring and renamed the argument to `parent_id_after_prune`, hope it's clearer now!

taxonomy/taxonomy_service/src/taxonomy_service/api.py
maverbiest: I think I'll leave it like this for now: the endpoint is specific for SILO and SILO has this inherent limit of 3MB. If we ever get into custom configuration of SILO instances wrt. the yaml size limit (you can configure this in Jackson, as far as I can tell), we can change it

taxonomy/taxonomy_service/src/taxonomy_service/helpers.py
```diff
@@ -0,0 +1,271 @@
...
+    return existing
+
+
+def get_spanning_tree(
+    db_conn: sqlite3.Connection, tax_ids: set[int]
+) -> tuple[list[Taxon], int]:
+    """Run a recursive CTE against the database to collect the path
+    from each taxon in tax_ids to their most recent common ancestor.
+
+    NOTE: `tax_ids` is assumed to have at least one ID, and all IDs are assumed to exist!
```
maverbiest: The reason I'm not calling `find_existing_ids()` here in this function is that I need the missing ids in the caller to handle them there.

An alternative would be to `find_existing_ids()` here in this function and then return a set of missing IDs, along with the spanning tree and the root ID.

taxonomy/taxonomy_service/src/taxonomy_service/api.py
```diff
@@ -145,6 +82,51 @@ def get_taxon(
...
+        prune: bool     If False, return a lineage that contains all provided tax_ids,
+                        as well as all taxa on the path from the provided tax_ids to the root.
+                        If True, return a lineage that only contains the provided tax_ids
+                        (and root), but with the overall hierarchy preserved by connecting
+                        the children of pruned taxa to the parent taxon.
+        allow_large: bool
+                        If False, return status 413 if the generated yaml file is
+                        larger than `LARGE_FILE_THRESHOLD`
+    """
+    tax_ids = {ROOT_TAX_ID} | set(payload.tax_ids)
```
anna-parker: you already add the ROOT_TAX_ID to the existing nodes in the spanning tree so you dont need to add it here as well

taxonomy/taxonomy_service/src/taxonomy_service/helpers.py
```diff
@@ -0,0 +1,261 @@
...
+
+    return existing
+
+
+def get_spanning_tree(
+    db_conn: sqlite3.Connection, tax_ids: set[int]
+) -> tuple[list[Taxon], set[int]]:
+    """Run a recursive CTE against the database to collect the path
+    from each taxon in tax_ids to the taxonomic root
+    NOTE: root itself is only included in output if it is part of `tax_ids`
```
anna-parker: root shouldnt be needed here - the recursion will stop when it it hits a node without an ancestor which is the root

taxonomy/taxonomy_service/src/taxonomy_service/api.py
anna-parker: update: sorry that was removed but I still dont think it is needed here

taxonomy/taxonomy_service/src/taxonomy_service/helpers.py
```diff
@@ -0,0 +1,261 @@
...
+    placeholders = ",".join("?" * len(existing_ids))
+    rows = db_conn.execute(
+        f"""
+      WITH RECURSIVE ancestors AS (
+          SELECT *
+          FROM taxonomy WHERE tax_id IN ({placeholders})
+          UNION
+          SELECT t.*
+          FROM taxonomy t JOIN ancestors a ON t.tax_id = a.parent_id
+          WHERE t.tax_id != t.parent_id
```
anna-parker: would need to also delete this line
```suggestion
```

taxonomy/taxonomy_service/src/taxonomy_service/helpers.py
anna-parker:
```suggestion
```

taxonomy/taxonomy_service/src/taxonomy_service/api.py
```diff
@@ -145,6 +82,51 @@ def get_taxon(
...
+    if tax_ids == {ROOT_TAX_ID}:
+        return Response(content="{}\n", media_type="application/yaml")
```
anna-parker:
```suggestion
    tax_ids = set(payload.tax_ids)
    if not tax_ids:
        return Response(content="{}\n", media_type="application/yaml")
```

taxonomy/taxonomy_service/src/taxonomy_service/helpers.py
```diff
@@ -0,0 +1,261 @@
...
+) -> list[Taxon]:
+    """Return a list of Taxa pruned to `keep_ids`, rooted at `root_id`.
+    Pruned taxa are skipped; their children reattach to the nearest kept ancestor
+    to preserve the overall hierarchy.
+    """
+    indexed_by_id: dict[int, Taxon] = {t.tax_id: t for t in tree}
+    if root_id not in indexed_by_id:
+        raise ValueError(f"root_id {root_id} does not exist in the provided tree")
+    children = map_child_nodes(tree)
+    return list(_prune(indexed_by_id, keep_ids, children, root_id, root_id).values())
```
anna-parker:
```suggestion
    return list(_prune(indexed_by_id, keep_ids | {root_id}, children, root_id, root_id).values())
```

anna-parker (approved):

taxonomy/taxonomy_service/src/taxonomy_service/helpers.py
maverbiest: The root has itself as ancestor. If we donn't check for the root node, the only reason the recursion stops is because `UNION` stops once it's no longer generating new rows (it deduplicates). That's why I kind of liked the explicit `WHERE t.tax_id != t.parent_id` guardrail as we're guaranteed to stop when we hit the root.

If we ever modified the query to be `UNION ALL` (which doesn't deduplicate) it would get stuck in an infinite loop  without the `WHERE t.tax_id != t.parent_id` check

anna-parker: Im not sure why we would ever switch to UNION ALL so I dont think this is needed, personally I dont find it to logical to have to add the root into this but I wont block

maverbiest: I think we did UNION ALL when we had the MRCA query, for example. But tbf that was a whole different query, not really a modification of this one.

I do agree that it's annoying to inject the root manually. Maybe I'll just warn in the docstring that UNION is needed to prevent infinite recursion and get rid of adding root

anna-parker: If we added back the option for the MRCA I think it would probably be more sth like: https://github.com/loculus-project/loculus/commit/86181dd49fce70bc2e7784cdfcfc7d014e5ac27d which splits up the steps quite neatly and we wouldnt use UNION ALL

anna-parker (approved):
looks good!

taxonomy/taxonomy_service/src/taxonomy_service/api.py
```diff
@@ -145,6 +254,64 @@ def get_taxon(
...
+                        as well as all taxa on the path from the provided tax_ids to the root.
+                        If True, return a lineage that only contains the provided tax_ids
+                        (and root), but with the overall hierarchy preserved by connecting
+                        the children of pruned taxa to the parent taxon.
+        allow_large: bool
+                        If False, return status 413 if the generated yaml file is
+                        larger than `LARGE_FILE_THRESHOLD`
+    """
+    taxon_ids = {1}  # always include the root node
+    for i in payload.tax_ids:
```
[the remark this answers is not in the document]
anna-parker: probably ok

taxonomy/taxonomy_service/src/taxonomy_service/api.py
```diff
@@ -107,6 +116,106 @@ def fetch_common_name(db_conn: sqlite3.Connection, taxon: Taxon) -> Taxon | None
...
+    if current_taxon.tax_id in keep_ids:
+        result[node_id] = current_taxon.model_copy(
+            update={"parent_id": parent_in_output}
+        )
+        next_parent = current_taxon.tax_id
+    else:
+        next_parent = parent_in_output
+
+    for child_id in children.get(node_id, []):
+        result.update(_prune(tree_by_id, keep_ids, children, child_id, next_parent))
```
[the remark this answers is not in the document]
anna-parker: this is ok for now as the depth for the taxonomies shouldnt be this deep

PR #6298, opened 2026-04-21
feat(various): Add read-only mode support for maintenance and archival

theosanderson:
## Description

Implements a read-only mode feature that allows Loculus instances to be put into a maintenance or archival state where write operations are disabled while read access remains available.

### Changes

**Backend:**
- Added `ReadOnlyModeInterceptor` that rejects mutating HTTP requests (POST, PUT, PATCH, DELETE) with a 503 Service Unavailable response when read-only mode is enabled
- Added `ServiceUnavailableException` and corresponding exception handler to return proper error responses
- Integrated the interceptor into the Spring Web configuration
- Added `readOnlyMode` configuration property to `BackendConfig`

**Website:**
- Added `readOnlyMode` configuration property to website config
- Modified `authMiddleware` to redirect users attempting to access login-enforced pages to a read-only notice page instead of Keycloak
- Updated navigation to hide Login and Submit items when read-only mode is active
- Added submission pages disabling middleware trigger for read-only mode
- Created dedicated 503 page with read-only specific messaging
- Added read-only banner message to `BaseLayout`

**Configuration:**
- Added `readOnlyMode` to Helm values schema and default values
- Updated Kubernetes template to pass read-only mode configuration to website
- Added configuration validation in website config spec

### Testing

- Added comprehensive unit tests in `ReadOnlyModeInterceptorTest` covering:
  - All HTTP methods pass when read-only mode is disabled
  - Read methods (GET, HEAD, OPTIONS) pass when read-only mode is enabled
  - Write methods (POST, PUT, PATCH, DELETE) throw `ServiceUnavailableException` with appropriate error message

### Documentation

- Added read-only mode section to Helm chart configuration documentation explaining the feature behavior and use cases

## PR Checklist
- [x] Manual testing: Read-only mode can be enabled via configuration; write requests are rejected with 503; read requests work normally; UI hides login/submission entry points and shows read-only banner.

https://claude.ai/code/session_01ED1KDPF4ZQ5ghobteqs1BN

🚀 Preview: https://claude-add-readonly-mode.loculus.org

### DIFF
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -33,6 +33,7 @@ disablePreprocessing: false
 disableIngest: false
 disableEnaSubmission: true
 disableTaxonomyService: false
+readOnlyMode: false
 website:
   websiteConfig:
     enableLoginNavigationItem: true
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -33,7 +33,7 @@ disablePreprocessing: false
 disableIngest: false
 disableEnaSubmission: true
 disableTaxonomyService: false
-readOnlyMode: false
+readOnlyMode: true # TODO: disable once migration/maintenance is complete
 website:
   websiteConfig:
     enableLoginNavigationItem: true
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -33,7 +33,7 @@ disablePreprocessing: false
 disableIngest: false
 disableEnaSubmission: true
 disableTaxonomyService: false
-readOnlyMode: true # TODO: disable once migration/maintenance is complete
+readOnlyMode: false
 website:
   websiteConfig:
     enableLoginNavigationItem: true
--- a/website/src/layouts/BaseLayout.astro
+++ b/website/src/layouts/BaseLayout.astro
@@ -19,9 +19,12 @@ const {
     bannerMessage: configBannerMessage,
     additionalHeadHTML,
     gitHubMainUrl,
+    readOnlyMode,
 } = websiteConfig;
 const remoteBannerMessage = await getRemoteBannerMessage();
-const bannerMessage = remoteBannerMessage ?? configBannerMessage;
+const bannerMessage = readOnlyMode
+    ? 'This instance is in read-only mode. Submissions and login are disabled.'
+    : (remoteBannerMessage ?? configBannerMessage);
 
 interface Props {
     title: string;
--- a/website/src/layouts/BaseLayout.astro
+++ b/website/src/layouts/BaseLayout.astro
@@ -22,9 +22,10 @@ const {
     readOnlyMode,
 } = websiteConfig;
 const remoteBannerMessage = await getRemoteBannerMessage();
-const bannerMessage = readOnlyMode
+const readOnlyBannerMessage = readOnlyMode
     ? 'This instance is in read-only mode. Submissions and login are disabled.'
-    : (remoteBannerMessage ?? configBannerMessage);
+    : undefined;
+const bannerMessage = remoteBannerMessage ?? configBannerMessage ?? readOnlyBannerMessage;
 
 interface Props {
     title: string;
--- a/website/src/middleware/authMiddleware.ts
+++ b/website/src/middleware/authMiddleware.ts
@@ -5,7 +5,7 @@ import JwksRsa from 'jwks-rsa';
 import { err, ok, ResultAsync } from 'neverthrow';
 import { type BaseClient, type TokenSet } from 'openid-client';
 
-import { getConfiguredOrganisms, getRuntimeConfig } from '../config.ts';
+import { getConfiguredOrganisms, getRuntimeConfig, getWebsiteConfig } from '../config.ts';
 import { getInstanceLogger } from '../logger.ts';
 import { KeycloakClientManager } from '../utils/KeycloakClientManager.ts';
 import { getAuthUrl } from '../utils/getAuthUrl.ts';
@@ -64,6 +64,19 @@ async function getValidTokenAndUserInfoFromParams(context: APIContext, client: B
 export const authMiddleware = defineMiddleware(async (context, next) => {
     let token: TokenCookie | undefined;
     let userInfo;
+
+    if (getWebsiteConfig().readOnlyMode) {
+        const enforceLogin = shouldMiddlewareEnforceLogin(
+            context.url.pathname,
+            getConfiguredOrganisms().map((it) => it.key),
+        );
+        if (enforceLogin) {
+            return context.redirect('/503?service=readonly');
+        }
+        deleteCookie(context);
+        context.locals.session = { isLoggedIn: false };
+        return next();
+    }
 
     const client = await KeycloakClientManager.getClient();
     if (client !== undefined) {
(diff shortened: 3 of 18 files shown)


website/src/layouts/BaseLayout.astro
```diff
@@ -19,9 +19,12 @@ const {
     bannerMessage: configBannerMessage,
     additionalHeadHTML,
     gitHubMainUrl,
+    readOnlyMode,
 } = websiteConfig;
 const remoteBannerMessage = await getRemoteBannerMessage();
-const bannerMessage = remoteBannerMessage ?? configBannerMessage;
+const bannerMessage = readOnlyMode
```
theosanderson: This should use the custom banner message if set, in preference to the read only one

theosanderson: @theosanderson-agent implement pls

theosanderson-agent: Done in eb03f6388 — now `bannerMessage = remoteBannerMessage ?? configBannerMessage ?? readOnlyBannerMessage`, so a custom banner (remote or config) takes precedence over the read-only one.

kubernetes/loculus/values.yaml
```diff
@@ -33,6 +33,7 @@ disablePreprocessing: false
 disableIngest: false
 disableEnaSubmission: true
 disableTaxonomyService: false
+readOnlyMode: true # TODO: disable once migration/maintenance is complete
```
theosanderson: disable

theosanderson: [This will fix the integration tests]

theosanderson:
@theosanderson-agent address the two points below
<details>

website/src/pages/503.astro — submission pages redirect to 404 in read-only mode

submissionPagesDisablingMiddleware rewrites submission URLs to /404 even when read-only mode is the reason. Users get a confusing "not found" page instead of the read-only explanation. Consider redirecting to /503?service=readonly when readOnlyMode is true.



website/src/types/config.ts line 261 — readOnlyMode not optional

readOnlyMode: z.boolean() without .default(false) means any existing config file that doesn't include this key will fail validation. Since the Helm template injects it unconditionally via the | default false filter this is safe for Kubernetes deployments, but local/test configs (like the one in config.spec.ts) have to explicitly add the field. Consider z.boolean().default(false) to be more resilient.

</details>

theosanderson-agent:
Both addressed in 01ceb43 on this branch:

- **`submissionPagesDisablingMiddleware`** now checks `readOnlyMode` before rewriting and routes submission URLs to `/503?service=readonly` instead of `/404`, so users hit the read-only explanation page.
- **`readOnlyMode`** in `website/src/types/config.ts` is now `z.boolean().default(false)`, so configs omitting the key still validate. Tweaked `readTypedConfigFile` to use `<Schema extends z.ZodTypeAny>` so the distinct input/output types a default introduces do not break callers.

`check-types`, `format`, and the full vitest suite (622 tests) all pass locally.

theosanderson:
@theosanderson-agent but wait - won't the submissionPageDisabling thing only come in if submission pages are fully disabled?

theosanderson-agent:
Good catch to check — but the PR extends the registration condition in `website/src/middleware.ts` exactly for this reason:

```ts
if (!(websiteConfig?.enableSubmissionPages ?? false) || (websiteConfig?.readOnlyMode ?? false)) {
    middlewares.push(submissionPagesDisablingMiddleware);
}
```

So the middleware is now added whenever `enableSubmissionPages` is false **OR** `readOnlyMode` is true. That means in read-only mode with submission pages otherwise enabled, the middleware still runs and rewrites submission URLs to `/503?service=readonly`.

theosanderson:
@theosanderson-agent could the auth-pages disabling thing be enough for the readonly bit?

theosanderson-agent:
Not quite, as things stand. `shouldMiddlewareEnforceLogin` only matches `/user`, `/{organism}/user`, and `/{organism}/my_sequences` — submission routes (`/{organism}/submission/*`) are not in that list. So in read-only mode the `authMiddleware` branch would set `isLoggedIn: false` and `next()` through; the submission pages then render their existing `NeedToLogin` component. Functional, but the user sees "You need to be logged in to access the submission portal" while the login link is hidden, which is a confusing dead end.

Two ways to collapse it:

1. **Keep the `submissionPagesDisablingMiddleware` path** (current PR) — explicit 503 with the read-only explanation. Costs one extra middleware registration.
2. **Add `^/{organism}/submission` to `shouldMiddlewareEnforceLogin`** — then the auth redirect alone handles it in read-only mode. But this also changes normal-mode behaviour: anonymous users hitting submission URLs would be sent to Keycloak instead of seeing the in-page `NeedToLogin`. Probably fine (arguably better), but it is a behaviour change outside read-only mode that should be decided on its own merits.

Happy to go with (2) if you want the simplification — let me know.

website/src/middleware/authMiddleware.ts
```diff
@@ -65,6 +65,19 @@ export const authMiddleware = defineMiddleware(async (context, next) => {
     let token: TokenCookie | undefined;
     let userInfo;
 
+    if (getWebsiteConfig().readOnlyMode) {
```
anna-parker, line 68: is this to ensure that users will be logged out - will they also be logged out if they are on the auth page?

website/src/layouts/BaseLayout.astro
```diff
@@ -19,9 +19,13 @@ const {
     bannerMessage: configBannerMessage,
     additionalHeadHTML,
     gitHubMainUrl,
+    readOnlyMode,
 } = websiteConfig;
 const remoteBannerMessage = await getRemoteBannerMessage();
-const bannerMessage = remoteBannerMessage ?? configBannerMessage;
+const readOnlyBannerMessage = readOnlyMode
```
anna-parker, line 25: I dont see this banner on the preview

website/src/middleware/authMiddleware.ts
theosanderson: It means users will be logged out whatever page they are on (I don't have strong feelings that that is especially essential, but it seems reasonable) - the redirect to the 503 bit is essential

website/src/layouts/BaseLayout.astro
theosanderson: If we configure a special banner message, as we have for this preview ("this is a demonstration instance"), then I've made that take priority. That's because in the migration, for example, we don't just want a generic message saying "This instance is in read only mode", we want to say "We are migrating servers, submission and authentication features are not available at this time, this should be completed by X UTC, thank you for your patience" or whatever.

theosanderson:
@theosanderson-agent investigate test failure of Dev server test

theosanderson-agent:
The Dev server test failure was unrelated to this PR. The `config-processor.py` step failed because `https://corneliusroemer.github.io/seqs/artefacts/sars-cov-2/M.fasta` returned **HTTP 503 Service Unavailable** — a transient GitHub Pages outage.

```
ValueError: Problem downloading URL: https://corneliusroemer.github.io/seqs/artefacts/sars-cov-2/M.fasta, Status Code: 503, Reason: Service Unavailable
```

I verified the URL is now reachable (200 OK), the Dev server test has been green on `main` immediately before this PR, and nothing in this PR touches config processing or external URL fetches. I've triggered a rerun of the failed job.

website/src/middleware/authMiddleware.ts
anna-parker: I think what I am curious about is what about the case where a user is editing theri keycloak login information - I assume they will not be signed out there and we wouldnt migrate those changes?

theosanderson: Yeah, interesting point - I mean they wouldn't be able to get to that page through the website or anything - to me it's an edge case we could not worry about for now

theosanderson: (obviously yes, they might have clicked through just before - but it's unlikely)

anna-parker (approved):
approving - but we should switch to readOnly disabled and confirm integration tests before merging :-)

PR #6302, opened 2026-04-23
feat!(silo-import, helm, website): add hierarchical filters for metadata fields

maverbiest:
## BREAKING CHANGES
This PR changes the `hostTaxonId` field from an `int` to a `string`. This is needed to be able to use `hostTaxonId` as a lineage in SILO.

### How to address
Update the type of the `hostTaxonId` metadata field to a string in your `values.yaml`:
```yaml
metadata:
  - name: hostTaxonId
    type: string
...
```

And then reprocess data via pipeline version bumps.

## Summary

Introduces a new "hierarchical filter" concept for metadata fields whose values form a hierarchy resolved dynamically at SILO import time (rather than statically pinned per pipeline version, the way lineageSystem works). The first use case is hostTaxonId: at import time the SILO importer collects the set of host taxon IDs observed in the released data, asks the taxonomy service to render a SILO-format lineage YAML for those taxa, and writes it into SILO's input directory. SILO then indexes the column as a lineage, which gives the website a hierarchical search UI ("Culicidae" matches records of all mosquito species, etc.).

The taxonomy-service endpoint that returns the SILO YAML lives on a separate branch that this PR will be merged into; the changes here cover the helm chart, the SILO importer, the preprocessing flip of hostTaxonId to string, and the website.

## Helm / config

- New metadata-field property hierarchicalFilter: <name>. Mutually exclusive with lineageSystem (enforced in values.schema.json). When set, the field is treated like a lineage system for  SILO indexing (generateLineageIndex) and surfaces a new hierarchicalSearch: true flag in the website config.
- New top-level hierarchicalFilterSources map: <name> → service base URL. Helm template fails if a metadata field references a name that has no source configured.
- _hierarchical-filters-for-organism.tpl mirrors the existing lineageSystemForOrganism helper. The SILO database config combines both lists into lineageDefinitionFilenames.
- silo-deployment.yaml builds and injects the HIERARCHICAL_FILTERS env var (JSON map) consumed by the importer.
- hostTaxonId in config switched from type: int to type: string This is required because SILO lineage-indexed columns are strings and YAML keys in the lineage file are strings; preprocessing changes coerce values to string at the boundary.
- West-nile given hierarchicalFilter: hostTaxon to observe the new feature on previews.
- New section in helm-chart-config.mdx documenting the feature.

SILO import

- HIERARCHICAL_FILTERS env var parsed into dict[HierarchicalFilterKind, HierarchicalFilterConfig] at startup. The HierarchicalFilterKind enum (currently only HOST_TAXON) defines supported kinds. Adding a new HierarchicalFilterKind is intentionally a code change: different external services will need different handlers to get the information needed to build a lineage.
- update_hierarchical_filters dispatches each configured kind to its handler. For HOST_TAXON, update_taxonomic_lineage POSTs the observed taxa to <url>/silo-lineage and writes the
response to <input_dir>/<kind>.yaml. If the service responds 413 (file exceeds SILO's Jackson YAML size limit), the importer retries with prune=true.
- ImporterRunner caches the last observed value set per kind in memory and skips re-posting when the set is unchanged across runs.
- New tests in tests/test_runner.py cover both the happy path and the skip-when-unchanged path; new tests in tests/test_config.py cover env-var parsing.

Frontend

- LineageField gains a mode prop with two values: 'lineage' (existing behavior, unchanged) and 'hierarchical' (new). A MODE_CONFIGS table makes the per-mode behavior diff explicit:

|                           | lineage               | hierarchical      |
  | ------------------------- | --------------------- | ----------------- |
  | defaultIncludeSublineages | false                 | true              |
  | includeZeroCounts         | true                  | false             |
  | showAliasOnly             | false                 | true              |
  | checkBoxLabel             | "include sublineages" | "include subtaxa" |

- AutoCompleteOptions extended with two flags. includeZeroCounts: false skips taxa with no records (so the dropdown only shows what's actually present in the data). showAliasOnly: true displays the human-readable alias (e.g. "Taxon 9606: Homo sapiens; human") instead of the raw tax ID, while keeping the canonical lineage as the option value.
- SearchForm renders LineageField with mode='hierarchical' for fields whose config has hierarchicalSearch: true. 
- New LineageField.spec.tsx cases cover hierarchical mode: default checkbox state, sublineage wildcard appending, alias display, and value resolution from alias back to canonical.

## Screenshots & manual testing
I configured a hierarchical filter for hostTaxonId and set up a preview. On the preview, I tested filtering west-nile sequences using the new hierarchical filter autocomplete. It seems to work as intended, and also the interplay with the existing host name - common / scientific looks reasonable: if I set a filter for the species 'Culex' using host name-scientific, all the options with zero count (birds etc.) no longes show up in the hierarchical filter autocomplete.

Here is the filter looks in action:

grafik

grafik

### PR Checklist

🚀 Preview: https://silo-pruned-tax-lineage.loculus.org

### DIFF
--- a/docs/src/content/docs/reference/helm-chart-config.mdx
+++ b/docs/src/content/docs/reference/helm-chart-config.mdx
@@ -81,6 +81,23 @@ lineageSystemDefinitions:
 
 Note: Lineages should be determined by the preprocessing pipeline and not the user, set the lineage field to `noInput: true` to reflect this or alternatively only allow users to submit valid lineage values by adding an `options` list.
 
+### Hierarchical filter sources
+
+For metadata fields whose values form a hierarchy that's resolved at SILO import
+time (rather than fetched as a static file per pipeline version), set
+`hierarchicalFilter: <name>` on the metadata field and add a corresponding entry under
+`hierarchicalFilterSources`. Each entry maps a name to the base URL of a
+service that provides the SILO lineage file. The importer collects observed
+values for the field, sends them to that service, and writes the returned YAML
+into SILO's input directory as `<name>.yaml`.
+
+```yaml
+hierarchicalFilterSources:
+  hostTaxon: http://loculus-taxonomy-service:5000
+```
+
+<SchemaDocs group='hierarchical-filter-sources' fieldColumnClass='w-48' />
+
 ### Organism (type)
 
 Each organism object has the following fields:
@@ -182,5 +199,6 @@ For [our ingest pipeline](https://github.com/loculus-project/loculus/tree/main/i
 <SchemaDocs group='gene' fieldColumnClass='w-48' />
 
 ### Image spec (type)
+
 
 <SchemaDocs group='image' fieldColumnClass='w-48' />
--- a/docs/src/content/docs/reference/helm-chart-config.mdx
+++ b/docs/src/content/docs/reference/helm-chart-config.mdx
@@ -85,18 +85,10 @@ Note: Lineages should be determined by the preprocessing pipeline and not the us
 
 For metadata fields whose values form a hierarchy that's resolved at SILO import
 time (rather than fetched as a static file per pipeline version), set
-`hierarchicalFilter: <name>` on the metadata field and add a corresponding entry under
-`hierarchicalFilterSources`. Each entry maps a name to the base URL of a
-service that provides the SILO lineage file. The importer collects observed
+`hierarchicalFilter: <URL>`, where URL is the URL of the service that will create the
+hierarchical file for SILO. The importer collects observed
 values for the field, sends them to that service, and writes the returned YAML
-into SILO's input directory as `<name>.yaml`.
-
-```yaml
-hierarchicalFilterSources:
-  hostTaxon: http://loculus-taxonomy-service:5000
-```
-
-<SchemaDocs group='hierarchical-filter-sources' fieldColumnClass='w-48' />
+into SILO's input directory as `<MetadataName>.yaml`.
 
 ### Organism (type)
 
--- a/docs/src/content/docs/reference/helm-chart-config.mdx
+++ b/docs/src/content/docs/reference/helm-chart-config.mdx
@@ -200,5 +200,4 @@ For [our ingest pipeline](https://github.com/loculus-project/loculus/tree/main/i
 
 ### Image spec (type)
 
-
 <SchemaDocs group='image' fieldColumnClass='w-48' />
--- a/kubernetes/loculus/templates/_hierarchical-filters-for-organism.tpl
+++ b/kubernetes/loculus/templates/_hierarchical-filters-for-organism.tpl
@@ -1,10 +1,10 @@
 {{- define "loculus.hierarchicalFiltersForOrganism" -}}
 {{- $organism := . -}}
 {{- $schema := $organism.schema | include "loculus.patchMetadataSchema" | fromYaml }}
-{{- $filters := list }}
+{{- $filters := dict }}
 {{- range $entry := $schema.metadata }}
   {{- if hasKey $entry "hierarchicalFilter" }}
-    {{- $filters = append $filters (dict "name" $entry.hierarchicalFilter "metadataField" $entry.name) }}
+    {{- $_ := set $filters $entry.name $entry.hierarchicalFilter }}
   {{- end }}
 {{- end }}
 {{- $filters | toYaml -}}
--- a/kubernetes/loculus/templates/_hierarchical-filters-for-organism.tpl
+++ b/kubernetes/loculus/templates/_hierarchical-filters-for-organism.tpl
@@ -1,0 +1,11 @@
+{{- define "loculus.hierarchicalFiltersForOrganism" -}}
+{{- $organism := . -}}
+{{- $schema := $organism.schema | include "loculus.patchMetadataSchema" | fromYaml }}
+{{- $filters := list }}
+{{- range $entry := $schema.metadata }}
+  {{- if hasKey $entry "hierarchicalFilter" }}
+    {{- $filters = append $filters (dict "name" $entry.hierarchicalFilter "metadataField" $entry.name) }}
+  {{- end }}
+{{- end }}
+{{- $filters | toYaml -}}
+{{- end }}
--- a/kubernetes/loculus/templates/_siloDatabaseConfig.tpl
+++ b/kubernetes/loculus/templates/_siloDatabaseConfig.tpl
@@ -1,12 +1,13 @@
 {{- define "loculus.siloDatabaseShared" -}}
 {{- $type := default "string" .type -}}
+{{- $lineageName := default .hierarchicalFilter .lineageSystem -}}
 - type: {{ ($type | eq "timestamp") | ternary "int" (($type | eq "authors") | ternary "string" $type) }}
   {{- if .generateIndex }}
   generateIndex: {{ .generateIndex }}
   {{- end }}
-  {{- if .lineageSystem }}
+  {{- if $lineageName }}
   generateIndex: true
-  generateLineageIndex: {{ .lineageSystem }} {{- /* must match the file name in the lineageDefinitionFilenames */}}
+  generateLineageIndex: {{ $lineageName }} {{- /* must match the file name in the lineageDefinitionFilenames */}}
   {{- end }}
 {{- end }}
 
--- a/kubernetes/loculus/templates/_siloDatabaseConfig.tpl
+++ b/kubernetes/loculus/templates/_siloDatabaseConfig.tpl
@@ -1,6 +1,6 @@
 {{- define "loculus.siloDatabaseShared" -}}
 {{- $type := default "string" .type -}}
-{{- $lineageName := default .hierarchicalFilter .lineageSystem -}}
+{{- $lineageName := ternary .name (default .hierarchicalFilter .lineageSystem) (not (empty .hierarchicalFilter)) -}}
 - type: {{ ($type | eq "timestamp") | ternary "int" (($type | eq "authors") | ternary "string" $type) }}
   {{- if .generateIndex }}
   generateIndex: {{ .generateIndex }}
--- a/kubernetes/loculus/templates/_siloDatabaseConfig.tpl
+++ b/kubernetes/loculus/templates/_siloDatabaseConfig.tpl
@@ -1,6 +1,6 @@
 {{- define "loculus.siloDatabaseShared" -}}
 {{- $type := default "string" .type -}}
-{{- $lineageName := ternary .name (default .hierarchicalFilter .lineageSystem) (not (empty .hierarchicalFilter)) -}}
+{{- $lineageName := ternary .name (default "" .lineageSystem) (not (empty .hierarchicalFilter)) -}}
 - type: {{ ($type | eq "timestamp") | ternary "int" (($type | eq "authors") | ternary "string" $type) }}
   {{- if .generateIndex }}
   generateIndex: {{ .generateIndex }}
--- a/kubernetes/loculus/templates/silo-deployment.yaml
+++ b/kubernetes/loculus/templates/silo-deployment.yaml
@@ -84,11 +84,13 @@ spec:
             {{- if $lineageSystem }}
               {{- $defsBySystem := dict -}}
               {{- range $ls := $lineageSystem }}
-                {{- $def := index $.Values.lineageSystemDefinitions $ls -}}
-                {{- if not $def -}}
-                  {{- fail (printf "lineageSystemDefinitions missing entry for lineage system '%s'" $ls) -}}
-                {{- end -}}
-                {{- $defsBySystem = set $defsBySystem $ls $def -}}
+                {{- if ne $ls "host_taxonomy" }} {{- /* host_taxonomy is populated from the taxonomy service at import time, not via lineageSystemDefinitions */ -}}
+                  {{- $def := index $.Values.lineageSystemDefinitions $ls -}}
+                  {{- if not $def -}}
+                    {{- fail (printf "lineageSystemDefinitions missing entry for lineage system '%s'" $ls) -}}
+                  {{- end -}}
+                  {{- $defsBySystem = set $defsBySystem $ls $def -}}
+                {{- end }}
               {{- end }}
             - name: LINEAGE_DEFINITIONS
               value: {{ $defsBySystem | toJson | quote }}
--- a/kubernetes/loculus/templates/silo-deployment.yaml
+++ b/kubernetes/loculus/templates/silo-deployment.yaml
@@ -103,6 +103,8 @@ spec:
               value: "/usr/local/bin/silo"
             - name: PREPROCESSING_CONFIG
               value: "/app/preprocessing_config.yaml"
+            - name: TAXONOMY_SERVICE_URL
+              value: {{ $.Values.siloImport.taxonomyServiceUrl | quote }}
           volumeMounts:
             - name: lapis-silo-database-config-processed
               mountPath: /preprocessing/input/reference_genomes.json
--- a/kubernetes/loculus/templates/silo-deployment.yaml
+++ b/kubernetes/loculus/templates/silo-deployment.yaml
@@ -5,6 +5,7 @@
 {{- $key := $item.key }}
 {{- $organismContent := $item.contents }}
 {{- $lineageSystem := $organismContent | include "loculus.lineageSystemForOrganism" | fromYamlArray }}
+{{- $hierarchicalFilters := $organismContent | include "loculus.hierarchicalFiltersForOrganism" | fromYamlArray }}
 ---
 apiVersion: apps/v1
 kind: Deployment
@@ -84,16 +85,26 @@ spec:
             {{- if $lineageSystem }}
               {{- $defsBySystem := dict -}}
               {{- range $ls := $lineageSystem }}
-                {{- if ne $ls "host_taxonomy" }} {{- /* host_taxonomy is populated from the taxonomy service at import time, not via lineageSystemDefinitions */ -}}
-                  {{- $def := index $.Values.lineageSystemDefinitions $ls -}}
-                  {{- if not $def -}}
-                    {{- fail (printf "lineageSystemDefinitions missing entry for lineage system '%s'" $ls) -}}
-                  {{- end -}}
-                  {{- $defsBySystem = set $defsBySystem $ls $def -}}
-                {{- end }}
+                {{- $def := index $.Values.lineageSystemDefinitions $ls -}}
+                {{- if not $def -}}
+                  {{- fail (printf "lineageSystemDefinitions missing entry for lineage system '%s'" $ls) -}}
+                {{- end -}}
+                {{- $defsBySystem = set $defsBySystem $ls $def -}}
               {{- end }}
             - name: LINEAGE_DEFINITIONS
               value: {{ $defsBySystem | toJson | quote }}
+            {{- end }}
+            {{- if $hierarchicalFilters }}
+              {{- $hfMap := dict -}}
+              {{- range $hf := $hierarchicalFilters }}
+                {{- $url := index $.Values.hierarchicalFilterSources $hf.name -}}
+                {{- if not $url -}}
+                  {{- fail (printf "hierarchicalFilterSources missing entry for filter '%s'" $hf.name) -}}
+                {{- end -}}
+                {{- $hfMap = set $hfMap $hf.name (dict "url" $url "metadataField" $hf.metadataField) -}}
+              {{- end }}
+            - name: HIERARCHICAL_FILTERS
+              value: {{ $hfMap | toJson | quote }}
             {{- end }}
             - name: SILO_RUN_TIMEOUT_SECONDS
               value: {{ $.Values.siloImport.siloTimeoutSeconds | quote }}
@@ -105,8 +116,6 @@ spec:
               value: "/usr/local/bin/silo"
             - name: PREPROCESSING_CONFIG
               value: "/app/preprocessing_config.yaml"
-            - name: TAXONOMY_SERVICE_URL
-              value: {{ $.Values.siloImport.taxonomyServiceUrl | quote }}
           volumeMounts:
             - name: lapis-silo-database-config-processed
               mountPath: /preprocessing/input/reference_genomes.json
--- a/kubernetes/loculus/templates/silo-deployment.yaml
+++ b/kubernetes/loculus/templates/silo-deployment.yaml
@@ -5,7 +5,7 @@
 {{- $key := $item.key }}
 {{- $organismContent := $item.contents }}
 {{- $lineageSystem := $organismContent | include "loculus.lineageSystemForOrganism" | fromYamlArray }}
-{{- $hierarchicalFilters := $organismContent | include "loculus.hierarchicalFiltersForOrganism" | fromYamlArray }}
+{{- $hierarchicalFilters := $organismContent | include "loculus.hierarchicalFiltersForOrganism" | fromYaml }}
 ---
 apiVersion: apps/v1
 kind: Deployment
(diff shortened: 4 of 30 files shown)


theosanderson:
(nit: I'd title the PR `feat(silo-import)` - SILO is a different piece of software [as I know you know :) )

loculus-silo/src/silo_import/lineage.py
```diff
@@ -44,12 +45,71 @@ def update_lineage_definitions(
...
+        response = _post_taxonomic_lineage(url, sorted_taxa)
+        if response.status_code == codes.request_entity_too_large:
+            logger.warning(
+                "Unpruned %s lineage exceeds size threshold; retrying with prune=true", file_base
+            )
+            response = _post_taxonomic_lineage(url, sorted_taxa, prune=True)
+        response.raise_for_status()
+        _write_text(destination, response.text)
+    except requests.RequestException as exc:
+        msg = f"Failed to fetch host taxonomy from {url}: {exc}"
```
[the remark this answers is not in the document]
maverbiest: This handler is specifically for getting a taxonomic lineage yaml from the taxonomy service... it is not "called for any filter kind"

loculus-silo/src/silo_import/runner.py
```diff
@@ -79,7 +80,21 @@ def run_once(self) -> None:
...
+                if new_values.get(kind, set()) != self.hierarchical_filter_values.get(kind, set())
+            }
+            if filters_with_new_values:
+                update_hierarchical_filters(
+                    new_values,
+                    self.config,
+                    self.paths,
+                    subset=filters_with_new_values,
+                )
+                self.hierarchical_filter_values = new_values
```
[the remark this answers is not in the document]
maverbiest: This won't ever be relevant. Even if there are no values for a filterkind, it will still be present in the `new_values` as `{kind: set()}`. The only way for a kind to be dropped from the dict is if the config changes, which will never happen for a running SILO importer instance.

docs/src/content/docs/reference/helm-chart-config.mdx
```diff
@@ -81,6 +81,23 @@ lineageSystemDefinitions:
...
+service that provides the SILO lineage file. The importer collects observed
+values for the field, sends them to that service, and writes the returned YAML
+into SILO's input directory as `<name>.yaml`.
+
+```yaml
+hierarchicalFilterSources:
+  hostTaxon: http://loculus-taxonomy-service:5000
+```
+
+<SchemaDocs group='hierarchical-filter-sources' fieldColumnClass='w-48' />
```
corneliusroemer: I'm not sure I get why we need the separate top level `hierarchicalFilterSources` when one can put the source just directly onto the hierarchicalFilter itself.

The extra name seems to just add extra indirection and config complexity.

What about this instead:


```suggestion
For metadata fields whose values form a hierarchy that's resolved at SILO import
time (rather than fetched as a static file per pipeline version), set
`hierarchicalFilter: <URL>`. The URL points at where the SILO lineage file is provided. 

The importer collects observed values for the field, sends them to that service, and writes the returned YAML
into SILO's input directory as `<metadataFieldName>.yaml`.

```yaml
hierarchicalFilterSources:
  hostTaxon: http://loculus-taxonomy-service:5000
```
```

File: kubernetes/loculus/templates/_hierarchical-filters-for-organism.tpl
Comment (corneliusroemer) [inline kubernetes/loculus/templates/_hierarchical-filters-for-organism.tpl:1]: Here you'd then output a dictionary `metadataFieldName -> fileURL` instead of the list of arbitrary names

File: kubernetes/loculus/templates/silo-deployment.yaml [hunk 4]
```diff
@@ -93,6 +94,18 @@ spec:
...
             {{- end }}
+            {{- if $hierarchicalFilters }}
+              {{- $hfMap := dict -}}
+              {{- range $hf := $hierarchicalFilters }}
+                {{- $url := index $.Values.hierarchicalFilterSources $hf.name -}}
+                {{- if not $url -}}
+                  {{- fail (printf "hierarchicalFilterSources missing entry for filter '%s'" $hf.name) -}}
+                {{- end -}}
+                {{- $hfMap = set $hfMap $hf.name (dict "url" $url "metadataField" $hf.metadataField) -}}
+              {{- end }}
```
corneliusroemer: And here you would already get the map that you need and not have to construct it - seems simpler in all regards to me: less code, less config, fewer chances to mess up config

loculus-silo/src/silo_import/config.py
```diff
@@ -1,11 +1,27 @@
...
 from dataclasses import dataclass
 from pathlib import Path
 
 
+class HierarchicalFilterKind(enum.StrEnum):
+    """If we add new hierarchical filters, they must be
+    registered as enum variants here, as well as in `lineage.update_hierarchical_filters`
+    """
+
+    HOST_TAXON = "hostTaxon"
```
corneliusroemer: Why do we hardcode taxonomy as a type of filter? I thought the feature was general for any hierarchy?

kubernetes/loculus/values.yaml
```diff
@@ -1235,7 +1237,7 @@ defaultOrganismConfig: &defaultOrganismConfig
           displayGroup: lengthCompleteness
           label: Length
       - name: hostTaxonId
-        type: int
+        type: string
```
corneliusroemer, line 1238: Why switch to string?

loculus-silo/src/silo_import/config.py
corneliusroemer, line 1: The diff here is rather large, a lot of defensive programming that might not be worth the extra lines to wade through in the future

docs/src/content/docs/reference/helm-chart-config.mdx
anna-parker: @corneliusroemer in your example `hierarchicalFilterSources:` can be deleted

website/src/components/SearchPage/fields/LineageField.tsx
```diff
@@ -4,12 +4,37 @@ import { SingleChoiceAutoCompleteField } from './SingleChoiceAutoCompleteField';
...
+}
+
+const MODE_CONFIGS: Record<LineageFieldMode, ModeConfig> = {
+    lineage: {
+        defaultIncludeSublineages: false,
+        includeZeroCounts: true,
+        showAlias: false,
+        checkBoxLabel: 'include sublineages',
+    },
+    hierarchical: {
```
anna-parker: so here hierarchical is actually hardcoding taxonomy

website/src/components/SearchPage/SearchForm.tsx
```diff
@@ -439,6 +439,18 @@ const SearchField = ({ field, lapisUrl, fieldValues, setSomeFieldValues, lapisSe
                     />
                 );
             }
+            if (field.hierarchicalSearch) {
+                return (
+                    <LineageField
```
anna-parker: I would have probably refactored this slightly differently - lineage and taxon are a subtype of hierarchical fields - we should call this HierarchicalField and then create the subtypes lineage and taxonomy

loculus-silo/src/silo_import/lineage.py
corneliusroemer, line 1: I wonder if we can't just keep it general by doing the tree pruning here?

Then we don't need any specialization for taxonomy. All this `HierarchicalFilterKind` goes away.

website/src/components/SearchPage/fields/AutoCompleteOptions.ts
```diff
@@ -26,6 +26,15 @@ type LineageOptionsProvider = {
...
     includeSublineages: boolean;
+    /**
+     * When true, show only the alias for a lineage (when one exists) instead of the canonical
+     * name. Used by hierarchical filters where the alias is the user-facing label.
+     */
+    showAlias?: boolean;
+    /**
+     * When false, only show options with a count higher than zero
+     */
+    includeZeroCounts?: boolean;
```
anna-parker, line 37: maybe we can ask about this in a channel but I actually think we should just change the code to always not show when teh count is zero

kubernetes/loculus/values.yaml
maverbiest: SILO needs it to be a string to build a lineage index on

loculus-silo/src/silo_import/runner.py
```diff
@@ -79,7 +80,30 @@ def run_once(self) -> None:
...
+            }
+            if filters_to_update:
+                update_hierarchical_filters(
+                    new_values,
+                    self.config,
+                    self.paths,
+                    subset=filters_to_update,
+                )
+                for kind in filters_to_update:
+                    self.hierarchical_filter_values[kind] = new_values.get(kind, set())
```
corneliusroemer: This seems to be doing caching of the results - is this code path expensive enough that caching is worth the extra complexity? If we save less than a second I'd say no.

website/src/components/SearchPage/fields/AutoCompleteOptions.ts
anna-parker: Ok from https://loculus.slack.com/archives/C06JCAZLG14/p1778154037130259 lets make this configurable just make sure to add (0) when things are zero to be clear

preprocessing/nextclade/tests/test_host_name_validation.py
```diff
@@ -66,7 +66,7 @@ def test_host_processing_direct_submission(mock_session: MagicMock) -> None:
     result = process_all([entry], "temp", config)
     metadata = result[0].processed_entry.data.metadata
 
-    assert metadata["hostTaxonId"] == 7159
+    assert metadata["hostTaxonId"] == "7159"
```
corneliusroemer, line 69: Do I understand correctly that SILO can only handle lineage hierarchies for strings hence we need to make taxon id a string?

maverbiest: yep that's it. But to be fair I don't hate having a string since it's an identifier and you're not going to do arithmetic with it or anything

docs/src/content/docs/reference/helm-chart-config.mdx
maverbiest: I'm definitely open to another approach here.

~My reasoning for not just doing `hierarchicalFilter: <URL>` on the metadata field is that there's a specific request that needs to be made to the taxonomy service in order to get the lineage (POST /silo-lineage with request body {tax_ids: [...]}). I thought having a keyword to match on was nicer than parsing the url itself. If we add for example a geographical filter the lineage will come from somewhere else with probably some different handling logic~

Got it - we just check the metadatafield name that the filter is defined on

website/src/components/SearchPage/SearchForm.tsx
maverbiest: Ahhh yea I agree that's nicer

loculus-silo/src/silo_import/lineage.py
maverbiest: I'm not sure I follow; do you mean building the lineage here in silo-import itself? Because for that we'd need to duplicate the taxonomy DB for each of these pods. We decided against this since there's one pod per organism and we already have the taxonomy service

website/src/components/SearchPage/fields/AutoCompleteOptions.ts
maverbiest: Sounds good! I think I'll just open a separate PR for that then since it's not specific to this feature

loculus-silo/src/silo_import/config.py
anna-parker: I agree I dont think `HierarchicalFilterKind` is required

maverbiest: ~What I'm not fully clear on is how we'd determine which kind of hierarchical filter we should be handling if we get rid of `HierarchicalFilterKind`.~

~Something like `if "taxonomy-service/silo-lineage" in url: call out to taxonomy service`?~

website/src/components/SearchPage/fields/LineageField.tsx
corneliusroemer: Should likely be called `taxonomy` if it's specialized to taxons which it is per `checkBoxLabel` being `include subtaxa`

loculus-silo/src/silo_import/runner.py
corneliusroemer: I think the caching here could be done with a lot less code by using a decorator on `update_hierarchical_filters`

corneliusroemer: the caching should definitely be pushed into update_hierarchical_filtes and not in the high level "controller"/"main" function here

kubernetes/loculus/values.yaml
```diff
@@ -1632,7 +1632,9 @@ defaultOrganisms:
               # - Identifier is the second to last field (extracted through named capture group)
               regex_pattern: '^(?:[^/]+/)?[^/]+/(?P<identifier>[^/]+)/\d{4}(?:-\d{2}){0,2}$'
         - name: hostTaxonId
-          type: int
+          type: string
+          hierarchicalFilter: *taxonomy_service_url
```
corneliusroemer, line 1658: This is mostly unchanged from the default organism, might be clearer and more succinct to use yaml anchor here and only override the few changed fields - `require: true` - this predates this PR

loculus-silo/src/silo_import/lineage.py
anna-parker: I think its ok to keep this in the taxonomy pod for now - otherwise we would still need to define the tree format that is sent to loculus-silo which can then be pruned, this way we use the silo lineage's format which is I guess "defined somewhat"

kubernetes/loculus/templates/_hierarchical-filters-for-organism.tpl
anna-parker: done

loculus-silo/src/silo_import/config.py
anna-parker: think this is now more streamlined

anna-parker (commented):
I think this looking really good! I have a few comments before we merge: 
1. As this adds more work to the silo-importer can you quickly check it is not much **slower** now?
2. In future we would like to **enable incremental upload**s in SILO, i.e. have silo-importer send silo a diff. I think we can extend this to the hierarchical lineages as the importer caches all the values seen and can thus update the values using a diff and request a new lineages file if the values have changed - but let me know if you think there are some potential issues with this
3. Potential blocker: **How does this work when the taxonomy hierarchy changes**? 
 - I know we download and store the taxonomy in a sqllite file. What happens when this is updated? Do we always use the most recent version? We should switch to only using a newer version manually if that is the case. We will need to add a slack notification for this case. 
 - I think the idea was that we would reprocess all the entries when the taxonomy changes by just using the new taxonomy hierarchy and increasing the pipeline version. However, there is then an intermediate period where new entries will use the new taxonomy and the silo importer will call the new taxonomy service and get a new hierarchy, although not all data has yet been processed with the new hierarchies. How should we handle this? Can we have two different taxonomy services running and in the config say which pipeline version uses which one? -> I am fine with changing this in future as I dont think the taxonomy hierarchy will change a lot as long as we ensure the taxonomy hierarchy cant update automatically and cause issues for us.

anna-parker:
Ok! we are just using the latest image for the taxonomy, we should fix this to the current date before rolling out these changes and add a slack notification when it is updated

maverbiest:
Thanks for the feedback @anna-parker! Regarding your questions:

1. I don't have any hard numbers, sorry. I wanted to check in the logs how long a download -> update SILO cycle takes, but no data is coming up on the preview (seems to be the ingest change). But briefly, the additional work we're doing now is:
             I) POST request to taxonomy service. This should take well < 1 second, and only happens if the values change (which will be very rare, I expect)
             II) Every 30 seconds, compare old_metadata_values to new_values, which is negligable
             III) SILO needs to build an additional lineage index for the hierarchical filter field. Here I'm unsure how long it takes, but I can't imagine it will kill our performance to do this over at most a few hundred distinct values.

2. Here I'd need to dive a bit more into how SILO works. You are right in that since the silo-importer keeps track of which values exist, we should be able to tell when there is a new value and then request a new lineage. What I don't know is how a running SILO instance handles it if we update a lineage file while it's running. Maybe that's already what we're doing now and it's not an issue, but I'm a bit ignorant on the inner workings of SILO. Something to look into!
 
3. Yes as you already mentioned, we currently take the latest version of the taxonomy DB. If the taxonomy changes in a way that affects the organisms in a running SILO instance, this will be an issue. So yea fully agree that we should pin the DB version before we merge. I think your sense that we would temporarily need two taxonomy-services (one old, one new) if we want to update is correct (or maybe one service with two version of the DB?). I don't have an immediate intuition for how we'd easily manage this, but for sure something we need to keep in mind.

anna-parker:
@maverbiest it probably makes sense to do a small regression test locally to see if the time to run silo import increases, not perfect but will flag any giant issues: https://github.com/loculus-project/loculus/pull/5764, from your response Im fairly confident my second point is resolved. For point number 3 we should pin the taxonomy image with this PR and then it should be safe to rollout - we can then make a PR to track updating the taxonomy in future.

maverbiest:
> @maverbiest it probably makes sense to do a small regression test locally to see if the time to run silo import increases, not perfect but will flag any giant issues

@anna-parker  does this tell us what we need? I realize I'm not actually interacting with a taxonomy-service here, but the base performance seems the same for mpox as in the PR you linked.

```sh
(.venv) ➜  loculus-silo git:(silo-pruned-tax-lineage) docker run --rm -it --entrypoint bash silo-importer
root@2a12ca7fbae1:/app# BACKEND_BASE_URL=https://backend.pathoplexus.org/mpox ROOT_DIR=. silo-import
2026-05-08 15:05:56,811 [INFO] silo_import.__main__: Importer configuration: ImporterConfig(backend_base_url='https://backend.pathoplexus.org/mpox', lineage_definitions=None, hard_refresh_interval=3600, poll_interval=30, silo_run_timeout=3600, root_dir=PosixPath('/app'), silo_binary=PosixPath('/usr/local/bin/silo'), preprocessing_config=PosixPath('/app/preprocessing_config.yaml'), hierarchical_filters=None)
2026-05-08 15:05:56,812 [INFO] silo_import.runner: Hard refresh triggered: time_since_last=1778252756.8s, interval=3600s,last_refresh_time=0
2026-05-08 15:05:56,812 [INFO] silo_import.download_manager: Starting download from backend with ETag: 0
2026-05-08 15:05:56,812 [INFO] silo_import.download_manager: Requesting released data from https://backend.pathoplexus.org/mpox/get-released-data?compression=zstd
2026-05-08 15:07:54,234 [INFO] silo_import.decompressor: Starting analyze_and_transform_ndjson
2026-05-08 15:08:00,578 [INFO] silo_import.download_manager: Downloaded 15565 records (ETag "2026-05-08Z14:11:25.827681")
2026-05-08 15:08:10,570 [INFO] silo_import.download_manager: Actual record count matches expected record count
2026-05-08 15:08:10,570 [INFO] silo_import.lineage: LINEAGE_DEFINITIONS not provided; skipping lineage configuration
2026-05-08 15:08:10,579 [INFO] silo_import.instruct_silo: Starting SILO preprocessing
```

anna-parker:
to test the affect of having the taxonomy service you will have to run this on an organism/example with a taxonomy service - for example the one in this PR

maverbiest:
@anna-parker 50ms to fetch the taxonomy lineage file and serialize it to `hostTaxonIds.yaml`

```sh
root@8232fe0ec1dc:/app# BACKEND_BASE_URL=https://backend-silo-pruned-tax-lineage.loculus.org/west-nile HIERARCHICAL_FILTERS='{"hostTaxonId":"http://host.docker.internal:5000"}'  ROOT_DIR=. python -m silo_import
2026-05-11 09:12:04,839 [INFO] __main__: Importer configuration: ImporterConfig(backend_base_url='https://backend-silo-pruned-tax-lineage.loculus.org/west-nile', lineage_definitions=None, hard_refresh_interval=3600, poll_interval=30, silo_run_timeout=3600, root_dir=PosixPath('/app'), silo_binary=PosixPath('/usr/local/bin/silo'), preprocessing_config=PosixPath('/app/preprocessing_config.yaml'), hierarchical_filters={'hostTaxonId': 'http://host.docker.internal:5000'})
2026-05-11 09:12:04,839 [INFO] silo_import.runner: Hard refresh triggered: time_since_last=1778490724.8s, interval=3600s,last_refresh_time=0
2026-05-11 09:12:04,839 [INFO] silo_import.download_manager: Starting download from backend with ETag: 0
2026-05-11 09:12:04,840 [INFO] silo_import.download_manager: Requesting released data from https://backend-silo-pruned-tax-lineage.loculus.org/west-nile/get-released-data?compression=zstd
2026-05-11 09:12:12,131 [INFO] silo_import.decompressor: Starting analyze_and_transform_ndjson
2026-05-11 09:12:12,410 [INFO] silo_import.download_manager: Downloaded 9029 records (ETag "2026-05-11Z08:57:42.90666")
2026-05-11 09:12:12,851 [INFO] silo_import.download_manager: Actual record count matches expected record count
2026-05-11 09:12:12,851 [INFO] silo_import.lineage: LINEAGE_DEFINITIONS not provided; skipping lineage configuration
2026-05-11 09:12:12,851 [INFO] silo_import.lineage: Fetching hostTaxonId hierarchy over 304 values
2026-05-11 09:12:12,901 [INFO] silo_import.instruct_silo: Starting SILO preprocessing
2026-05-11 09:12:12,908 [ERROR] silo_import.instruct_silo: SILO preprocessing stderr:
2026-05-11 09:12:12,908 [ERROR] silo_import.instruct_silo: SILO preprocessing stdout: [2026-05-11 09:12:12.903] [logger] [info] [main.cpp:107] Starting SILO (version 0.11.0)
[2026-05-11 09:12:12.903] [logger] [info] [main.cpp:129] preprocessing_config = {"initialization_files":{"database_config_file":"database_config.yaml","directory":"/preprocessing/input","lineage_definition_files":["hostTaxonId.yaml"],"phylogenetic_tree_file":null,"reference_genome_file":"reference_genomes.json"},"input_file":"data.ndjson.zst"}
[2026-05-11 09:12:12.903] [logger] [info] [preprocessing.cpp:15] preprocessing - initializing Database
[2026-05-11 09:12:12.907] [logger] [info] [database_config.cpp:219] Reading database config from /preprocessing/input/database_config.yaml
[2026-05-11 09:12:12.907] [logger] [info] [reference_genomes.cpp:156] Read reference genomes from file: /preprocessing/input/reference_genomes.json
```

anna-parker:
Great stuff! I will take a quick look now!

kubernetes/loculus/templates/_siloDatabaseConfig.tpl
```diff
@@ -1,12 +1,13 @@
 {{- define "loculus.siloDatabaseShared" -}}
 {{- $type := default "string" .type -}}
+{{- $lineageName := ternary .name (default .hierarchicalFilter .lineageSystem) (not (empty .hierarchicalFilter)) -}}
```
anna-parker:
```suggestion
{{- $lineageName := ternary .name (default "" .lineageSystem) (not (empty .hierarchicalFilter)) -}}
```

loculus-silo/src/silo_import/decompressor.py
```diff
@@ -17,34 +19,53 @@ class NdjsonAnalysis:
...
     """
     Decompress and analyze a zstd-compressed NDJSON file.
 
     Args:
         path: Path to the compressed NDJSON file
+        hierarchical_filters: Map of hierarchical metadata field names to service URLs
 
     Returns:
-        NdjsonAnalysis with record count and pipeline versions found
+        NdjsonAnalysis with record count, pipeline version, and per-kind
```
anna-parker:
```suggestion
        NdjsonAnalysis with record count, pipeline version, and
```

loculus-silo/tests/test_runner.py
```diff
@@ -25,6 +27,7 @@ def make_config(
     tmp_path: Path,
     lineage_definitions: dict[str, dict[int, str]] | None = None,
     hard_refresh_interval: int = 1,
+    hierarchical_filters: dict[MetadataField, str] | None = None,
```
anna-parker: technically its HierarchicalServiceUrl but not too important
```suggestion
    hierarchical_filters: dict[MetadataField, HierarchicalServiceUrl] | None = None,
```

anna-parker (approved):
Looks good! Can you make sure we have an issue to track **How should reprocessing and silo import handle the case when the taxonomy hierarchy needs to be updated?** and also an issue to add slack notifications if the taxon hierarchy has changed when we download it?

kubernetes/loculus/templates/_siloDatabaseConfig.tpl
maverbiest: ah yeah good catch

theosanderson:
@theosanderson-agent create a PR to revert this commit

theosanderson-agent:
Opened revert PR: https://github.com/loculus-project/loculus/pull/6424

Notes:
- Created from `main` at `2ebde668a`, reverting merge commit `dbda10cd4`.
- One conflict in `loculus-silo/src/silo_import/lineage.py` from the follow-up typing fix in #6404; resolved by dropping `_post_silo_lineage` (and its now-unused `from typing import Any` import) entirely, which effectively also reverts #6404.
- Restores `hostTaxonId` to `int` — anyone who already migrated `values.yaml` to `string` will need to revert that.

PR #6304, opened 2026-04-27
feat!(website, backend): SeqSet citation tracking

tombch:
### Breaking Changes
- This PR is marked as breaking due to removal of the `get-seqset-cited-by-publication` endpoint, as it is replaced with the endpoint `get-seqset-citations`. The former endpoint used the `CitedBy` return type, and the new one uses `List<SeqSetCitation>`. 

### Schema Changes

- Adds a `seqset_citation_source` table: a 'citation source' being any publication (or possibly other item) that references one or more SeqSets. A citation source must have a DOI, title, year and contributors, and can have an origin either from Crossref or manually curated (manual curation endpoints will be added in a follow-up PR).
- Adds a `seqset_to_citation_source` table for database joins between citation sources and the seqsets they cite.

### Backend Changes

- Adds a scheduled task to update SeqSet citation sources with results from Crossref. The task has an initial delay of one minute, then runs every 6 hours. It first checks if the Crossref backend service is active, then queries the [Crossref Cited-by API](https://www.crossref.org/documentation/cited-by/retrieve-citations/) to retrieve all citations for the DOI prefix configured in Loculus. If the returned XML cannot be parsed, or doesn't have a `crossref_result` tag, an `IllegalStateException` is thrown. If individual `forward_links` fail validation (missing dois/titles/years) these are added as `CrossRefValidationError` objects and logged in the task output. Validated citations are then merged into individual citation sources, conflicts logged, and results upserted in the database. If a curated citation source exists for a DOI found on Crossref, **it is updated with the results from Crossref**.
- Adds the backend endpoint `/get-seqset-citations`: for a Seqset ID and version, retrieves citations and returns their source DOIs, titles, years and contributors for each citation of the SeqSet. 
- Removes the `get-seqset-cited-by-publication` endpoint as this is now made redundant.
- *There aren't any endpoints yet for manual curation of citations, or for aggregating citations for groups, sequences etc - but I think these would be better in a follow-up PR*.
- *Also - there should probably be some cleanup task for removing citation sources if all their SeqSets all get deleted - if that is still a supported feature*.

### Website Changes

- Adds a `View Citations` modal to the SeqSet details page which lists citations for the seqset.
- Updates the total count of citations and the citations over time graph on the SeqSet details page to use the `get-seqset-citations` endpoint.

#### Additional website changes
- Updates the other modals on the SeqSet details page to use the `BaseDialog` component for consistency.

### Screenshot
image

### PR Checklist

🚀 Preview: https://seqset-citations.loculus.org


backend/src/main/kotlin/org/loculus/backend/service/crossref/CrossRefService.kt
```diff
@@ -59,6 +62,50 @@
...
+        }
+    }
+
+    fun getCrossRefCitedBy(doi: String): List<SeqSetCitation> {
+        checkIsActive()
+
+        val endDate = LocalDate.now().format(DateTimeFormatter.ISO_LOCAL_DATE)
+        val connection = URI(
+            properties.endpoint + "/servlet/getForwardLinks?usr=${properties.username}&pwd=${properties.password}&doi=${doi}&endDate=${endDate}&include_postedcontent=true"
+        ).toURL().openConnection() as HttpURLConnection
```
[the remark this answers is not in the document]
tombch: This alert was dismissed after switching to match a valid `seqSetId` and `version` first, and then retrieving the DOI from that.

website/src/components/SeqSetCitations/SeqSetItemActions.tsx
```diff
@@ -95,17 +105,41 @@ const SeqSetItemActionsInner: FC<SeqSetItemActionsProps> = ({
...
+                onClose={() => setExportModalVisible(false)}
+                title=''
+                fullWidth={false}
+                className='min-h-[60vh]'
+            >
+                <div className='min-w-[1000px]'></div>
                 <ExportSeqSet seqSet={seqSet} seqSetRecords={seqSetRecords} databaseName={databaseName} />
-            </Modal>
+            </BaseDialog>
+            <BaseDialog
```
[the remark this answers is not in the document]
tombch: SeqSets without a DOI may still have citations (e.g. added manually by a curator)

website/src/components/SeqSetCitations/SeqSetItem.tsx
```diff
@@ -82,10 +79,31 @@ const SeqSetItemInner: FC<SeqSetItemProps> = ({
...
 
     const {
-        data: seqSetCrossRefWork,
-        isLoading: isCrossRefWorkLoading,
-        isError: isCrossRefWorkError,
-    } = useCrossRefWork(seqSet.seqSetDOI);
+        isLoading: isSeqSetCitationsLoading,
+        error: seqSetCitationsError,
+        data: seqSetCitations,
+    } = seqSetCitationClientHooks(clientConfig).useGetSeqSetCitedBy({
```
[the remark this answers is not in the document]
tombch: I fixed the issue with the hook called on every render, but the duplication is still there - this could be solved by either moving the fetching into the server-side, or update the `withQueryProvider` to use a shared query client so we have deduping of requests, but that's a change for outside this PR

website/src/components/SeqSetCitations/SeqSetCitationsList.tsx
```diff
@@ -0,0 +1,81 @@
...
+        error: seqSetCitationsError,
+        data: seqSetCitations,
+    } = seqSetCitationClientHooks(clientConfig).useGetSeqSetCitedBy({
+        params: { seqSetId: seqSet.seqSetId, version: seqSet.seqSetVersion },
+    });
+
+    return (
+        <div className='flex flex-col items-center w-full'>
+            <div className='flex justify-start items-center py-5'>
+                <h1 className='text-xl font-semibold py-4'>SeqSet Citations</h1>
```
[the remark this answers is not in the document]
tombch: The other modals use h1 so leaving this

backend/src/main/kotlin/org/loculus/backend/service/crossref/CrossRefService.kt
```diff
@@ -49,16 +53,86 @@ class CrossRefService(final val properties: CrossRefServiceProperties) {
...
+    }
+
+    fun getCrossRefCitedBy(doiPrefix: String): List<SeqSetCitingSource> {
+        checkIsActive()
+
+        // End date is the current date at time of request
+        val endDate = LocalDate.now().format(DateTimeFormatter.ISO_LOCAL_DATE)
+        val connection = URI(
+            properties.endpoint +
+                "/servlet/getForwardLinks?usr=${properties.username}&pwd=${properties.password}&doi=$doiPrefix&endDate=$endDate&include_postedcontent=true",
```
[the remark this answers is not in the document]
tombch: I did already try this, but Crossref cited-by doesn't seem to support it (401: Unsupported authorization type: Basic)

backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCrossRefCitationsTask.kt
```diff
@@ -0,0 +1,78 @@
...
+    /**
+     * Runs every six hours, with an initial delay of one minute.
+     * Adds citing sources from CrossRef, and connects to SeqSets via their DOI.
+     */
+    @Scheduled(
+        initialDelay = 1,
+        fixedDelay = 360,
+        timeUnit = java.util.concurrent.TimeUnit.MINUTES,
+    )
+    fun task() {
```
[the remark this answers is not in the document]
tombch: IMO this isn't necessary

backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCitationsDatabaseService.kt
```diff
@@ -434,18 +485,31 @@ class SeqSetCitationsDatabaseService(
...
+            )
 
-        return citedBy
+        return SeqSetToCitingSourceTable.innerJoin(
+            SeqSetCitingSourceTable,
+        ).selectAll()
+            .where {
+                (SeqSetToCitingSourceTable.seqSetId eq seqSet[SeqSetsTable.seqSetId]) and
+                    (SeqSetToCitingSourceTable.seqSetVersion eq seqSet[SeqSetsTable.seqSetVersion])
+            }.orderBy(SeqSetCitingSourceTable.year to SortOrder.DESC).map {
```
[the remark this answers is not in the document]
tombch: I've since migrated the year field so this is no longer an issue

backend/src/main/kotlin/org/loculus/backend/service/crossref/CrossRefService.kt
```diff
@@ -49,16 +53,94 @@ class CrossRefService(final val properties: CrossRefServiceProperties) {
...
+                seqSetDOIs = setOf(seqSetDOI),
+            )
+        }
+    }
+
+    fun getCrossRefCitedBy(doiPrefix: String): List<SeqSetCitingSource> {
+        checkIsActive()
+
+        // End date is the current date at time of request
+        val endDate = LocalDate.now().format(DateTimeFormatter.ISO_LOCAL_DATE)
```
maverbiest: I think we use `org.loculus.backend.utils.DateProvider` for this

tombch: Thanks, I have updated this now

theosanderson:
@codex review

backend/src/main/kotlin/org/loculus/backend/api/SeqSetCitations.kt
```diff
@@ -57,6 +56,28 @@ data class SeqSet(
...
+
+data class SeqSetCitingSource(
+    val sourceDOI: String,
+    val title: String,
+    val year: Int,
+    val contributors: List<CitationContributor>,
+    val seqSetDOIs: Set<String> = emptySet(),
+)
+
+data class SeqSetCitation(
```
maverbiest: Would it make sense to compose this into `SeqSetCitingSource`?

I also find that the names of the two classes here don't convey their meaning very clearly (at least to me). Would something like `SeqSetCitingPublication` and `Publication` work? (or `Record` or w/e)

```kotlin
data class SeqSetCitingPublication(
    val publication: Publication,
    val seqSetDOIs: Set<String> = emptySet(),
)

data class Publication(
    val sourceDOI: String,
    val title: String,
    val year: Int,
    val contributors: List<Contributor>,
)
```

backend/src/main/kotlin/org/loculus/backend/api/SeqSetCitations.kt
```diff
@@ -57,6 +56,28 @@ data class SeqSet(
     val seqSetDOI: String?,
 )
 
+data class CitationContributor(val givenName: String, val surname: String)
```
maverbiest: Similar to my comment below, to me the thing that people contribute to is the Publication (or CitingSource etc.), not to the act of citing. The citation is a connection between two publications, not really it's own entity

(of course, feel free to disagree if you think I'm bikeshedding)

backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetsTables.kt
```diff
@@ -28,3 +31,20 @@ object SeqSetToRecordsTable : Table("seqset_to_records") {
     val seqSetVersion = long("seqset_version") references SeqSetsTable.seqSetVersion
     override val primaryKey = PrimaryKey(seqSetRecordId, seqSetId, seqSetVersion)
 }
+
+object SeqSetCitingSourceTable : Table("seqset_citing_source") {
+    val citingSourceId = long("citing_source_id").autoIncrement()
+    val sourceDOI = text("source_doi").uniqueIndex()
+    val origin = enumerationByName<CitationOrigin>("origin", 10)
```
maverbiest: Why do we need to limit the max size to 10 here? In the db the field is just unbounded text, correct? (well okay, it's an enum, but we could make arbitrarily large emum variants in the future)

backend/src/test/kotlin/org/loculus/backend/service/crossref/CrossRefServiceTest.kt
```diff
@@ -16,13 +20,268 @@ class CrossRefServiceTest(@Autowired private val crossRefService: CrossRefServic
...
+        """.trimIndent()
+
+        val ex = assertThrows<IllegalStateException> {
+            crossRefService.parseCrossRefCitedByXML(xml)
+        }
+        assertTrue(ex.message!!.contains("missing DOI"))
+    }
+
+    @Test
+    fun `parseCrossRefCitedByXML returns empty list for completely malformed input`() {
```
maverbiest: I feel like this should actually fail quite loudly.

It seems a bit inconsistent that atm we reject a whole batch of citations if one of them is missing a field, but for completely malformed XML we return silently and the task logs "Fetched 0 citing source(s)..." at INFO

tombch: Yeah I agree, I left it like this as it seems to be the default behaviour for the Jsoup xml parser but I will update

backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCrossRefCitationsTask.kt
```diff
@@ -0,0 +1,78 @@
...
+@ConditionalOnProperty(BackendSpringProperty.ENABLE_SEQSETS, havingValue = ENABLE_SEQSETS_TRUE_VALUE)
+class SeqSetCrossRefCitationsTask(
+    private val crossRefService: CrossRefService,
+    private val seqSetCitationsDatabaseService: SeqSetCitationsDatabaseService,
+) {
+    /**
+     * Runs every six hours, with an initial delay of one minute.
+     * Adds citing sources from CrossRef, and connects to SeqSets via their DOI.
+     */
+    @Scheduled(
```
maverbiest, line 46: Here you hard-code how the task is scheduled, which is consistent with the `CleanUpAuxTableTask`. However, for the `CleanUpStaleSequencesInProcessingTask` the frequency is set in `application.properties` and then threaded through `BackendSpringProperty`.

It's not directly related to this PR but might make sense to do this consistently for scheduled tasks, perhaps someone else would like to weigh in?

backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetsTables.kt
tombch: The max size argument is required for `enumerationByName`- exposed orm thinks the column is a `varchar` but the actual column is `text`. So yeah the 10 isn't actually doing anything in practice, somewhat arbitrary I agree

tombch: Just had a look and exposed has a `customEnumeration` so we can do this instead:
```
    val origin = customEnumeration(
        name = "origin",
        sql = "text",
        fromDb = { CitationOrigin.valueOf(it as String) },
        toDb = { it.name },
    )
```

which i guess removes the arbitrary length and keeps the enum type checking - I have switched to use this

backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCrossRefCitationsTask.kt
```diff
@@ -0,0 +1,78 @@
...
+import org.loculus.backend.config.BackendSpringProperty
+import org.loculus.backend.config.ENABLE_SEQSETS_TRUE_VALUE
+import org.loculus.backend.service.crossref.CrossRefService
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+import org.springframework.scheduling.annotation.Scheduled
+import org.springframework.stereotype.Component
+
+private val log = mu.KotlinLogging.logger {}
+
+internal fun mergeCitingSources(citingSources: List<SeqSetCitingSource>): Set<SeqSetCitingSource> {
```
maverbiest: Are `SeqSetCitingSource`s guaranteed to be ordered by date here? To me it would make sense to overwrite information from older entries with that from newer entries with the same `sourceDOI` (but maybe that's already happening)

backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetsTables.kt
maverbiest: > The max size argument is required for enumerationByName

Ahhh I see thanks for explaining! The `customEnumeration` implementation looks nice

backend/src/main/kotlin/org/loculus/backend/api/SeqSetCitations.kt
tombch: yeah I agree in the current layout this one should be `CitingSourceContributor`

backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCrossRefCitationsTask.kt
tombch: The Crossref cited-by api doesn't return any indicator of when a citation was made, so there is no ordering by date. But (and I might be wrong here) all forward links with the same `sourceDOI` should be from the same time, as its all the same publication. So title/year/authors _should_ be identical

backend/src/test/kotlin/org/loculus/backend/service/crossref/CrossRefServiceTest.kt
tombch: Updated this now to fail loudly for completely malformed xml (e.g. `<<crossref_result>...`) as well as xml missing `<crossref_result>`

backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetsTables.kt
tombch: Thanks!

backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCitationsDatabaseService.kt
```diff
@@ -367,6 +377,52 @@ class SeqSetCitationsDatabaseService(
         )
     }
 
+    fun updateCitingSourcesFromCrossRef(citingSources: Set<SeqSetCitingSource>): CitingSourcesUpdateResult {
+        // Map of database seqSet DOIs to their ID and version
+        val doiToSeqSet = SeqSetsTable
```
maverbiest, line 383: This loads all seqsets we have in the DB. Would it make sense to get only those for which the DOI is mentioned by the `citingSources`? 

e.g., `val seqSetDOIs = citingSources.flatMap { it.seqSetDOIs }.toSet()` and filter by those?

backend/src/main/kotlin/org/loculus/backend/api/SeqSetCitations.kt
tombch: I'm going to mark this resolved as I've now updated to this structure:

```
data class CitationSource(
    val sourceDOI: String,
    val title: String,
    val year: Int,
    val contributors: List<CitationContributor>,
)

data class SeqSetCitationSource(val source: CitationSource, val seqSetDOIs: Set<String> = emptySet())

data class SeqSetCitation(val source: CitationSource)
```

Which has the citation source object composed

backend/src/main/kotlin/org/loculus/backend/api/SeqSetCitations.kt
tombch: IMO now with `CitationSource`, this one can just stay as `CitationContributor`

backend/src/main/kotlin/org/loculus/backend/service/crossref/CrossRefService.kt
```diff
@@ -49,16 +54,94 @@ class CrossRefService(final val properties: CrossRefServiceProperties) {
...
             throw RuntimeException("The CrossRefService is not active as it has not been configured.")
         }
     }
 
+    fun parseCrossRefCitedByXML(citedByXML: String): List<SeqSetCitingSource> {
+        val doc = Jsoup.parse(citedByXML, "", Parser.xmlParser())
+
+        return doc.select("forward_link").map { forwardLink ->
+            val seqSetDOI = forwardLink.attr("doi").takeIf { it.isNotBlank() }
+                ?: throw IllegalStateException("CrossRef forward_link missing SeqSet DOI: $forwardLink")
```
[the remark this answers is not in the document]
tombch: This has now been updated - critical errors related to invalid XML or missing `crossref_result` tag will throw, but when iterating the forward links, any forward link with missing dois/title/year will be appended as a `CrossRefValidationError` and logged by the task.

backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCitationsDatabaseService.kt
maverbiest: Update: as we discussed offline, the alternative has a potential risk of running into the limit of how many parmeter placeholders are supported. For a task that runs once every 6 hours or so it also shouldn't be that big of a deal to load the table.

Feel free to resolve my comment!

backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCrossRefCitationsTask.kt
```diff
@@ -0,0 +1,89 @@
...
+    fun task() {
+        log.info { "Updating SeqSet CrossRef citations..." }
+
+        if (!crossRefService.isActive) {
+            log.info { "CrossRef service is not active, skipping SeqSet citation update." }
+            return
+        }
+
+        val doiPrefix = crossRefService.doiPrefix
+        if (doiPrefix.isNullOrBlank()) {
```
maverbiest, line 60: I wonder if it wouldn't be nicer to make the doiPrefix non-nullable so you don't have to check this. Is it not essentially a configuration error if the CrossRefService is active but there's no doiPrefix set? Or am I missing something?

backend/src/main/kotlin/org/loculus/backend/api/SeqSetCitations.kt
maverbiest: Sorry to push on this again, but why not plain `Contributor`?

backend/src/test/kotlin/org/loculus/backend/service/crossref/CrossRefServiceTest.kt
```diff
@@ -16,13 +21,355 @@ class CrossRefServiceTest(@Autowired private val crossRefService: CrossRefServic
...
+    @Test
+    fun `parseCrossRefCitedByXML returns empty result when no forward_link elements present`() {
+        val xml = "<crossref_result></crossref_result>"
+        val result = crossRefService.parseCrossRefCitedByXML(xml)
+        assertTrue(result.sources.isEmpty())
+        assertTrue(result.validationErrors.isEmpty())
+    }
+
+    @Test
+    fun `parseCrossRefCitedByXML records validation error when forward_link is missing the seqSet DOI attribute`() {
```
maverbiest: There's ~7 tests here with the exact same skeleton: invalid XML -> parse -> test for error string

We should look into whether we can do this with a single parametrized test, although I've never done this in kotlin, looks like it exists in this junit we use: https://www.baeldung.com/parameterized-tests-junit-5

backend/src/main/kotlin/org/loculus/backend/api/SeqSetCitations.kt
anna-parker: I think this is fine :-)

anna-parker: I think this is fine :-) If anything `CitingSourceContributor` maybe is a bit better

backend/src/main/kotlin/org/loculus/backend/controller/SeqSetCitationsController.kt
```diff
@@ -103,9 +104,9 @@ class SeqSetCitationsController(
             submissionDatabaseService.getApprovedUserAccessionVersions(authenticatedUser),
         )
 
-    @Operation(description = "Get count of SeqSet cited by publications")
+    @Operation(description = "Get SeqSet citations from publications")
     @GetMapping("/get-seqset-cited-by-publication")
```
anna-parker: This predates this PR but I think the name of this endpoint is terrible and we should fix it while doing this update. If I read `get-seqset-cited-by-publication` I would think I would upload a publication and get the seqSets cited by it. Instead this returns a list of publications that cite the seqSet, I think it should thus be called `/get-publications-citing-seqset` (or in true REST fashion: `GET /seqsets/{seqSetId}/citing-publications`)

backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCrossRefCitationsTask.kt
maverbiest: (I realise it's not something you implemented in this PR, but I think it'd be nice to put the CrossRefService behind a single `enabled` flag in the config like other services are, instead of checking these individual properties)

backend/src/main/kotlin/org/loculus/backend/api/SeqSetCitations.kt
```diff
@@ -57,6 +56,24 @@ data class SeqSet(
...
+data class CitationSource(
+    val sourceDOI: String,
+    val title: String,
+    val year: Int,
+    val contributors: List<CitationContributor>,
+)
+
+data class SeqSetCitationSource(val source: CitationSource, val seqSetDOIs: Set<String> = emptySet())
+
+data class SeqSetCitation(val source: CitationSource)
```
anna-parker, line 98: as this is the new return type of a public endpoint we should add it to the schema so it is visible on the swagger page (like CitedBy below)

Image

website/src/components/SeqSetCitations/CitationTable.tsx
```diff
@@ -0,0 +1,74 @@
+import { type FC } from 'react';
+
+import { type SeqSetCitation } from '../../types/seqSetCitation';
+
+interface CitationTableProps {
+    isLoading: boolean;
+    error: Error | null;
+    citations: SeqSetCitation[];
```
maverbiest, line 8: Should this use the new`SeqSetCitations` type from ../../types/seqSetCitation?

backend/src/main/resources/db/migration/V1.30__add_seqset_citations_table.sql
```diff
@@ -0,0 +1,26 @@
+create table seqset_citation_source (
+    citation_source_id   bigserial,
+    source_doi         text not null unique,
```
anna-parker, line 3: I might have asked this before, sorry for forgetting, I thought that not all citations might have created a DOI and we wanted to still be able to add them here - so I think we would need to allow this to still be null?

website/src/components/SeqSetCitations/CitationTable.tsx
```diff
@@ -0,0 +1,74 @@
...
+                        className='text-primary-700'
+                        href={`https://doi.org/${citation.source.sourceDOI}`}
+                        target='_blank'
+                        rel='noopener noreferrer'
+                    >
+                        {citation.source.title}
+                    </a>
+                    <div className='text-sm text-gray-500'>
+                        {citation.source.contributors
+                            .map((contributor) => `${contributor.givenName} ${contributor.surname}`)
```
maverbiest: The backend allows for contributors without a first name. This formatting with the join on ', ' will insert two leading whitespaces in these cases

backend/src/main/kotlin/org/loculus/backend/api/SeqSetCitations.kt
```diff
@@ -57,6 +56,24 @@ data class SeqSet(
...
+}
+
+data class CitationSource(
+    val sourceDOI: String,
+    val title: String,
+    val year: Int,
+    val contributors: List<CitationContributor>,
+)
+
+data class SeqSetCitationSource(val source: CitationSource, val seqSetDOIs: Set<String> = emptySet())
```
anna-parker, line 95: I find the naming of `SeqSetCitationSource` and `SeqSetCitation` confusing - why do we actually need `SeqSetCitation` cant we just use a bare `CitationSource`, it seems weird to me to create a class with only one object

backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCrossRefCitationsTask.kt
```diff
@@ -0,0 +1,89 @@
...
+    )
+    fun task() {
+        log.info { "Updating SeqSet CrossRef citations..." }
+
+        if (!crossRefService.isActive) {
+            log.info { "CrossRef service is not active, skipping SeqSet citation update." }
+            return
+        }
+
+        val doiPrefix = crossRefService.doiPrefix
```
anna-parker, line 59: sorry if I am missing sth, dont we have to do this call for every doi in the db? why are we only calling this on one DOI?

website/src/components/common/BaseDialog.tsx
```diff
@@ -9,17 +9,25 @@ interface BaseDialogProps {
...
-export const BaseDialog: React.FC<BaseDialogProps> = ({ title, isOpen, onClose, children, fullWidth = true }) => {
+export const BaseDialog: React.FC<BaseDialogProps> = ({
+    title,
+    isOpen,
+    onClose,
+    children,
+    fullWidth = true,
+    className,
+}) => {
     const fullWidthClasses = fullWidth ? 'w-full w-max-5xl' : '';
```
maverbiest:
```suggestion
    const fullWidthClasses = fullWidth ? 'w-full max-w-5xl' : '';
```

maverbiest (commented):
I was finally able to go through the whole PR, apologies that it took so long!

Looks great overall, no major blockers on my side. I think we could make `CrossRefServiceTest.kt` a bit cleaner with a parametrized test. I would also like to interact with the front-end changes on a preview or locally. Do you have any tips for setting this up so there's some test seqsets to play around with?

backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCrossRefCitationsTask.kt
anna-parker: ah ok all our seqSets will have the DOI prefix that we were issued: https://www.crossref.org/services/content-registration/ - maybe it would be good to add this to the docs for people who are new to this like me

backend/src/main/kotlin/org/loculus/backend/service/crossref/CrossRefService.kt
```diff
@@ -49,16 +62,139 @@ class CrossRefService(final val properties: CrossRefServiceProperties) {
...
+    val dateTimeFormatterdd: DateTimeFormatter = DateTimeFormatter.ofPattern("dd")
+    val dateTimeFormatteryyyy: DateTimeFormatter = DateTimeFormatter.ofPattern("yyyy")
 
     private fun checkIsActive() {
         if (!isActive) {
             throw RuntimeException("The CrossRefService is not active as it has not been configured.")
         }
     }
 
+    fun parseCrossRefCitedByXML(citedByXML: String): CrossRefCitedByResult {
```
anna-parker, line 76: We should link the docs here for future reference: https://www.crossref.org/documentation/cited-by/retrieve-citations/

backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCitationsDatabaseService.kt
```diff
@@ -367,6 +378,55 @@ class SeqSetCitationsDatabaseService(
...
+            }
+            .flatMap { result ->
+                val citationSourceId = result[SeqSetCitationSourceTable.citationSourceId]
+                val seqSetDOIs = matchedSourceToSeqSetDOIs.getValue(result[SeqSetCitationSourceTable.sourceDOI])
+                seqSetDOIs.mapNotNull { doi ->
+                    doiToSeqSet[doi]?.let { (seqSetId, version) ->
+                        SeqSetToCitationSourceEntry(citationSourceId, seqSetId, version)
+                    }
+                }
+            }
```
anna-parker, line 414: I think this can be done more efficiently and had some fun thinking about it:

```suggestion

        // Collect only the seqSet DOIs referenced by the incoming citation sources
        val referencedSeqSetDOIs = citationSources.flatMapTo(mutableSetOf()) { it.seqSetDOIs }
        if (referencedSeqSetDOIs.isEmpty()) return CitationSourcesUpdateResult(emptySet())

        // Map of database seqSet DOIs to their ID and version, limited to referenced DOIs
        val doiToSeqSet = SeqSetsTable
            .select(SeqSetsTable.seqSetId, SeqSetsTable.seqSetVersion, SeqSetsTable.seqSetDOI)
            .where { SeqSetsTable.seqSetDOI inList referencedSeqSetDOIs }
            .associate { it[SeqSetsTable.seqSetDOI]!! to (it[SeqSetsTable.seqSetId] to it[SeqSetsTable.seqSetVersion]) }

        // Map sourceDOI → only the seqSet DOIs confirmed present in the database
        val matchedSourceToSeqSetDOIs = citationSources.mapNotNull { source ->
            val matchedDOIs = source.seqSetDOIs.filterTo(mutableSetOf()) { it in doiToSeqSet }
            if (matchedDOIs.isEmpty()) null else source.source.sourceDOI to matchedDOIs
        }.toMap()

        val matchedSources = citationSources.filter { it.source.sourceDOI in matchedSourceToSeqSetDOIs }

        // Upsert matched citation sources based on their DOI
        // Build entries for the join table from the returned source DOIs + primary keys
        val joinEntries = SeqSetCitationSourceTable
            .batchUpsert(matchedSources, SeqSetCitationSourceTable.sourceDOI) {
                this[SeqSetCitationSourceTable.sourceDOI] = it.source.sourceDOI
                this[SeqSetCitationSourceTable.origin] = CitationOrigin.CROSSREF
                this[SeqSetCitationSourceTable.title] = it.source.title
                this[SeqSetCitationSourceTable.year] = it.source.year
                this[SeqSetCitationSourceTable.contributors] = it.source.contributors
            }
            .flatMap { result ->
                val citationSourceId = result[SeqSetCitationSourceTable.citationSourceId]
                val seqSetDOIs = matchedSourceToSeqSetDOIs.getValue(result[SeqSetCitationSourceTable.sourceDOI])
                seqSetDOIs.map { doi ->
                    val (seqSetId, version) = doiToSeqSet.getValue(doi)
                    SeqSetToCitationSourceEntry(citationSourceId, seqSetId, version)
                }
            }
```

anna-parker: ah I see now the thread on sql param limits... not sure how likely we are to start hitting the 50k param limit but I guess we can stay with the current state :-)

backend/src/test/kotlin/org/loculus/backend/controller/seqsetcitations/CitationEndpointsTest.kt
```diff
@@ -56,43 +73,179 @@ class CitationEndpointsTest(@Autowired private val client: SeqSetCitationsContro
...
-        client.getSeqSetCitedByPublication()
+    fun `WHEN calling get user cited by seqSet for user sequences in a seqSet THEN returns results`() {
+        val seqSetResult = client.createSeqSet()
+            .andExpect(status().isOk)
+            .andExpect(content().contentType(MediaType.APPLICATION_JSON_VALUE))
+            .andExpect(jsonPath("\$.seqSetId").isString)
+            .andExpect(jsonPath("\$.seqSetVersion").value(1))
+            .andReturn()
+
+        client.getUserCitedBySeqSet()
```
anna-parker, line 84: sorry I know this isnt sth you added, but again I really dislike the name of this endpoint, IMO it again sounds like it does sth different it should be sth like `@GetMapping("/user-sequence-seqset-citations-by-year")
fun getUserSequenceCitationsByYear(...)`

backend/src/test/kotlin/org/loculus/backend/controller/seqsetcitations/CitationEndpointsTest.kt
```diff
@@ -56,43 +73,179 @@ class CitationEndpointsTest(@Autowired private val client: SeqSetCitationsContro
...
+                jsonPath(
+                    "\$.detail",
+                    containsString("SeqSet $seqSetId, version $seqSetVersion has a DOI and cannot be deleted"),
+                ),
+            )
+    }
+
+    @Test
+    fun `WHEN multiple crossref citation runs link the same citation source THEN all citations are recorded`() {
+        val seqSetAResult = client.createSeqSet().andExpect(status().isOk).andReturn()
```
anna-parker: I guess this could be extracted into a function but not a blocker :-)

website/src/components/SeqSetCitations/CitationTable.tsx
anna-parker: .map((contributor) =>
  [contributor.givenName, contributor.surname].filter(Boolean).join(' ')
)
.join(', ')

anna-parker (commented):
Hey Tom! Sorry only finished looking through this today, its a bit hard to test without CrossRef examples - I added a few comments on the backend but it would be great to put this on staging on Monday and then take a closer look, let me know if you need any help :-)

backend/src/main/kotlin/org/loculus/backend/controller/SeqSetCitationsController.kt
tombch: Yeah I agree it should be renamed, at first I was thinking something like `get-seqset-citations` as an option

website/src/components/common/BaseDialog.tsx
tombch: I removed the incorrect `w-max-5xl` as if we use `max-w-5xl` it makes the full width mode not full width anymore

website/src/components/SeqSetCitations/CitationTable.tsx
tombch: Thanks, updated!

backend/src/main/kotlin/org/loculus/backend/controller/SeqSetCitationsController.kt
tombch: This has now been renamed to `get-seqset-citations`

backend/src/main/kotlin/org/loculus/backend/service/crossref/CrossRefService.kt
tombch: Added docs for this in the `getCrossRefCitedBy` function

backend/src/test/kotlin/org/loculus/backend/controller/seqsetcitations/CitationEndpointsTest.kt
tombch: I agree - I was thinking to rename/replace this in a follow-up PR when adding group citations?

backend/src/main/kotlin/org/loculus/backend/api/SeqSetCitations.kt
tombch: We discussed this and have decided to keep as-is, as having the structure with source as an object is useful for sequence citations and keeps things consistent in #6491. The same component can be used for both seqset and sequence citations as they follow the same 'shape'

backend/src/main/kotlin/org/loculus/backend/api/SeqSetCitations.kt
tombch: Thanks, this has been added!

website/src/components/SeqSetCitations/CitationTable.tsx
tombch: I guess this is an opinion thing but imo its nicer to just import `SeqSetCitation` and use it singular and in an array rather than need to import `SeqSetCitation` and `SeqSetCitations`

backend/src/test/kotlin/org/loculus/backend/service/crossref/CrossRefServiceTest.kt
tombch: I have now parameterized these tests! There's a few cases of it being used elsewhere e.g. in `DataUseTermsControllerTest`

backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCrossRefCitationsTask.kt
tombch: I updated the task description to include more information about what its doing: f2aa6498e7d1e5a059d082bd95a43e4dd3b93874. I didn't include a link to the content registration page there though as I didn't think this is the best place for it - I have added it to the description of the `DOIPrefix` in the `values.schema.json`: e2fb4a4bda1beb234ef41bbedb1b17f22e9e6557

theosanderson:
@codex review

backend/src/main/kotlin/org/loculus/backend/service/crossref/CrossRefService.kt
```diff
@@ -49,16 +62,142 @@ class CrossRefService(final val properties: CrossRefServiceProperties) {
...
+                ?: run {
+                    validationErrors.add(
+                        CrossRefValidationError(
+                            "CrossRef citation source missing DOI for SeqSet $seqSetDOI: $citationElement",
+                        ),
+                    )
+                    return@mapNotNull null
+                }
+
+            val title = citationElement.selectFirst("title")?.text()?.takeIf { it.isNotBlank() }
```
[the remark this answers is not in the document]
theosanderson: Hm, presumably this is a false positive?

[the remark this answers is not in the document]
tombch: I think this is valid - all the citations so far have been from posted content (preprints), which uses `title` instead of `article_title`. I had not noticed they were different 🙃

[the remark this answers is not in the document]
theosanderson: Ah, thanks be to Codex!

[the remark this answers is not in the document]
tombch: Yeah bit of an oversight on my part, thanks be to codex 😅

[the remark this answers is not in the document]
tombch: This has now been updated to handle different title tags

backend/src/test/kotlin/org/loculus/backend/controller/seqsetcitations/CitationEndpointsTest.kt
tombch: Marking this as resolved as it will be addressed in a follow-up PR

backend/src/main/kotlin/org/loculus/backend/api/SeqSetCitations.kt
tombch: IMO `CitationContributor` is clear enough in this case and also not too verbose so im going to mark this as resolved (could be altered in the future)

backend/src/test/kotlin/org/loculus/backend/controller/seqsetcitations/CitationEndpointsTest.kt
tombch: Done!

backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCrossRefCitationsTask.kt
tombch: Discussed with Anya and we have said this could be potentially addressed in a follow-up PR!

website/src/components/SeqSetCitations/SeqSetItem.tsx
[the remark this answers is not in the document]
tombch: Switching this can be addressed in a follow up PR

backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCrossRefCitationsTask.kt
tombch: Discussed with Anya and we agreed this can be addressed in a follow-up PR

backend/src/main/resources/db/migration/V1.30__add_seqset_citations_table.sql
tombch: Me and Anya discussed this - the `source_doi` is not the primary key, so making this nullable in the future, and adding an additional field (if we want to add citation sources without a DOI) should be straightforward

backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCitationsDatabaseService.kt
```diff
@@ -367,6 +377,47 @@ class SeqSetCitationsDatabaseService(
...
+            .where { SeqSetsTable.seqSetDOI.isNotNull() }
+            .associate { it[SeqSetsTable.seqSetDOI]!! to (it[SeqSetsTable.seqSetId] to it[SeqSetsTable.seqSetVersion]) }
+
+        // Citing sources with at least one SeqSet present in the database
+        val matchedSources = citingSources.filter { source ->
+            source.seqSetDOIs.any { it in doiToSeqSet }
+        }
+
+        // Update or insert the matched citing sources
+        SeqSetCitingSourceTable.batchUpsert(matchedSources) {
```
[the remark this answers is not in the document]
tombch: Currently the approach is crossref should take precedence over curated data - crossref will contain the up-to-date entry. Also this can be changed in future when curated citations are properly supported

anna-parker (approved):
tested on staging and looks great!

PR #6307, opened 2026-04-28
feat(deposition): only add timestamp to alias for integration tests - not when testing on staging

anna-parker:
resolves #

### New ena deposition parameter
`random_alias` - when true add a timestamp+entropy to the biosample/bioproject submission alias (this is similar to our submissionId), this is only set for the integration tests

### Screenshot
ENA enforces that the alias (created by us from the loculus accession) of each submitted biosample is unique - this is important as it is used for revising biosamples.

However, when testing ENA submissions e.g. for the integration tests we want to be able to submit the same biosample multiple times to ENA dev. 

Previously I just added a timestamp to the alias for all submissions to ENA dev to allow multiple submissions of the same biosample on ENA dev which was directly used by the integration tests - but this means e.g. if I want to test a revision on staging (which also submits to ENA dev/test) this fails as the alias is different. 

This PR decouples adding a timestamp to the alias from if the ena deposition pod is talking to ENA test (it should have never been coupled). This keeps the current functionality for integration tests (here we add a timestamp to the alias) and additionally keeps the current functionality for production (no timestamp to alias).

### PR Checklist

### DIFF
--- a/ena-submission/scripts/test_ena_submission_integration.py
+++ b/ena-submission/scripts/test_ena_submission_integration.py
@@ -428,7 +428,7 @@ def _test_successful_sample_submission(
     create_sample_sync_state_with_submission_table(db_engine, config=config)
     check_sample_submission_started(db_engine, sequences_to_upload)
 
-    sample_table_create(db_engine, config, test=config.test)
+    sample_table_create(db_engine, config)
     create_sample_sync_state_with_submission_table(db_engine, config)
     check_sample_submission_submitted(db_engine, sequences_to_upload)
 
@@ -439,7 +439,7 @@ def _test_successful_project_submission(
     create_project_sync_state_with_submission_table(db_engine, config)
     check_project_submission_started(db_engine, sequences_to_upload)
 
-    project_table_create(db_engine, config, test=config.test)
+    project_table_create(db_engine, config)
     create_project_sync_state_with_submission_table(db_engine, config)
     check_project_submission_submitted(db_engine, sequences_to_upload)
 
@@ -588,6 +588,9 @@ class TestSubmission:
             f"ENA submission URL is {self.config.ena_submission_url} instead of https://wwwdev.ebi.ac.uk/ena/submit/drop-box/submit/"
         )
         assert self.config.test, "Test mode is not enabled."
+        assert self.config.random_alias, (
+            "Random alias is not enabled, this will case conflicts in ENA dev if tests are run simultaneously."  # noqa: E501
+        )
 
 
 class TestFirstPublicUpdate(TestSubmission):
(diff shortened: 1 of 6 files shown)


ena-submission/scripts/test_ena_submission_integration.py
```diff
@@ -588,6 +588,9 @@ def setup_method(self) -> None:
             f"ENA submission URL is {self.config.ena_submission_url} instead of https://wwwdev.ebi.ac.uk/ena/submit/drop-box/submit/"
         )
         assert self.config.test, "Test mode is not enabled."
+        assert self.config.random_alias, (
+            "Random alias is not enabled, this will case conflicts in ENA dev if tests are run simultaneously."  # noqa: E501
```
maverbiest:
```suggestion
            "Random alias is not enabled, this will cause conflicts in ENA dev if tests are run simultaneously."  # noqa: E501
```

ena-submission/src/ena_deposition/ena_submission_helper.py
```diff
@@ -149,7 +149,7 @@ def get_project_xml(project_set: ProjectSet) -> dict[str, str]:
     }
 
 
-def get_alias(prefix: str, test=False, set_alias_suffix: str | None = None) -> XmlAttribute:
+def get_alias(prefix: str, random_alias=False, set_alias_suffix: str | None = None) -> XmlAttribute:
```
maverbiest, line 152: It's not super clear to me yet in what cases you'd pass a `set_alias_suffix`, but from the logic in the function seems it's mutually exclusive with `random_alias`. E.g., I could now pass `random_alias=True` and `set_alias_suffix=some_suffix` and I won't get a random alias. Is it worth making this explicit somehow?

Again, maybe the current setup makes total sense but it's not immediately clear to me as I'm still getting familiar with this service

maverbiest (commented):
One comment (more of a question, really) about alias creation

ena-submission/src/ena_deposition/ena_submission_helper.py
anna-parker: Ah yes they are only used for testing atm and they are used fro separate things.

I set the alias suffix for testing revisions in the integration tests - this is because the alias is a unique ID for each submission and is required to revise a sequence - so I need to store the alias I used before. In production code the alias is always created from the pathoplexus accession in a standardized way (also to avoid duplicate submissions), however for testing we need to ensure we can retry tests but still store the alias used for revising.

anna-parker: In case you are worried about revising a sequence multiple times, for biosamples this is not possible - although for assemblies it is sadly - because you create a new sample with ADD and revise with MODIFY (you cannot ADD a biosample with an alias that already exists)

maverbiest: Ahhh I see so it's sort of an override you use when testing to ensure you can use the same alias again to test revisions. Thanks for clearing that up!

maverbiest (approved):

PR #6341, opened 2026-05-04
feat!(deployment): add option to run ingest as a staggered cronjob twi…

anna-parker:
…ce a day instead of a deployment

resolves #

### PR Checklist

### DIFF
--- a/kubernetes/loculus/templates/ingest-deployment.yaml
+++ b/kubernetes/loculus/templates/ingest-deployment.yaml
@@ -1,9 +1,91 @@
 {{- $dockerTag := include "loculus.dockerTag" .Values }}
 {{- if not .Values.disableIngest }}
-{{- range $_, $item := (include "loculus.enabledOrganisms" . | fromJson).organisms }}
+{{- range $i, $item := (include "loculus.enabledOrganisms" . | fromJson).organisms }}
 {{- $key := $item.key }}
 {{- $organismContent := $item.contents }}
 {{- if $organismContent.ingest }}
+{{- if $.Values.ingestAsCronJob }}
+{{- $hourOffset := mod $i 12 }}
+---
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: loculus-ingest-cronjob-{{ $key }}
+  annotations:
+    argocd.argoproj.io/sync-options: Replace=true
+spec:
+  schedule: {{ printf "0 %d,%d * * *" $hourOffset (add $hourOffset 12) | quote }}
+  concurrencyPolicy: Forbid
+  startingDeadlineSeconds: 3600
+  jobTemplate:
+    spec:
+      activeDeadlineSeconds: {{ $.Values.ingestLimitSeconds }}
+      template:
+        metadata:
+          labels:
+            app: loculus
+            component: loculus-ingest-cronjob-{{ $key }}
+        spec:
+          {{- include "possiblePriorityClassName" $ | nindent 10 }}
+          restartPolicy: OnFailure
+          initContainers:
+            - name: version-check
+              image: busybox
+              {{- include "loculus.resources" (list "ingest-init" $.Values) | nindent 14 }}
+              command: ['sh', '-c', '
+                CONFIG_VERSION=$(grep "verify_loculus_version_is:" /package/config/config.yaml | sed "s/verify_loculus_version_is: //;");
+                DOCKER_TAG="{{ $dockerTag }}";
+                echo "Config version: $CONFIG_VERSION";
+                echo "Docker tag: $DOCKER_TAG";
+                if [ "$CONFIG_VERSION" != "$DOCKER_TAG" ]; then
+                  echo "Version mismatch: ConfigMap version $CONFIG_VERSION does not match docker tag $DOCKER_TAG";
+                  exit 1;
+                else
+                  echo "Version match confirmed";
+                fi
+              ']
+              volumeMounts:
+                - name: loculus-ingest-config-volume-{{ $key }}
+                  mountPath: /package/config/config.yaml
+                  subPath: config.yaml
+          containers:
+            - name: ingest-{{ $key }}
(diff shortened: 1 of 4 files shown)


kubernetes/loculus/values.yaml
```diff
@@ -31,6 +31,7 @@ disableWebsite: false
 disableBackend: false
 disablePreprocessing: false
 disableIngest: false
+ingestAsCronJob: false
```
```suggestion
ingestAsCronJob: true
```

kubernetes/loculus/values.yaml
```diff
@@ -31,6 +31,7 @@ disableWebsite: false
 disableBackend: false
 disablePreprocessing: false
 disableIngest: false
+ingestAsCronJob: true
```
anna-parker: revert before merging

kubernetes/loculus/templates/ingest-deployment.yaml
```diff
@@ -1,9 +1,91 @@
...
+  name: loculus-ingest-cronjob-{{ $key }}
+  annotations:
+    argocd.argoproj.io/sync-options: Replace=true
+spec:
+  schedule: {{ printf "0 %d,%d * * *" $hourOffset (add $hourOffset 12) | quote }}
+  concurrencyPolicy: Forbid
+  startingDeadlineSeconds: 3600
+  jobTemplate:
+    spec:
+      activeDeadlineSeconds: {{ $.Values.ingestLimitSeconds }}
```
theosanderson, line 22: I believe this will stop the job after 30 mins at which point it will no longer approve stuff still coming back from prepro?

theosanderson: I basically think we do need at minimum some deployment-y thing for approvals which should always be running?

kubernetes/loculus/templates/ingest-deployment.yaml
```diff
@@ -1,9 +1,91 @@
 {{- $dockerTag := include "loculus.dockerTag" .Values }}
 {{- if not .Values.disableIngest }}
-{{- range $_, $item := (include "loculus.enabledOrganisms" . | fromJson).organisms }}
+{{- range $i, $item := (include "loculus.enabledOrganisms" . | fromJson).organisms }}
 {{- $key := $item.key }}
 {{- $organismContent := $item.contents }}
 {{- if $organismContent.ingest }}
+{{- if $.Values.ingestAsCronJob }}
```
theosanderson, line 7: Ideally it would be great to avoid doing things two different ways as otherwise we basically lose our coverage of one branch from previews etc

kubernetes/loculus/templates/ingest-deployment.yaml
anna-parker: Ah yeah we could extend the limit for a bit longer to ensure all is approved, but tbf if we have to wait 12h for newly processed ingested INSDC sequences to be approved on prod its not too bad IMO.

kubernetes/loculus/templates/ingest-deployment.yaml
anna-parker: that is true... but I recall we tried really hard to have cronjobs on previews and this didnt really work... teh alternative would be to have deployments and we add an `ingestWaitTillStartSeconds` parameter to ensure that deployments can be properly staggered

anna-parker: However, this will mean that we will have deployments for all organisms running all the time where we request the same amount of mem/cpu (even after the initial ingest which takes only minutes is finished) - so the actual usage will be less but I worry it wont affect how much we pay, even if we decrease the actual ingest frequency

theosanderson: If our goal is just to be able to have less nodes we can just allocate less CPU requests for ingest and it will just mean that if we saturate our reduced number of nodes then ingest will run a bit slowly. So in terms of saving money, we can basically do that without any of these changes.

I'm not saying staggering couldn't be good though - it could reduce the intensity of load on the backend, which might be nice. But IMO if we do it it should be as part of a holistic thing. (I do think there are various ways one can imagine doing that.) Otherwise I just worry we'll break ingest specifically for PPX and not know, which would be pretty bad and not too hard to do.

anna-parker: > then ingest will run a bit slowly

Sadly ingest will fail without enough ram (I saw this on GS)

anna-parker: But yeah I completely agree that we would ideally not have a different set up between production and previews for testing purposes

anna-parker: What I kinda dislike is now saying we should keep ingest a deployment on production because we had to make it a deployment on previews as a hack. 

Because IMO ingest should just be a simple cronjob that runs multiple times a day and can easily be staggered

theosanderson: Yeah I think there's a couple of different things:
- Approval IMO should be a deployment (or we should implement the auto-approval feature in the backend or whatever). Having to wait 12 hrs in some circs sounds not great for me. We could potentially just have one deployment that would do all the organsisms.
- I agree that otherwise, ingest should be a cronjob. I think most of the problems we had with the cronjob related to the reloader - I do think we could get something working both for previews and for prod

anna-parker: > I do think we could get something working both for previews and for prod

Lets think about this cause yeah I also think it would be the best solution! Lets say we switch to cronjobs in previews, we can define when they are run (to allow staggered and less frequent runs on prod) but we would still need to ensure the cronjob runs when the preview is launched. For previews I guess we could get cronjobs to start every minute unless a cronjob is already running? We would have to check this works

anna-parker: I modified this to now reuse as much code as possible to hopefully not diverge between options - I will also test out making ingest a cronjob

PR #6346, opened 2026-05-05
feat(deployment): Ingest as cronjob/job

theosanderson:
## Description

Changes ingest to use cronjobs rather than deployments. So that ingest runs immediately on argoCD sync we have a job that runs postsync and which starts jobs from the cronjob. It uses a new RBAC. There is also an RBAC used to ensure two jobs never run in parallel for the same organism (`wait-for-no-other-ingest` init container).

- Prevents concurrent ingest runs for the same organism via pod locking (also prevents the revoke and regroup cronjob from running at the same time as an ingest deployment - this is an improvement to the current state)
- Staggered CronJob scheduling reduces resource contention (an organism is ingested every 30min to prevent overlapping ingest and resource overuse)

## Documentation
Docs actually said we used a cronjob already:
image.

### Potential changes
Developers should not experience any changes on previews, however as ingest now only runs AFTER a successful sync if any pod on your preview does not become healthy after a sync ingest will only run as a staggered cronjob (unless you trigger it yourself)/

🚀 Preview: https://claude-organism-ingest-cr.loculus.org


kubernetes/loculus/templates/ingest.yaml
```diff
@@ -0,0 +1,297 @@
...
+            app: loculus
+            component: loculus-ingest-cronjob-{{ $key }}
+        spec:
+          {{- include "possiblePriorityClassName" $ | nindent 10 }}
+          restartPolicy: Never
+          containers:
+            - name: ingest-{{ $key }}
+              image: {{ $organismContent.ingest.image}}:{{ $dockerTag }}
+              imagePullPolicy: {{ $.Values.imagePullPolicy }}
+              resources:
```
anna-parker: an existing issue but its weird we have resources defined explicitly here - it should really use the same ones as ingest for that organism

theosanderson: > the same ones as ingest for that organism

I think atm we don't have the concept of specific resources for ingest per organism? (Maybe that's what you're saying - I guess I wouldn't add this in this PR as there's enough going on)

theosanderson: ah sorry, I misread where this comment was made, yeah, I understand now

theosanderson:
@codex review

kubernetes/loculus/templates/ingest.yaml
```diff
@@ -0,0 +1,297 @@
...
+                    fi
+                    # No peer in Running. Tie-break among Pending pods (typically all in init):
+                    # lowest pod name by lexicographic order wins, others wait until it transitions.
+                    LEADER=$(echo "$ALL" | awk '$2=="Pending" {print $1}' | sort | head -n 1)
+                    if [ -z "$LEADER" ] || [ "$LEADER" = "$HOSTNAME" ]; then
+                      echo "No conflicting ingest pods for {{ $key }}, proceeding."
+                      exit 0
+                    fi
+                    echo "Tie-break for {{ $key }}: yielding to lower-named peer $LEADER"
+                    sleep 10
```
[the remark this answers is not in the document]
theosanderson: I think this is OK, they can time out. It's better than having any potential for race conditions where we get double ingests

kubernetes/loculus/templates/ingest.yaml
```diff
@@ -0,0 +1,297 @@
+{{- $dockerTag := include "loculus.dockerTag" .Values }}
+{{- if not .Values.disableIngest }}
+{{- $ingestOrganisms := list }}
+{{- range $_, $item := (include "loculus.enabledOrganisms" . | fromJson).organisms }}
+  {{- if $item.contents.ingest }}
+  {{- $ingestOrganisms = append $ingestOrganisms $item }}
+  {{- end }}
+{{- end }}
+{{- $ingestCount := len $ingestOrganisms }}
+{{- if gt $ingestCount 0 }}
```
anna-parker: in the approve job we only used `disableIngest` as a gate - maybe we should do the same here (just to make it a tad simpler - also its unlikely we would enable ingest and have no organisms)

kubernetes/loculus/templates/ingest.yaml
```diff
@@ -0,0 +1,297 @@
...
+{{/*
+RBAC so each ingest pod can list other ingest pods to avoid the CronJob
```
anna-parker, line 11: note: by default ServiceAccounts are namespace scoped (thats good)

kubernetes/loculus/templates/ingest.yaml
```diff
@@ -0,0 +1,297 @@
...
+metadata:
+  name: loculus-ingest-trigger
+subjects:
+  - kind: ServiceAccount
+    name: loculus-ingest-trigger
+roleRef:
+  kind: Role
+  name: loculus-ingest-trigger
+  apiGroup: rbac.authorization.k8s.io
+{{/*
```
anna-parker, line 75: shouldnt the scheduling be in a new "YAML document? i.e. shoudlnt we add `---` here?

kubernetes/loculus/templates/ingest.yaml
```diff
@@ -0,0 +1,297 @@
...
+{{- range $i, $item := $ingestOrganisms }}
+  {{- $key := $item.key }}
+  {{- $organismContent := $item.contents }}
+  {{- $minute := mul (mod $i 2) 30 }}
+  {{- $hourOffset := div $i 2 }}
+  {{- $hourExpr := "*" }}
+  {{- if gt $periodHours 1 }}
+    {{- $hourExpr = printf "%d-23/%d" $hourOffset $periodHours }}
+  {{- end }}
+  {{- $schedule := printf "%d %s * * *" $minute $hourExpr }}
```
anna-parker, line 90: note: %d means format as a digit, %s means format as a string

kubernetes/loculus/templates/ingest.yaml
```diff
@@ -0,0 +1,297 @@
...
+subjects:
+  - kind: ServiceAccount
+    name: loculus-ingest-trigger
+roleRef:
+  kind: Role
+  name: loculus-ingest-trigger
+  apiGroup: rbac.authorization.k8s.io
+{{/*
+Schedule one organism per 30-minute slot so ingest runs are staggered.
+Period (hours) = ceil(N/2); organism i runs at minute (i%2)*30 of hour
```
anna-parker, line 77: Note: I guess the only thing to consider here is that we have hardcoded how many organisms are ingested in an hour but not the frequency that an organisms is ingested in a day, so this means that the frequency an organism is ingested will become lower and lower the more organisms we add, instead of keeping the frequency a single organism is ingested stable (e.g. ingest once a day) and then changing time between ingesting to match this

kubernetes/loculus/templates/ingest.yaml
```diff
@@ -0,0 +1,297 @@
...
+            - |
+              set -eu
+              JOB_NAME="loculus-ingest-bootstrap-{{ $key }}-$(date +%s)"
+              echo "Creating $JOB_NAME from cronjob/loculus-ingest-cronjob-{{ $key }}"
+              kubectl create job --from=cronjob/loculus-ingest-cronjob-{{ $key }} "$JOB_NAME"
+---
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: loculus-revoke-and-regroup-cronjob-{{ $key }}
```
anna-parker, line 160: maybe this should also get the ServiceAccountName tag as it would also be good to not run this at the same time as a different ingest job

kubernetes/loculus/templates/ingest.yaml
```diff
@@ -0,0 +1,297 @@
...
+    argocd.argoproj.io/sync-options: Replace=true
+spec:
+  schedule: {{ $schedule | quote }}
+  concurrencyPolicy: Forbid
+  startingDeadlineSeconds: 60
+  successfulJobsHistoryLimit: 1
+  failedJobsHistoryLimit: 2
+  jobTemplate:
+    spec:
+      activeDeadlineSeconds: {{ $.Values.ingestLimitSeconds }}
```
anna-parker, line 106: I dont see `ingestLimitSeconds` in the values.yaml?

theosanderson: https://github.com/loculus-project/loculus/blob/92b6d20faa70a6424017bf6b96574afcd54dbf5a/kubernetes/loculus/values.yaml#L46 ?

kubernetes/loculus/templates/ingest.yaml
theosanderson: yeah this wasn't a real decision made here, would be happy to use whatever logic you had!

kubernetes/loculus/templates/ingest.yaml
theosanderson: @theosanderson-agent can you do this and push to this branch

theosanderson-agent: Done in dfdec4407 — dropped the `gt $ingestCount 0` gate so `disableIngest` is the only gate, matching the autoapprove deployment. helm lint passes.

kubernetes/loculus/templates/ingest.yaml
anna-parker: ah true sorry - was searching in the PPX repo

kubernetes/loculus/templates/ingest.yaml
```diff
@@ -0,0 +1,297 @@
...
+                      name: slack-notifications
+                      key: slack-hook
+                - name: NCBI_API_KEY
+                  valueFrom:
+                    secretKeyRef:
+                      name: ingest-ncbi
+                      key: api-key
+              args:
+                - snakemake
+                - results/submitted
```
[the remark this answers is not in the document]
anna-parker: we should probably drop the pause rule

kubernetes/loculus/templates/ingest.yaml
```diff
@@ -0,0 +1,295 @@
...
+                  subPath: config.yaml
+          volumes:
+            - name: loculus-ingest-config-volume-{{ $key }}
+              configMap:
+                name: loculus-ingest-config-{{ $key }}
+          {{- end }}
+{{/*
+Sync-hook trigger: kicks off a one-off ingest Job from the CronJob's
+template whenever argocd syncs. The trigger itself exits in seconds
+(so it doesn't block the sync); the spawned ingest Job runs untracked
```
anna-parker: this sounds a bit concerning: `the spawned ingest Job runs untracked by argocd`

kubernetes/loculus/templates/ingest.yaml
```diff
@@ -0,0 +1,297 @@
...
+      serviceAccountName: loculus-ingest-trigger
+      containers:
+        - name: trigger
+          image: alpine/kubectl:1.36.0
+          command:
+            - sh
+            - -c
+            - |
+              set -eu
+              JOB_NAME="loculus-ingest-bootstrap-{{ $key }}-$(date +%s)"
```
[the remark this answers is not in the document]
anna-parker: I dont think this is a problem

kubernetes/loculus/templates/ingest.yaml
```diff
@@ -0,0 +1,295 @@
...
```
anna-parker, line 153: Can we call this loculus-ingest-triggered- instead of bootstrap?

kubernetes/loculus/templates/ingest.yaml
theosanderson: hmm good point - I recall seeing the job in argoCD - @theosanderson-agent any thoughts?

kubernetes/loculus/templates/ingest.yaml
```diff
@@ -0,0 +1,295 @@
...
+              echo "Creating $JOB_NAME from cronjob/loculus-ingest-cronjob-{{ $key }}"
+              kubectl create job --from=cronjob/loculus-ingest-cronjob-{{ $key }} "$JOB_NAME"
+---
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: loculus-revoke-and-regroup-cronjob-{{ $key }}
+  annotations:
+    argocd.argoproj.io/sync-options: Replace=true
+spec:
```
anna-parker, line 163: In general I wonder if we can reuse some of the code from loculus-ingest-cronjob-{{ $key }}

kubernetes/loculus/templates/ingest.yaml
theosanderson-agent: Going through the chain:

- **Trigger Job** (`loculus-ingest-trigger-{key}`) is a proper argocd `Sync` hook — argocd creates it, watches it, and reports its status. So argocd sees this one.
- **CronJob** is a regular tracked manifest, and any Jobs the CronJob controller spawns on schedule carry an `ownerReference` back to the CronJob, so argocd surfaces those under the CronJob in its tree (this is what you usually see in the UI).
- **Bootstrap Job** (`loculus-ingest-bootstrap-{key}-{ts}`, the one the trigger runs via `kubectl create job --from=cronjob/...`) is the one the comment is about. `kubectl create --from=cronjob` does *not* set an `ownerReference`, and we are not adding any argocd tracking annotation, so this Job is not in the desired state and not linked to a tracked parent. Argocd will still *show* it in the namespace view (and the pod spec carries `app: loculus` etc., so it tends to render alongside the rest of the app), but it is not part of `argocd app diff`/`sync`/health.

So the comment is technically accurate but maybe a bit alarmist — what is untracked is just the one-shot startup run; the steady-state CronJob-spawned Jobs are tracked normally via owner refs.

If we want it to appear cleanly under the CronJob in the argocd tree, the options are:

1. **Add an explicit argocd tracking annotation** to the bootstrap Job at creation time (e.g. `kubectl annotate` after `kubectl create`). Keeps it visible without K8s GC implications.
2. **Set `ownerReferences`** on the bootstrap Job pointing at the CronJob — neat tree, but risky because the CronJob has `argocd.argoproj.io/sync-options: Replace=true`, so a sync would cascade-delete an in-flight bootstrap Job.
3. **Drop the trigger+bootstrap split** and make the trigger Job itself do the ingest as a `PostSync` hook. Cleanest from argocd visibility, but the sync would then block on the full ingest run — which is exactly what we wanted to avoid.

Happy to soften the comment in the template and/or implement (1) if Anna wants stricter visibility — let me know which direction you want to go.

kubernetes/loculus/templates/ingest.yaml
theosanderson: @theosanderson-agent - address this and push to this PR, use same resources as the other job

theosanderson-agent: Done in 2a59325eb — replaced the hard-coded `resources:` block in the revoke-and-regroup CronJob with `{{ include "loculus.resources" (list "ingest" $.Values) }}`, so it now uses the same resources as the main ingest CronJob.

anna-parker: thanks @theosanderson !!!

kubernetes/loculus/templates/ingest.yaml
anna-parker: I think this should be ok then, thanks for clarifying :-)

kubernetes/loculus/templates/ingest.yaml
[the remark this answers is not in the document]
anna-parker: @theosanderson-agent could you make this change?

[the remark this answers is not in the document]
theosanderson-agent: Done in 3c357a6 — dropped the `pause` rule, removed `results/paused` from `rule all`, and removed the now-unused `cycle_sleep_seconds` default.

kubernetes/loculus/templates/ingest.yaml
theosanderson: The `---` separates kubernetes things, so it needs to be within the `range`. But not disagreeing that we could have more spacing to give separation from the RBAC

kubernetes/loculus/templates/ingest.yaml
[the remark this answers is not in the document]
anna-parker: @theosanderson-agent can you add the `serviceAccountName` and try to reuse some of the code between cronjobs?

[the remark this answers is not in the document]
theosanderson-agent: Done in 7db802e8c. Re-extracted `_ingest-pod-spec.tpl` (the helper had been inlined back in bad6f9b2e when the sync-hook Job was the only other consumer; it now uses `kubectl create job --from` so it does not need the spec). Both CronJobs now `include` it — `args` is the only parameter that differs.

This means the revoke-and-regroup CronJob now also gets:
- `serviceAccountName: loculus-ingest-lock`
- the `wait-for-no-other-ingest` init container
- the `ingest-organism` / `loculus-ingest-runner` labels

so a manual revoke run will block on (and be visible to) any concurrent scheduled ingest for the same organism.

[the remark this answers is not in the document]
theosanderson: it's not really adding the serviceAccountName per se - more adding the whole locking mechanism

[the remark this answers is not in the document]
theosanderson: (but the agent seems to have got the right idea afaict)

kubernetes/loculus/templates/ingest.yaml
```diff
@@ -0,0 +1,297 @@
...
+      ttlSecondsAfterFinished: 86400
+      template:
+        metadata:
+          labels:
+            app: loculus
+            component: loculus-ingest-cronjob-{{ $key }}
+            ingest-organism: {{ $key }}
+            loculus-ingest-runner: "true"
+        spec:
+          {{- include "possiblePriorityClassName" (dict "Values" $.Values) | nindent 10 }}
```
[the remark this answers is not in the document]
theosanderson: @theosanderson-agent let's address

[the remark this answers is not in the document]
theosanderson-agent: Done in c2ebb62d2 — switched to `include "possiblePriorityClassName" .` in `_ingest-pod-spec.tpl`. The dict passed into the helper already has `Values`, so forwarding `.` matches the convention used everywhere else.

kubernetes/loculus/templates/ingest.yaml
anna-parker: ah ok then no worries we can keep as is

kubernetes/loculus/templates/ingest.yaml
anna-parker: I guess an alternative would be - but I guess this isnt too important
```
N = # organisms
K = daily ingest frequency on an organism level 
total daily runs = N * K
slot size = 30 minutes
available slots per day = 48
capacity per slot = ceil((N * K) / 48)
```

anna-parker:
@theosanderson-agent the ingest jobs that are triggered by the sync will fail multiple times before completing leading to the pods being red on argocd. The ingest pods fail because they cannot talk to the backend ro the ena deposition pod. Would it be possible to use `argocd.argoproj.io/sync-wave: "20"` to ensure that the ingest sync triggers after the backend and ena deposition pods complete? or would it be better to enable retries in ingest for backend and ena deposition calls?

theosanderson-agent:
Good catch — switched the argocd hook from `Sync` to `PostSync` in 15992ee. PostSync only fires once argocd reports the whole application as Healthy, so the spawned ingest job no longer starts before the backend / ena-submission rollouts have settled. Pushed to this branch.

Sync-wave `"20"` would also work, but it relies on every other resource defaulting to wave 0 (currently true here, but easy to break by accident). PostSync is more explicit about the "wait for everything to be healthy" semantic.

Adding retries inside ingest for backend / ena calls is still a nice defense-in-depth — useful for transient blips outside of deploy time — but I would treat that as a separate change rather than rolling it into this PR.

anna-parker (approved):
awesome! we can maybe still modify the exact logic for how we stagger ingest for all organisms later but for now this is great and works well on the preview from looking at argocd

kubernetes/loculus/templates/ingest.yaml
```diff
@@ -0,0 +1,297 @@
...
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: loculus-ingest-cronjob-{{ $key }}
+  annotations:
+    argocd.argoproj.io/sync-options: Replace=true
+spec:
+  schedule: {{ $schedule | quote }}
+  concurrencyPolicy: Forbid
+  startingDeadlineSeconds: 60
```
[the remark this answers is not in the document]
theosanderson: @theosanderson-agent go for 600 seconds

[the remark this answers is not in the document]
theosanderson-agent: Bumped both `startingDeadlineSeconds` to 600 in c408d94 (also applied it to the revoke-and-regroup CronJob since the same concern applies there).

kubernetes/loculus/templates/ingest.yaml
anna-parker: I created a follow up issue and will close this for now

Issue #6348, opened 2026-05-05
Consider making ingest lower priority than other stuff


theosanderson:
Ingest is a bit less time-sensitive than other components of Loculus, so we could potentially use a slightly lower priority class. This would help to smooth out roll-outs by letting everything else start up before ingest kicks into gear (where necessary due to resource limitations)

theosanderson:
@theosanderson-agent what are some options for this - does it have to mean defining a new priority class?

theosanderson-agent:
No — defining a new `PriorityClass` is just one option, and probably not the simplest. Today the chart applies a single global `podPriorityClassName` to every component via [`_possiblePriorityClassName.tpl`](https://github.com/loculus-project/loculus/blob/main/kubernetes/loculus/templates/_possiblePriorityClassName.tpl). A few ways we could differentiate ingest:

1. **Per-component override (no new PriorityClass needed if the cluster already has a lower class).** Add e.g. `ingestPodPriorityClassName` (falling back to `podPriorityClassName` when unset), and have ingest + autoapprove use it. Deployments that already define a low-priority class can point this at it; deployments that do not get unchanged behavior. This is the smallest chart change.

2. **Just omit the global priority for ingest.** Same idea but inverted: drop the `possiblePriorityClassName` include from the ingest/autoapprove deployments (or gate it behind a flag). Ingest then runs at the cluster default priority while everything else runs elevated — achieves "lower than other stuff" without anyone needing to author a new class. Note that Kubernetes ships only `system-cluster-critical` / `system-node-critical` out of the box, which are higher, so "lower" generally means either a deployment-defined class or no class at all.

3. **Define a new low-priority `PriorityClass` in the chart.** Cleanest from a "its the same everywhere" perspective, but `PriorityClass` is cluster-scoped, so it needs cluster-admin RBAC at install time and can collide with cluster-operator-managed classes. Probably overkill given (1)/(2) work.

4. **Sidestep priority classes entirely with ArgoCD sync-waves.** The deployments already have ArgoCD annotations, so we could put ingest + autoapprove in a later sync-wave (`argocd.argoproj.io/sync-wave: "1"` etc.). That directly addresses the stated motivation — letting everything else come up first during rollouts — without touching scheduler priority. Downside: it only affects rollout ordering, not steady-state preemption under resource pressure.

5. **Reduce ingests resource requests / add a startup delay.** Indirect; not really what was asked but cheapest.

[...]

PR #6349, opened 2026-05-05
feat!(deployment): create PriorityClass dynamically per namespace

theosanderson-agent:
## Summary

Refs #6348.

This restructures how pod priority is configured for loculus deployments. Previously the chart referenced an externally-supplied `PriorityClass` by name through `podPriorityClassName`. That meant every cluster running loculus had to provision a `PriorityClass` out-of-band before installing the chart. [Theo edit: only if they wanted to use the priority class feature]

This PR drops `podPriorityClassName` entirely and instead lets the chart create its own `PriorityClass` resource:

- New value `podPriorityValue` (integer) in `values.yaml`/`values.schema.json`. When set, the chart renders a `PriorityClass` whose **name matches `.Release.Namespace`** and whose `value` is `podPriorityValue`. All loculus pods then reference that class via `priorityClassName: {{ .Release.Namespace }}`.
- When `podPriorityValue` is unset (the default), no `PriorityClass` is created and pods run at the cluster default priority — this is the new "no priority configured" behavior.
- The `PriorityClass` is annotated with `argocd.argoproj.io/sync-wave: "-10"` so ArgoCD creates it before the deployments that reference it.
- Tied to the namespace name so multiple loculus installs on the same cluster get distinct, non-colliding `PriorityClass` resources without any per-install naming config.
- The existing `_possiblePriorityClassName.tpl` helper now renders `priorityClassName: {{ .Release.Namespace }}` (gated on `podPriorityValue` being set), so all of the deployments that already include the helper (`backend`, `website`, `lapis`, `silo`, `keycloak`, `preprocessing`, `ingest`, `autoapprove`, `taxonomy`, `ena-submission`) are wired up automatically.

This is groundwork for #6348: now that we can manage our own `PriorityClass`, a follow-up can add a second, lower-valued class for ingest/autoapprove without requiring cluster-admin coordination.

### Migration notes for chart users

- Anyone currently passing `podPriorityClassName: foo` (e.g. in PPX `production.yaml`) will need to switch to `podPriorityValue: <integer>` to keep getting a `PriorityClass` applied. The class will then live inside the namespace's chart release rather than being externally managed.
- ArgoCD needs RBAC to create cluster-scoped `PriorityClass` objects. If that's not granted, the simplest path is to leave `podPriorityValue` unset and not use a `PriorityClass` at all.

### One subtle behavior change

The old check was `{{- if .Values.podPriorityClassName }}` (truthiness — empty string falsy). The new check is `hasKey .Values "podPriorityValue"` (presence). This means setting `podPriorityValue: 0` is now an explicit choice — it creates a `PriorityClass` with value 0 (cluster default) rather than being treated as "unset". This seemed more correct for an integer field; happy to switch to a different sentinel if preferred.

## Test plan

- [x] `helm lint kubernetes/loculus -f kubernetes/loculus/values.yaml` passes.
- [x] `helm template` with the default values: no `PriorityClass` rendered, no `priorityClassName` lines on any deployment.
- [x] `helm template ... --set podPriorityValue=100 --namespace my-test-ns`: a single `PriorityClass` named `my-test-ns` with `value: 100` is rendered, and every deployment template gets `priorityClassName: my-test-ns`.
- [x] Negative values render correctly (`value: -100`).
- [x] Verify on a real cluster that ArgoCD can create the `PriorityClass` with the `sync-wave: "-10"` annotation before the deployments that reference it.
- [x] Verify that uninstall cleans up the `PriorityClass` (it is cluster-scoped but owned by the chart release).

🚀 Preview: https://dynamic-priority-class.loculus.org

### DIFF
--- a/kubernetes/loculus/templates/priorityclass.yaml
+++ b/kubernetes/loculus/templates/priorityclass.yaml
@@ -2,7 +2,7 @@
 apiVersion: scheduling.k8s.io/v1
 kind: PriorityClass
 metadata:
-  name: {{ .Release.Namespace }}
+  name: {{ .Release.Namespace }}-priority
   annotations:
     argocd.argoproj.io/sync-wave: "-10"
 value: {{ .Values.podPriorityValue }}
--- a/kubernetes/loculus/templates/priorityclass.yaml
+++ b/kubernetes/loculus/templates/priorityclass.yaml
@@ -1,0 +1,11 @@
+{{- if hasKey .Values "podPriorityValue" }}
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: {{ .Release.Namespace }}
+  annotations:
+    argocd.argoproj.io/sync-wave: "-10"
+value: {{ .Values.podPriorityValue }}
+globalDefault: false
+description: "Priority class for namespace {{ .Release.Namespace }}, managed by the loculus Helm chart."
+{{- end }}
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -2822,4 +2822,3 @@ defaultResources:
 ingest:
   ncbiGatewayUrl: null
   mirrorBucket: null
-podPriorityValue: 10
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -2822,3 +2822,4 @@ defaultResources:
 ingest:
   ncbiGatewayUrl: null
   mirrorBucket: null
+podPriorityValue: 10
--- a/kubernetes/loculus/templates/_possiblePriorityClassName.tpl
+++ b/kubernetes/loculus/templates/_possiblePriorityClassName.tpl
@@ -1,5 +1,5 @@
 {{- define "possiblePriorityClassName" -}}
-{{- if .Values.podPriorityClassName }}
-priorityClassName: {{ .Values.podPriorityClassName }}
+{{- if hasKey .Values "podPriorityValue" }}
+priorityClassName: {{ .Release.Namespace }}
 {{- end -}}
 {{- end -}}
--- a/kubernetes/loculus/templates/_possiblePriorityClassName.tpl
+++ b/kubernetes/loculus/templates/_possiblePriorityClassName.tpl
@@ -1,5 +1,5 @@
 {{- define "possiblePriorityClassName" -}}
 {{- if hasKey .Values "podPriorityValue" }}
-priorityClassName: {{ .Release.Namespace }}
+priorityClassName: {{ .Release.Namespace }}-priority
 {{- end -}}
 {{- end -}}
--- a/kubernetes/loculus/values.schema.json
+++ b/kubernetes/loculus/values.schema.json
@@ -1931,9 +1931,9 @@
       "type": ["string", "integer"],
       "description": "TODO - this is required/set by ArgoCD"
     },
-    "podPriorityClassName": {
-      "type": "string",
-      "description": "TODO - found in PPX production.yaml"
+    "podPriorityValue": {
+      "type": "integer",
+      "description": "If set, the chart creates a PriorityClass named after the release namespace with this value, and assigns it to all loculus pods. Higher values mean higher scheduling priority; can be negative for lower-than-default priority. If unset, no PriorityClass is created and pods run at the cluster default priority."
     },
     "testconfig": {
       "description": "TODO"


theosanderson:
I think this is much nicer in itself, it was always a faff that you had to create this manually. It also unlocks doing dynamic stuff for ingest later

kubernetes/loculus/templates/priorityclass.yaml
```diff
@@ -0,0 +1,11 @@
+{{- if hasKey .Values "podPriorityValue" }}
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: {{ .Release.Namespace }}
+  annotations:
+    argocd.argoproj.io/sync-wave: "-10"
+value: {{ .Values.podPriorityValue }}
```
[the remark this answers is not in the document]
theosanderson: I think this is ok as is

kubernetes/loculus/values.yaml
```diff
@@ -2822,3 +2822,4 @@ defaultResources:
 ingest:
   ncbiGatewayUrl: null
   mirrorBucket: null
+podPriorityValue: 10
```
theosanderson: revert

anna-parker: I wonder if we should still set to the default 0 here

kubernetes/loculus/templates/priorityclass.yaml
```diff
@@ -0,0 +1,11 @@
+{{- if hasKey .Values "podPriorityValue" }}
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: {{ .Release.Namespace }}
```
theosanderson: @theosanderson-agent let's name it [namespace]-priority and update throughout

kubernetes/loculus/values.yaml
theosanderson: I don't think so - it adds complexity by creating a priority class when we don't really need one in this case. As the PR desc highlights some ArgoCDs may not like creating cluster-level resources so it's nice to provide an option which doesn't need that

kubernetes/loculus/templates/priorityclass.yaml
anna-parker: as discussed with @theosanderson async this should be name `{{ .Release.Namespace }}-priority` to prevent name collision with existing priority classes and issues with rollout

theosanderson-agent: Done in 14149471. Renamed the PriorityClass to `{{ .Release.Namespace }}-priority` and updated the `possiblePriorityClassName` helper accordingly. All `priorityClassName` references go through the helper, so no other call sites needed changes. `helm lint` passes and `helm template -n loctest` renders `loctest-priority` consistently.

theosanderson:
@theosanderson-agent can you draft a PR for the pathoplexus/pathoplexus repo to go with this change

theosanderson-agent:
Drafted as pathoplexus/pathoplexus#924 — switches the three env files (`demo.yaml`, `staging.yaml`, `production.yaml`) from `podPriorityClassName: <env>` to `podPriorityValue: 10` (matching the new chart default). Flagged in the PR description that the `10` value should be confirmed against whatever the externally-managed `demo` / `staging` / `production` PriorityClasses currently have, and that it should not be merged until this PR lands and `loculusVersion` is bumped.

theosanderson:
@theosanderson-agent fix up with values from https://github.com/pathoplexus/toplevel_gitops/blob/main/priority-classes.yaml

[edit: actually agent probs can't access this repo]

theosanderson-agent:
Confirmed I can't access `pathoplexus/toplevel_gitops` (private to me). Could you paste the relevant `value:` fields from `priority-classes.yaml` here, or just tell me the integers for `demo` / `staging` / `production`? I'll then update pathoplexus/pathoplexus#924 accordingly.

theosanderson:
@theosanderson-agent 
<details>
apiVersion: scheduling.k8s.io/v1
kind: PriorityClass
metadata:
  name: high-priority
value: 5000
globalDefault: false
description: "This priority class should be used by high priority pods which are required for debugging and system functionality."
---
apiVersion: scheduling.k8s.io/v1
kind: PriorityClass
metadata:
  name: production
value: 2000
globalDefault: false
description: "This priority class should be used for production environment pods only."
---
apiVersion: scheduling.k8s.io/v1
kind: PriorityClass
metadata:
  name: staging
value: 1000
globalDefault: false
description: "This priority class should be used for staging environment pods only."
---
apiVersion: scheduling.k8s.io/v1
kind: PriorityClass
metadata:
  name: demo
value: 500
globalDefault: false
description: "This priority class should be used for demo environment pods only."
</details>

theosanderson-agent:
Updated pathoplexus/pathoplexus#924 with the matching integer values from your existing `PriorityClass` resources:

- `demo.yaml`: `podPriorityValue: 500`
- `staging.yaml`: `podPriorityValue: 1000`
- `production.yaml`: `podPriorityValue: 2000`

So the chart-created `<namespace>-priority` classes will have the same numeric priority as the externally-managed `demo`/`staging`/`production` classes they replace, preserving relative scheduling order against any non-loculus workloads. Left `high-priority` (5000) alone — that's a separate cluster-debug class loculus doesn't use.

anna-parker (approved):
thanks!

PR #6351, opened 2026-05-06
chore(deploy): download urls in parallel in config-processor

### LINKED ISSUE #6350: deploy.py via config-processor.py takes very long (~2m) for Pathoplexus config
Just ran `../generate_local_test_config.sh --values ../loculus_values/values.yaml --from-live --live-host preview-main.pathoplexus.org` in the Pathoplexus repo and was surprised this took 2m4s.

There might be simple ways to speed it up. Gut feeling is that the slow part is making dozens of network requests to fetch sequences from Github pages - this could be parallelized.

~IIRC, deploy.py is run on each deployment on Argo, so we also need to wait 2m there right now, so~ speeding up could be beneficial not just for local dev but also for deployment smoothness (through faster init container processing)

Image

corneliusroemer:
Resolves #6350

Reduces time to process PPX config down from 2m to 6s:

```
./generate_local_test_config.sh --values ~/code/pathoplexus/loculus_values/values.yaml --from-live --live-host preview-main.pathoplexus.org
```

Some savings also transfer to faster pod startup (via reduced time of init container) of backend/website, where this script also does lots of downloading. For example backend init container on Loculus takes only 6s vs 18s. On PPX this is even more marked, we might save almost a minute.

Before (on Loculus main):

```
../generate_local_test_config.sh --from-live  6.46s user 0.72s system 39% cpu 18.193 total
```

After:

```
../generate_local_test_config.sh --from-live  3.99s user 0.44s system 100% cpu 4.403 total
```

### Screenshot

image

Also works in init container on ArgoCD:

image

🚀 Preview: https://speed-up-deploy.loculus.org

### DIFF
--- a/kubernetes/config-processor/config-processor.py
+++ b/kubernetes/config-processor/config-processor.py
@@ -1,8 +1,13 @@
 import os
 import re
 import shutil
+import threading
+from concurrent.futures import ThreadPoolExecutor, as_completed
 
 import requests
+
+DEFAULT_MAX_WORKERS = 16
+thread_local = threading.local()
 
 
 def copy_structure(input_dir, output_dir):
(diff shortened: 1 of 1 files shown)


theosanderson:
@codex review

theosanderson (approved):

PR #6362, opened 2026-05-07
feat(silo-importer): refactor code and simplify config

anna-parker:
resolves #

### PR Checklist

🚀 Preview: https://anya-refactor.loculus.org

### DIFF
--- a/docs/src/content/docs/reference/helm-chart-config.mdx
+++ b/docs/src/content/docs/reference/helm-chart-config.mdx
@@ -85,7 +85,7 @@ Note: Lineages should be determined by the preprocessing pipeline and not the us
 
 For metadata fields whose values form a hierarchy that's resolved at SILO import
 time (rather than fetched as a static file per pipeline version), set
-`hierarchicalFilter: <URL>`, where URL is the URL of the service that will create the 
+`hierarchicalFilter: <URL>`, where URL is the URL of the service that will create the
 hierarchical file for SILO. The importer collects observed
 values for the field, sends them to that service, and writes the returned YAML
 into SILO's input directory as `<MetadataName>.yaml`.
--- a/docs/src/content/docs/reference/helm-chart-config.mdx
+++ b/docs/src/content/docs/reference/helm-chart-config.mdx
@@ -85,16 +85,11 @@ Note: Lineages should be determined by the preprocessing pipeline and not the us
 
 For metadata fields whose values form a hierarchy that's resolved at SILO import
 time (rather than fetched as a static file per pipeline version), set
-`hierarchicalFilter: <name>` on the metadata field and add a corresponding entry under
-`hierarchicalFilterSources`. Each entry maps a name to the base URL of a
-service that provides the SILO lineage file. The importer collects observed
+`hierarchicalFilter: <URL>`, where URL is the URL of the service that will create the 
+hierarchical file for SILO. on the metadata field and add a corresponding entry under
+`hierarchicalFilterSources`. The importer collects observed
 values for the field, sends them to that service, and writes the returned YAML
-into SILO's input directory as `<name>.yaml`.
-
-```yaml
-hierarchicalFilterSources:
-  hostTaxon: http://loculus-taxonomy-service:5000
-```
+into SILO's input directory as `<MetadataName>.yaml`.
 
 <SchemaDocs group='hierarchical-filter-sources' fieldColumnClass='w-48' />
 
--- a/docs/src/content/docs/reference/helm-chart-config.mdx
+++ b/docs/src/content/docs/reference/helm-chart-config.mdx
@@ -86,8 +86,7 @@ Note: Lineages should be determined by the preprocessing pipeline and not the us
 For metadata fields whose values form a hierarchy that's resolved at SILO import
 time (rather than fetched as a static file per pipeline version), set
 `hierarchicalFilter: <URL>`, where URL is the URL of the service that will create the 
-hierarchical file for SILO. on the metadata field and add a corresponding entry under
-`hierarchicalFilterSources`. The importer collects observed
+hierarchical file for SILO. The importer collects observed
 values for the field, sends them to that service, and writes the returned YAML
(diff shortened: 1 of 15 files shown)


loculus-silo/src/silo_import/runner.py
```diff
@@ -79,7 +80,30 @@ def run_once(self) -> None:
...
+                )
+            }
+            if filters_to_update:
+                update_hierarchical_filters(
+                    new_values,
+                    self.config,
+                    self.paths,
+                    metadata_to_update=filters_to_update,
+                )
+                for field in filters_to_update:
```
anna-parker: this means we are storing all of the values in the silo importer

docs/src/content/docs/reference/helm-chart-config.mdx
```diff
@@ -85,16 +85,10 @@ Note: Lineages should be determined by the preprocessing pipeline and not the us
 
 For metadata fields whose values form a hierarchy that's resolved at SILO import
 time (rather than fetched as a static file per pipeline version), set
-`hierarchicalFilter: <name>` on the metadata field and add a corresponding entry under
-`hierarchicalFilterSources`. Each entry maps a name to the base URL of a
-service that provides the SILO lineage file. The importer collects observed
+`hierarchicalFilter: <URL>`, where URL is the URL of the service that will create the
```
maverbiest, line 88: Ahh okay so the metadata field name determines which handler will be used in the end to get the lineage file? I like it!

loculus-silo/src/silo_import/lineage.py
```diff
@@ -46,28 +46,27 @@ def update_lineage_definitions(
...
-        values = values_by_kind.get(kind, set())
-        match kind:
-            # If we add new filter kinds, we should implement a corresponding
-            # handler function and call it here
-            case HierarchicalFilterKind.HOST_TAXON:
-                update_taxonomic_lineage(kind.value, values, hf.url, paths)
+        url = config.hierarchical_filters.get(field)
+        if not url:
+            continue
+        update_taxonomic_lineage(field, new_values, url, paths)
```
maverbiest, line 67: So is the idea to now just always call the taxonomy service, and add a match statement here if/when we decide to add further types of hierarchical filter?

loculus-silo/src/silo_import/config.py
```diff
@@ -109,26 +96,4 @@ def _parse_hierarchical_filters(
...
-                kind=kind,
-                url=entry["url"],
-                metadata_field=entry["metadataField"],
-            )
-        except (KeyError, TypeError) as exc:
-            msg = f"HIERARCHICAL_FILTERS entry for '{filter_name}' is malformed: {entry}"
-            raise RuntimeError(msg) from exc
-
-    return filters
+    return data or None
```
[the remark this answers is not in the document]
anna-parker: eh shoudl be fine

loculus-silo/src/silo_import/lineage.py
```diff
@@ -46,28 +46,25 @@ def update_lineage_definitions(
...
             continue
-        values = values_by_kind.get(kind, set())
-        match kind:
-            # If we add new filter kinds, we should implement a corresponding
-            # handler function and call it here
-            case HierarchicalFilterKind.HOST_TAXON:
-                update_taxonomic_lineage(kind.value, values, hf.url, paths)
+        if not url:
+            continue
+        update_taxonomic_lineage(field, new_values, url, paths)
```
maverbiest, line 67:
```suggestion
        # If we add  more hierarchical filters in the future, we should match on the metadata field 
        # name and call the appropriate handler
        update_taxonomic_lineage(field, new_values, url, paths)
```

loculus-silo/src/silo_import/lineage.py
```diff
@@ -46,28 +46,25 @@ def update_lineage_definitions(
...
 def update_hierarchical_filters(
-    values_by_kind: dict[HierarchicalFilterKind, set[str]],
+    metadata_values: dict[MetadataField, set[str]],
+    old_metadata_values: dict[MetadataField, set[str]],
     config: ImporterConfig,
     paths: ImporterPaths,
-    *,
-    subset: set[HierarchicalFilterKind] | None = None,
 ) -> None:
     """Dispatch each configured filter to its dedicated handler.
```
maverbiest:
```suggestion
    """Dispatch each configured filter to the taxonomy filter handler.
```

maverbiest (commented):
The simplification to the config is great, thanks! 

I do have questions about `update_hierarchical_filters`. Now we're just always calling `update_taxonomic_lineage` for each field with a hierarchicalFilter. This works of course since we have the one, but eventually we'll have to add something to match the approriate handler to each field, right?

Otherwise I'm happy to merge this into my branch

loculus-silo/src/silo_import/lineage.py
anna-parker: its only called when there is a change in the values of that field (it should be exactly what you did in run_once just moved to this function)

loculus-silo/src/silo_import/lineage.py
anna-parker: I dont think this is required? the url is filter specific - the only thing we assume is that the endpoint `/silo-lineage` exists, which yes might have to be changed in future but I think thats ok

maverbiest: Well we also assume that we get the lineage from the endpoint by providing a request body with {tax_ids: [1, 2, ...]} (happens in `_post_taxonomic_lineage`). This won't make sense for e.g. a geographic filter, so somehow we should figure out which service we're calling?

What we could do is define some sort of contract that all /silo-lineage endpoints should adhere to, so we won't provide {tax_ids: [...]} but {values: [...]} or something. 

Won't work if we want to get values from a service we don't control but that's ofc a hurdle we can take if we ever get there (and quite unlikely)

anna-parker: I think this is fine for now, until we actually add more services this starts getting into premature optimization and adding code that we might not need

PR #6363, opened 2026-05-07
test(integration): fix racy URL construction by preventing double `?`

corneliusroemer:
It's buggy to add a URL param by appending `?param` as this requires the URL to not end in ? or have other params set:

```
    await page.goto(`${page.url()}?geoLocCountry=${NULL_QUERY_VALUE}`);
```

This caused flakes in tests as we automatically add trailing `?` in various places in react code. Tests would pass when playwright was faster than the addition of the `?` - but sometimes playwright was fast enough causing failures.

I spotted this by looking at playwright trace and seeing the URL being `http://localhost:3000/dummy-organism-with-files/submission/6/released??column_submissionId=true`

Now, we construct the full URL with params upfront or safely add extra params using a helper `setUrlParamsAndGoTo`

This fixes this test failure (or at least one way this would show up sporadically):

```
1) [firefox-without-dep] › tests/specs/features/file-sharing.spec.ts:102:5 › bulk revise 2 seqs with files

    Test timeout of 300000ms exceeded.

    Error: page.reload: Test timeout of 300000ms exceeded.
    Call log:
      - waiting for navigation until "load"

       at pages/search.page.ts:221

      219 |     async waitForSequences(role: 'link' | 'cell', name: string | RegExp) {
      220 |         while (!(await this.page.getByRole(role, { name: name }).isVisible())) {
    > 221 |             await this.page.reload();
          |                             ^
      222 |             await this.page.waitForTimeout(2000);
      223 |         }
      224 |     }
```

I tried to find all instances of this bug across the codebase and identified (and fixed) one more instance.

### DIFF
--- a/integration-tests/tests/utils/setUrlParamsAndGoTo.ts
+++ b/integration-tests/tests/utils/setUrlParamsAndGoTo.ts
@@ -3,18 +3,14 @@ import { Page } from '@playwright/test';
 type UrlParamValue = string | number | boolean;
 
 /**
- * Safe way to add/update/remove URL search params and navigate to the new URL. Handles relative URLs correctly.
+ * Safe way to add/update URL search params and navigate to the new URL.
  * Should be used in place of `page.goto(page.url() + '?column_submissionId=true');`
  */
 export async function setUrlParamsAndGoTo(page: Page, params: Record<string, UrlParamValue>) {
     const url = new URL(page.url());
 
     for (const [key, value] of Object.entries(params)) {
-        if (value === undefined || value === null) {
-            url.searchParams.delete(key);
-        } else {
-            url.searchParams.set(key, String(value));
-        }
+        url.searchParams.set(key, String(value));
     }
 
     await page.goto(url.toString());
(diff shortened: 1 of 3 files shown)


integration-tests/tests/utils/setUrlParamsAndGoTo.ts
```diff
@@ -0,0 +1,17 @@
+import { Page } from '@playwright/test';
+
+type UrlParamValue = string | number | boolean;
+
+/**
+ * Safe way to add/update URL search params and navigate to the new URL.
+ * Should be used in place of `page.goto(page.url() + '?column_submissionId=true');`
```
anna-parker, line 7: we have some code that uses this in the `buildUrlFromParams` - is this also a potential issue?

anna-parker (approved):
thanks!

integration-tests/tests/utils/setUrlParamsAndGoTo.ts
corneliusroemer: Yeah, where? I've grepped for `buildUrlFromParams` but couldn't find a single hit

corneliusroemer: I'm resolving so I can merge - will have a look if you point me at something concrete

anna-parker: @corneliusroemer the function buildUrlFromParams

PR #6364, opened 2026-05-07
feat(website): refactor website code

anna-parker:
only reason why there is more code is I modified the tests to use hostTaxa explicitly

### PR Checklist

### DIFF
--- a/kubernetes/loculus/templates/_common-metadata.tpl
+++ b/kubernetes/loculus/templates/_common-metadata.tpl
@@ -347,8 +347,8 @@ organisms:
   {{- if .hierarchicalFilter }}
   hierarchicalSearch: true
   {{- end }}
-  {{- if and .hierarchicalFilter .hierarchicalSearchText }}
-  hierarchicalSearchText: {{ .hierarchicalSearchText }}
+  {{- if and .hierarchicalFilter .hierarchicalSearchLabel }}
+  hierarchicalSearchLabel: {{ .hierarchicalSearchLabel }}
   {{- end }}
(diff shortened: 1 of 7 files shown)


kubernetes/loculus/templates/_common-metadata.tpl
```diff
@@ -347,6 +347,9 @@ organisms:
   {{- if .hierarchicalFilter }}
   hierarchicalSearch: true
   {{- end }}
+  {{- if and .hierarchicalFilter .hierarchicalSearchText }}
```
maverbiest: Is `hierarchicalSearchText` just for the checkbox label? If so I would prefer something like `checkBoxLabel`  (or even `hierarchicalSearchLabel`)

kubernetes/loculus/templates/_common-metadata.tpl
```diff
@@ -347,6 +347,9 @@ organisms:
...
+  hierarchicalSearchText: {{ .hierarchicalSearchText }}
```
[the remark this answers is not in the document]
maverbiest: Or we should make sure to quote it in the values.yaml I guess (does that carry over?)

maverbiest (approved):
This makes much more sense, thanks! The only thing I dislike is the name `hierarchicalSearchText`, as its function is a bit unclear to me based on the name

(and we need to run formatting it seems)

kubernetes/loculus/templates/_common-metadata.tpl
anna-parker: sounds good!

PR #6385, opened 2026-05-09
feat(ingest): Allow combining ingest grouping overrides

theosanderson-agent:
## Summary
- allow ingest to use both calculated NCBI assembly grouping and a downloaded `grouping_override_url` in one run
- write calculated and downloaded grouping files to separate intermediates, then merge them into `results/groups.json`
- add a tested merge helper that rejects conflicting group names and accessions assigned to multiple groups

## Context
This is motivated by Andes virus grouping work: NCBI assembly grouping rescues 4 groups that the heuristic would miss, while a static curated file can rescue release-date-only splits such as `BA02_SMM484`. Findings gist: https://gist.github.com/theosanderson-agent/ded53e709683a52063233c174aa7c0c0

## Validation
- `python -m pytest ingest/tests/test_merge_group_overrides.py`
- `python -m py_compile ingest/scripts/merge_group_overrides.py ingest/tests/test_merge_group_overrides.py`
- `git diff --check`
- Docker Snakemake dry run with both `calculate_groups_override_json` and `grouping_override_url` configured, confirming `download_groups` + `get_assembly_groups` feed `merge_groups`, and `override_group_segments` consumes merged `results/groups.json`
- Docker Snakemake dry runs for static-only and calculated-only modes, confirming backward-compatible direct `results/groups.json` behavior

### DIFF
--- a/ingest/scripts/merge_group_overrides.py
+++ b/ingest/scripts/merge_group_overrides.py
@@ -1,0 +1,63 @@
+"""Merge multiple ingest grouping override JSON files."""
+
+from __future__ import annotations
+
+import json
+import pathlib
+
+import click
+
+
+def merge_group_overrides(group_files: list[str]) -> dict[str, list[str]]:
+    merged_groups: dict[str, list[str]] = {}
+    accession_to_group: dict[str, str] = {}
+
+    for group_file in group_files:
+        with open(group_file, encoding="utf-8") as file:
+            groups: dict[str, list[str]] = json.load(file)
+
+        for group_name, accessions in groups.items():
+            if group_name in merged_groups:
+                if merged_groups[group_name] == accessions:
+                    continue
+                msg = f"Group name {group_name!r} is defined differently in multiple override files"
+                raise ValueError(msg)
+
+            duplicated_accessions = sorted(
+                accession for accession in accessions if accession in accession_to_group
+            )
+            if duplicated_accessions:
+                existing_groups = {
+                    accession: accession_to_group[accession] for accession in duplicated_accessions
+                }
+                msg = (
+                    f"Accessions in group {group_name!r} already appear in other override groups: "
+                    f"{existing_groups}"
+                )
+                raise ValueError(msg)
+
+            merged_groups[group_name] = accessions
+            accession_to_group.update({accession: group_name for accession in accessions})
+
+    return merged_groups
+
+
+@click.command()
+@click.option(
+    "--groups",
+    "group_files",
+    required=True,
+    multiple=True,
+    type=click.Path(exists=True),
+    help="Grouping override JSON file to merge. Can be provided multiple times.",
+)
+@click.option("--output-file", required=True, type=click.Path())
+def main(group_files: list[str], output_file: str) -> None:
+    merged_groups = merge_group_overrides(list(group_files))
+    output_path = pathlib.Path(output_file)
+    output_path.parent.mkdir(parents=True, exist_ok=True)
+    output_path.write_text(json.dumps(merged_groups, indent=4) + "\n", encoding="utf-8")
+
+
+if __name__ == "__main__":
+    main()
--- a/ingest/scripts/merge_group_overrides.py
+++ b/ingest/scripts/merge_group_overrides.py
@@ -18,7 +18,7 @@ def merge_group_overrides(group_files: list[str]) -> dict[str, list[str]]:
 
         for group_name, accessions in groups.items():
             if group_name in merged_groups:
-                if merged_groups[group_name] == accessions:
+                if set(merged_groups[group_name]) == set(accessions):
                     continue
                 msg = f"Group name {group_name!r} is defined differently in multiple override files"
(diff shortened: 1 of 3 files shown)


anna-parker:
Ah great stuff! This has been on my radar for a bit now, glad to see it being tackled!

theosanderson:
@theosanderson-agent format pls

theosanderson-agent:
Ran `ruff format` and `ruff check --fix` on the new ingest files. Two auto-fixes applied (FURB145 dict comprehension → `dict.fromkeys`, removed stray blank line). Pushed as ae40978e7.

Three `S101` (assert in tests) warnings remain — these are pytest idioms and the existing `tests/test_ingest.py` uses asserts the same way, so I left them.

theosanderson:
@theosanderson-agent get CI to pass

theosanderson-agent:
The failing check was `snakefmt` formatting in `ingest-tests`. Reformatted the `DOWNLOADED_GROUPS_JSON` ternary in `ingest/Snakefile` to span multiple lines as snakefmt prefers, verified locally with `snakefmt --compact-diff --check .` (passes), and pushed as `ef8dd11f9`.

theosanderson:
@codex review

ingest/scripts/merge_group_overrides.py
```diff
@@ -0,0 +1,63 @@
...
+    accession_to_group: dict[str, str] = {}
+
+    for group_file in group_files:
+        with open(group_file, encoding="utf-8") as file:
+            groups: dict[str, list[str]] = json.load(file)
+
+        for group_name, accessions in groups.items():
+            if group_name in merged_groups:
+                if merged_groups[group_name] == accessions:
+                    continue
```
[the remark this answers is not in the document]
theosanderson: @theosanderson-agent fix this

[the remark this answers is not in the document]
theosanderson-agent: Fixed in 542846076 — `merge_group_overrides` now compares group memberships as sets, so identical members in different orders no longer raise a conflict. Added a regression test for it.

ingest/scripts/merge_group_overrides.py
```diff
@@ -0,0 +1,63 @@
...
+    for group_file in group_files:
+        with open(group_file, encoding="utf-8") as file:
+            groups: dict[str, list[str]] = json.load(file)
+
+        for group_name, accessions in groups.items():
+            if group_name in merged_groups:
+                if set(merged_groups[group_name]) == set(accessions):
+                    continue
+                msg = f"Group name {group_name!r} is defined differently in multiple override files"
+                raise ValueError(msg)
```
anna-parker, line 46: we probably want to send a slack notification in this case as ingest will start failing but we will have no way of knowing it is failing

ingest/scripts/merge_group_overrides.py
```diff
@@ -0,0 +1,63 @@
...
+
+            duplicated_accessions = sorted(
```
anna-parker, line 48: so this will raise an error if we have the same group in the overrides and the assemblies but it just has a different name, as the name of a group is completely arbitrary and not used for anything we should really not error unless the groups contain different entries

ingest/scripts/merge_group_overrides.py
theosanderson: Yeah I was thinking that more generally we should probably have a health monitor that checks if ingest has run successfully in last 24 hrs and lets us know if not (similar for preprocessing) but yes, for now Slack sounds good!

theosanderson: @theosanderson-agent can you do the Slack thing?

theosanderson-agent: Done in 1d88d9ea8 — `merge_group_overrides` now reads `slack_hook` and `organism` from the ingest config and posts a Slack notification before re-raising on either failure mode (group-name conflict or shared accessions). The Snakefile passes `--config-file results/config.yaml` to the merge step, and I added tests covering the notification path and the no-hook-configured case.

Left for follow-up: Anna’s second comment about not erroring when the same accessions appear under a different group name (since group names are arbitrary) — that’s a behavior change to the merge itself, so happy to do it here or in a separate PR — let me know.

ingest/scripts/merge_group_overrides.py
anna-parker: I added this in here: https://github.com/loculus-project/loculus/pull/6390 - I want to test the ingest changes still a bit

anna-parker: @theosanderson if you fix this I think this PR can be merged

anna-parker (approved):

ingest/scripts/merge_group_overrides.py
anna-parker: @theosanderson I added this change now - let me know if you have any issues with this, otherwise I will try to merge this by end of day to clean up all the grouping issues you found!

PR #6390, opened 2026-05-11
Ingest grouping dates

anna-parker:
see [@theosanderson](https://github.com/theosanderson)'s report https://gist.github.com/theosanderson-agent/ded53e709683a52063233c174aa7c0c0

### Screenshot
New groupings
- AY389361.2.L/AF467768.2.M/U88410.1.S (three sequences merged)
- NC_005301.3.L/NC_005300.2.M/NC_005302.1.S (three sequences merged)
- KT384388.1.L and KT384397.1.S split
- PQ463982.1.L, PQ463983.1.M and PQ463984.1.S split

### PR Checklist

🚀 Preview: https://ingest-grouping-dates.loculus.org

### DIFF
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -135,7 +135,7 @@ defaultOrganismConfig: &defaultOrganismConfig
           function: parse_date_into_range
           inputs:
             date: sampleCollectionDate
-            releaseDate: earliestReleaseDate
+            releaseDate: processed.earliestReleaseDate
           args:
             fieldType: dateRangeString
         required: true
@@ -152,7 +152,7 @@ defaultOrganismConfig: &defaultOrganismConfig
           function: parse_date_into_range
           inputs:
             date: sampleCollectionDate
-            releaseDate: earliestReleaseDate
+            releaseDate: processed.earliestReleaseDate
           args:
             fieldType: dateRangeLower
         rangeOverlapSearch:
@@ -171,7 +171,7 @@ defaultOrganismConfig: &defaultOrganismConfig
           function: parse_date_into_range
           inputs:
             date: sampleCollectionDate
-            releaseDate: earliestReleaseDate
+            releaseDate: processed.earliestReleaseDate
           args:
             fieldType: dateRangeUpper
         rangeOverlapSearch:
@@ -196,6 +196,7 @@ defaultOrganismConfig: &defaultOrganismConfig
         definition: Date the viral nucleotide accession was first released on the INSDC.
         type: date
         header: "INSDC"
+        oneHeader: true
         orderOnDetailsPage: 1000
         preprocessing:
           function: parse_timestamp
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -135,7 +135,7 @@ defaultOrganismConfig: &defaultOrganismConfig
           function: parse_date_into_range
           inputs:
             date: sampleCollectionDate
-            releaseDate: ncbiReleaseDate
+            releaseDate: earliestReleaseDate
           args:
             fieldType: dateRangeString
         required: true
@@ -152,7 +152,7 @@ defaultOrganismConfig: &defaultOrganismConfig
           function: parse_date_into_range
           inputs:
             date: sampleCollectionDate
-            releaseDate: ncbiReleaseDate
+            releaseDate: earliestReleaseDate
           args:
             fieldType: dateRangeLower
         rangeOverlapSearch:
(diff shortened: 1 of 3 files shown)


anna-parker:
@theosanderson-agent figure out why earliestReleaseDate is wrong for CCHF after these changes - it is showing the loculus release data and not the earliest NCBI release date

theosanderson-agent:
Found it. Root cause is in the preprocessing config templating, not the grouping itself.

This PR makes `ncbiReleaseDate` `perSegment: true` and points CCHF's `earliestReleaseDate.externalFields` at `ncbiReleaseDate_L/_M/_S`. The website/backend schemas and the ingest grouping all correctly produce the segment-suffixed fields, so the submitted metadata contains `ncbiReleaseDate_L/_M/_S` and no bare `ncbiReleaseDate`.

But `loculus.preprocessingSpecs` (`_preprocessingFromValues.tpl`) generated the per-segment preprocessing specs with the **unsuffixed** input name:

```yaml
ncbiReleaseDate_L:
  function: parse_timestamp
  inputs:
    timestamp: ncbiReleaseDate   # should be ncbiReleaseDate_L
  args:
    segment: L
```

`add_input_metadata` looks the input up verbatim in the submitted metadata, doesn't find `ncbiReleaseDate`, and returns `None`. So the processed `ncbiReleaseDate_L/_M/_S` come out null, `EarliestReleaseDateFinder` sees only nulls, and falls back to `releasedAtTimestamp` — the Loculus internal release date. Exactly the symptom you saw.

This only bites fields that are both `perSegment` *and* have an explicit `preprocessing.inputs` block referencing a plain submitted field. `ncbiUpdateDate` has had the same latent bug since 2024 but it's only a display column so nobody noticed. `nextclade.*`-input perSegment fields were fine because those resolve per-segment via the `segment` arg.

Fix in #6455 (draft, based off this branch): `loculus.preprocessingSpecs` now suffixes any input that references a perSegment field with the segment name when generating per-segment specs. Verified with `helm template` (the `ncbiReleaseDate_*` / `ncbiUpdateDate_*` specs now read the matching `_L/_M/_S` input, `nextclade.*` inputs unchanged) and `helm lint`.

kubernetes/loculus/values.yaml
```diff
@@ -171,7 +171,7 @@ defaultOrganismConfig: &defaultOrganismConfig
           function: parse_date_into_range
           inputs:
             date: sampleCollectionDate
-            releaseDate: earliestReleaseDate
+            releaseDate: processed.earliestReleaseDate
```
anna-parker, line 174: this seems to be using the PPX submission date and not the INSDC release date as an upper bound... @theosanderson-agent can you figure out why?

anna-parker: uggh... I guess its because earliestReleaseDate is actually calculated by the backend and not preprocessing

theosanderson-agent: Dug into this — the upper bound falls back to the PPX submission date because `earliestReleaseDate` simply isn't available during preprocessing, so neither `earliestReleaseDate` nor `processed.earliestReleaseDate` ever has a value here.

Chain of reasoning:

1. `earliestReleaseDate` is a backend-*derived* field. It's only computed at release time in `ReleasedDataModel.kt` (via `EarliestReleaseDateFinder`, which mixes the internal Loculus `releasedAtTimestamp` with the external fields). It is **not** part of the unprocessed metadata that the backend hands to preprocessing.

2. The helm template `_preprocessingFromValues.tpl` generates a spec for *every* field. Since `earliestReleaseDate` has no `preprocessing:` block, it gets `function: identity` reading the **raw** `earliestReleaseDate` input. During preprocessing that raw field is absent, so `processed.earliestReleaseDate` = `identity(None)` = `None`. The old `releaseDate: earliestReleaseDate` (raw) was also `None`. So switching to `processed.` doesn't actually change anything here — both resolve to empty.

3. In `parse_date_into_range` (`processing_functions.py:510-536`), an empty `releaseDate` means `release_date = None`, so it falls to `max_upper_limit = submitted_at`, where `submitted_at` comes from `args["submittedAt"]` — the Loculus/PPX submission timestamp. That's the PPX submission date you're seeing.

To bound by the INSDC release date instead, reference `ncbiReleaseDate`, which ingest *does* supply as raw input (mapped from NCBI's `releaseDate` in `ingest/config/defaults.yaml`) and is therefore available during preprocessing:

```yaml
inputs:
  date: sampleCollectionDate
  releaseDate: ncbiReleaseDate   # or processed.ncbiReleaseDate
```

Note `ncbiReleaseDate` is per-segment, so for segmented organisms the template will suffix it (`ncbiReleaseDate_L`, etc.) automatically.

anna-parker: yeah makes sense this field isnt available before preprocessing... I guess I need to proceed differently

Issue #4522, opened 2025-06-26
For revocation entries there is no way to review the version comment I added before approving the entry


anna-parker:
![Image](https://github.com/user-attachments/assets/c3f1b818-7570-4c72-8713-a5ef54ff77ec)

PR #6395, opened 2026-05-11
feat(website): display version comment on review cards for revocations

### LINKED ISSUE #6394: Display version comment on review cards
It would be nice to be able to review this

theosanderson-agent:
## Summary

image

Fixes #6394 (and addresses #4522).

Previously, when reviewing a revocation entry on the review page, there was no way to see the version comment the submitter entered when creating the revocation. The card only showed a generic message: *"This is a revocation entry, which will create a new version that revokes this accession"*. This made it impossible to double-check the revocation reason before approving the entry.

This PR surfaces the version comment alongside the existing revocation marker on the review card.

### Backend
- Add `versionComment: String?` to `SequenceEntryStatus` ([SubmissionTypes.kt](https://github.com/loculus-project/loculus/blob/main/backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt)).
- In `getSequences` ([SubmissionDatabaseService.kt](https://github.com/loculus-project/loculus/blob/main/backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt)), extract the value from `unprocessed_data.metadata.versionComment` using a JSONB scalar path. After the [V1.27 migration](https://github.com/loculus-project/loculus/blob/main/backend/src/main/resources/db/migration/V1.27__move_version_comment_to_metadata.sql) the comment lives there for revocations.

### Website
- Add `versionComment` to the `sequenceEntryStatus` zod schema.
- Render a `Version comment` `KeyValueComponent` on the review card whenever the entry is a revocation and a comment was provided.
- Add a Vitest case asserting the comment is rendered for revocation entries.

## Design notes

- The existing `/get-data-to-edit` endpoint (used by `useGetMetadataAndAnnotations` to populate the review card's metadata) explicitly errors for revocations, so it was not appropriate to repurpose. Returning the comment on the list response means no extra HTTP round-trip per card.
- For non-revocation revisions, `versionComment` is already part of the processed metadata and rendered via the existing `MetadataList`. The new card field is gated on `isRevocation` to avoid showing it twice.

## Test plan

- [ ] Manual check: submit a revocation with a version comment, open the review page, confirm the comment is shown on the card.
- [ ] Manual check: submit a revocation without a comment — no "Version comment" row appears (only the existing revocation marker).
- [ ] Manual check: revisions still show their `versionComment` via the normal metadata list (unchanged behavior).
- [x] `CI=1 npm run test` (website) — 646 tests passing including new `should show the version comment for revocation entries` case.
- [x] `CI=1 npm run check-types` (website) — clean.
- [x] `npm run format` (website) — clean.
- [x] `./gradlew compileKotlin && ./gradlew ktlintFormat` (backend) — clean.
- [ ] Backend integration tests — could not run locally (test infra needs Docker/postgres setup); rely on CI.

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
@@ -276,6 +276,12 @@ data class SequenceEntryStatus(
     val isRevocation: Boolean = false,
     val submissionId: String,
     val dataUseTerms: DataUseTerms,
+    @Schema(
+        description = "Comment supplied by the submitter for this version. " +
+            "For revocations this is the revocation reason; otherwise it is the value of the user-supplied " +
+            "`versionComment` metadata field. Null if no comment was provided.",
+    )
+    val versionComment: String? = null,
 ) : AccessionVersionInterface
 
 data class EditedSequenceEntryData(
--- a/backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
@@ -827,6 +827,10 @@ class SubmissionDatabaseService(
                 not(SequenceEntriesView.statusIs(Status.PROCESSED))
         }
 
+        val versionCommentExpression = SequenceEntriesView.unprocessedDataColumn
+            .extract<String>("metadata", "versionComment", toScalar = true)
+            .alias("version_comment")
+
         val entries = SequenceEntriesView
             .join(
                 DataUseTermsTable,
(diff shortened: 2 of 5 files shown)


theosanderson:
<details>
> The existing /get-data-to-edit endpoint (used by useGetMetadataAndAnnotations to populate the review card's metadata) explicitly errors for revocations, so it was not appropriate to repurpose. Returning the comment on the list response means no extra HTTP round-trip per card.

@theosanderson-agent Try making an alternative PR which: renames this endpoint to /get-original-data-for-entry and makes it not error for revocations and uses that instead
</details

theosanderson-agent:
Opened #6412 as the alternative you suggested: `/get-data-to-edit/{accession}/{version}` is renamed to `/get-original-data-for-entry/{accession}/{version}`, the endpoint no longer throws 422 for revocations, and the review card now reads `versionComment` from `originalData.metadata` instead of a new field on `/get-sequences`. `SequenceEntryVersionToEdit.processedData` becomes nullable (`null` for revocations) and `isRevocation` is added to the response. CLI / integration auth test / website client all updated to the new path.

Happy to close this PR in favour of #6412 if that approach looks right.

PR #6403, opened 2026-05-12
fix(website): Prevent text from going under 'X' button in autocomplete input

maverbiest:
In https://github.com/loculus-project/loculus/pull/6302 we added hierarchical filters. That PR modified the autocomplete we use to select lineages to also be usable for general hierarchical filters. The first of these filters we added is a filter on host organism. 

Since these can have long names (longer than lineage identifiers) the names now go under the 'X' button that is used to deselect the currently selected autocomplete option, which is ugly:

grafik

This PR makes it so the text is cut off before the 'X' button:

grafik

Longer term, we should decide on whether we want to limit the length of text for autocomplete options, add line wrapping, or come up with something else to make this part of the UI cleaner.

### PR Checklist

🚀 Preview: https://fix-autocomplete-text.loculus.org

### DIFF
--- a/website/src/components/SearchPage/fields/SingleChoiceAutoCompleteField.tsx
+++ b/website/src/components/SearchPage/fields/SingleChoiceAutoCompleteField.tsx
@@ -19,16 +19,17 @@ import MdiChevronUpDown from '~icons/mdi/chevron-up-down';
 import MdiTick from '~icons/mdi/tick';
 
 const CustomInput = forwardRef<HTMLInputElement, InputHTMLAttributes<HTMLInputElement>>((props, ref) => (
-    <TextField
-        ref={ref}
-        fieldValue={props.value}
-        onChange={props.onChange}
-        onFocus={props.onFocus}
-        disabled={props.disabled}
-        autoComplete='off'
-        label={props.placeholder ?? ''}
-        className='pr-16'
-    />
+    <div className='[&_input]:pr-16'>
+        <TextField
+            ref={ref}
+            fieldValue={props.value}
+            onChange={props.onChange}
+            onFocus={props.onFocus}
+            disabled={props.disabled}
+            autoComplete='off'
+            label={props.placeholder ?? ''}
+        />
+    </div>
 ));
 
 const logger = getClientLogger('SingleChoiceAutoCompleteField');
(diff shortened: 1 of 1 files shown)


website/src/components/SearchPage/fields/TextField.tsx
```diff
@@ -121,7 +121,15 @@ export const TextField = forwardRef<HTMLInputElement | HTMLTextAreaElement, Text
...
             step: props.type === 'int' ? 1 : undefined,
         };
-        return <FloatingLabel {...inputProps} variant='outlined' type={inputType} value={fieldValue} />;
+        return (
+            <FloatingLabel
+                {...inputProps}
+                variant='outlined'
+                type={inputType}
+                value={fieldValue}
+                className={className}
```
corneliusroemer: This ends up adding the class to all the children, not just the <input> which is what you rather want.

This is how you can add it to just the input, by wrapping:

```
<div className='[&_input]:pr-16'>
  <FloatingLabel .../>
</div>
```

See https://github.com/loculus-project/loculus/pull/6409/changes

corneliusroemer: The tests are failing because of this - you make the label so wide that Playwright starts hitting the label and the label intercepts the poniters. This explains #6409 - I didn't realize I was debugging your branch's test failures - now it all makes sense :)

The changes in #6409 still make sense for consistency - and they would help similar surprises in the future.

maverbiest: Thanks for figuring this out!

website/src/components/SearchPage/fields/SingleChoiceAutoCompleteField.tsx
```diff
@@ -19,16 +19,18 @@ import MdiChevronUpDown from '~icons/mdi/chevron-up-down';
...
-    />
+    <div className='[&_input]:pr-16'>
+        <TextField
+            ref={ref}
+            fieldValue={props.value}
+            onChange={props.onChange}
+            onFocus={props.onFocus}
+            disabled={props.disabled}
+            autoComplete='off'
+            placeholder={props.placeholder ?? ''}
```
tombch: I think the placeholder prop here can be removed, as it is being removed for TextField in this PR: #6472

tombch (approved):
Aside from the one comment it looks good to me!

website/src/components/SearchPage/fields/SingleChoiceAutoCompleteField.tsx
maverbiest: Ah thanks! I'll just wait until that's merged then and rebase after

corneliusroemer: Now merged so you can rebase

PR #6410, opened 2026-05-12
fix(website): don't send false positive errors in lineage autocomplete options hooks

corneliusroemer:
A bug in `createLineageOptionsHook` meant that the hook was always returning an Error - even if there was no error: `new AggregateError([aggregateError, defError].filter(Boolean))` is always an `AggregateError([])` even if it doesn't contain an interior error.

This cause lots of post requests to `/admin/logs.txt` whenever a user interacted with the lineage search field that includes sublineage functionality.

@maverbiest helped investigating this bug.

### Screenshot

Currently in production we fire off dozens of posts to `/admin/logs.txt` due to the false positive error:

Google Chrome 2026-05-12 15 10 06

Each mpox search page load fires a dozen errors, looks like this in Grafana:

image

### Manual testing

Testing via `npm run dev`, we no longer fire off false errors.

### PR Checklist

### DIFF
--- a/website/src/components/SearchPage/fields/AutoCompleteOptions.ts
+++ b/website/src/components/SearchPage/fields/AutoCompleteOptions.ts
@@ -2,8 +2,8 @@ import { useCallback } from 'react';
 
 import { lapisClientHooks } from '../../../services/serviceHooks.ts';
 import type { LineageDefinition } from '../../../types/lapis.ts';
-import { stringifyMaybeAxiosError } from '../../../utils/stringifyMaybeAxiosError.ts';
 import { NULL_QUERY_VALUE } from '../../../utils/search.ts';
+import { stringifyMaybeAxiosError } from '../../../utils/stringifyMaybeAxiosError.ts';
 import type { LapisSearchParameters } from '../DownloadDialog/SequenceFilters.tsx';
 
 export type Option = {
--- a/website/src/components/SearchPage/fields/AutoCompleteOptions.ts
+++ b/website/src/components/SearchPage/fields/AutoCompleteOptions.ts
@@ -173,7 +173,7 @@ const createLineageOptionsHook = (
     return function hook() {
         const {
             data,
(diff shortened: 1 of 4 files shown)


website/src/components/SearchPage/fields/AutoCompleteOptions.ts
```diff
@@ -182,14 +185,14 @@ const createLineageOptionsHook = (
         const {
             data,
             isPending: aggregateIsPending,
```
maverbiest:
```suggestion
            isPending: aggregatedIsPending,
```

maybe to be fully consistent?

website/src/components/SearchPage/fields/AutoCompleteOptions.ts
```diff
@@ -233,10 +236,19 @@ const createLineageOptionsHook = (
...
+            aggregatedEndpointError && `aggregated endpoint: ${stringifyMaybeAxiosError(aggregatedEndpointError)}`,
+            definitionEndpointError &&
+                `lineage definition endpoint: ${stringifyMaybeAxiosError(definitionEndpointError)}`,
+        ].filter(Boolean);
+
         return {
             options,
-            isPending: aggregateIsPending || defIsLoading,
-            error: new AggregateError([aggregateError, defError].filter(Boolean)),
+            isPending: aggregateIsPending || definitionIsLoading,
```
maverbiest:
```suggestion
            isPending: aggregatedIsPending || definitionIsLoading,
```

website/src/components/SearchPage/fields/AutoCompleteOptions.ts
corneliusroemer: i've renamed to be consistent

maverbiest (approved):
Looks good to me!

I also ran the website locally both on `main` and on this branch and confirmed the error spam is gone

PR #6413, opened 2026-05-12
fix(prepro): update alignment requirement ANY

anna-parker:
The ANY alignment requirement is currently only used by Genspectrum where ingest is the only submitter (and thus only submits sequences that can be assigned to a segment using nextclade sort). However, in order to use the alignment requirement ANY for other users we need to ensure that users cannot submit one valid sequence (e.g. segment M) and an invalid sequence (e.g. an rsv sequence for CCHF). The current ANY alignment requirement would still allow such a case to be submitted - which we need to prevent.

### Why use the ANY alignment requirement?

This is useful for multi-segmented viruses with many sequences on the INSDC. INSDC does not group segments and we group segments in ingest prior to submitting to loculus by assigning the segment in ingest. `nextclade sort` is an excellent alternative to aligning all sequences against every single segment and segment subtype. But `nextclade sort` can assign more sequences than will later be aligned (e.g. shorter sequences and recombinants), if we use the default ALL alignment requirement a group with one segment that can not be aligned will fail preprocessing and all other segments (even if they align) will not be visible.


theosanderson:
@codex review

preprocessing/nextclade/src/loculus_preprocessing/nextclade.py
```diff
@@ -457,18 +443,6 @@ def assign_segment(  # noqa: C901
...
-        and (not has_unaligned_sequence or config.alignment_requirement == AlignmentRequirement.ANY)
-    ):
-        sequence_assignment.alert.errors.append(
-            sequence_annotation(
-                "No sequence data could be classified - "
-                "check you are submitting to the correct organism.",
-            )
-        )
-
     return sequence_assignment
```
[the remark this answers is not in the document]
anna-parker: If classification fails then nothing will align and the function `alignment_errors_warnings` will add an error so this should be fine

kubernetes/loculus/values.schema.json
```diff
@@ -686,7 +686,7 @@
                     "docsIncludePrefix": false,
                     "type": "string",
                     "enum": ["ALL", "ANY"],
-                    "description": "If multi-segmented viruses should require ALL segments (that the user provides) align or ANY segment aligns"
+                    "description": "If multi-segmented viruses should require ALL segments (that the user provides) align or ANY segment aligns (in case ANY it must still be possible to assign the segments)"
```
anna-parker:
```suggestion
                    "description": "Applies only to multi-segmented organisms. Determines whether ALL or ANY segments that are submitted must align. In the case ANY it must still be possible to assign the segment."
```

preprocessing/nextclade/src/loculus_preprocessing/config.py
```diff
@@ -44,8 +44,8 @@ class ProcessingSpec(BaseModel):
 
 class AlignmentRequirement(StrEnum):
     # Determines whether ALL or ANY segments that a user provides must align.
-    # ANY: warn if some segments fail and some segments align
-    # ALL: error if any segment fails even if some segments align
+    # ANY: warn if some segments can be assigned and not aligned while other segments align
```
```suggestion
    # ANY: warn if not all segments can be aligned (all must be assigned) and at least one must fully align
```

kubernetes/loculus/values.schema.json
```diff
@@ -686,7 +686,7 @@
                     "docsIncludePrefix": false,
                     "type": "string",
                     "enum": ["ALL", "ANY"],
-                    "description": "If multi-segmented viruses should require ALL segments (that the user provides) align or ANY segment aligns"
+                    "description": "Applies only to multi-segmented organisms. Determines whether ALL or ANY segments that are submitted must align. In the case ANY it must still be possible to assign the segment."
```
anna-parker, line 689: I am also open to changing to the options ALIGN and ASSIGN, where in one option every segment must align and in the other every segment must only be ASSIGNED.

theosanderson (approved):

PR #6216, opened 2026-04-01
Fix Keycloak database service naming for ephemeral dev deployments

theosanderson:
### PR Checklist
- [x] All necessary documentation has been adapted.
- [x] The implemented feature is covered by appropriate, automated tests.
- [x] Any manual testing that has been done is documented (i.e. what exactly was tested?)

🚀 Preview: Add `preview` label to enable

PR #6415, opened 2026-05-12
fix(deployment): use negative sync-wave for development databases

### LINKED ISSUE #6414: Use sync-waves to prevent race conditions with development databases
We get race conditions where development databases get deleted after something (e.g. keycloak) has connected to them. We should be able to solve this with a negative sync wave on them. This is simpler than another approach we have made a PR with already.

theosanderson-agent:
## Summary

Fixes #6414.

Adds `argocd.argoproj.io/sync-wave: "-5"` to the development standin database deployments (main + keycloak) and their services so they sync ahead of consumers like keycloak and the backend.

### Why

The standin deployments use `Replace=true` and re-create on every docker tag change (via the `LOCULUS_VERSION` env var). Without an explicit sync wave, ArgoCD may process the standin database and its consumers (keycloak/backend) in the same wave. That means:

1. Consumers can start a connection to the soon-to-be-replaced DB pod.
2. The standin DB is then deleted and recreated empty.
3. The consumer holds connections to the dying DB and/or reads the new (empty) one mid-handshake, leading to flaky data loss in dev environments.

By placing the standin DBs and their services in sync wave `-5` (between `-10` used by `PriorityClass` and the default `0`), ArgoCD ensures the new DB pod is up and the service routes to it before any consumer is touched in the same sync.

This is the simpler alternative mentioned in the issue compared to the approach taken in #6216.

### DIFF
--- a/kubernetes/loculus/templates/keycloak-database-service.yaml
+++ b/kubernetes/loculus/templates/keycloak-database-service.yaml
@@ -3,6 +3,8 @@ apiVersion: v1
 kind: Service
 metadata:
   name: loculus-keycloak-database-service
+  annotations:
+    argocd.argoproj.io/sync-wave: "-5"
 spec:
   type: ClusterIP
   selector:
--- a/kubernetes/loculus/templates/keycloak-database-standin.yaml
+++ b/kubernetes/loculus/templates/keycloak-database-standin.yaml
@@ -6,6 +6,7 @@ metadata:
   name: loculus-keycloak-database
   annotations:
     argocd.argoproj.io/sync-options: Replace=true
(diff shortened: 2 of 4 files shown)


theosanderson:
@theosanderson-agent fix the issue

theosanderson-agent:
Added the `argocd.argoproj.io/sync-wave: "-5"` annotation to the `loculus-database-init` ConfigMap in `kubernetes/loculus/templates/loculus-database-standin.yaml` so it syncs in the same wave as the Deployment that mounts it. Pushed to `dev-db-sync-wave` in 1772aed98. `helm lint` passes.

tombch (approved):

PR #6433, opened 2026-05-13
refactor(deployment): replace `{{ now }}` annotation with values hash

theosanderson:
## Summary

Replaces the `timestamp: {{ now | quote }}` pod template annotation on every Deployment in the Helm chart with `valuesHash: {{ .Values | toYaml | sha256sum }}`.

## Motivation

Refs #6431.

`{{ now }}` re-evaluates on every Helm render, so:

- Every ArgoCD sync rolls every pod, even when nothing has actually changed.
- `helm diff` / dry-runs are never clean — there is always at least one phantom field change per Deployment.
- Two different `{{ now }}` annotations in the same chart are guaranteed to disagree by a few milliseconds, which is exactly the class of bug behind #6431: the dev Keycloak DB Deployment and the Keycloak Deployment had separate restart triggers and ended up out of sync after a redeploy with the same image tag, leading to *“Unexpected error when handling authentication request to identity provider”*.

## Approach

Hash the chart's input values instead:

```yaml
annotations:
  valuesHash: {{ .Values | toYaml | sha256sum }}
```

Properties of this approach:

- **Stable**: identical inputs produce the same hash, so re-syncing without changes does not roll pods.
- **Sensitive**: any change to `values.yaml` (image tag, env, replica count, feature flag, …) produces a new hash and triggers a rolling restart.
- **Consistent across templates**: every Deployment annotated this way will agree on whether a roll is needed — no more drift between e.g. the Keycloak DB and Keycloak itself.

## Files changed

The annotation is updated in all 10 Deployment templates that previously used `{{ now }}`:

- `autoapprove-deployment.yaml`
- `docs-preview.yaml`
- `ena-submission-deployment.yaml`
- `keycloak-database-standin.yaml`
- `lapis-deployment.yaml`
- `loculus-backend.yaml`
- `loculus-preprocessing-deployment.yaml`
- `loculus-website.yaml`
- `silo-deployment.yaml`
- `taxonomy-deployment.yaml`

## Behaviour change to be aware of

Previously, *any* sync would restart all pods. After this change, a sync with unchanged values will **not** restart pods. This is the intended behaviour, but it means workflows that relied on “re-syncing to get a fresh pod” (e.g. to refresh a `latest`-tagged image, or to recover from an in-cluster issue) will need to be done explicitly via `kubectl rollout restart` or by bumping a value.

For Keycloak specifically (the #6431 root cause), the Keycloak DB standin and Keycloak Deployment now share an annotation source. If we deploy something that wipes the dev DB, the hash will change for both, so Keycloak will roll alongside its DB.

## Test plan

- [ ] `helm lint kubernetes/loculus -f kubernetes/loculus/values.yaml` passes
- [ ] `helm template` renders without errors and the annotation appears as a 64-char hex string on each Deployment
- [ ] Deploy to a dev cluster; re-sync with no changes and confirm pods are **not** rolled
- [ ] Change a value (e.g. an env var or image tag) and confirm the relevant Deployments roll
- [ ] Verify the dev-Keycloak-DB issue from #6431 no longer occurs after a same-tag redeploy

https://claude.ai/code/session_01SevbUTUMLU9wGCYw6f1shL

---
_Generated by [Claude Code](https://claude.ai/code/session_01SevbUTUMLU9wGCYw6f1shL)_

🚀 Preview: https://claude-explore-now-altern.loculus.org

### DIFF
--- a/kubernetes/loculus/templates/autoapprove-deployment.yaml
+++ b/kubernetes/loculus/templates/autoapprove-deployment.yaml
@@ -18,7 +18,7 @@ spec:
   template:
     metadata:
       annotations:
-        timestamp: {{ now | quote }}
+        valuesHash: {{ .Values | toYaml | sha256sum }}
       labels:
         app: loculus
         component: loculus-autoapprove-deployment
--- a/kubernetes/loculus/templates/autoapprove-deployment.yaml
+++ b/kubernetes/loculus/templates/autoapprove-deployment.yaml
@@ -18,7 +18,7 @@ spec:
   template:
     metadata:
       annotations:
-        valuesHash: {{ .Values | toYaml | sha256sum }}
+        valuesHash: {{ .Values | toJson | sha256sum | quote }}
       labels:
         app: loculus
         component: loculus-autoapprove-deployment
--- a/kubernetes/loculus/templates/keycloak-deployment.yaml
+++ b/kubernetes/loculus/templates/keycloak-deployment.yaml
@@ -14,6 +14,8 @@ spec:
       component: keycloak
   template:
     metadata:
+      annotations:
+        valuesHash: {{ .Values | toYaml | sha256sum }}
       labels:
         app: loculus
         component: keycloak
--- a/kubernetes/loculus/templates/keycloak-deployment.yaml
+++ b/kubernetes/loculus/templates/keycloak-deployment.yaml
@@ -14,8 +14,10 @@ spec:
       component: keycloak
   template:
     metadata:
+      {{- if and .Values.runDevelopmentKeycloakDatabase (not .Values.developmentDatabasePersistence) }}
       annotations:
         valuesHash: {{ .Values | toYaml | sha256sum }}
+      {{- end }}
       labels:
         app: loculus
         component: keycloak
--- a/kubernetes/loculus/templates/keycloak-deployment.yaml
+++ b/kubernetes/loculus/templates/keycloak-deployment.yaml
@@ -16,7 +16,7 @@ spec:
     metadata:
       {{- if and .Values.runDevelopmentKeycloakDatabase (not .Values.developmentDatabasePersistence) }}
       annotations:
-        valuesHash: {{ .Values | toYaml | sha256sum }}
+        valuesHash: {{ .Values | toJson | sha256sum | quote }}
       {{- end }}
       labels:
         app: loculus
--- a/kubernetes/loculus/templates/docs-preview.yaml
+++ b/kubernetes/loculus/templates/docs-preview.yaml
@@ -17,7 +17,7 @@ spec:
   template:
     metadata:
       annotations:
-        valuesHash: {{ .Values | toYaml | sha256sum }}
+        valuesHash: {{ .Values | toJson | sha256sum | quote }}
       labels:
         app: loculus
         component: docs
--- a/kubernetes/loculus/templates/docs-preview.yaml
+++ b/kubernetes/loculus/templates/docs-preview.yaml
@@ -17,7 +17,7 @@ spec:
   template:
     metadata:
       annotations:
-        timestamp: {{ now | quote }}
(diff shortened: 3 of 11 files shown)


corneliusroemer:
For CI/previews, we want all pods to roll whenever anything changes - any container image, values etc. The reason we saw the keycloak issue is as Claude says that we don't have the annotation on the keycloak pod. We can switch to hashing values but then we also need to add the hash to keycloak - which isn't happening in this PR AFAICT.

One potential issue in particular for prod: the values are not all state. There's some stuff that can change without `Values` changing like secrets or external resources downloaded by config processor.

For most rollouts in prod, `Values` should change, so issues should be rare, but it's something to keep in mind.

Main benefit of switching away from `now` is that we don't get the unnecessary rollouts every 24hr.

For sparse rollouts where we only roll pods that need to, we could hash relevant subtrees of values but that seems a lot of work. 

We might want to use `toJson | sha256sum | quote` instead of `toYaml | sha256sum`. Quoting for type safety and json due to being more compact and deterministic.

We should probably also add the hash annotation to keycloak, at least for previews like done in https://github.com/loculus-project/loculus/pull/6432

theosanderson:
>  The reason we saw the keycloak issue is as Claude says that we don't have the annotation on the keycloak pod. We can switch to hashing values but then we also need to add the hash to keycloak - which isn't happening in this PR AFAICT.

Keycloak is annotated with the Loculus commit tag so it does restart on rollouts - not saying we shouldn't do something else as well. The downside is that it logs people out on PPX (but that will be happening anyway when we bump Loculus).

theosanderson:
> One potential issue in particular for prod: the values are not all state. There's some stuff that can change without Values changing like secrets or external resources downloaded by config processor.

Secrets are values. But yep, this is a good reason to version external resources.

kubernetes/loculus/templates/autoapprove-deployment.yaml
```diff
@@ -18,7 +18,7 @@ spec:
   template:
     metadata:
       annotations:
-        timestamp: {{ now | quote }}
+        valuesHash: {{ .Values | toYaml | sha256sum }}
```
anna-parker: Question for @corneliusroemer and @theosanderson could we switch to using the commit hash here instead? then we would trigger a restart everytime a preview gets a new commit pushed (which we want) and everytime we rollout changes but not every 24 hours. But maybe I am missing sth important

theosanderson: wouldn't update when we roll-out new set of values to PPX

anna-parker: just to confirm as we use dockerTags in the deployment image on a preview new changes to code (not values.yaml) will still be synced on the preview>

kubernetes/loculus/templates/keycloak-deployment.yaml
```diff
@@ -14,6 +14,10 @@ spec:
       component: keycloak
   template:
     metadata:
+      {{- if and .Values.runDevelopmentKeycloakDatabase (not .Values.developmentDatabasePersistence) }}
+      annotations:
+        valuesHash: {{ .Values | toYaml | sha256sum }}
+      {{- end }}
```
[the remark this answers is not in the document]
anna-parker: I think this is what we want on previews keycloak will now restart with other pods however on prod it will only restart when the dockerTag changes (code change) or a field in the keycloak template changes (which we want), it will not restart for changes that do not affect it which is what we want.

kubernetes/loculus/templates/autoapprove-deployment.yaml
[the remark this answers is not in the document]
anna-parker: done :-)

anna-parker (approved):

corneliusroemer (approved):
I think this should work and help, let's keep an eye on Loculus argo to see if this works after merge

theosanderson:
@codex would you have seen any problems with this?

PR #6447, opened 2026-05-15
feat(deployment,website): Split website organism configs into separate ConfigMaps

theosanderson-agent:
## Summary

Splits the website organism configuration out of the monolithic `loculus-website-config` ConfigMap into one ConfigMap per organism. The website now reads `website_config.json` plus optional `organisms/*.json` files and merges them at startup.

This keeps generated website organism config data out of a single Kubernetes object so instances with many or large organisms stay comfortably below the ConfigMap/Secret object size limit.

## Implementation

- Render `loculus-website-config` with global website config only and one `loculus-web-org-config-*` ConfigMap per organism.
- Mount the website config through a projected volume so `organisms/*.json` lands under the same `CONFIG_DIR`.
- Update `deploy.py config` to generate split local test config files.
- Document the split config layout and ignore generated split config fixtures.

## Size check

Using current Pathoplexus values with the Bundibugyo organism branch:

- old monolithic `loculus-website-config`: 976,493 bytes of `website_config.json`
- new global `loculus-website-config`: 2,568 bytes of `website_config.json` in the main render
- largest split organism ConfigMap: 81,638 bytes (`loculus-web-org-config-mpox`)
- largest ConfigMap/Secret data block across demo/main/production/staging: ~123.2 KB (`loculus-backend-config`), about 11.75% of 1 MiB

The reconstructed split website config matches the old monolithic website config for the Pathoplexus main render.


theosanderson:
@codex review

kubernetes/loculus/templates/_config-processor.tpl
```diff
@@ -62,3 +62,7 @@
 - name: {{ .name }}-processed
   emptyDir: {}
 {{- end }}
+
+{{- define "loculus.websiteOrganismConfigMapName" -}}
+{{- printf "loculus-web-org-config-%s" . | trunc 63 | trimSuffix "-" -}}
```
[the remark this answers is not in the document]
theosanderson: I think this will generally be a problem throughout our code - we can add an issue to restrict the size on organism technical names

theosanderson:
@corneliusroemer @maverbiest @tombch would one of you have a chance to review? I think this is quite important as I think it could currently be blocking adding any more organisms

maverbiest:
@theosanderson I might have some time over the weekend/early next week so will do my best to review then

.gitignore
```diff
@@ -28,6 +28,9 @@ __pycache__/
 
 .astro/
 
+/website/tests/config/*.json
+/website/tests/config/*.yaml
+
```
tombch: Feel free to skip as this is not an issue but I guess these could be moved into the websites `.gitignore`?

kubernetes/loculus/templates/_config-processor.tpl
tombch, line 67: Is there a reason why the truncation limits to 63 characters, rather than the limit for config map names (253 characters)?

tombch: Ah, I have had a look around and organism names are used in a bunch of labels which limit to 63 characters, so I guess this is already a thing that needs restricting to lower than 253 chars (as mentioned in the resolved comment, this is already an issue)

tombch (approved):
Tested it locally as well, looks good to me!

.gitignore
theosanderson: resolved

PR #6454, opened 2026-05-18
refactor(deposition): rename variables for clarity, avoid mutating

corneliusroemer:
We came across this code when discussing what happens if submitters have multiple custom bioprojects.

Just a few small tweaks to variable names and avoiding mutation for clarity. Things that could be >1 had variable names that implied it was at most one, etc.


ena-submission/src/ena_deposition/create_project.py
anna-parker, line 201: I guess in theory we should send a notification if we get to this point as we dont know which one to add - but not too important as we will be notified the submission is taking long and will see this warning in the logs

anna-parker (approved):

PR #6462, opened 2026-05-19
feat(backend): get rid of warning

anna-parker:
```
 WARNING: A restricted method in java.lang.System has been called
  WARNING: java.lang.System::loadLibrary has been called by com.github.luben.zstd.util.Native$1 in an unnamed module
  (jar:nested:/app/app.jar/!BOOT-INF/lib/zstd-jni-1.5.7-8.jar!/)
  WARNING: Use --enable-native-access=ALL-UNNAMED to avoid a warning for callers in this module
```

resolves #

### PR Checklist

🚀 Preview: https://backend-warning.loculus.org

### DIFF
--- a/backend/entrypoint.sh
+++ b/backend/entrypoint.sh
@@ -3,10 +3,11 @@ JVM_OPTS=${JVM_OPTS:-}
 # Take script arguments
 ARGS="${*}"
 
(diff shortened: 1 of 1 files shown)


corneliusroemer (approved):
Could have also added to JVM_OPTS and gotten rid of the if/else as we now always have JVM_OPTS

PR #6466, opened 2026-05-19
feat(config): cchf - use same config as ppx

anna-parker:
The grouping override json used on loculus and ppx is different, I think its good to make them the same to avoid confusion in future.

### PR Checklist


maverbiest (approved):
Not sure how the grouping works but I can see that it's the same as the PPX config so lgtm

PR #6472, opened 2026-05-20
fix(website): remove duplicate placeholder on lineage search fields

### LINKED ISSUE #6470: Duplicate placeholder text on lineage search field causing small visual bug
We seem to put 2 placeholders on top of each other in the lineage search field. The lower one's letter `g`s just show up behind the front label `g`s, notice the tiny specks below the gs:

Image

The lower label becomes visible when one clicks on the field:

Image

corneliusroemer:
Resolves #6470

CustomInput in SingleChoiceAutoCompleteField.tsx passes both placeholder and label with the same value to TextField, causing duplicate visible text in the input field

Fix: remove placeholder prop from CustomInput's TextField call

### Screenshot

No more double placeholder:

image

### DIFF
--- a/website/src/components/SearchPage/fields/SingleChoiceAutoCompleteField.spec.tsx
+++ b/website/src/components/SearchPage/fields/SingleChoiceAutoCompleteField.spec.tsx
@@ -63,6 +63,7 @@ describe('SingleChoiceAutoCompleteField', () => {
 
         const input = screen.getByLabelText('Test Field');
         expect(input).toBeInTheDocument();
+        expect(input).toHaveAttribute('placeholder', ' ');
 
         await userEvent.click(input);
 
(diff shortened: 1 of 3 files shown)


tombch (approved):
Looks good to me! I think this bug was introduced when I implemented the placeholder prop on the `TextField` in https://github.com/loculus-project/loculus/pull/6416

corneliusroemer:
@tombch ah good spot, it's the line here:

https://github.com/loculus-project/loculus/pull/6416/changes#diff-d454f69c592f26029ccd3b68f3a0819e507d3a434f1abda0a76ab5d5fb7574bbR126

Not a big deal, just a tiny visual issue but still good to fix I think.

corneliusroemer:
Now looking at your PR, the better fix might be to hardcode the placeholder at the component itself to `' '` instead of what I'm doing here at the caller.

tombch:
Yeah I'd agree with that, i think this was the only place the placeholder was added anyway

corneliusroemer:
@tombch I've made the change as suggested, undoing your placeholder addition and hardcoding `' '`

tombch (approved):
Looks good to me!

PR #6485, opened 2026-05-21
fix(website): stop date range strictness checkbox stuck on strict when no date entered

### LINKED ISSUE #6483: Strictness checkbox in collection date range filter jumps on page load
The strictness checkbox in collection date range filter jumps on page load which isn't very professional and no other element on the page jumps like this so we should probably get rid of the animation.

Image

theosanderson-agent:
Fixes #6483

## What was happening

When you toggled the strictness checkbox in a `DateRangeField` (e.g. the Collection date filter) without having entered any dates, was stuck on. This matches the user reports in #6483:

- The checkbox "actually seems stuck enabled"

## Fix

In the read-in effect, skip the `setStrictMode` call when none of the four bound fields are defined — there is no signal to derive from in that case, so we preserve the user's manual selection. When at least one field is defined the existing `isStrictMode` logic still runs and stays in charge.

The write-out effect is unchanged. External (URL-driven) changes still flow in correctly because they always come with at least one bound defined.

🚀 Preview: https://fix-date-range-strictness.loculus.org

### DIFF
--- a/website/src/components/SearchPage/fields/DateRangeField.spec.tsx
+++ b/website/src/components/SearchPage/fields/DateRangeField.spec.tsx
@@ -158,6 +158,38 @@ describe('DateRangeField', () => {
         );
     });
 
+    it('does not snap back to strict when user toggles without having entered dates', async () => {
+        function Wrapper() {
+            const [values, _setValues] = useState<FieldValues>({});
+
+            const setValues: SetSomeFieldValues = useCallback((...fieldValuesToSet) => {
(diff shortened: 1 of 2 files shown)


corneliusroemer:
@theosanderson-agent Not sure it fixes the checkbox jump - it's a different issue you're solving here, of it being stuck

corneliusroemer (approved):
Looks reasonable to set strict mode via URL params only when defined. I also tested and seems to work.

PR #6487, opened 2026-05-21
feat!(config, prepro): Change hostTaxonId from int to string

maverbiest:
## BREAKING CHANGES
This PR changes the type of metadata field `hostTaxonId` from int to string.

(will only affect deployments that configure a `hostTaxonId: int` metadata field)

### How to address
Update the type of the hostTaxonId metadata field to a string in your `values.yaml`:

```yaml
metadata:
  - name: hostTaxonId
    type: string
```

and run the following SQL command against your database once you've rolled out:

```sql
UPDATE public.sequence_entries_preprocessed_data
SET processed_data = jsonb_set(
	processed_data,
	'{metadata, hostTaxonId}',
	to_jsonb(processed_data -> 'metadata' ->> 'hostTaxonId')
)
WHERE processed_data->'metadata' ? 'hostTaxonId';
```

## Summary
We want to add a hierarchical filter on the hostTaxonId field, this was initially merged in https://github.com/loculus-project/loculus/pull/6302 and has since been reverted since it was a breaking change that we hadn't properly tested in a live deployment.

We have since tested this change on staging in combination with a SQL command to change the type of hostTaxonId to string in the DB. This worked smoothly. The current PR just changes the type of hostTaxonId in the config, preprocessing function, and ENA tests. The actual implementation of hierarchical filters will be reintroduced in a follow-up PR.

This keeps the PRs more focused and separates the concerns.

### PR Checklist

🚀 Preview: https://hosttaxonid-int-to-string.loculus.org


anna-parker (approved):
LGTM

PR #6491, opened 2026-05-21
feat(website, backend): Sequence citations 

tombch:
### Backend Changes
- Adds a `get-sequence-citations` endpoint, which returns citing sources for seqsets that include the sequence entry.
- The endpoint requires an `accession` for the sequence entry, with `version` as an optional parameter. If only the accession is provided, the endpoint returns citing sources for seqsets which use any version `{accession}.*`, as well as unversioned `{accession}`, of the sequence entry. If both an `accession` and `version` are provided, the endpoint returns citing sources which reference that exact `{accession}.{version}`.

### Website Changes
- Adds a `CitationList` component to display the first N citations (in this case, currently three) directly on the sequence details page, and adds a 'view all citations modal', clickable if they exceed this number. 
- The displayed citations also include the seqset they came from, and how the sequence was used (i.e. pinned version or unversioned).

### Screenshot

image

### PR Checklist


website/src/components/SequenceDetailsPage/SequenceCitations.tsx
theosanderson, line 1: (haven't seen a preview or anything but just trying to give some early feedback: IMO we could surface some of these even without opening a modal, in the main page)

backend/src/main/kotlin/org/loculus/backend/controller/SeqSetCitationsController.kt
```diff
@@ -109,6 +111,11 @@ class SeqSetCitationsController(
     fun getSeqSetCitations(@RequestParam seqSetId: String, @RequestParam version: Long): List<SeqSetCitation> =
         seqSetCitationsService.getSeqSetCitations(seqSetId, version)
 
+    @Operation(description = "Get sequence citations from publications or other sources")
+    @GetMapping("/get-sequence-citations")
```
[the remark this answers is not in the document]
tombch: It has one?

backend/src/test/kotlin/org/loculus/backend/controller/seqsetcitations/CitationEndpointsTest.kt
```diff
@@ -251,9 +287,52 @@ class CitationEndpointsTest(
...
+            val citationsMap: Map<String, String>,
+            val accession: String,
+            val version: Long?,
+            val expectedCitingSourceDOIs: List<String>,
+        ) {
+            override fun toString() = description
+        }
+
         @JvmStatic
         fun authorizationTestCases(): List<Scenario> = listOf(
```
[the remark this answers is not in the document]
tombch: IMO this is unnecessary

theosanderson:
@codex review

website/src/components/SequenceDetailsPage/DataTable.tsx
```diff
@@ -103,6 +109,28 @@ const DataTableComponent: React.FC<Props> = ({
                 </div>
             )}
 
+            {hasSequenceCitations && <hr className='my-8 border-t-2 border-gray-200' />}
```
theosanderson: I wonder if it could just be folded into the sections above, with no horizontal rule?

backend/src/main/kotlin/org/loculus/backend/api/SeqSetCitations.kt
```diff
@@ -97,6 +97,11 @@ data class SeqSetCitationSource(val source: CitationSource, val seqSetDOIs: Set<
 @Schema(description = "A citation of a SeqSet.")
 data class SeqSetCitation(val source: CitationSource)
 
+data class SeqSetCitingSequence(val seqSetAccession: String, val sequenceAccession: String)
```
anna-parker: we might have a specific type for loculus accessions and seqSetAccessions in the backend

backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCitationsDatabaseService.kt
```diff
@@ -523,6 +530,52 @@ class SeqSetCitationsDatabaseService(
             }
     }
 
+    fun getSequenceCitations(accession: String, version: Long?): List<SequenceCitation> {
+        log.info { "Get sequence citations for accession $accession, version $version" }
```
anna-parker: we should probably not log " version " when version is empty

backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCitationsDatabaseService.kt
```diff
@@ -523,6 +530,52 @@ class SeqSetCitationsDatabaseService(
...
+
+        // If a version is provided, return citations pinned to that exact accession and version
+        // Otherwise, return all citations for the accession, regardless of version, including unversioned
+        val accessionQuery: Op<Boolean> = if (version != null) {
+            val accessionVersion = AccessionVersion(accession, version)
+            SeqSetRecordsTable.accession eq accessionVersion.displayAccessionVersion()
```
anna-parker, line 549: maybe you could leave a comment somewhere that accession can be accession and accessionVersion - I wonder if we should actually refactor the table to have an accession and version column instead of having to parse the contents of the accession column

backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCitationsDatabaseService.kt
```diff
@@ -523,6 +530,52 @@ class SeqSetCitationsDatabaseService(
...
+        } else {
+            (SeqSetRecordsTable.accession eq accession) or
+                (SeqSetRecordsTable.accession like (LikePattern.ofLiteral("$accession.") + "%"))
+        }
+
+        return SeqSetCitationSourceTable.innerJoin(
+            SeqSetToCitationSourceTable,
+        ).innerJoin(
```
anna-parker, line 557: we discussed that the seqsettorecordstable can actually be dropped and the seqsetrecordstable refactored as:
```
object SeqSetRecordsTable : Table("seqset_to_records") {
    val seqSetRecordId = long("seqset_record_id").autoIncrement()
    val accession = varchar("accession", 255)
    val type = varchar("type", 255)
    val isFocal = bool("is_focal").default(true)
    val seqSetId = varchar("seqset_id", 255) references SeqSetsTable.seqSetId
    val seqSetVersion = long("seqset_version") references SeqSetsTable.seqSetVersion
    override val primaryKey = PrimaryKey(seqSetRecordId, seqSetId, seqSetVersion)
}
```

website/src/services/seqSetCitationApi.ts
```diff
@@ -30,6 +37,21 @@ const getSeqSetCitationsEndpoint = makeEndpoint({
...
+        {
+            name: 'version',
+            type: 'Query',
+            schema: z.number().optional(),
+        },
+    ],
+    alias: 'getSequenceCitations',
+    response: sequenceCitations,
+    errors: [notAuthorizedError],
+});
```
anna-parker, line 53: might be more consistent:
```suggestion
const getSequenceCitationsEndpoint = makeEndpoint({
    method: 'get',
    path: '/get-sequence-citations',
    parameters: [
        {
            name: 'accession',
            type: 'Query',
            schema: z.string(),
        },
        {
            name: 'version',
            type: 'Query',
            schema: z.number().optional(),
        },
    ],
    alias: 'getSequenceCitations',
    response: sequenceCitations,
    errors: [notAuthorizedError],
});
```

website/src/components/SequenceDetailsPage/DataTable.tsx
anna-parker: would be nice to reuse the code for the general sections as much as possible as the styling here is basically the same

website/src/pages/seq/[accessionVersion]/index.astro
```diff
@@ -9,14 +9,21 @@ import ErrorBox from '../../../components/common/ErrorBox.tsx';
...
+import { parseAccessionVersionFromString } from '../../../utils/extractAccessionVersion.ts';
 import { getAccessToken } from '../../../utils/getAccessToken';
 import { getMyGroups } from '../../../utils/getMyGroups';
 import { getSegmentNames } from '../../../utils/sequenceTypeHelpers';
 
 const accessionVersion = Astro.params.accessionVersion!;
+const { accession } = parseAccessionVersionFromString(accessionVersion);
+const sequenceCitationsPromise = SeqSetCitationClient.create().call('getSequenceCitations', {
+    params: { accession }, // Display citations across all accession versions
+});
```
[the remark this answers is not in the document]
Comment (anna-parker) [reply website/src/pages/seq/[accessionVersion]/index.astro]: sounds good to add!

backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCitationsDatabaseService.kt
anna-parker: this can be done in a later refactor if needbe

backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCitationsDatabaseService.kt
tombch: Added!

backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCitationsDatabaseService.kt
tombch: Added a comment for this!

website/src/services/seqSetCitationApi.ts
tombch: IMO I'm going to leave this one as-is, its less code and the other endpoints have their required params in the path string so its consistent with that. Feel free to unresolve if you disagree!

backend/src/main/kotlin/org/loculus/backend/api/SeqSetCitations.kt
tombch: These are a bit awkward because we have the versioned SeqSet, but the `sequenceAccession` can be versioned or unversioned - I've updated this with schema annotations to be clearer on what exactly is being returned

website/src/pages/seq/[accessionVersion]/index.astro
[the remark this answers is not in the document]
Comment (tombch) [reply website/src/pages/seq/[accessionVersion]/index.astro]: This has been added!

website/src/components/SequenceDetailsPage/DataTable.tsx
tombch: I have folded this in, and will open a separate PR with the styling consolidated!

tombch:
@claude review

theosanderson:
@codex review

website/src/pages/seq/[accessionVersion]/index.astro
```diff
@@ -6,17 +6,25 @@ import { SequenceDataUI } from '../../../components/SequenceDetailsPage/Sequence
...
 const accessionVersion = Astro.params.accessionVersion!;
+const { accession } = parseAccessionVersionFromString(accessionVersion);
+const sequenceCitationsPromise = seqSetsAreEnabled()
+    ? SeqSetCitationClient.create().call('getSequenceCitations', {
+          params: { accession }, // Display citations across all accession versions
+      })
+    : undefined;
 
 const sequenceDetailsTableData = await findOrganismAndData(accessionVersion);
+const sequenceCitations = (await sequenceCitationsPromise)?.unwrapOr(undefined);
```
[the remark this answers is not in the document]
Comment (tombch) [reply website/src/pages/seq/[accessionVersion]/index.astro]: Updated in both cases of `findOrganismAndData` - doesn't handle the promise resolution until after any redirects or not-found responses. I realise this is currently duplicated code that would be best incorporated into the `findOrganismAndData` function. I'm thinking in a follow up to refactor this function to include citations, as well as remove what looks like a duplicated fetch for data use terms across all organisms

anna-parker (approved):
approving as this seems to be working well on staging!

PR #6495, opened 2026-05-22
fix(website): use consistent checkbox css in search filters

### LINKED ISSUE #6483: Strictness checkbox in collection date range filter jumps on page load

corneliusroemer:
We used different css for segment filters and strictness checkbox.

The checkbox was the only one to use DaisyUI CSS classes. Replacing with the standard pattern we use for checkboxes across search UI.

Tested locally: no more jumping, consistent look.

Resolves #6483

### Screenshots

Before:

Google Chrome 2026-05-22 15 57 21

After:

Google Chrome 2026-05-22 15 57 12

### DIFF
--- a/website/src/components/SearchPage/fields/DateRangeField.tsx
+++ b/website/src/components/SearchPage/fields/DateRangeField.tsx
@@ -121,7 +121,7 @@ export const DateRangeField = ({ field, fieldValues, setSomeFieldValues }: DateR
                     <span className='text-gray-400 text-sm mr-2'>strict</span>
                     <input
                         type='checkbox'
-                        className='checkbox checkbox-sm text-3xl [--chkbg:white] [--chkfg:theme(colors.gray.700)] checked:border-gray-300'
+                        className='h-4 w-4 rounded border-gray-300 text-primary-600 focus:ring-primary-600 cursor-pointer'
                         checked={strictMode}
                         onChange={(event) => setStrictMode(event.target.checked)}
                     />


theosanderson:
I like avoiding the animation (and generally I think we should move away from daisy-ui, as we have been) but I actually quite like the existing style otherwise. I think this is a bit too prominent and a bit too similar to the triangles below.

image

corneliusroemer:
@theosanderson You'd like to keep the inconsistent styles - or you think the inconsistency is ok because it's different semantics? Or change all checkbox styles to be more similar to DaisyUI rather than Tailwind?

This is current inconsistent state:

Google Chrome 2026-05-27 10 25 12

Vs:

Google Chrome 2026-05-27 10 25 27

anna-parker:
I also think its nice to use the same checkbox style! I dont mind which one we choose though

theosanderson:
we have `strict` written in grey to make it less salient. In the current version, the checkbox largely matches that:

image

Similarly, for required segments it matches (black text, salient checkbox) - although this one doesn't matter so much because people won't tick it super often so it will mostly be seen in unchecked state.

image


In the new version

image

The checkbox, to me, is the most salient thing there - even though it's the least relevant thing for people to interact with (leaving it on strict is mostly what people will want) - so to me it's both wrong that it's so salient and also it doesn't match with the grey text.

(so yeah, semantically different IMO)

PR #6499, opened 2026-05-26
fix!(prepro): Make host field curatable for INSDC-ingested sequences

maverbiest:
## BREAKING CHANGE

A prepro function has changed its name for clarity, please change
```
preprocessing:
   function: validate_host
```
to:
```
preprocessing:
   function: resolve_host_taxon_id
```

### Summary 

For INSDC ingested sequences, we currently skip host validation and just take the `hostTaxonId` as-is, if it's provided.

This was done because we did not want to overwrite INSDC-ingested information during preprocessing, as it would be a type of curation. If a hypothetical INSDC sequence has a `hostTaxonId` pointing to organism A, but a scientific name pointing to organism B, our preprocessing would overwrite this so both fields point to organism A. While this may be correct, it would mean that this sequence looks different on our end versus in INSDC.

However, we're now running into issues where we can't perform desired curations on INSDC-ingested data, specifically for sequences where a `hostNameScientific` is present but the `hostTaxonId` is missing. This PR aims to correct this.

### PR Checklist
I signed in as superuser and confirmed I could revise/curate a sequence by editing the host field:
image

🚀 Preview: https://remove-insdc-special-case.loculus.org

### DIFF
--- a/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
@@ -1646,7 +1646,9 @@ class ProcessingFunctions:
                 message=f"Host validation for '{unvalidated}' failed with code {response.status_code}: {body.get('detail', '')}",
             )
             return ProcessingResult(
-                datum=unvalidated if args["is_insdc_ingest_group"] else None,
+                datum=unvalidated
+                if args["is_insdc_ingest_group"] and unvalidated.isdigit()
+                else None,
                 warnings=[message] if args["is_insdc_ingest_group"] else [],
                 errors=[message] if not args["is_insdc_ingest_group"] else [],
             )
@@ -1687,16 +1689,6 @@ class ProcessingFunctions:
         input_fields: list[str],
         args: FunctionArgs,
     ) -> ProcessingResult:
-        # for INSDC-ingested sequences we just return the name INSDC gives us (if any)
-        # if we ever want to change this behaviour, we just have to remove this short-circuit,
-        # the rest of the function is set up to validate INSDC-ingested data as well
-        if args["is_insdc_ingest_group"]:
-            return ProcessingResult(
-                datum=input_data.get("hostNameScientific"),
-                warnings=[],
-                errors=[],
-            )
-
         tax_service = args.get("taxonomy_service_url")
         if not tax_service:
             return missing_taxonomy_service_error(input_fields, output_field)
@@ -1717,17 +1709,18 @@ class ProcessingFunctions:
 
         body = response.json()
         if response.status_code != requests.codes.ok:
+            message = ProcessingAnnotation.from_fields(
+                input_fields,
+                [output_field],
+                AnnotationSourceType.METADATA,
+                message=f"Could not map '{tax_id}' to scientific name. Code {response.status_code}: {body.get('detail', '')}",
+            )
             return ProcessingResult(
-                datum=None,
-                warnings=[],
-                errors=[
-                    ProcessingAnnotation.from_fields(
-                        input_fields,
-                        [output_field],
-                        AnnotationSourceType.METADATA,
-                        message=f"Internal error: could not map '{tax_id}' to scientific name. Code {response.status_code}: {body.get('detail', '')}",
-                    )
-                ],
+                datum=input_data.get("hostNameScientific")
+                if args["is_insdc_ingest_group"]
+                else None,
+                warnings=[message] if args["is_insdc_ingest_group"] else [],
+                errors=[message] if not args["is_insdc_ingest_group"] else [],
             )
 
         scientific_name = body.get("scientific_name")
--- a/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
@@ -1646,7 +1646,7 @@ class ProcessingFunctions:
                 message=f"Host validation for '{unvalidated}' failed with code {response.status_code}: {body.get('detail', '')}",
             )
             return ProcessingResult(
-                datum=None,
+                datum=unvalidated if args["is_insdc_ingest_group"] else None,
                 warnings=[message] if args["is_insdc_ingest_group"] else [],
                 errors=[message] if not args["is_insdc_ingest_group"] else [],
             )
--- a/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
@@ -1605,23 +1605,6 @@ class ProcessingFunctions:
         return the tax_id of the most generic taxon (i.e., the one that's closest to
         the root of the taxonomy)
         """
-        # for INSDC-ingested sequences we just return the id INSDC gives us (if any)
-        # if we ever want to change this behaviour, we just have to remove this short-circuit,
-        # the rest of the function is set up to validate INSDC-ingested data as well
-        if args["is_insdc_ingest_group"]:
-            tax_id = input_data.get("hostTaxonId")
-            if not isinstance(tax_id, str) or not tax_id.isdigit():
-                return ProcessingResult(
-                    datum=None,
-                    warnings=[],
-                    errors=[],
-                )
-            return ProcessingResult(
-                datum=tax_id,
-                warnings=[],
-                errors=[],
-            )
-
         tax_service = args.get("taxonomy_service_url")
         if not tax_service:
             return missing_taxonomy_service_error(input_fields, output_field)
--- a/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
@@ -1589,7 +1589,7 @@ class ProcessingFunctions:
         )
 
     @staticmethod
-    def validate_host(
+    def resolve_host_taxon_id(
         input_data: InputMetadata,
         output_field: str,
         input_fields: list[str],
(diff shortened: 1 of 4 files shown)


preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
anna-parker: We should have just modified the code here to return `unvalidated` for the ingest ingest user. This will have the same behavior as currently (no fields will be removed) but allow curators to supply data via `input_data.get("host")`. 
```suggestion
                datum=unvalidated if args["is_insdc_ingest_group"] else None,
```

anna-parker: this is also inline with how we handle invalid countries for the INSDC ingest user

maverbiest: What about this situation, though: INSDC sequence has no `hostTaxonId` but has a `hostNameScientific` (this is what we saw yesterday). This means that `unvalidated` will take the value of `hostNameScientific`, if this scientific name is incorrect and fails hostname validation, we will now attempt to set the value of `hostTaxonId` to the value provided as `hostNameScientific` by the INSDC sequence.

maverbiest: I guess we could do:
```python
            return ProcessingResult(
                datum=unvalidated
                if args["is_insdc_ingest_group"] and unvalidated.isdigit()
                else None,
                warnings=[message] if args["is_insdc_ingest_group"] else [],
                errors=[message] if not args["is_insdc_ingest_group"] else [],
            )
```

anna-parker: I actually wonder if we should rename this to host_to_hostId

anna-parker: sounds good, could you give me an example of a sequence from the INSDC without the host taxonId but with a host scientific name I dont think I have ever seen this

maverbiest: It came up when discussing this sequence: https://pathoplexus.org/andv/search?accession=PP_006VLTW&selectedSeq=PP_006VLTW.1

Here is the thread on slack: https://loculus.slack.com/archives/C06JCAZLG14/p1779736209859899

anna-parker: If I understand the thread, this sequence doesnt seem to have any host information on INSDC, I believe the homo sapiens host was added by the curator

anna-parker: we can do this later

anna-parker: but again lets add the isdigit check and then we can merge :-)

maverbiest: Yes I believe it's the case that this was because of a curation, I also never saw something like this in data on INSDC when we were implementing this feature originally.

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -1732,17 +1707,18 @@ def scientific_name_from_id(
 
         body = response.json()
         if response.status_code != requests.codes.ok:
+            message = ProcessingAnnotation.from_fields(
```
corneliusroemer: Here we definitely want to log something

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
corneliusroemer, line 1641: We should distinguish between invalid organism, i.e. the taxon id/name doesn't exist - which is going to be a very specific HTTP code 422, and other errors like internal server error etc. Right now it treats invalid user data the same as user did everything right but we did something wrong.

corneliusroemer: Function name should be changed to make it clear that this function outputs a taxon id. It doesn't just validate.

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -1732,17 +1707,18 @@ def scientific_name_from_id(
 
         body = response.json()
```
corneliusroemer: This is dangerous - response might not be valid json, e.g. you could get something like 

```
HTTP/1.1 500 Internal Server Error

Content-Type: text/html

<html>Internal Server Error</html>
```

preprocessing/nextclade/tests/test_host_name_validation.py
```diff
@@ -90,14 +93,19 @@ def test_host_processing_insdc(mock_session: MagicMock) -> None:
 
     result = process_all([entry], "temp", config)
     metadata = result[0].processed_entry.data.metadata
+    print(result)
```
corneliusroemer: Probably want to get rid of print here

preprocessing/nextclade/tests/test_host_name_validation.py
corneliusroemer: I'm not sure this comment accurately describes what the test does?

preprocessing/nextclade/tests/test_host_name_validation.py
```diff
@@ -116,12 +124,11 @@ def test_host_processing_invalid_hostname(mock_session: MagicMock) -> None:
     metadata = result[0].processed_entry.data.metadata
 
     assert metadata["hostTaxonId"] is None
-    assert metadata["hostNameScientific"] == "not a real species"
+    assert metadata["hostNameScientific"] == None
```
corneliusroemer: Why the change here?

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
anna-parker, line 1697: we are now returning too early here (then no test change required
```suggestion
        if not tax_id and not args["is_insdc_ingest_group"]:
```

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
anna-parker: true - probably best to put it inside the try catch statement

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
anna-parker: this can be done in a follow up PR :-)

preprocessing/nextclade/tests/test_host_name_validation.py
maverbiest: Ah yes thanks

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
maverbiest: Ahh yes good catch thanks

maverbiest: I've renamed the function to `resolve_host_taxon_id`, but am open to suggestions if you don't like it!

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
maverbiest: I'm now logging a warning message (the same message that gets attached as a ProcessingAnnotation)

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
maverbiest: Thanks! I now do this in the try/catch statement as suggested. It should be caught by `except requests.exceptions.RequestException`

preprocessing/nextclade/tests/test_host_name_validation.py
maverbiest: You're right, it had drifted from the actual implementation. I've updated now

preprocessing/nextclade/tests/test_host_name_validation.py
maverbiest: It was because the function was returning too early after removing the short-circuit for INSDC ingest user from `scientific_name_from_id` (see @anna-parker 's comment below)

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
anna-parker: made issue: https://github.com/loculus-project/loculus/issues/6523

anna-parker, line 1656: I just realized if it doesnt return a list its in an unexpected format and might not be a dict right? so we cant call get on it, and we probably need to error? e.g. `AttributeError: 'NoneType' object has no attribute 'get'`

anna-parker:
the preview is having issues, I am removing the tag and after it is killed I will add the label back

anna-parker (approved):
the preview is still having issues but I think this should be fine - we really need to figure out what is up with all the sync issues

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
anna-parker: https://github.com/loculus-project/loculus/issues/6532

maverbiest: It can be a list of taxa or just a plain taxon, depending on whether `host` is a scientific name or a taxon id, so it's not unexpected for it to not be a list:

If `unvalidated` is a scientific name, prepro calls `GET /taxa?scientific_name=<name>`, and the taxonomy service uses `fetch_by_sci_name` to see find taxa with that name. This returns a list of taxa (or None) since multiple taxa can have the same scientific name. If `unvalidated` is a taxon id, prepro calls `GET /taxa/<id>`, andtaxonomy service uses `fetch_by_id` to validate, which just returns one taxon (or None) since ids are unique.

(if either of these return None, the taxonomy service returns a 404, which we already guard against in the block above)

PR #6501, opened 2026-05-27
feat!(prepro): accept ISO and lucene ranges

### LINKED ISSUE #6468: Prepro: accept "Lucene style" date range
The NCBI datasets CLI is now returning date ranges (yay!) but these use the Lucerne style date range (date TO date) which prepro doesnt accept, leading to data revisions erroring with 

Image

We need to update the prepro pipeline to accept this range and then reprocess all data afterwards.

See details in https://loculus.slack.com/archives/C05G172HL6L/p1779197286791419

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/6468

Prepro now accepts `sampleCollectionDate` date ranges in the ISO and lucene format (the `parse_date_into_range` parses this input into a range). The new range is used for calculating the upper and lower bound of the sampleCollectionDate, additionally `dateRangeString` (or string version of the range) is returned by this function. The `dateRangeString` format is a compressed ISO inspired format (see below)

```
Output of the new `dateRangeString` function:
    - Single day:           2026-05-28
    - Full month:           2026-05
    - Full year:            2026
    - Multiple full years:  2024/2026 (ISO range)
    - Multiple full months: 2025-01/2026-03 (ISO range)
    - Arbitrary range:      2026-05-10/2026-05-28 (ISO range)
```

Note the sampleCollectionDate (in dateRangeString format) is used as input for the displayName. The displayName uses "/" to separate fields, thus using the ISO format for ranges is not ideal. Here I opted to use the lucene format for displayNames and updated the concatenate function to handle a new input format: `dateRangeString`.

### BREAKING CHANGES

This PR should be relatively backward compatible - the only change is that where we restricted the year/month possible sampleCollectionDate using releasedDate (this was seen in the upper and lower bounds) we now also see this restriction in the date string (see example below). 

To use this PR on PPX we need to update the config and reprocess all submissions: https://github.com/pathoplexus/pathoplexus/pull/980

Ideally users of the `concatenate` function that are using a date as an input type might want to consider switching to a `dateRangeString` if their date can also be a date range (i.e. if their input is the `dateRangeString` output of the `parse_date_into_range` function).

### PR Checklist
All 50 erroring CCHF sequences on main, now are correctly processed.
Example erroring sequence on main: 
image
Now succeeding sequence on this branch:
image

- [x] run regression test against main - the 50 CCHF sequences that were missing are now back, additionally the releaseDate now adds a nice upperBound which is visible for a couple sequences e.g.
image
instead of:
image

🚀 Preview: https://accept-ranges.loculus.org


kubernetes/loculus/values.yaml
```diff
@@ -124,7 +124,7 @@ defaultOrganismConfig: &defaultOrganismConfig
       - name: sampleCollectionDate
         displayName: Collection date
         definition: "The date on which the sample was collected."
-        guidance: "Please format the date as `YYYY-MM-DD`. If day is unknown, use `YYYY-MM` (e.g., `2020-03`). If both day and month are unknown, use `YYYY` (e.g., `2020`). Please provide at least a year."
+        guidance: "Please format the date as `YYYY-MM-DD`. If day is unknown, use `YYYY-MM` (e.g., `2020-03`). If both day and month are unknown, use `YYYY` (e.g., `2020`). If only a range is known please supply the range in ISO format `date_lower/date_upper` where `date_lower` and `date_upper` are in the format `YYYY-MM-DD`, `YYYY-MM` or `YYYY`. Please provide at least a year."
```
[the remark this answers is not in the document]
anna-parker: we shouldn't overwhelm the users with options

preprocessing/nextclade/tests/test_metadata_processing_functions.py
```diff
@@ -966,6 +966,151 @@ def test_parse_date_into_range() -> None:
...
+            {"date": "[2021-01-01 TO 2021-12-31]"},
+            "field_name",
+            ["field_name"],
+            {
+                "fieldType": "dateRangeUpper",
+                "submittedAt": ts_from_ymd(2021, 6, 15),
+            },
+        ).datum
+        == "2021-06-15"
+    ), "dateRangeUpper: lucerne range upper bound should be tightened by submittedAt."
```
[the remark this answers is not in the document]
anna-parker: 2 and 3 can be ignored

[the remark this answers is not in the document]
anna-parker: oh no this is actually covered

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -450,131 +505,123 @@ def parse_date_into_range(
...
-                        [output_field],
-                        AnnotationSourceType.METADATA,
-                        message=(
-                            f"Metadata field {output_field}:'{input_date_str}' is in the future."
-                        ),
+                    datum = DateRange(
+                        date_range_string=date_range_string,
+                        date_range_lower=lower_date.date_range_lower,
+                        date_range_upper=upper_date.date_range_upper,
                     )
```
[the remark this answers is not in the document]
anna-parker: we test this later on

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -450,131 +506,125 @@ def parse_date_into_range(
                 errors=[],
             )
 
-        formats_to_messages = {
```
anna-parker, line 453: this has been moved into a separate function as it is reused

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -450,131 +506,125 @@ def parse_date_into_range(
...
-            if message:
-                warnings.append(
-                    ProcessingAnnotation.from_fields(
-                        input_fields,
-                        [output_field],
-                        AnnotationSourceType.METADATA,
-                        message=f"Metadata field {output_field}:'{input_date_str}' - " + message,
+                    if lower_date is None or upper_date is None:
+                        msg = "Could not parse lower or upper date in range"
+                        raise ValueError(msg)
```
corneliusroemer: You raise a message and throw it away later

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -450,131 +506,125 @@ def parse_date_into_range(
...
 
-            if message:
-                warnings.append(
-                    ProcessingAnnotation.from_fields(
-                        input_fields,
-                        [output_field],
-                        AnnotationSourceType.METADATA,
-                        message=f"Metadata field {output_field}:'{input_date_str}' - " + message,
+                    if lower_date is None or upper_date is None:
+                        msg = "Could not parse lower or upper date in range"
```
anna-parker: log this and just break out dont throw an error

preprocessing/nextclade/tests/test_metadata_processing_functions.py
```diff
@@ -966,6 +966,151 @@ def test_parse_date_into_range() -> None:
...
+            {"date": "[2021 TO 2021-06]"},
+            "field_name",
+            ["field_name"],
+            {
+                "fieldType": "dateRangeString",
+                "submittedAt": ts_from_ymd(2022, 1, 1),
+            },
+        ).datum
+        == "2021/2021-06"
+    ), "dateRangeString: lucerne range should be returned in ISO format."
```
corneliusroemer: `s/lucerne/lucene`

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
anna-parker: updated and added a test :-)

theosanderson:
@theosanderson-agent for interest, try to make a PR where the dateString is a downstream property of the upper and lower dates

<details>
Here's the conversation with names stripped:

Remind me how NCBI is using this - is it mostly entire years/months? [2021-01-01_TO_2021-12-31]
they do both
yeah but do they do anything which is neither?
Im not sure I understand what you mean... I accept date like YYYY YYYY-MM and YYYY-MM-DD
the examples in the PR are year to year
I was asking if NCBI has any ranges which are not a full year or full month
Basically I would say we should:

if the range spans at most a month we should format in the display name as 2012-01
if the range spans more than a month but at most a year we should format in the display name as: 2012
if the display name spans multiple years - idk

augur only gives an example of year to year but as lucene can have full day to day think its fine to include it: https://github.com/nextstrain/augur/pull/2001/changes
if the range spans more than a month but at most a year we should format in the display name as: 2012
I feel like this starts getting quite custom...
if the display name spans multiple years - idk
these are the only cases we will have atm
these are the only cases we will have atm
ok this was my question
well in that case there's little reason to have my custom logic for now
https://pathoplexus.org/ebola-bdbv/search?order=ascending&page=1&selectedSeq=PP_006X6A7.1 would you be proposing to change this display name to Lucene style?
no no
this will stay as is
however in theory if someone submitted it as [2021-01-01 TO 2021-12-31] I would keep it as lucene
Yeah, that seems odd I guess
to not have that case we need a lot more logic
I feel like the logic should be that 2021 gets converted internally to the range (which it is, in part), and the display name comes from the range
sure I mean we can have a solution we don't think is optimal for pragmatism
the range would be [2021-01-01 TO 2021-12-31]
so you would need to add a function to compress the range to a single year, alll doable but we just need to account for leap years
hah not sure how leap years come in?
well the number of days in a month changes
we can only convert compress to a month if the start and end of the range are the start and end of a month
sure you were talking about year
... but the string version of 2012-01 would be [2012-01-01  to 2012-01-31]
so you would need to do both, I just think this is unnecessarily complex
yep I get how february for leap years needs a date library
I think its unlikely we will hit this edge case but if you think its important I can change
I think in terms of the eventual goal for what is optimal, it should definitely be that [2021-01-01 TO 2021-12-31] and 2021 give consistent outcomes, but as above
sure I mean we can have a solution we don't think is optimal for pragmatism
So basically the dateString would be a derived property of the lower and the upper , rather an a parallel thing
but yeah, happy with anything for now - great to unblock NCBI
</details>

theosanderson-agent:
Anya edit/update: this has been added to this PR (with further improvements :-) )

Took a stab at this in #6507 (draft, based on `accept_ranges`).

`DateRange` no longer carries `date_range_string` — it's a `@property` derived from `(lower, upper)` via `derive_date_range_string`:

- `lower == upper` → `YYYY-MM-DD`
- spans a full calendar month (leap-year aware via `calendar.monthrange`) → `YYYY-MM`
- spans a full calendar year → `YYYY`
- otherwise → `<lower>/<upper>` where each side picks the most compact form

So `2021` and `[2021-01-01 TO 2021-12-31]` now both yield `"2021"`. Some other knock-on changes:
- `[2021-05-01 TO 2021-06-30]` → `"2021-05/2021-06"` (compact)
- `2021-12` with `submittedAt=2021-12-15` → `"2021-12/2021-12-15"` (the derived string now reflects the tightened upper bound)

All 32 unit tests pass. Happy to fold into #6501 if you'd prefer to ship as one — left it as a draft so it's easy to compare/revert.

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -272,6 +272,95 @@ def taxonomy_network_error(
...
+    if (
+        lower.month == 1
+        and lower.day == 1
+        and upper.month == 12  # noqa: PLR2004
+        and upper.day == 31  # noqa: PLR2004
+    ):
+        return f"{lower.year}/{upper.year}"
+    if lower.day == 1 and upper.day == calendar.monthrange(upper.year, upper.month)[1]:
+        return f"{lower.year}-{lower.month:02d}/{upper.year}-{upper.month:02d}"
+    return f"{lower.strftime('%Y-%m-%d')}/{upper.strftime('%Y-%m-%d')}"
```
tombch, line 361: I might be misunderstanding, but with this function to compact a date range, don't we lose the information of exact dates that _were_ the first or last of the month? Could an approach be to also attach the raw lower and upper dates as properties on the `DateRange` object so we know when it did have those exact dates?

tombch: Me and Anya discussed this in person and determined this is how it should be!

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -826,6 +931,18 @@ def add_errors():
                         if null_per_backend(processed.datum)
                         else str(processed.datum)
                     )
+                elif field_types[i] == "dateRangeString":
+                    # Assumes date ranges are in ISO format
+                    raw_value = str(input_data[order[i]])
```
tombch: Claude flagged this - the `input_data[order[i]]` is type `InputMetadataValue` which could potentially be `None`, this would get converted to `"None"`, which wouldn't be picked up by `null_per_backend` and then make it into the date range string

tombch: Suggested fix:

```
raw = input_data[order[i]]
if null_per_backend(raw):
    formatted_input_data.append(fallback_value)
else:
    raw_value = str(raw)
...
```

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
corneliusroemer, line 1: Might make sense to move the date stuff into a new file, this file is getting ridiculously long

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -272,6 +272,95 @@ def taxonomy_network_error(
...
+    """
+    if lower == upper:
+        return lower.strftime("%Y-%m-%d")
+    if (
+        lower.day == 1
+        and upper.day == calendar.monthrange(upper.year, upper.month)[1]
+        and lower.month == upper.month
+        and lower.year == upper.year
+    ):
+        return f"{lower.year}-{lower.month:02d}"
```
corneliusroemer: Couldn't this also be written more succinctly as `lower.strftime("%Y-%m")`?

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -272,6 +272,95 @@ def taxonomy_network_error(
...
+    if (
+        lower.month == 1
+        and lower.day == 1
+        and upper.month == 12  # noqa: PLR2004
+        and upper.day == 31  # noqa: PLR2004
+        and lower.year == upper.year
+    ):
+        return f"{lower.year}"
```
corneliusroemer:
```suggestion
        return lower.strftime("%Y")
```

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -272,6 +272,95 @@ def taxonomy_network_error(
     )
 
 
+@dataclass
+class DateRange:
+    date_range_lower: datetime
```
corneliusroemer, line 277: Since we only support date we might use `date` throughout instead of `datetime` for clearer semantics

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -272,6 +272,95 @@ def taxonomy_network_error(
...
+                    date_range_upper=parsed_date.replace(month=12, day=31),
+                    message=msg,
+                )
+        break
+    return None
+
+
+def derive_date_range_string(lower: datetime, upper: datetime) -> str:
+    """
+    Convert a date range into the most compact string representation possible.
```
corneliusroemer, line 326: One question is what to do with mixed precision like `2024/2025-04-01` right now this would end up as `2024-01-01/2025-04-01` IIUC

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -272,6 +272,95 @@ def taxonomy_network_error(
...
+        return f"{lower.year}"
+    if (
+        lower.month == 1
+        and lower.day == 1
+        and upper.month == 12  # noqa: PLR2004
+        and upper.day == 31  # noqa: PLR2004
+    ):
+        return f"{lower.year}/{upper.year}"
+    if lower.day == 1 and upper.day == calendar.monthrange(upper.year, upper.month)[1]:
+        return f"{lower.year}-{lower.month:02d}/{upper.year}-{upper.month:02d}"
```
corneliusroemer:
```suggestion
        return lower.strftime("%Y-%m") + "/" + upper.strftime("%Y-%m")
```

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -394,25 +483,30 @@ def check_date(
...
+        - YYYY-MM-DD
+        Or date string as a range formatted as:
+        - ISO format: date/date (where date is in one of the above formats)
+        - lucene format: [date TO date] (where date is in one of the above formats)
+        Uses (`input.releaseDate`) as an upper bound when parsing into a range.
         Return value determined FunctionArgs:
         fieldType: "dateRangeString" | "dateRangeLower" | "dateRangeUpper"
-        Default fieldType is "dateRangeString"
+        Default fieldType is "dateRangeString" - in the case of a range the date is returned
+        as a single month/year or an ISO date range format
```
corneliusroemer:
```suggestion
        as a single month/year or an ISO date range format by function `derive_date_range_string`
```

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
anna-parker: yup that is the current behavior, I think its fine for now as mixing precision does look a bit odd (and it would require even more logic), but open to changing if you think it would be an improvement

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
anna-parker: ah yeah that works think! merged these changes in one commit

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -450,131 +544,145 @@ def parse_date_into_range(
...
+                    return ProcessingResult(
+                        datum=None,
+                        warnings=[],
+                        errors=[
+                            ProcessingAnnotation.from_fields(
+                                input_fields,
+                                [output_field],
+                                AnnotationSourceType.METADATA,
+                                message=f"Metadata field {output_field}: "
+                                f"Detected data range but could not parse date: {input_date_str}.",
```
tombch: Not blocking but should `data range` be `date range`

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
anna-parker: thanks!

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
anna-parker:
```suggestion
                                f"Detected date range but could not parse date: {input_date_str}.",
```

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -450,131 +544,145 @@ def parse_date_into_range(
...
+                    ),
+                )
+            )
 
-        # If all parsing attempts fail, it's an unrecognized format
-        return ProcessingResult(
-            datum=None,
-            warnings=[],
-            errors=[
+        if datum.date_range_lower > datum.date_range_upper:
```
tombch: Having this check occur after the clamping of `date_range_upper` to max of release date means this error could also fire even if the users range was ok (but after the release date). E.g.:

```
2025-08-01/2025-12-31 with release_date = 2025-06-01:

Upper gets clamped from 2025-12-31 → 2025-06-01.
The existing "is after release date" error fires (lower=2025-08-01 > release_date).
The new "invalid date range" error also fires (lower=2025-08-01 > clamped upper=2025-06-01), even though the user's input range was perfectly well-ordered.
```

(Shoutout to Claude for spotting this)

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
anna-parker: I think we did this to have an easy comparison with the datetime object max_upper_limit (predates this PR) that is created from a timestamp so think its ok to keep as a datetime object for now

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
corneliusroemer: Yes it's None if there's no date supplied, in practice this happens for ingested data, e.g. https://pathoplexus.org/seq/PP_000SW83.1

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
anna-parker: thanks - renamed!

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
corneliusroemer, line 940: Why do we still need this date field type? Isn't this covered by dateRangeString now? Is it actually used anywhere?

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -826,6 +931,18 @@ def add_errors():
...
+                    if raw_value.count("/") > 1:
+                        date_string = None
+                    elif raw_value.count("/") == 1:
+                        date_string = "[" + raw_value.replace("/", " TO ") + "]"
+                    else:
+                        date_string = raw_value
+                    formatted_input_data.append(
+                        fallback_value if null_per_backend(date_string) else str(date_string)
+                    )
                 elif field_types[i] == "timestamp":
```
corneliusroemer, line 964: Is this used anywhere? Not by PPX so maybe we can remove to simplify?

This is how we use it in PPX, the only place we use `concatenate`

```
      - name: displayName
        displayName: Display name
        definition: A human-readable label for the sequence record, with the format `{geoLocCountry}/{accessionVersion}/{sampleCollectionDate}`.
        preprocessing:
          function: concatenate
          inputs:
            geoLocCountry: geoLocCountry
            sampleCollectionDate: sampleCollectionDate
          args:
            order: [geoLocCountry, ACCESSION_VERSION, sampleCollectionDate]
            type: [string, ACCESSION_VERSION, string]
        noInput: true
```

anna-parker: date is used by Genspectrum

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
corneliusroemer, line 847: The concatenate function is quite complex and we only seem to use it to derive a display name. Should we replace it with a much simpler display name function that does one job and get rid of the code we don't actually use here?

anna-parker: sorry date is used by date is used by Genspectrum probably easiest to keep for now to not break them

anna-parker: I would be ok with simplifying it in a later PR :-)

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
anna-parker: thanks thanks! I moved the error up :-)

corneliusroemer (commented):
Thanks, this is a good feature to have.

Unclear from your PR description:
- Why do you say `dateRangeString` "us[es] the ISO format as a default."? Can it also output something else? Maybe we should name it `dateRangeISO`
- What do we use `dateRangeString` for? Is this what we now show for `collectionDate` - i.e. the single string field we show on seq details and output into metadata?
- I'm not sure we need the square brackets around the date string in the display name? Why `Turkey/LOC_XYZ/[2012_TO_2015]` if we can also just do `Turkey/LOC_XYZ/2012_TO_2015`
- Short summary of what is breaking in PR desc would be good. IIUC this should be backwards compatible, we just accept more types of dates, nothing existing should change?
- Let's add a TODO checklist of follow-up things like changing PPX docs, reprocessing etc.

Should we maybe add some coverage for this in integration tests? We can change one of the standard Ebola sequences to use a date range as input and then verify display name etc works correctly?

Going through the code the main thoughts are:
- I'm not a fan of the concatenate function. It's 160 lines long and only used to derive the display name so we could just make a simple display name function instead with less complexity
- Do we want mixed precision in date strings? Right now we won't get this output: `2025/2026-01-06` but this instead: `2025-01-01/2026-01-06` IIUC

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
anna-parker: but yes we could probably remove the timestamp option (but lets do that in a different PR)

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -826,6 +931,22 @@ def add_errors():
                         if null_per_backend(processed.datum)
                         else str(processed.datum)
                     )
+                elif field_types[i] == "dateRangeString":
+                    # Assumes date ranges are in ISO format
+                    raw = input_data[order[i]]
+                    if null_per_backend(raw):
+                        formatted_input_data.append(fallback_value)
+                        break
```
tombch: Shouldn't this `break` be a `continue`? currently with this it would exit the loop if any null value is encountered right? Which I guess isn't an issue with the date at the end currently but could trip up in future

anna-parker:
@corneliusroemer 

1. Why do you say dateRangeString "us[es] the ISO format as a default."? Can it also output something else? Maybe we should name it dateRangeISO

I will update the description but we use ISO and a compressed version of ISO i.e. yyyy instead of 2022-01-01/2022-12-31 - no other option is available.

2. What do we use dateRangeString for? Is this what we now show for collectionDate - i.e. the single string field we show on seq details and output into metadata?

exactly we use it here

3. I'm not sure we need the square brackets around the date string in the display name? Why Turkey/LOC_XYZ/[2012_TO_2015] if we can also just do Turkey/LOC_XYZ/2012_TO_2015

I also wasnt sure about this- I added them to be more lucene style but Im happy to remove

4. Short summary of what is breaking in PR desc would be good. IIUC this should be backwards compatible, we just accept more types of dates, nothing existing should change?

Should be backward compatible, but a config change is required and reprocessing, hence the breaking change (I think I say that)


5. I'm not a fan of the concatenate function. It's 160 lines long and only used to derive the display name so we could just make a simple display name function instead with less complexity

Nothing against this - I would just prefer to do this in a separate function.


6. Do we want mixed precision in date strings? Right now we won't get this output: 2025/2026-01-06 but this instead: 2025-01-01/2026-01-06 IIUC

I'm not a fan of mixed precision so would vote to keep as is but no string feelings (we can always change later)

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
anna-parker: omg yes that was a brain fart

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -450,131 +544,145 @@ def parse_date_into_range(
...
-            logger.debug(f"parsed_date: {datum}")
-
-            if datum.date_range_upper > max_upper_limit:
-                logger.debug(
-                    "Tightening upper limit due to release date or current date. "
-                    f"Original upper limit: {datum.date_range_upper},"
-                    f"new upper limit: {max_upper_limit}"
+                datum = DateRange(
+                    date_range_lower=lower_date.date_range_lower,
+                    date_range_upper=upper_date.date_range_upper,
```
tombch, line 598: Not blocking as this might be intentional, just to note - this discards the warning messages for the `lower_date` and `upper_date`

anna-parker: no this is probably nice to have - I have added some changes :-)

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
anna-parker: created a follow up issue: https://github.com/loculus-project/loculus/issues/6526

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
anna-parker: https://github.com/loculus-project/loculus/issues/6526

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
tombch: Nice! Only thing i can see is the warning messages would have a double full stops as the upper and lower date messages also have full stops

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -588,12 +579,36 @@ def parse_date_into_range(  # noqa: C901, PLR0912, PLR0915
...
                 )
 
+        if datum and datum.message:
+            warnings.append(
+                ProcessingAnnotation.from_fields(
+                    input_fields,
+                    [output_field],
+                    AnnotationSourceType.METADATA,
+                    message=f"Metadata field {output_field}:'{input_date_str}' - " + datum.message,
+                )
```
tombch, line 609: Also this would lead to a double concatenated `Metadata field {output_field}:'{input_date_str}' - Metadata field {output_field}:'{input_date_str}' - ...` I think?

anna-parker: good catch!

tombch (approved):
Nice, this looks good to me!

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
anna-parker: closing this for now, we can do this in a follow up PR

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
anna-parker: closing this for now, we can change the logic in future if desired

PR #6502, opened 2026-05-27
chore(docs): migrate astro to v6, tailwind to v4 and fix any arising issues

corneliusroemer:
@anna-parker asked me to look into astro upgrade for docs as dependabot PRs were failing for months.

This PR upgrades astro to v6 (released March 26) and tailwind to v4 (released Nov 24, more than a year ago)

I tested the docs build locally. There are small visual changes due to starlight version upgrade but I haven't seen any bugs. New tailwind version appears to work as well.

~Unfortunately, I don't think we can use previews for docs.~ Learned something new, prepend `docs-` to the URL: https://docs-docs-tailwind-vite.loculus.org

### Screenshot

image

🚀 Preview: https://docs-tailwind-vite.loculus.org


theosanderson:
> Unfortunately, I don't think we can use previews for docs.

We can `docs-x.loculus.org` or something

corneliusroemer:
> We can docs-x.loculus.org or something

Nice, that works, now preview is up

docs/astro.config.mjs
```diff
@@ -1,9 +1,10 @@
 import { defineConfig } from 'astro/config';
 import starlight from '@astrojs/starlight';
-import tailwind from '@astrojs/tailwind';
+import tailwindcss from '@tailwindcss/vite';
 
 // https://astro.build/config
 export default defineConfig({
+    site: 'https://loculus.org',
```
anna-parker, line 7: discussed this is necessary for the site map, it will mean previews the generated canonical URLs and sitemap entries will still point to loculus.org (which seems desired)

anna-parker (approved):
dont fully understand all the changes but clicked around and it looks good to me! thanks Cornelius!

PR #6507, opened 2026-05-27
refactor(prepro): derive dateRangeString from (lower, upper)

theosanderson-agent:
## Summary

Per the suggestion from @theosanderson on #6501, this branch has been reset to `accept_ranges` and now contains the minimal derived-string follow-up.

`DateRange` no longer stores `date_range_string` as parallel state. It stores only `date_range_lower` and `date_range_upper`, and `date_range_string` is computed from those final bounds on demand.

Consistency win: `2021` and `[2021-01-01 TO 2021-12-31]` both yield `"2021"`.

## How it works

`derive_date_range_string(lower, upper)` keeps the existing compression policy from #6501:

| Range | Output |
|---|---|
| `lower == upper` | `YYYY-MM-DD` |
| exact full calendar month | `YYYY-MM` |
| exact full calendar year | `YYYY` |
| otherwise | `YYYY-MM-DD/YYYY-MM-DD` |

This avoids the extra endpoint-level compaction from the previous version of this PR.

## Behavior changes vs #6501

- `[2021-01-01 TO 2021-12-31]` -> `"2021"` instead of preserving `"2021-01-01/2021-12-31"`.
- `2021-12` with `submittedAt=2021-12-15` -> `"2021-12-01/2021-12-15"`, because tightening the upper bound is reflected in the derived string.
- `2021-01/2021-06-30` -> `"2021-01-01/2021-06-30"`, because the original endpoint spelling is no longer preserved.
- `[2021-05-01 TO 2021-06-30]` remains `"2021-05-01/2021-06-30"`; no additional endpoint-shortening rules are added.

## Test plan

- [x] `PYTHONPATH=src pytest tests/test_metadata_processing_functions.py` - 32 passed
- [x] `ruff check --diff` clean (only pre-existing unsafe fixes available)
- [x] `ruff format --check` clean

## Notes

Targets `accept_ranges` so this can be reviewed as a small follow-up on top of #6501.


anna-parker:
changes incorporated in the other PR

PR #6516, opened 2026-05-27
chore(dependabot): ignore dtolnay/rust-toolchain updates as we want to stick to v1.92

corneliusroemer:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,6 +6,7 @@ updates:
       interval: weekly
     ignore:
       - dependency-name: "dtolnay/rust-toolchain"
+        # See https://github.com/loculus-project/loculus/pull/6516#issuecomment-4564199837
   - package-ecosystem: npm
     directory: website/
     schedule:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,8 @@ updates:
     directory: /
     schedule:
       interval: weekly
+    ignore:
+      dependency-name: "dtolnay/rust-toolchain"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -4,6 +4,8 @@ updates:
     directory: /
     schedule:
       interval: weekly
+    ignore:
```
anna-parker, line 7: @corneliusroemer can you link to an issue why we want to pin this? (can be an issue in a rust github)

corneliusroemer: I can link this PR? But the git blame should also show.

corneliusroemer:
Reason for pinning: maintainer uses unconventional release strategy of tagging things in advance of them working. E.g. Rust 1.100 doesn't exist yet so if you use @1.100 it will fail, yet the tag exists. So opting out of dependabot here seems wise. Also, we use a specific rust version to build the image, it's 1.92 so we should also just use that here.

.github/dependabot.yml
```diff
@@ -4,6 +4,8 @@ updates:
...
+      - dependency-name: "dtolnay/rust-toolchain"
```
corneliusroemer, line 8:
```suggestion
      - dependency-name: "dtolnay/rust-toolchain"
        # See https://github.com/loculus-project/loculus/pull/6516#issuecomment-4564199837
```

.github/dependabot.yml
corneliusroemer: done

anna-parker (approved):
thanks!

Issue #6295, opened 2026-04-21
Even when `host` is a required field, I can omit it for bulk-submissions if I have a `hostNameScientific` or `hostTaxonId` column


maverbiest:
Since https://github.com/loculus-project/loculus/pull/6203, users specifiy the host organism for their submissions via the `host` field, and `hostNameScientific` and `hostTaxonId` are `noInput`. Currently, `host` is a required input field for direct dengue and yellow-fever submissions.

In `processing_functions.py`, `validate_host()` is used to validate user-supplied host organisms. Since the `host` field is a new addition, and legacy direct submissions didn't have a `host` field when this was implemented, we get the unvalidated input as follows:

```python
        # host will exist only for direct submissions
        # hostTaxonId and hostNameScientific are noInput, so the only case where they exist is
        # for INSDC ingested sequences or for legacy direct submissions
        unvalidated = (
            input_data.get("host")
            or input_data.get("hostTaxonId")
            or input_data.get("hostNameScientific")
        )
```

The problem here is that, even though `hostTaxonId` and `hostNameScientific` are `noInput`, this only really applies to submissions of single sequences through the form. For bulk submissions, we just take all columns from the input metadata file. If someone provides metadata with a `hostNameScientific` column, this information will happily be used for host validation. 

In principle this is fine, as we validate this input and it will still reject sequences if no `hostNameScientific` and `hostTaxonId` are provided. It is, however, undocumented behaviour and it's inconsistent between bulk submissions and single submissions (as the form will refuse to submit if `host` is not set).

## The solution
We now store `unprocessed_data` in addition to `original_data`. Since we ran a migration to always have a `host` field in the `unprocessed_data` (see https://github.com/pathoplexus/pathoplexus/pull/899), I think we we should now be able to just check the `host` field for direct submissions in `validate_host`, and no longer look at the `hostNameScientific` and `hostTaxonId`. This would mean that direct submissions for organisms with a required `host` field are actually rejected.

theosanderson:
BTW we need to update the PPX example data with this change - definitely before fixign this bug :)

anna-parker:
The proposed solution will break for newly INSDC ingested sequences unless we modify what INSDC ingest submits (i.e. INSDC ingest submits `hostTaxonId` and `hostNameScientific` - we would need to change this and ensure the calculated hash is the same for everything or else all ingested sequences will get a version bump).

Personally I think this isn't that big a problem

maverbiest:
[...]

PR #6534, opened 2026-05-29
feat!(ingest): Add host field for INSDC sequences during ingest

### LINKED ISSUE #6533: Inconstent host processing for INSDC and direct submissions
INSDC-ingested submissions and direct submissions host information is processed differently:
  - Direct submissions provide a single host input field (a taxon id or scientific name). Preprocessing's `resolve_host_taxon_id` validates it against the taxonomy service, and this is then used to derive `hostNameScientific`, `hostNameCommon`.
  - INSDC ingest instead submits `hostTaxonId`, `hostNameScientific` directly as input metadata, and preprocessing short-circuits for the ingest group (passes the values through unvalidated). `hostNameCommon` is still derived using the taxonomy_service in this path.

This divergence means curators editing an ingested sequence never see/edit a host field. The host inputs are `noInput: true`.

We should make INSDC-ingested sequences carry a single host input field (populated during ingest from the taxon id, falling back to the scientific name) and be processed uniformly with direct submissions.

maverbiest:
resolves https://github.com/loculus-project/loculus/issues/6533 https://github.com/loculus-project/loculus/issues/6295

This PR modifies ingest to ensure host organism information is processed uniformly for INSDC-ingested sequences and direct submissions. Because of this, we can also remove INSDC specific behaviour from preprocessing. This will be done in a separate PR once we know this ingest change is correct and stable.

Concretely, ingest now collapses submissions with `hostTaxonId` and/or `hostNameScientific` into a single `host` field (removing `hostTaxonId` and`hostNameScientific`). The `hostNameCommon` field is also removed. 

## Breaking changes
Not strictly breaking, but when rolling out this PR, this DB surgery needs to be run to make sure existing INSDC submissions fit the new model:

<details>
<summary>An old command with a WHERE clause that was too strict </summary>

The command to run:

```sql
UPDATE sequence_entries
SET unprocessed_data = jsonb_set(
    unprocessed_data,
    '{metadata}',
    ((unprocessed_data -> 'metadata') - 'hostTaxonId' - 'hostNameScientific' - 'hostNameCommon')
    || CASE
         WHEN COALESCE(
                NULLIF(unprocessed_data -> 'metadata' ->> 'hostTaxonId', ''),
                NULLIF(unprocessed_data -> 'metadata' ->> 'hostNameScientific', '')
              ) IS NOT NULL
         THEN jsonb_build_object('host', COALESCE(
                NULLIF(unprocessed_data -> 'metadata' ->> 'hostTaxonId', ''),
                NULLIF(unprocessed_data -> 'metadata' ->> 'hostNameScientific', '')))
         ELSE '{}'::jsonb
       END
)
WHERE unprocessed_data IS NOT NULL
  AND (unprocessed_data -> 'metadata' ? 'hostTaxonId'
       OR unprocessed_data -> 'metadata' ? 'hostNameScientific'
       OR unprocessed_data -> 'metadata' ? 'hostNameCommon')
  AND NOT (unprocessed_data -> 'metadata' ? 'host');
```
</details>

```sql
UPDATE sequence_entries
SET unprocessed_data = jsonb_set(
    unprocessed_data,
    '{metadata}',
    ((unprocessed_data -> 'metadata') - 'hostTaxonId' - 'hostNameScientific' - 'hostNameCommon')
    || CASE
         WHEN COALESCE(
                NULLIF(unprocessed_data -> 'metadata' ->> 'hostTaxonId', ''),
                NULLIF(unprocessed_data -> 'metadata' ->> 'hostNameScientific', '')
              ) IS NOT NULL
         THEN jsonb_build_object('host', COALESCE(
                NULLIF(unprocessed_data -> 'metadata' ->> 'hostTaxonId', ''),
                NULLIF(unprocessed_data -> 'metadata' ->> 'hostNameScientific', '')))
         ELSE '{}'::jsonb
       END
)
WHERE unprocessed_data IS NOT NULL
  AND (unprocessed_data -> 'metadata' ? 'hostTaxonId'
       OR unprocessed_data -> 'metadata' ? 'hostNameScientific'
       OR unprocessed_data -> 'metadata' ? 'hostNameCommon')
  AND NULLIF(unprocessed_data -> 'metadata' ->> 'host', '') IS NULL;
```

### PR Checklist


anna-parker:
@maverbiest I would do this in three steps:
1. have ingest submit new sequences with host and not hostTaxonId and hostScientificName (modify hash code correctly so that the hash doesnt change)
2. run a db surgery with an sql command to modify the data in the code (this shouldnt be in db migrations IMO as this is code for data that not every loculus user will have)
3. remove prepro code that handles the INSDC special case.

maverbiest:
@anna-parker okay cool, thanks! I will get rid of the DB migration then and we can run it as a surgery. Regarding you point 1., could you clarify what you mean by 'modify the hash'? I thought the aim was to ensure the hash doesn't change so we don't trigger revisions? What I'm currently doing on this branch is collapsing the taxon id and scientific name fields into a single `host` field after the hash is computed, but could be I misunderstood

anna-parker:
> 'modify the hash'? I thought the aim was to ensure the hash doesn't change so we don't trigger revisions? 

yes I was unclear, I meant to modify the code for the hash calculation so that it doesnt change

ingest/scripts/prepare_metadata.py
```diff
@@ -151,6 +170,11 @@ def main(
 
         record["hash"] = hashlib.md5(prehash.encode(), usedforsecurity=False).hexdigest()
 
+        # for segmented organisms, this has to happen in `heuristic_group_segments.py`
```
anna-parker, line 173: Im not sure this is true? I think we can just modify host here after we calculate the hash

anna-parker: yeah Im pretty sure the grouping code doesnt need to be touched - we will continue to ingest taxonId and scientificName and these should be the same across segments. Only when calculating the hash do we need to take this into consideration

anna-parker: also (sorry another thing), would be nice to see the prepro changes required after this in a follow up PR already so we can test the full process on staging :-)

maverbiest: Alright I made some changes! I was a bit confused because now we have some test tsv files that we check against where there's still `hostTaxonId` and `hostNameScientifc`. But I guess in a real run those sequences would go through `prepare_metadata.py` first before being grouped so they'd get the host name resolved, right? But maybe I should read the ingest flow a bit more deeply

maverbiest: Hmm but now if I remove this check for `config.segmented` in prepare_metadat.py, `test_ingest` fails. Do I need to update hashes in the test files? Will look into this!
```python
        if not config.segmented:
            record = resolve_host_information(record)
```

maverbiest: Okay I don't see how to do this without also touching code in the grouping files. The `submissionId` <-> `id` change also happens in `prepare_metadata`, `heuristic_group_segments`, and `override_group_segments`. But maybe we can discuss offline!

anna-parker: uggh yeah you are correct we still need it there as we calculate the hash again, its very ugly that this is split up between files - I will think about maybe refactoring this for clarity.

maverbiest: I put things back to the way they were before. But agreed it's a not so nice to have to do this three times in different files

ingest/scripts/prepare_metadata.py
```diff
@@ -36,6 +36,25 @@ class Config:
...
+    and hostNameCommon fields on the record will be removed.
+
+    This should be done after computing the hash for a record to not trigger
+    revisions for all INSDC data
+    """
+    host = record.get("hostTaxonId") or record.get("hostNameScientific")
+    record.pop("hostTaxonId", None)
+    record.pop("hostNameScientific", None)
+    record.pop("hostNameCommon", None)
+    if host:
```
anna-parker: maybe just set this to None as a default

anna-parker (approved):
this has now been tested on staging and looks safe to merge- lets maybe mark it as breaking (its technically not breaking but just to remind people to run the db surgery with this)

PR #6567, opened 2026-06-01
fix(deployment): Increase TTL for ingest trigger jobs a lot but add values hash to ens…

theosanderson:
Currently we are getting issues with jobs appearing out of sync, and probably constantly being recreated, because the job is deleted after 10 mins then argo tries to recreate it

### DIFF
--- a/kubernetes/loculus/templates/ingest.yaml
+++ b/kubernetes/loculus/templates/ingest.yaml
@@ -124,10 +124,11 @@ metadata:
   name: loculus-ingest-trigger-{{ $key }}
   annotations:
     argocd.argoproj.io/sync-wave: "10"
+    valuesHash: {{ .Values | toJson | sha256sum | quote }}
 spec:
   activeDeadlineSeconds: 120
   backoffLimit: 0
-  ttlSecondsAfterFinished: 600
+  ttlSecondsAfterFinished: 900000
   template:
     metadata:
       labels:


theosanderson:
@theosanderson-agent - add the replace true thing

theosanderson-agent:
Done — added `argocd.argoproj.io/sync-options: Replace=true` to the ingest trigger Job.

Since the long `ttlSecondsAfterFinished` means the finished Job lingers between syncs, and Job specs are largely immutable, ArgoCD can't patch the existing Job in place when the `valuesHash` annotation changes. `Replace=true` makes ArgoCD delete and recreate the Job instead, so it re-triggers reliably on sync.

`sync` is your branch (not created by the agent), so I pushed to a new branch and opened #6569 targeting `sync` — merge it in or cherry-pick `070c90372`.

kubernetes/loculus/templates/ingest.yaml
```diff
@@ -124,10 +124,11 @@ metadata:
   name: loculus-ingest-trigger-{{ $key }}
   annotations:
     argocd.argoproj.io/sync-wave: "10"
+    valuesHash: {{ .Values | toJson | sha256sum | quote }}
 spec:
   activeDeadlineSeconds: 120
   backoffLimit: 0
-  ttlSecondsAfterFinished: 600
+  ttlSecondsAfterFinished: 900000
```
anna-parker, line 131: lets maybe just make this 24hours? 86400

PR #6597, opened 2026-06-02
feat(backend): scope released-data ETag to organism and pipeline version [with trigger]

theosanderson:
## Summary

The released-data ETag is currently derived from a single `last_time_updated` timestamp per table in `table_update_tracker`. Any write to `sequence_entries_preprocessed_data` — e.g. a reprocessing run for a single organism, or a background run of a not-yet-current pipeline version — bumps that one shared timestamp and therefore invalidates the `get-released-data` ETag for **every** organism. With continuous preprocessing this means the SILO import for organism A is needlessly re-triggered by processing activity on organism B (and by processing for pipeline versions that aren't even released yet).

This PR implements the trigger-based, per-organism + per-pipeline-version tracker discussed in the etag thread (the "second version"), so that preprocessed-data writes only invalidate the ETag of the affected organism and pipeline version.

## Approach

`table_update_tracker` gains two dimensions, `organism` and `pipeline_version`, with the primary key becoming `(table_name, organism, pipeline_version)`.

- **`sequence_entries_preprocessed_data`** gets organism/pipeline-aware per-statement triggers. The organism is resolved via the `(accession, version)` foreign key into `sequence_entries`; `pipeline_version` is a native column of the preprocessed-data table. Rows are grouped so each affected `(organism, pipeline_version)` pair gets one upsert per statement.
- **All other tables** keep writing table-wide rows, now using the `''` (organism) and `0` (pipeline_version) sentinels. The generic `update_table_tracker()` function was updated so its `ON CONFLICT` target matches the new composite key.
- **The ETag query** unions the sentinel rows with the organism/pipeline-specific rows: `organism IN ('', :organism) AND pipeline_version IN (0, :currentPipelineVersion)`, then takes the max `last_time_updated`. This preserves correctness (table-wide writes such as submissions, data-use-terms changes, external metadata and pipeline switches still always count) while removing the cross-organism and cross-pipeline-version churn from the dominant source of updates: preprocessing.

`getReleasedData` now passes the organism and its current processing pipeline version into `getLastDatabaseWriteETag`. The argument-less `getLastDatabaseWriteETag()` used by `extract-unprocessed-data` is unchanged in behaviour (it still takes the max over all rows regardless of organism/pipeline).

### Notes on edge cases
- **Data use terms / external metadata** have no version (and DUT has no organism), so they keep writing table-wide sentinel rows and continue to invalidate the ETag as before — this addresses the concern raised in the thread about the DUT table.
- **Cascading deletes**: deleting a `sequence_entries` row cascades to `sequence_entries_preprocessed_data`. By the time the preprocessed-data `AFTER DELETE` statement trigger fires, the parent rows are already gone, so the join finds nothing — but `sequence_entries`' own table-wide trigger has already bumped the sentinel row, so the ETag still changes.
- A transition table (`REFERENCING NEW/OLD TABLE`) cannot be attached to a trigger covering more than one event, so INSERT, UPDATE and DELETE each get their own trigger.

## Tests
- `GIVEN data processed for a non-current pipeline version THEN etag is unchanged until it becomes current` — verifies pipeline-version scoping.
- `GIVEN reprocessing of another organism THEN this organism's released-data etag is unchanged` — verifies organism scoping.
- Existing ETag / released-data / extract / submit-processed / revise / pipeline-switch tests still pass.

## ⚠️ Schema docs
`backend/docs/db/schema.sql` is **not** updated in this PR — it is regenerated by the `Generate and Validate Flyway Schema` workflow. Please add the **`update_db_schema`** label so CI commits the regenerated schema dump.

## PR Checklist
- [x] Any manual testing that has been done is documented (ran the backend test suites for released-data, extract, submit-processed, revise, and pipeline-switch locally against Dockerised Postgres).


backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
```diff
@@ -90,13 +91,27 @@ open class ReleasedDataModel(
...
-            }
-        }
+    open fun getLastDatabaseWriteETag(
+        tableNames: List<String>? = null,
+        organism: Organism? = null,
+        pipelineVersion: Long? = null,
+    ): String {
+        val query = UpdateTrackerTable.select(UpdateTrackerTable.lastTimeUpdatedDbColumn)
+        tableNames?.let { query.andWhere { UpdateTrackerTable.tableNameColumn inList it } }
+        organism?.let { query.andWhere { UpdateTrackerTable.organismColumn inList listOf("", it.name) } }
```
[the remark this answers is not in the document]
theosanderson: `NULL` would be much clearer than `0` and ``

backend/docs/db/schema.sql
```diff
@@ -50,6 +50,29 @@ CREATE FUNCTION public.jsonb_concat(a jsonb, b jsonb) RETURNS jsonb
...
+
+CREATE FUNCTION public.update_preprocessed_data_tracker() RETURNS trigger
+    LANGUAGE plpgsql
+    AS $$
+BEGIN
+    INSERT INTO table_update_tracker (table_name, organism, pipeline_version, last_time_updated)
+    SELECT TG_TABLE_NAME, se.organism, cr.pipeline_version, timezone('UTC', CURRENT_TIMESTAMP)
+    FROM changed_rows cr
+    JOIN sequence_entries se
+      ON se.accession = cr.accession AND se.version = cr.version
```
anna-parker, line 111: Thoughts: 
 - Do we want to restrict this to only sequences are released, i.e. where se.released_at IS NOT NULL?
 - lets check we have an index on this join
- I dont quite understand the meaning of REFERENCING NEW TABLE and REFERENCING OLD TABLE -> means that this is executed/triggered once per sql statement not per row, has to be different for deletes

backend/docs/db/schema.sql
```diff
@@ -853,24 +878,38 @@ CREATE TRIGGER update_tracker_trigger AFTER INSERT OR DELETE OR UPDATE OR TRUNCA
...
+--
+
+CREATE TRIGGER update_tracker_trigger_ins AFTER INSERT ON public.sequence_entries_preprocessed_data REFERENCING NEW TABLE AS changed_rows FOR EACH STATEMENT EXECUTE FUNCTION public.update_preprocessed_data_tracker();
+
+
+--
+-- Name: sequence_entries_preprocessed_data update_tracker_trigger_upd; Type: TRIGGER; Schema: public; Owner: postgres
+--
+
+CREATE TRIGGER update_tracker_trigger_upd AFTER UPDATE ON public.sequence_entries_preprocessed_data REFERENCING NEW TABLE AS changed_rows FOR EACH STATEMENT EXECUTE FUNCTION public.update_preprocessed_data_tracker();
```
anna-parker, line 1021: these can be merged? 
```suggestion
CREATE TRIGGER update_tracker_trigger_upd AFTER UPDATE OR INSERT ON public.sequence_entries_preprocessed_data REFERENCING NEW TABLE AS changed_rows FOR EACH STATEMENT EXECUTE FUNCTION public.update_preprocessed_data_tracker();
```

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
```diff
@@ -314,7 +314,9 @@ open class SubmissionController(
         ) @RequestHeader(value = HttpHeaders.IF_NONE_MATCH, required = false) ifNoneMatch: String?,
     ): ResponseEntity<StreamingResponseBody> {
         val lastDatabaseWriteETag = releasedDataModel.getLastDatabaseWriteETag(
-            RELEASED_DATA_RELATED_TABLES,
+            tableNames = RELEASED_DATA_RELATED_TABLES,
```
anna-parker, line 317: RELEASED_DATA_RELATED_TABLES includes the aux tables and I believe they can be removed

backend/docs/db/schema.sql
anna-parker: ah also the `data_use_terms_table` and `EXTERNAL_METADATA_TABLE_NAME` lets maybe do a join there to add the organism?

backend/docs/db/schema.sql
anna-parker: ah postgres doesnt allow this - apologies its because of the specific REFERENCING

backend/docs/db/schema.sql
theosanderson: Yeah I did think about this but thought it would make less difference to our practical situation atm so didn't bother, but it's a fair point, do feel free to push anything if motivated to!

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
```diff
@@ -314,7 +314,9 @@ open class SubmissionController(
...
+            organism = organism,
+            pipelineVersion = submissionDatabaseService.getCurrentProcessingPipelineVersion(organism),
```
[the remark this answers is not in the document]
anna-parker: true we can get the pipelineversion in the `getLastDatabaseWriteETag` function

backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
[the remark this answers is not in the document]
anna-parker: ah that would be nice but we need not NULL values for a primary key...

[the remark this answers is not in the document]
theosanderson: Claude says:

```
Drop the PK, replace it with a unique constraint that treats NULLs as equal:
  ALTER TABLE table_update_tracker
    ADD CONSTRAINT table_update_tracker_uq
    UNIQUE NULLS NOT DISTINCT (table_name, organism, pipeline_version);
  Now table-wide rows store real NULLs, ON CONFLICT (table_name, organism, pipeline_version) still dedupes them, and the ETag query reads naturally:
  (organism IS NULL OR organism = :organism)
  AND (pipeline_version IS NULL OR pipeline_version = :version)
  or with IS NOT DISTINCT FROM. This is the cleanest — you get exactly the semantics you wanted ("no organism / no version applies") with no sentinels, and it directly
  answers Anna: you keep uniqueness, you just don't call it a PK. By default a plain UNIQUE treats NULLs as distinct, which would silently break the dedup — NULLS NOT
  DISTINCT is the piece that makes it work.
```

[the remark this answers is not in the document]
anna-parker: ah nice!

anna-parker (approved):
it would be good to merge in https://github.com/loculus-project/loculus/pull/6616 as I tested this PR with that additional change then this should be safe to merge I think!

backend/docs/db/schema.sql
anna-parker: done in the merged PR :-)

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
anna-parker: done :-)

backend/src/test/kotlin/org/loculus/backend/controller/submission/GetReleasedDataEndpointTest.kt
```diff
@@ -229,6 +230,60 @@ class GetReleasedDataEndpointTest(
...
+                PreparedProcessedData.successfullyProcessedOtherOrganismData(
+                    accession = it.accession,
+                    version = it.version,
+                )
+            },
+            organism = OTHER_ORGANISM,
+            pipelineVersion = 2,
+        )
+
+        submissionControllerClient.getReleasedData(organism = DEFAULT_ORGANISM, ifNoneMatch = initialEtag)
```
anna-parker, line 319: here we should still check that the etag for default organism changes

backend/src/test/kotlin/org/loculus/backend/controller/submission/GetReleasedDataEndpointTest.kt
```diff
@@ -229,6 +230,60 @@ class GetReleasedDataEndpointTest(
...
+        // Once version 2 becomes the current pipeline, the released data (and therefore the etag) changes.
+        submissionDatabaseService.useNewerProcessingPipelineIfPossible()
+
+        submissionControllerClient.getReleasedData(ifNoneMatch = initialEtag)
+            .andExpect(status().isOk)
+            .andExpect(header().string(ETAG, greaterThan(initialEtag)))
+    }
+
+    @Test
+    fun `GIVEN reprocessing of another organism THEN this organism's released-data etag is unchanged`() {
```
anna-parker, line 297: ah I guess we should actually add a test that submitting new data to default organism doesnt bump the etag for other organism only for default organism

anna-parker: I added these tests here: https://github.com/loculus-project/loculus/pull/6622 - not blocking but I guess good to only merge this with the tests

backend/src/test/kotlin/org/loculus/backend/controller/submission/GetReleasedDataEndpointTest.kt
anna-parker: done

backend/src/test/kotlin/org/loculus/backend/controller/submission/GetReleasedDataEndpointTest.kt
anna-parker: done

PR #6611, opened 2026-06-03
fix(website): center svg icon on review card

### LINKED ISSUE #6612: Note icon on processed data on review page is misaligned
Image

This is the HTML

```html
<span class="text-base">
    <span class="text-red-600 truncate max-w-xs inline-block"
        data-tooltip-id="text-tooltip-DateLOC_001JADU.1">2025-07-02</span>
    <svg viewBox="0 0 24 24" width="1.2em" height="1.2em" class="text-yellow-500 inline-block"
        data-tooltip-id="note-tooltip-DateLOC_001JADU.1">
        <path fill="currentColor"
            d="M17.75 ..."></path>
    </svg>
</span>
```

corneliusroemer:
Resolves #6612 svg icon not centered.

Before:

iTerm2 2026-06-03 11 30 54

After:

Google Chrome 2026-06-03 11 36 06

Also fixes that we add the literal css class `undefined` when `extraStyle` is not passed to `KeyValueComponent` (and hence `undefined`)  on the review page.

🚀 Preview: https://fix-undefined-style.loculus.org

### DIFF
--- a/website/src/components/ReviewPage/ReviewCard.tsx
+++ b/website/src/components/ReviewPage/ReviewCard.tsx
@@ -460,7 +460,7 @@ const KeyValueComponent: FC<KeyValueComponentProps> = ({
     return (
         <div className={`flex flex-col m-2 `}>
             <span className={keyStyle ?? 'text-gray-500 uppercase text-xs'}>{keyName}</span>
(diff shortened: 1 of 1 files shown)


theosanderson (approved):
Thanks

PR #6628, opened 2026-06-07
fix(website): fix tooltip on edit page

### LINKED ISSUE #6627: Descriptions ToolTip on revise and submit edit page is broken (no longer displaying when hovering over the info button)

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/6627

From claude:
I found the root cause. It's a bug in react-tooltip v6's resolveDataTooltipAnchor function:

⏺ There's the bug. Line 259 checks currentElement instanceof HTMLElement — but SVG elements are SVGElement, not
  HTMLElement. So when data-tooltip-id is on the <MaterialSymbolsInfoOutline> (which renders as an <svg>), the library
  walks up the DOM tree and skips the SVG element entirely, never finding the anchor.
  The fix is to put data-tooltip-id on a wrapping <span> instead of on the SVG icon directly — exactly the pattern HoverTooltip already uses.

### PR Checklist

🚀 Preview: https://tooltip.loculus.org

### DIFF
--- a/integration-tests/tests/specs/features/single-submit.spec.ts
+++ b/integration-tests/tests/specs/features/single-submit.spec.ts
@@ -1,3 +1,4 @@
+import { expect } from '@playwright/test';
 import { test } from '../../fixtures/group.fixture';
 import { SingleSequenceSubmissionPage } from '../../pages/submission.page';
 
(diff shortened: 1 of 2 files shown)


theosanderson (approved):
Thanks!

PR #6633, opened 2026-06-08
feat(ingest, backend): improve ingest curation warning - add diff

### LINKED ISSUE #3084: Ingest should not treat curator revisions as latest and revise

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/3084

Adds the option to filter by AccessionVersion on the `get-unprocessed-metadata` endpoint. Have ingest call this endpoint to get all previous metadata (includes sequence hashes, so it will indirectly tell us if a sequence has been changed) when we get into the case where ingest would like to revise a curated sequence. 

Ingest will then add a diff of the curated metadata and the metadata it would like to revise to the notification so that curators can decide better to to handle the situation.

### PR Checklist
- [x] All necessary documentation has been adapted for backend endpoints
- [x] The backend endpoint updates are covered by unit tests
1. Ingest sequences from the INSDC with initial hash
2. Curate a sequence from the INSDC as a superuser (add a revision, do not change hash)
3. Ingest ignores revision as the hash has not changed
4. Log into database and manually modify the hash of the curated entry (unprocessed data not original data)
5. Run the ingest cronjob and verify that an informative message is sent to the loculus notifications slack:
image

🚀 Preview: https://curation-warning-add-diff.loculus.org

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
@@ -418,6 +418,10 @@ open class SubmissionController(
             description = "The metadata fields that should be returned. If not provided, all fields are returned.",
         ) @RequestParam(required = false) fields: List<String>?,
         @Parameter(
+            description = "Filter by accession versions in 'accession.version' format. " +
+                "If not provided, all accession versions are considered.",
+        ) @RequestParam(required = false) accessionVersionsFilter: List<String>?,
+        @Parameter(
             description = "Filter by group ids. If not provided, all groups are considered.",
         ) @RequestParam(required = false) groupIdsFilter: List<Int>?,
         @Parameter(
@@ -432,11 +436,22 @@ open class SubmissionController(
             headers.add(HttpHeaders.CONTENT_ENCODING, compression.compressionName)
         }
 
+        val parsedAccessionVersions = accessionVersionsFilter?.takeIf { it.isNotEmpty() }?.map {
+            val lastDot = it.lastIndexOf('.')
+            if (lastDot <= 0 || lastDot == it.length - 1) {
+                throw BadRequestException("Invalid accession version format '$it', expected 'accession.version'")
+            }
+            val version = it.substring(lastDot + 1).toLongOrNull()
+                ?: throw BadRequestException("Invalid version in accession version '$it', expected a number")
+            AccessionVersion(it.substring(0, lastDot), version)
+        }
+
         val totalRecords = submissionDatabaseService.countUnprocessedMetadata(
             authenticatedUser,
             organism,
             groupIdsFilter?.takeIf { it.isNotEmpty() },
             statusesFilter?.takeIf { it.isNotEmpty() },
+            parsedAccessionVersions,
         )
         headers.add(X_TOTAL_RECORDS, totalRecords.toString())
         // TODO(https://github.com/loculus-project/loculus/issues/2778)
@@ -452,6 +467,7 @@ open class SubmissionController(
                 groupIdsFilter?.takeIf { it.isNotEmpty() },
                 statusesFilter?.takeIf { it.isNotEmpty() },
                 fields?.takeIf { it.isNotEmpty() },
+                parsedAccessionVersions,
             )
         }
 
--- a/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
@@ -437,13 +437,7 @@ open class SubmissionController(
         }
 
         val parsedAccessionVersions = accessionVersionsFilter?.takeIf { it.isNotEmpty() }?.map {
-            val lastDot = it.lastIndexOf('.')
-            if (lastDot <= 0 || lastDot == it.length - 1) {
-                throw BadRequestException("Invalid accession version format '$it', expected 'accession.version'")
-            }
-            val version = it.substring(lastDot + 1).toLongOrNull()
-                ?: throw BadRequestException("Invalid version in accession version '$it', expected a number")
-            AccessionVersion(it.substring(0, lastDot), version)
+            AccessionVersion.fromString(it)
         }
 
         val totalRecords = submissionDatabaseService.countUnprocessedMetadata(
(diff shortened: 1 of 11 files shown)


ingest/scripts/loculus_client.py
```diff
@@ -447,33 +447,34 @@ def get_sequence_status(config: Config):
...
+        params["fields"] = fields
+    if accessionVersionsFilter:
+        params["accessionVersionsFilter"] = accessionVersionsFilter
 
     while True:
         logger.info("Getting previously submitted sequences")
 
         response = make_request(HTTPMethod.GET, url, config, params=params)
+        logger.info(f"Params: {params}")
+        logger.info(f"Response: {response}")
```
anna-parker:
```suggestion
```

ingest/scripts/compare_hashes.py
```diff
@@ -114,12 +170,15 @@ def process_hashes(
...
+        metadata_diff = calculate_metadata_diff(
+            update_manager.config, new_metadata, previously_submitted_entry
+        )
         # Sequence has been curated before - special case
         notification = (
             f"Ingest: Sequence {corresponding_loculus_accession} with INSDC "
             f"accession {ingested_insdc_accession} has been curated before "
             f"- do not know how to proceed. New hash: {newly_ingested_hash}, "
-            f"old hash: {previously_submitted_entry.hash}."
+            f"old hash: {previously_submitted_entry.hash}. Metadata diff: {metadata_diff}"
```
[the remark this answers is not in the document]
anna-parker: this can be ignored for now

ingest/scripts/compare_hashes.py
```diff
@@ -86,16 +95,63 @@ def sample_out_hashed_records(
     return not keep
 
 
+def calculate_metadata_diff(
+    config: Config, new_metadata: dict[str, Any], previous_entry: LatestLoculusVersion
+) -> dict[str, Any]:
+    previous_metadata_list = get_submitted(
+        config,  # type: ignore
```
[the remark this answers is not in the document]
anna-parker: I now import the config so it doesnt have to be ignored

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
```diff
@@ -432,11 +436,22 @@ open class SubmissionController(
             headers.add(HttpHeaders.CONTENT_ENCODING, compression.compressionName)
         }
 
+        val parsedAccessionVersions = accessionVersionsFilter?.takeIf { it.isNotEmpty() }?.map {
```
corneliusroemer, line 439: Surprised we don't have a reusable parser for this already - apparently we don't, but makes sense to make one and centralize

anna-parker: was also surprised

corneliusroemer: I've got a PR to centralize coming

ingest/scripts/compare_hashes.py
```diff
@@ -20,10 +22,7 @@
 
 
 @dataclass
-class Config:
-    organism: str
-    segmented: str
-    nucleotide_sequences: list[str]
+class CompareHashesConfig(Config):
```
corneliusroemer: Why don't we add slack_hook to config so we can reuse the full config from `loculus_client` and don't need to subtype?

ingest/scripts/compare_hashes.py
```diff
@@ -86,16 +85,61 @@ def sample_out_hashed_records(
...
+            diff[key] = {"new": new_metadata[key], "old": None}
+
+        else:
+            new_value = new_metadata[key]
+            old_value = previous_metadata[key]
+
+            if new_value != old_value:
+                diff[key] = {"new": new_value, "old": old_value}
+
+    return diff
```
corneliusroemer: More succinct:

```suggestion
    return {
        key: {
            "old": previous_metadata.get(key),
            "new": new_metadata.get(key),
        }
        for key in new_metadata.keys() | previous_metadata.keys()
        if new_metadata.get(key) != previous_metadata.get(key)
    }
```

ingest/scripts/compare_hashes.py
```diff
@@ -302,9 +349,12 @@ def main(
...
         full_config = yaml.safe_load(file)
         relevant_config = {
-            key: full_config[key] for key in Config.__annotations__ if key in full_config
+            key: full_config[key]
+            for key in CompareHashesConfig.__annotations__
+            if key in full_config
         }
-        config = Config(**relevant_config)
+        relevant_config = {f.name: full_config.get(f.name, []) for f in dataclasses.fields(CompareHashesConfig)}
+        config = CompareHashesConfig(**relevant_config)
```
corneliusroemer: This has gotten a lot more complicated and I'm not sure it's the same behavior as previously.

Do you remember why we do this subsetting of config? Why not just reuse the original config?

ingest/scripts/compare_hashes.py
```diff
@@ -302,9 +349,12 @@ def main(
     with open(config_file, encoding="utf-8") as file:
         full_config = yaml.safe_load(file)
         relevant_config = {
-            key: full_config[key] for key in Config.__annotations__ if key in full_config
+            key: full_config[key]
+            for key in CompareHashesConfig.__annotations__
+            if key in full_config
         }
```
corneliusroemer: Delete this

ingest/scripts/compare_hashes.py
anna-parker: discussed that we subset for consistency and that there was an unnecessary duplication

ingest/scripts/compare_hashes.py
anna-parker: done

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
anna-parker: thank you!

corneliusroemer (approved):

Issue #6636, opened 2026-06-09
XLSX template with dropdown choices for fields


theosanderson:
We could have much more complicated XLSX templates that enforce dropdown options for fields. This gets a bit complicated because we don't want all the fields present by default. We need this if we add more regulated fields.

PR #6637, opened 2026-06-09
feat(website): XLSX submission template with Data + Guidance sheets and dropdown enforcement

theosanderson:
Implements #6636 — a richer XLSX submission template with enforced dropdown choices for controlled-vocabulary fields.

## What it does

The XLSX template (`/[organism]/submission/template?fileType=xlsx`) is now a three-sheet workbook:

| Sheet | Visible | Protected | Purpose |
|---|---|---|---|
| **`Data`** | yes | no | What the submitter fills in. |
| **`Guidance`** | yes | read-only | Human-readable reference for every field. |
| **`_lists`** | hidden | read-only | Lookup source backing the dropdowns. |

### Data sheet
- **All input fields** as columns, **priority (default/`metadataTemplate`) fields first, then the rest**. Headers stay machine field names so the file round-trips through upload.
- **Enforced dropdowns** on every field with a controlled vocabulary (`options`): a value outside the list is **rejected** (`errorStyle=stop`). Free-text columns get no validation, so they accept anything.
- Dropdowns are **per-column and header-driven**: each validation reads its own column header and resolves that field's options in `_lists` (`MATCH`/`OFFSET`), so the dropdown follows the field if the column is renamed or moved.
- **Header hover notes** carrying each field's definition, guidance and example.
- **Colour-coded headers** by tier — required / priority / optional — with a matching legend on `Guidance`.
- **Frozen header row**; date-typed columns formatted `yyyy-mm-dd` (what upload expects); columns sized to the field name.

### Guidance sheet (read-only)
One row per field: Field name, Display name, Required, Definition, Guidance, Example, Allowed values — plus the colour legend.

## Other changes
- **Upload coupling:** the parser reads the first sheet (`Data`); its multi-sheet warning now ignores the `Guidance`/`_lists` reference sheets, so a freshly-downloaded template uploads without a spurious warning.
- **Library:** generation uses **ExcelJS** (`@lokalise/xlsx`/SheetJS-CE cannot write data validations, hidden sheets or sheet protection); upload parsing is unchanged. **TSV output is unchanged.**
- **Upload help text** now leads with the XLSX template (recommended) and labels the TSV one minimal.

## Previews
- loculus: https://excel-val.loculus.org
- Pathoplexus context (real organisms/vocabularies): https://preview-excel.pathoplexus.org (pathoplexus/pathoplexus#1028)

🚀 Preview: https://excel-val.loculus.org

### DIFF
--- a/website/src/components/Submission/FileUpload/fileProcessing.spec.ts
+++ b/website/src/components/Submission/FileUpload/fileProcessing.spec.ts
@@ -96,8 +96,8 @@ describe('fileProcessing', () => {
         expect(processedFile.fastaHeader()).toBe('fooid description');
     });
 
-    test('template reference sheets (Config, _lists) do not trigger a multi-sheet warning', async () => {
-        const file = await buildWorkbookFile(['Config', '_lists']);
+    test('template reference sheets (Guidance, _lists) do not trigger a multi-sheet warning', async () => {
+        const file = await buildWorkbookFile(['Guidance', '_lists']);
         const processingResult = await METADATA_FILE_KIND.processRawFile(file);
 
         expect(processingResult.isOk()).toBe(true);
@@ -105,7 +105,7 @@ describe('fileProcessing', () => {
     });
 
     test('an unexpected extra sheet still triggers a multi-sheet warning', async () => {
-        const file = await buildWorkbookFile(['Config', 'My other data']);
+        const file = await buildWorkbookFile(['Guidance', 'My other data']);
         const processingResult = await METADATA_FILE_KIND.processRawFile(file);
 
         expect(processingResult.isOk()).toBe(true);
--- a/website/src/components/Submission/FileUpload/fileProcessing.spec.ts
+++ b/website/src/components/Submission/FileUpload/fileProcessing.spec.ts
@@ -1,9 +1,22 @@
 import { fail } from 'assert';
 import { promises as fs } from 'fs';
 
+import ExcelJS from 'exceljs';
 import { describe, expect, test } from 'vitest';
 
 import { METADATA_FILE_KIND, PLAIN_SEGMENT_KIND } from './fileProcessing';
+
+async function buildWorkbookFile(extraSheetNames: string[]): Promise<File> {
+    const workbook = new ExcelJS.Workbook();
+    const dataSheet = workbook.addWorksheet('Data');
+    dataSheet.addRow(['submissionId', 'country']);
+    dataSheet.addRow(['sample1', 'Germany']);
+    for (const sheetName of extraSheetNames) {
+        workbook.addWorksheet(sheetName).addRow(['ignored']);
+    }
+    const buffer = await workbook.xlsx.writeBuffer();
+    return new File([buffer], 'template.xlsx');
+}
 
 async function loadTestFile(fileName: string): Promise<File> {
     const path = `${import.meta.dirname}/test_files/${fileName}`;
@@ -81,5 +94,21 @@ describe('fileProcessing', () => {
         const processedText = await processedFile.text();
         expect(processedText).toBe('ACTGACTGACTG');
         expect(processedFile.fastaHeader()).toBe('fooid description');
+    });
+
+    test('template reference sheets (Config, _lists) do not trigger a multi-sheet warning', async () => {
+        const file = await buildWorkbookFile(['Config', '_lists']);
+        const processingResult = await METADATA_FILE_KIND.processRawFile(file);
+
+        expect(processingResult.isOk()).toBe(true);
+        expect(processingResult._unsafeUnwrap().warnings()).toHaveLength(0);
+    });
+
+    test('an unexpected extra sheet still triggers a multi-sheet warning', async () => {
+        const file = await buildWorkbookFile(['Config', 'My other data']);
+        const processingResult = await METADATA_FILE_KIND.processRawFile(file);
+
+        expect(processingResult.isOk()).toBe(true);
+        expect(processingResult._unsafeUnwrap().warnings()).toHaveLength(1);
     });
 });
--- a/website/src/components/Submission/FileUpload/fileProcessing.spec.ts
+++ b/website/src/components/Submission/FileUpload/fileProcessing.spec.ts
@@ -111,4 +111,24 @@ describe('fileProcessing', () => {
         expect(processingResult.isOk()).toBe(true);
         expect(processingResult._unsafeUnwrap().warnings()).toHaveLength(1);
     });
+
+    test('parses the Data sheet by name even when it is not the first sheet', async () => {
+        // Reference sheet dragged in front of Data — the Data sheet must still be the one parsed.
+        const workbook = new ExcelJS.Workbook();
+        workbook.addWorksheet('Guidance').addRow(['Field name', 'Display name']);
+        const data = workbook.addWorksheet('Data');
+        data.addRow(['submissionId', 'country']);
+        data.addRow(['sample1', 'Germany']);
+        const buffer = await workbook.xlsx.writeBuffer();
+        const file = new File([buffer], 'template.xlsx');
+
+        const processingResult = await METADATA_FILE_KIND.processRawFile(file);
+        expect(processingResult.isOk()).toBe(true);
+        const processedFile = processingResult._unsafeUnwrap();
+
+        expect(processedFile.warnings()).toHaveLength(0);
+        const text = await processedFile.text();
+        expect(text.split('\n')[0]).toContain('submissionId'); // parsed Data...
+        expect(text).not.toContain('Field name'); // ...not the Guidance reference sheet
+    });
 });
--- a/website/src/components/Submission/FileUpload/fileProcessing.ts
+++ b/website/src/components/Submission/FileUpload/fileProcessing.ts
@@ -19,6 +19,11 @@ export type FileKind<F extends ProcessedFile> = {
 
 const COMPRESSION_EXTENSIONS = ['zst', 'gz', 'zip', 'xz'];
 
+// Sheets that the downloadable XLSX template adds alongside the `Data` sheet (see
+// `pages/[organism]/submission/template/index.ts`). Kept in sync manually rather than imported, to
+// avoid pulling the server-only template endpoint (and its dependencies) into the client bundle.
+const TEMPLATE_REFERENCE_SHEET_NAMES = new Set(['Config', '_lists']);
+
 export const METADATA_FILE_KIND: FileKind<ProcessedFile> = {
     type: 'metadata',
     icon: MaterialSymbolsLightDataTableOutline,
@@ -265,7 +270,12 @@ export class ExcelFile implements ProcessedFile {
         // filename needs to end in 'tsv' for the uploaded file
         const tsvFile = new File([tsvBlob], 'converted.tsv', { type: 'text/tab-separated-values' });
         this.tsvFile = tsvFile;
-        if (workbook.SheetNames.length > 1) {
+        // Sheets that the downloadable template adds for reference/lookup purposes are expected and
+        // should not trigger the "you have unprocessed sheets" warning.
+        const unexpectedSheets = workbook.SheetNames.slice(1).filter(
+            (sheetName) => !TEMPLATE_REFERENCE_SHEET_NAMES.has(sheetName),
+        );
+        if (unexpectedSheets.length > 0) {
             this.processingWarnings.push(
                 `The file contains ${workbook.SheetNames.length} sheets, only the first sheet (${firstSheetName}; ${rowCount} rows) was processed.`,
             );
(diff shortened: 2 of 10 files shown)


theosanderson:
@codex review

website/src/components/Submission/FileUpload/fileProcessing.ts
```diff
@@ -241,8 +248,13 @@ export class ExcelFile implements ProcessedFile {
...
         });
 
-        const firstSheetName = workbook.SheetNames[0];
-        let sheet = workbook.Sheets[firstSheetName];
+        // Parse the `Data` sheet by name when present, so reordering the template's sheets (e.g.
+        // dragging `Guidance` to the front) does not cause the reference sheet to be read as
+        // metadata. Fall back to the first sheet for arbitrary, non-template workbooks.
+        const dataSheetName = workbook.SheetNames.includes(TEMPLATE_DATA_SHEET_NAME)
+            ? TEMPLATE_DATA_SHEET_NAME
+            : workbook.SheetNames[0];
```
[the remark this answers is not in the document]
theosanderson: Acceptable behaviour for the moment I think

website/src/pages/[organism]/submission/template/index.ts
Comment (anna-parker) [inline website/src/pages/[organism]/submission/template/index.ts:1]: I realized that we have a download link in the docs that links to a tsv file with definitions, it would be great to replace it with this

Image

Comment (anna-parker) [reply website/src/pages/[organism]/submission/template/index.ts:1]: Nice just confirmed I can download this excel file, add data and submit it and it is accepted and data is sent correctly!

Comment (theosanderson) [reply website/src/pages/[organism]/submission/template/index.ts:1]: Made #6644 for this - makes a lot of sense but not blocking I think

website/src/components/Submission/FileUpload/fileProcessing.spec.ts
```diff
@@ -1,10 +1,23 @@
...
 import { promises as fs } from 'fs';
 
+import ExcelJS from 'exceljs';
 import { describe, expect, test } from 'vitest';
 
 import { METADATA_FILE_KIND, PLAIN_SEGMENT_KIND } from './fileProcessing';
 
+async function buildWorkbookFile(extraSheetNames: string[]): Promise<File> {
+    const workbook = new ExcelJS.Workbook();
+    const dataSheet = workbook.addWorksheet('Data');
```
anna-parker, line 11: These tests assume that DATA_SHEET_NAME= 'Data' and TEMPLATE_REFERENCE_SHEET_NAMES = (Guidance, _lists), we should use the constants instead

website/src/config.ts
```diff
@@ -166,6 +167,52 @@ export function getMetadataTemplateFields(
...
+ * submission-detail fields first, then the default-enabled "template" fields
+ * (`schema.metadataTemplate`, or all input fields if unset), then the remaining opt-in fields.
+ * Fields enabled by default are tagged with `isTemplateField: true`.
+ */
+export function getOrderedTemplateInputFields(organism: string, action: 'submit' | 'revise'): TemplateInputField[] {
+    const schema = getConfig(organism).schema;
+
+    const submissionIdInputFields = getSubmissionIdInputFields(schema);
+    const accessionFields = action === 'revise' ? [getAccessionInputField()] : [];
+    const detailFields = [...accessionFields, ...submissionIdInputFields];
```
anna-parker: I might rename this as requiredInternalFields to be clearer when reading

anna-parker: I found this a bit hard to read, I would actually split up the case where the metadataTemplate is set and where it is not set more to make it easier to understand what is happening

```
export function getOrderedTemplateInputFields(organism: string, action: 'submit' | 'revise'): TemplateInputField[] {
    const schema = getConfig(organism).schema;

    const submissionIdInputFields = getSubmissionIdInputFields(schema);
    const accessionFields = action === 'revise' ? [getAccessionInputField()] : [];
    const requiredInternalFields = [...accessionFields, ...submissionIdInputFields];

    const requiredInternalFieldsNames = new Set(requiredInternalFields.map((field) => field.name));
    const nonInternalInputFields = schema.inputFields.filter((field) => !requiredInternalFieldsNames.has(field.name));

    const metadataTypeByName = new Map(schema.metadata.map((entry) => [entry.name, entry.type] as const));
    const decorate = (field: InputField, isTemplateField: boolean): TemplateInputField => ({
        ...field,
        isTemplateField,
        metadataType: metadataTypeByName.get(field.name),
    });

    if (schema.metadataTemplate === undefined || schema.metadataTemplate === null) {
        return [
            ...requiredInternalFields.map((field) => decorate(field, true)),
            ...nonInternalInputFields.map((field) => decorate(field, true)),
        ];
    }

    const fieldsByName = new Map(nonInternalInputFields.map((field) => [field.name, field]));

    const templateFieldNameSet = new Set(schema.metadataTemplate);
    const templateFields = schema.metadataTemplate
        .map((name) => fieldsByName.get(name))
        .filter((field): field is InputField => field !== undefined);

    const restFields = nonInternalInputFields.filter((field) => !templateFieldNameSet.has(field.name));

    return [
        ...requiredInternalFields.map((field) => decorate(field, true)),
        ...templateFields.map((field) => decorate(field, true)),
        ...restFields.map((field) => decorate(field, false)),
    ];
}
```

website/src/pages/[organism]/submission/template/index.ts
```diff
@@ -38,27 +48,216 @@ export const GET: APIRoute = ({ params, request }) => {
...
-        return new TextEncoder().encode(content).buffer;
+function createTsvTemplate(organism: string, action: UploadAction): ArrayBuffer {
+    const columnNames = Array.from(getMetadataTemplateFields(organism, action).keys());
+    const content = columnNames.join('\t') + '\n';
+    return new TextEncoder().encode(content).buffer;
+}
+
+/**
+ * Builds a workbook with three sheets:
+ *  - `Data`: the sheet submitters fill in. Columns are the machine field names (so the file
```
Comment (anna-parker) [inline website/src/pages/[organism]/submission/template/index.ts]: again we have the constant names here if we later refactor it will be hard to change - lets update the docs to use the constants

website/src/pages/[organism]/submission/template/index.ts
```diff
@@ -38,27 +48,216 @@ export const GET: APIRoute = ({ params, request }) => {
...
     return new Response(fileBuffer, {
         headers,
     });
 };
 
-function createTemplateFile(fileType: TemplateFileType, columnNames: string[]): ArrayBuffer {
-    if (fileType === 'tsv') {
-        const content = columnNames.join('\t') + '\n';
-        return new TextEncoder().encode(content).buffer;
+function createTsvTemplate(organism: string, action: UploadAction): ArrayBuffer {
```
Comment (anna-parker) [inline website/src/pages/[organism]/submission/template/index.ts:59]: I guess it might be a bit confusing that the tsv file only has a subset of all the fields shown in the excel file

website/src/pages/[organism]/submission/template/index.ts
```diff
@@ -38,27 +48,216 @@ export const GET: APIRoute = ({ params, request }) => {
...
+ * has options. The validation reads its column's own header and resolves that field's option list in
+ * `_lists`, so the dropdown follows the field when the column is renamed or moved.
+ */
+function addOptionsLookup(
+    workbook: ExcelJS.Workbook,
+    dataSheet: ExcelJS.Worksheet,
+    fields: TemplateInputField[],
+    optionFields: TemplateInputField[],
+): void {
+    const listsSheet = workbook.addWorksheet(LISTS_SHEET_NAME, { state: 'hidden' });
```
Comment (anna-parker) [inline website/src/pages/[organism]/submission/template/index.ts:127]: How does this work if we have multiple fields with options? wont it produce a new sheet each time?

Comment (anna-parker) [reply website/src/pages/[organism]/submission/template/index.ts:127]: ah its ok we only call this function once

website/src/pages/[organism]/submission/template/index.ts
```diff
@@ -38,27 +48,216 @@ export const GET: APIRoute = ({ params, request }) => {
...
+    // Empty-string password: protection is cosmetic (guard against accidental edits), not a security boundary.
+    void listsSheet.protect('', { selectLockedCells: false, selectUnlockedCells: false });
+
+    // `dataValidations.add` exists at runtime but is missing from ExcelJS' type definitions. Using it
+    // (rather than per-cell `cell.dataValidation`) writes one compact range rule per option field
+    // instead of materialising a validation object for every cell.
+    const dataValidations = (dataSheet as unknown as { dataValidations: WorksheetDataValidations }).dataValidations;
+    const sheet = `'${LISTS_SHEET_NAME}'`;
+    optionFields.forEach((field) => {
+        const dataColumn = columnLetter(fields.indexOf(field) + 1);
```
Comment (anna-parker) [inline website/src/pages/[organism]/submission/template/index.ts:143]: hmm... so this code it quite complex and involves creating subfunctions for converting indices to excel column names... chatGPT suggested using
`const matchOffset = `MATCH(INDIRECT(ADDRESS(1,COLUMN())),'_lists'!$1:$1,0)-1;` - all this requires it that the columns in both sheets have the same name. Not sure if this becomes even more magical tho

website/src/config.ts
```diff
@@ -166,6 +167,52 @@ export function getMetadataTemplateFields(
...
+ */
+export type TemplateInputField = InputField & { isTemplateField: boolean; metadataType?: MetadataType };
+
+/**
+ * Returns every submittable input field for the template download, in column order:
+ * submission-detail fields first, then the default-enabled "template" fields
+ * (`schema.metadataTemplate`, or all input fields if unset), then the remaining opt-in fields.
+ * Fields enabled by default are tagged with `isTemplateField: true`.
+ */
+export function getOrderedTemplateInputFields(organism: string, action: 'submit' | 'revise'): TemplateInputField[] {
```
anna-parker, line 184: looking at the examples it might be nice to have required fields first and then desired other fields in the templates (more like in the docs)

website/src/pages/[organism]/submission/template/index.ts
```diff
@@ -38,27 +48,216 @@ export const GET: APIRoute = ({ params, request }) => {
...
+    // --- Data sheet ---
+    const dataSheet = workbook.addWorksheet(DATA_SHEET_NAME);
+    dataSheet.addRow(fields.map((field) => field.name));
+    dataSheet.views = [{ state: 'frozen', ySplit: 1 }]; // keep the header row visible while scrolling
+
+    const headerRow = dataSheet.getRow(1);
+    fields.forEach((field, index) => {
+        dataSheet.getColumn(index + 1).width = columnWidthFor(field);
+        if (field.metadataType === 'date') {
+            dataSheet.getColumn(index + 1).numFmt = 'yyyy-mm-dd';
```
Comment (anna-parker) [inline website/src/pages/[organism]/submission/template/index.ts:96]: I am able to add non valid dates e.g. 2023-bb into the excel sheet and there is no error

Comment (anna-parker) [reply website/src/pages/[organism]/submission/template/index.ts:96]: ah never mind our input date is free text... and might have to stay free text to support date ranges?

website/src/components/Submission/FileUpload/fileProcessing.spec.ts
theosanderson: The current implementation (even though Claude decided it) is my preference for how tests should work. IMO tests should independently specify behaviour. Otherwise if we e.g. accidentally had "Data" and "Guidance" swapped in a PR the test would still pass. Our tests always say things like `click('Submit')` not `click(SUBMIT_LABEL)`, rightly IMO

website/src/pages/[organism]/submission/template/index.ts
Comment (theosanderson) [reply website/src/pages/[organism]/submission/template/index.ts]: (happy to change this one!)

website/src/pages/[organism]/submission/template/index.ts
Comment (theosanderson) [reply website/src/pages/[organism]/submission/template/index.ts:96]: Yep, known behaviour. Given we accept date ranges in various formats, we need to either have complex logic (which we can add later) or accept anything.

website/src/config.ts
theosanderson: We can already change this by reordering the template. I'd keep the flexibility to choose how we do it - because I think there's a good case for e.g. keeping all the `geoLoc` stuff together.

website/src/pages/[organism]/submission/template/index.ts
Comment (anna-parker) [reply website/src/pages/[organism]/submission/template/index.ts:96]: tested this works as intended on a date field (unaligned organism), I am able to upload data and dates are reformated and accepted correctly:
image

website/src/pages/[organism]/submission/template/index.ts
Comment (theosanderson) [reply website/src/pages/[organism]/submission/template/index.ts:143]: I initially wanted something that would apply to the whole sheet, not per-column, and investigated that a lot. But this wasn't possible to achieve whilst having dropdowns and validation (because it would have broken free-text fields). I'm not sure about your exact proposal (feel free to preview).

anna-parker (approved):
some small comments but overall looks great - amazing feature!

website/src/pages/[organism]/submission/template/index.ts
Comment (theosanderson) [reply website/src/pages/[organism]/submission/template/index.ts:143]: (will resolve this for now but we can revisit)

website/src/config.ts
theosanderson: addressed

website/src/pages/[organism]/submission/template/index.ts
Comment (theosanderson) [reply website/src/pages/[organism]/submission/template/index.ts]: addressed

PR #6641, opened 2026-06-10
feat(config): fix host name scientific search for CCHF

anna-parker:
resolves https://loculus.slack.com/archives/C05G172HL6L/p1781079401799899

### Screenshot
The CCHF config overwrote hostScientificName - assuming it would be in the ingest metadata, this is no longer true. Additionally, the config still passes hostScientificName and hostTaxonId as fallback values but these no longer will exist in standard metadata submissions and can thus be removed

### PR Checklist

🚀 Preview: https://fix-host.loculus.org


maverbiest (approved):
Thanks!

PR #6658, opened 2026-06-11
feat!(kubernetes): add fine grained pod scheduling configurations with nodeSelector, tolerations, and affinity 

florianzwagemaker:
As briefly discussed on Slack.
This PR adds the `podScheduling` root-level configuration property in the helm chart with three new child properties for kubernetes pod scheduling control;
- `nodeSelector` (object): maps the Loculus pod deployment directly to a Kubernetes node with a defined label for placement
- `tolerations` (array): allows pods to bypass specific node taints
- `affinity` (object): allows for the addition of fine grained affinity rules such as nodeAffinity or podAffinity as well as conditional scheduling.

Additionally moved the `podPriorityValue` key from a root property to be grouped underneath the `podScheduling` key.

### Example

This assumes a cluster is already created either in kubernetes or locally through k3s.
1. get the node name with `kubectl get nodes`
2. label the node where you wish to deploy the loculus pods with `kubectl label nodes my-nodepool workload-type=my-custom-label`
3. taint the node to ensure only the pods with toleration rules are allowed on this particular node with `kubectl taint nodes my-nodepool workload-type=my-custom-label:NoSchedule`
4. Define your node pool configuration in the helm chart like the following: 
```yaml
podScheduling:
  nodeSelector: {}
  tolerations:
    - key: "workload-type"
      operator: "Equal"
      value: "my-custom-label"
      effect: "NoSchedule"
  affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
          - matchExpressions:
              - key: "workload-type"
                operator: "In"
                values:
                  - "my-custom-label"
```
when deploying, all Loculus pods will now schedule on the node named `my-nodepool` because it matches the assigned label `my-custom-label` in the affinity rules. The tolerations rule states that the Loculus pods *are* allowed on this label while any other pod without this tolerations rule won't be allowed on this node during scheduling.

If a more direct mapping is desired then this can be achieved the yaml example below,  but this will not allow for more fine-grained control which is probably desired in production-like kubernetes clusters
```yaml
podScheduling:
  nodeSelector:
    workload-type: my-custom-label
```

### Breaking changes
Need to move `podPriorityValue` if set, to be under `podScheduling`


kubernetes/loculus/values.yaml
theosanderson, line 1: thanks a lot for this. I think it would be great to group these under a common key, `podScheduling` say

theosanderson: possibly `podPriorityClass` should live there too

florianzwagemaker: That seems logical yeah.
Grouped these properties now to be under the `podScheduling` key and moved the `podPriorityValue` key to this group as well (which then makes the podPriorityClass)
the rest is unchanged, ignored the copilot suggestions above for now under the (potentially incorrect) assumption that these configurations will only be changed by users that already have a fair amount of experience with kubernetes deployments.

theosanderson:
@codex review

kubernetes/loculus/values.yaml
theosanderson: I'm not really sure why CoPilot decided to get involved here, it doesn't normally :D

kubernetes/loculus/templates/ena-submission-deployment.yaml
```diff
@@ -22,6 +22,7 @@ spec:
         component: loculus-ena-submission
     spec:
       {{- include "possiblePriorityClassName" . | nindent 6 }}
+      {{- include "loculus.podScheduling" . | nindent 6 }}
```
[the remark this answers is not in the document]
theosanderson: (would be great to address)

theosanderson:
@claude review

florianzwagemaker:
Included the podScheduling block to the last two pod specs as suggested, all pods should now schedule correctly.
Let me know if any other changes are wanted or necessary :)

theosanderson (approved):
LGTM and CI passes in https://github.com/loculus-project/loculus/pull/6665, thanks

PR #6661, opened 2026-06-12
fix(website): stop collection date range fields bouncing between values while typing

theosanderson:
Relates to https://loculus.slack.com/archives/C05G172HL6L/p1781275386940369 - hard to reproduce the original issue to test but the new approach seems strictly better, and less code (apart from added tests)

Below written by AI

____________
## Problem

The **Collection date** From/To inputs (and any other field configured with `rangeOverlapSearch`) could intermittently **bounce back and forth between two values while typing** — a race condition reported during normal use of the search page.

## Root cause

In the default Loculus config, `collectionDate` uses `rangeOverlapSearch` (`kubernetes/loculus/values.yaml`), so the "Collection date / From / To" box is rendered by **`DateRangeField`**.

`DateRangeField` kept its **own** `lowerValue`/`upperValue` state and synced it **bidirectionally** with the global `fieldValues` via two `useEffect`s:

- **write effect:** local state → `setSomeFieldValues` → URL/query state → re-derived `fieldValues`
- **read effect:** `fieldValues` → `setLowerValue`/`setUpperValue`

`fieldValues` is recomputed from URL state on every change and **lags one round-trip behind** while typing (each keystroke goes through `setSomeFieldValues` + `setPage(1)` + a parent `useMemo`). When you type a second character before the first round-trip lands, the read effect fires with a **stale** `fieldValues` and resets the local state backward; the write effect then pushes that stale value back out — so the field ping-pongs between the freshly typed value and the lagging one. This is the classic two-effect mirror race.

This is a longstanding bug, not a regression from a recent change.

## Fix

Make `fieldValues` the **single source of truth**:

- Remove the `lowerValue`/`upperValue` local state and **both** mirroring effects.
- Derive the displayed values directly from `fieldValues`.
- Write only on **real user interaction** via a `commit(...)` helper invoked from the date inputs' `onChange` and the strict-mode toggle — never from an effect, so there is no feedback loop.
- Keep a single effect that re-derives `strictMode` from the populated fields (guarded so toggling strictness with no dates entered doesn't snap back). It is a pure function of which fields exist and cannot oscillate against the writes.

No behavioural change to how dates map onto the four underlying `Lower*/Upper*` fields, or to strict/non-strict mode switching.

## Tests

- Added a test asserting **no write happens on mount** (rendering must not push values back into query state).
- Rewrote the "updates query params" test to use a stateful wrapper that feeds values back through `fieldValues` (the old version passed only because the buggy local mirror accumulated edits without any feedback) and assert **both** edits are retained — i.e. neither value bounces back.
- All `DateRangeField` (12) and `DateField` (4) tests pass.

🚀 Preview: https://fix-date-range-field-boun.loculus.org

### DIFF
--- a/website/src/components/SearchPage/fields/DateRangeField.spec.tsx
+++ b/website/src/components/SearchPage/fields/DateRangeField.spec.tsx
@@ -101,7 +101,7 @@ describe('DateRangeField', () => {
         }
     });
 
-    it('derives switches mode correctly', async () => {
+    it('does not write to query params on mount', () => {
         render(
             <DateRangeField
                 field={field}
@@ -110,11 +110,19 @@ describe('DateRangeField', () => {
             />,
         );
 
-        expect(setSomeFieldValues).toHaveBeenCalledWith(
-            ['collectionDateRangeLowerFrom', '2024-01-01'],
-            ['collectionDateRangeUpperTo', '2024-12-31'],
-            ['collectionDateRangeUpperFrom', null],
-            ['collectionDateRangeLowerTo', null],
+        // Rendering must not push values back into the query state — only real user interaction
+        // (typing a date or toggling strictness) should. This avoids the feedback loop that made
+        // the date inputs bounce between two values while typing.
+        expect(setSomeFieldValues).not.toHaveBeenCalled();
+    });
+
+    it('derives switches mode correctly', async () => {
+        render(
+            <DateRangeField
+                field={field}
+                fieldValues={{ collectionDateRangeLowerFrom: '2024-01-01', collectionDateRangeUpperTo: '2024-12-31' }}
+                setSomeFieldValues={setSomeFieldValues}
+            />,
         );
 
         const strictCheckbox = screen.getByRole('checkbox');
@@ -131,13 +139,34 @@ describe('DateRangeField', () => {
     });
 
     it('updates query params if user types in new dates', async () => {
-        render(
-            <DateRangeField
-                field={field}
-                fieldValues={{ collectionDateRangeLowerFrom: '2024-01-01', collectionDateRangeUpperTo: '2024-12-31' }}
-                setSomeFieldValues={setSomeFieldValues}
-            />,
-        );
+        const lastValues: { current: FieldValues } = { current: {} };
+
+        function Wrapper() {
+            const [values, _setValues] = useState<FieldValues>({
+                collectionDateRangeLowerFrom: '2024-01-01',
+                collectionDateRangeUpperTo: '2024-12-31',
+            });
+            lastValues.current = values;
+
+            const setValues: SetSomeFieldValues = useCallback((...fieldValuesToSet) => {
+                _setValues((state) => {
+                    const newState = { ...state };
+                    fieldValuesToSet.forEach(([k, v]) => {
+                        // mirror the production behaviour of useSearchPageState: null/'' deletes
+                        if (v === null || v === '') {
+                            delete newState[k];
+                        } else {
+                            newState[k] = v;
+                        }
+                    });
+                    return newState;
+                });
+            }, []);
+
+            return <DateRangeField field={field} fieldValues={values} setSomeFieldValues={setValues} />;
+        }
+
+        render(<Wrapper />);
 
         const fromInput = screen.getByText('From').closest('div')?.querySelector('input');
         const toInput = screen.getByText('To').closest('div')?.querySelector('input');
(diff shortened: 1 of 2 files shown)


website/src/components/SearchPage/fields/DateRangeField.spec.tsx
```diff
@@ -131,13 +138,34 @@ describe('DateRangeField', () => {
...
-        render(
-            <DateRangeField
-                field={field}
-                fieldValues={{ collectionDateRangeLowerFrom: '2024-01-01', collectionDateRangeUpperTo: '2024-12-31' }}
-                setSomeFieldValues={setSomeFieldValues}
-            />,
-        );
+        const lastValues: { current: FieldValues } = { current: {} };
+
+        function Wrapper() {
```
anna-parker: the code looks good, I also did some testing of the preview. However, I have trouble understanding this test - why have you created this wrapper here?

anna-parker: ah I see... claude explained that `The simpler tests (like "renders the component" on line 55) use setSomeFieldValues = vi.fn(). That works fine when you only need to assert what setSomeFieldValues was called with. But vi.fn() doesn't actually update anything — so DateRangeField never receives new fieldValues props after the user types.`

Maybe this would still be better as an integration test though?
```
test('date range field retains typed values', async ({ page }) => {
      const searchPage = new SearchPage(page);
      await searchPage.ebolaSudan();

      const fromInput = page.getByLabel('From').first();
      const toInput = page.getByLabel('To').first();
  
      await fromInput.fill('1987-04-23');
      await toInput.fill('2014-10-13');

      await expect(fromInput).toHaveValue('1987-04-23');
      await expect(toInput).toHaveValue('2014-10-13');
  });
```

anna-parker (approved):
approving as this seems like an overall improvement and testing can be improved without further reviews

website/src/components/SearchPage/fields/DateRangeField.spec.tsx
theosanderson: @theosanderson-agent see if you can implement this (or comment if you disagree) [you can push to this PR]

theosanderson-agent: I dug into the integration-test route against the live preview, and I think a Playwright test is the wrong tool for *this specific* bug — but I took the readability feedback to heart instead. Pushed 3fa7fc081.

**Why the suggested integration test does not work as written.** The From/To inputs are rsuite `DatePicker`s (a segmented masked `<input>`), and I confirmed all three of these against `fix-date-range-field-boun.loculus.org`:
- `fromInput.fill("1987-04-23")` is a no-op — the input keeps its `yyyy-MM-dd` placeholder. rsuite ignores a directly-set value; it only reacts to per-keystroke parsing.
- Typing the date (`pressSequentially`) is flaky: typing `1987-04-23` yields `0003-02-dd`, and even digits-only `19870423` yields `yyyy-09-23`. The masked caret/auto-advance behaviour does not survive Playwright keystrokes reliably.
- `page.getByLabel("To")` is ambiguous — it matches **6** elements (several combobox floating-labels contain "To"), so `.first()` grabs the wrong field.

The reliable interaction in this repo is clicking a calendar day (see the existing `active-filters.dependent.spec.ts` date test), which I got working. **But** that does not catch this bug: I ran the same calendar-click test against both `main.loculus.org` (buggy) and the preview (fixed) and it **passes on both**. The bounce is a fast-typing race where `fieldValues` lags one round-trip behind; calendar `oneTap` fires a single change and closes, so there is no race to reproduce. (The original author also noted the bug is hard to reproduce even manually.)

So the unit test is the right level here — it reproduces the exact feedback loop deterministically.

**What I did change:** your underlying question was "why the wrapper?". I replaced the three duplicated inline `Wrapper`s with one documented `StatefulDateRangeField` helper. The doc comment explains the key point: a `vi.fn()` mock can not catch this bug, because the field renders whatever is in its `fieldValues` prop — a value can only "bounce" if edits actually feed back into that prop, which a mock never does. The helper mimics production `useSearchPageState.setSomeFieldValues` (null/"" deletes the key). All 12 DateRangeField tests still pass; types/format clean.

Happy to also add a general calendar-based integration test for broad coverage of the date-range field if you want it, but I would not label it a regression test for the bounce since it does not catch it.

theosanderson: @theosanderson-agent - I think probably the feedback is that we don't necessarily need a regression test for the specific issue - so lose this, and any unnecessary comments. If we lack any coverage of the date picker then we could add a test that uses the date picker.

theosanderson-agent: Done in a5a7478fa. Dropped the dedicated bounce regression test and the explanatory comments:

- Removed the `StatefulDateRangeField` harness and the typing-bounce test.
- The spec is now back to the `main` version, with the **only** change the fix actually forces: the field no longer writes to query params on mount, so the stale on-mount `toHaveBeenCalledWith` assertion in `derives switches mode correctly` is gone. (The two pre-existing stateful-wrapper tests for the strict-mode toggle and clearing are untouched.)

On date-picker coverage: we already have an integration test that uses the picker — `active-filters.dependent.spec.ts` → "date range filter can be removed by clicking the X" clicks a calendar day and asserts the resulting filter. So we are not lacking coverage there. Happy to extend it (e.g. assert the From value after selecting) if you want more, but per earlier findings, *typing* into the rsuite masked input via Playwright is flaky — the reliable interaction is clicking a calendar day, which that test already does.

10 DateRangeField unit tests pass; types/format clean.

PR #6678, opened 2026-06-15
feat(backend, config): S3 garbage collection without race condition

### LINKED ISSUE #3935: Build a garbage collector

anna-parker:
builds on https://github.com/loculus-project/loculus/pull/6543 with modifications to prevent a race condition and additional code improvements, tests were modified to use existing helper functions used elsewhere in the codebase 

resolves https://github.com/loculus-project/loculus/issues/3935
resolves https://github.com/loculus-project/loculus/pull/6543#discussion_r3414887935

## Config
Requires https://github.com/loculus-project/loculus/pull/6705 to be merged before GC can be enabled on instances with multiple backend replicas, the default config parameters will be applied without user changes:

```
s3:
  enabled: true
  gcInitialDelayMinutes: 15 # initial delay before first garbage collection task runs
  gcEnabled: false # have garbage collection log instead of actually delete files
  gcFrequencyMinutes: 1440 # frequency of garbage collection polling in minutes
  gcOrphanRetentionPeriodMinutes: 10080. # 7 days
```

## Garbage collection

This PR introduces a scheduled task in the backend that finds and deletes orphaned files on S3 older than `s3.gcOrphanRetentionPeriodMinutes`.

Orphan files are defined as files that have been uploaded to S3 and are not referenced by any sequence submission in the database. Concretely, orphan files are identified by comparing file ids from the `files` table to `files` referenced by `sequence_entries.submitted_data` and `sequence_entries_preprocessed_data.processed_data` and archived data (to do this the JSONB is parsed by the task)

## The Exact Race Window

  The current code has two separate steps with no lock between them:

```
  GC:         getOrphanedFileIds() ──────────────────── deleteFile(F) + deleteFileEntry(F)
                    ↑                                          ↑
  User:             │  validateFileIdsExist(F) → submit() ────┤
                    │  (passes, F still in DB)                │
                    └──────────────────────────────── F deleted, but submission committed
```

  validateFileIdsExist checks the DB, then the submission writes to sequence_entries later. Between those two events, GC can delete the file. The window is roughly: time to delete all the S3 objects + DB rows in the batch.

Additionally, if multiple GC tasks are run simultaneously one could mark a file for deletion while a user is referencing it in a submission and the other GC task could immedietly delete it (very small window) which is resolved with preventing parallel task runs.

This is resolved by having the task run in 2 steps, in the first files are marked for deletion, files marked for deletion cannot be referenced by a submission and in the second actually deleting files marked for deltion that have since not been referenced by a submission.

## Alternatives considered
- Have one transaction for finding the files to delete and deleting them, this transaction would need to apply table locks/isolation to prevent the file validation function (called when submitting) to read a file as existing while we are deleting it (getting into the same issue as above). We additionally tested that just reading all the jsonb in ppx takes 3min, this transaction time will only grow once we add actual files. Locking tables while doing this transaction is not feasible.

## Manual testing (by @maverbiest)
For testing, I made the GC run every 5 minutes and set `gcOrphanRetentionPeriodMinutes` to 0, so everything that is not referenced by a sequence is considered an orphan.

### PUTting files to S3

After logging in on the website as testuser, making a group with group id 2, and getting an auth token for the testuser:

```sh
RESP=$(curl -s -X POST "$BACKEND/files/request-upload?groupId=$GROUP_ID&numberFiles=1" -H "Authorization: Bearer $TOKEN")
FILE_ID=$(echo "$RESP" | jq -r '.[0].fileId')
URL=$(echo "$RESP" | jq -r '.[0].url')
echo "orphan test content" > /tmp/orphan.txt
curl -X PUT --upload-file /tmp/orphan.txt "$URL"
```

### Orphan files are deleted from S3

Directly after I upload a file to S3, I can stat it in the preview deployment:
```sh
➜  loculus git:(s3-garbage-collection) kubectl exec -n "$NS" "$POD" -- mc stat "local/loculus-preview-private/files/c49bd8da-6661-4063-8718-2d6703a6c8f9"
Name      : c49bd8da-6661-4063-8718-2d6703a6c8f9
Date      : 2026-06-09 12:17:01 UTC
Size      : 20 B
ETag      : 415ab09be58b32f2bd922650fa08e3d4
Type      : file
Metadata  :
  Content-Type: binary/octet-stream
```

5 minutes later, after I see that the GC task has run, the file is no longer there:

```sh
➜  loculus git:(s3-garbage-collection) kubectl exec -n "$NS" "$POD" -- mc stat "local/loculus-preview-private/files/c49bd8da-6661-4063-8718-2d6703a6c8f9"
mc: <ERROR> Unable to stat `local/loculus-preview-private/files/c49bd8da-6661-4063-8718-2d6703a6c8f9`. Object does not exist.
command terminated with exit code 1
```

### Orphan files are deleted from the DB

Similarly, I can follow the lifecycle of an orphan file in the database though pgAdmin by portforwarding to the database pod in the preview deployment:

```sh
kubectl port-forward loculus-database-659b795cbd-7f65x --namespace prev-s3-garbage-collection 5432:5432
```

Screenshot 2026-06-10 at 10 49 58

Screenshot 2026-06-10 at 10 50 24

Screenshot 2026-06-10 at 10 50 36

### Files attached to submissions are protected from GC

I also uploaded a test file for `Test organism (with files)` through the website. After submission, I monitored the backend logs to wait until I see a GC run. There are no logs of anything being deleted, and I can also still see the file in the DB:

grafik

I can also still `stat` the file on S3, confirming it was protected from garbage collection both in the database and on S3:

```sh
➜  loculus git:(s3-garbage-collection) kubectl exec -n "$NS" minio-6c565c64bf-wdks6 -- mc stat "local/loculus-preview-private/files/d2b5debc-dca9-4279-8c5d-9bd81bd046ba"
Name      : d2b5debc-dca9-4279-8c5d-9bd81bd046ba
Date      : 2026-06-10 09:20:36 UTC
Size      : 25 B
ETag      : 671ee2e3e61346d33e76305c0862bdc8-1
Type      : file
Metadata  :
  Content-Type       : binary/octet-stream
  X-Amz-Tagging-Count: 1
```

### PR Checklist

🚀 Preview: https://gc-anya.loculus.org


backend/docs/db/schema.sql
theosanderson, line 1: an alternative would be just to delete the entry before we delete the S3 file? worst case scenario one would get truly orphaned files in a crash, but one could add a job to find those if it became a real problem, which seems unlikely

theosanderson: (but no strong feelings, I don't know the code well)

anna-parker: Im not sure I understand how we would accomplish your suggestion as the db tables dont have a link from the s3 file to the table entries - we have to read all jsonb to find the ones linked so I dont think your suggestion is possible... but please correct me if I misunderstand!

theosanderson: Ah, sorry, I think I misunderstood your diagram! Thanks!

maverbiest:
Tanks a lot for this! Definitely better to guard against this :)

Do I understand correctly that in the current setup, the scan of all se and sepd is done twice per GC task? I'm wondering if it's not possible to modify the query to return file_id, upload_requested_at, and marked_for_deletion_at at the same time, and then figure out in kotlin what files need to be deleted and which to be marked for deletion. That way you'd only have to do the heavy SQL work once. Or is there something I'm missing that would make that impractical?

anna-parker:
@maverbiest I think I covered this option in `alternatives considered` in the PR description but let me know if you think I missed sth or it is unclear :-)

maverbiest:
@anna-parker hmm if I understand the description of the alternative correctly that is about having one query that both finds and deletes. That's not what I was thinking but I probably didn't explain very well; what I meant is to have one query to find unreferenced files, but instead of it just returning file_id, also have it return upload_requested_at, and marked_for_deletion_at. Something like:

```kotlin
  val unreferenced = filesDatabaseService.getUnreferencedFiles()
  val toDelete = unreferenced.filter { it.markedForDeletionAt != null }.map { it.id }
  val toMark   = unreferenced.filter { it.markedForDeletionAt == null && it.uploadRequestedAt < threshold }.map { it.id }
```

Then run another query to delete what files in `toDelete`, and then another query to mark the files in `toMark`. The first expensive query is still just a SELECT so doesn't lock the DB, the other two queries go by indexed columns so would be fast & not lock the DB for long.

anna-parker:
ah yes that would work! its a bit more custom code I guess but doable!

maverbiest:
I was wondering if the plan moving forward is to get this PR ready to merge, or to incorporate these improvements into https://github.com/loculus-project/loculus/pull/6543? 

I'm good either way, but if we're focusing on this PR I'll probably close my original one to avoid confusion & double work :)

backend/src/main/resources/db/migration/V1.33__update_files_table.sql
maverbiest, line 1: Maybe we could rename this file since we now also add a column in addition to an index

backend/src/main/kotlin/org/loculus/backend/service/files/FilesDatabaseService.kt
```diff
@@ -57,6 +69,83 @@ class FilesDatabaseService(private val dateProvider: DateProvider) {
...
+        if (fileIds.isEmpty()) return
+        val now = dateProvider.getCurrentDateTime()
+        fileIds.processInDatabaseSafeChunks { chunk ->
+            FilesTable.update({ FilesTable.idColumn inList chunk }) {
+                it[markedForDeletionAtColumn] = now
+            }
+        }
+    }
+
+    fun getMarkedForDeletionFileIds(fileIds: Set<FileId>): Set<FileId> = fileIds.chunkedForDatabase({ chunk ->
```
maverbiest: Not blocking but could we add a short docstring here explaining that the function filters a set of input FileIds (similar to `getUncheckedFileIds` below)? Upon first reading the function name I expected it to get the FileIds of all files marked for deletion in the DB

backend/src/main/kotlin/org/loculus/backend/service/submission/S3GarbageCollectionTask.kt
```diff
@@ -0,0 +1,115 @@
...
+
+        val threshold = dateProvider.getCurrentInstant()
+            .minus(orphanRetentionPeriod, DateTimeUnit.MINUTE, DateProvider.timeZone)
+            .toLocalDateTime(DateProvider.timeZone)
+
+        val orphans = filesDatabaseService.getOrphanedFileIds(threshold)
+        val markedOrphans = orphans.filterValues { it != null }.keys
+        val newOrphans = orphans.filterValues { it == null }.keys
+
+        // Phase 2: delete files that were marked in a previous run and are still unreferenced
```
maverbiest, line 69:
```suggestion
        // Phase 2: delete files that were marked in a previous run
```

Maybe remove this part? There is no way for a file to be 'rescued' once it's been marked for deletion, right? Or is there?

backend/src/main/kotlin/org/loculus/backend/service/submission/S3GarbageCollectionTask.kt
```diff
@@ -0,0 +1,115 @@
...
+            deleteFiles(markedOrphans, orphanRetentionPeriod)
+        }
+
+        // Phase 1: mark newly discovered orphans so they are rejected from submissions until the next run
+        if (newOrphans.isNotEmpty()) {
+            filesDatabaseService.markFilesForDeletion(newOrphans)
+            log.info { "S3 garbage collection task marked ${newOrphans.size} orphan(s) for deletion" }
+        } else {
+            log.info { "S3 garbage collection task identified no new orphan files" }
+        }
```
maverbiest, line 84: One thing I'm wondering here is whether marking files for deletion should be gated by `enabled` as well?

If I understand the current approach correctly, even when we disable the GC files will get marked for deletion which means they will cause submissions to be rejected. Not blocking in this one way or the other but might be worth flagging in a comment at least

backend/src/main/kotlin/org/loculus/backend/service/submission/S3GarbageCollectionTask.kt
```diff
@@ -0,0 +1,115 @@
...
+                    "submission after $orphanRetentionPeriod minutes"
+            }
+            auditLogger.log(
+                "CLEANUP",
+                "S3 garbage collection task deleted ${fileIds.size - deleteFailures} orphan(s) " +
+                    "not referenced by a submission after $orphanRetentionPeriod minutes",
+            )
+            if (deleteFailures > 0) {
+                log.warn {
+                    "S3 garbage collection task unsuccessfully attempted to delete $deleteFailures orphan file(s)"
```
maverbiest, line 117: Maybe we could keep track of `s3DeleteFailures` and `dbDeleteFailures` separately?

kubernetes/loculus/values.yaml
```diff
@@ -66,6 +59,10 @@ fileSharing:
   outputFileUrlType: backend
 s3:
   enabled: true
+  gcInitialDelayMinutes: 15 # initial delay before first garbage collection task runs
+  gcEnabled: false # have garbage collection log instead of actually delete files
+  gcFrequencyMinutes: 1440 # frequency of garbage collection polling in minutes
+  orphanRetentionPeriodMinutes: 7
```
maverbiest:
```suggestion
  orphanRetentionPeriodMinutes: 10080
```

maverbiest: I think I'd prefer this to also carry the `gc` prefix in line with the other gc-related config . We could do `gcOrphanRetentionPeriodMinutes` (or maybe `gcGracePeriodMinutes`)?

backend/src/main/kotlin/org/loculus/backend/service/submission/S3GarbageCollectionTask.kt
```diff
@@ -0,0 +1,115 @@
...
+
+@Component
+@ConditionalOnProperty("loculus.s3.enabled", havingValue = "true")
+class S3GarbageCollectionTask(
+    private val filesDatabaseService: FilesDatabaseService,
+    private val s3Service: S3Service,
+    private val dateProvider: DateProvider,
+    private val auditLogger: AuditLogger,
+    @Value("\${${BackendSpringProperty.S3_ORPHAN_RETENTION_PERIOD_MINUTES}}") private val orphanRetentionPeriod: Int,
+    @Value("\${${BackendSpringProperty.S3_GC_ENABLED}:false}") private val enabled: Boolean = false,
```
maverbiest: I find it confusing that we set the default value twice in one line. It's not obvious to me what would happen if they ever get out of sync. Then there's also a third default value in `application.json` (which actually defaults to `true` for enabled instead of `false`).

Could we keep all defaults in `application.json` so they all live in one place? Also for the initial delay and the frequency?

backend/src/test/kotlin/org/loculus/backend/service/files/S3GarbageCollectionTaskTest.kt
```diff
@@ -0,0 +1,158 @@
...
+    @Autowired val convenienceClient: SubmissionConvenienceClient,
+    @Autowired val dateProvider: DateProvider,
+    @Autowired val auditLogger: AuditLogger,
+) {
+    @MockkBean(relaxed = true)
+    lateinit var s3Service: S3Service
+
+    private var groupId = 0
+
+    private fun createProcessedSubmissionReferencingFile(file: UUID) {
```
maverbiest, line 51: This has a lot of overlap with the second test in `GetOrphanedFileIdsTest.kt`, could this be extracted and reused? Or would that be too much of a hassle since they're slightly different?

backend/src/main/kotlin/org/loculus/backend/service/submission/S3GarbageCollectionTask.kt
anna-parker: yes if it is used in a submission (referenced) then we will not delete so this is an accurate comment I believe

backend/src/main/kotlin/org/loculus/backend/service/submission/S3GarbageCollectionTask.kt
anna-parker: This is intentional! This will show us how GC works without actually deleting files

backend/src/main/kotlin/org/loculus/backend/service/files/FilesDatabaseService.kt
anna-parker: I renamed the function to make this clearer :-)

backend/src/main/kotlin/org/loculus/backend/service/submission/S3GarbageCollectionTask.kt
anna-parker: we do that above so I think that is ok

backend/src/main/kotlin/org/loculus/backend/service/submission/FileMappingPreconditionValidator.kt
```diff
@@ -91,6 +91,12 @@ class FileMappingPreconditionValidator(
                 "The following file IDs do not exist: " + nonExistentFileIds.joinToString(),
             )
         }
+        val markedForDeletion = filesDatabaseService.filterMarkedForDeletionFileIds(fileIds)
+        if (markedForDeletion.isNotEmpty()) {
+            throw UnprocessableEntityException(
+                "The following file IDs are no longer valid: " + markedForDeletion.joinToString(),
```
corneliusroemer: Explain why not valid: unreferenced for too long so garbage collected

backend/src/main/kotlin/org/loculus/backend/service/submission/S3GarbageCollectionTask.kt
```diff
@@ -0,0 +1,122 @@
...
+            "Running S3 garbage collection task to clean up orphan files at least $orphanRetentionPeriod " +
+                "minutes old (enabled = $enabled)"
+        }
+
+        val threshold = dateProvider.getCurrentInstant()
+            .minus(orphanRetentionPeriod, DateTimeUnit.MINUTE, DateProvider.timeZone)
+            .toLocalDateTime(DateProvider.timeZone)
+
+        val orphans = filesDatabaseService.getOrphanedFileIds(threshold)
+        val markedOrphans = orphans.filterValues { it != null }.keys
```
corneliusroemer: Maybe comment that the values here are timestamps of when it was marked

corneliusroemer (approved):
Looks great as discussed offline

backend/src/main/kotlin/org/loculus/backend/service/submission/S3GarbageCollectionTask.kt
anna-parker: yeah the name was confusing I updated!

backend/src/test/kotlin/org/loculus/backend/service/files/S3GarbageCollectionTaskTest.kt
anna-parker: Im going to keep as is just to get this merged but if you would like to improve in a later PR that is always great :-)

kubernetes/loculus/values.yaml
```diff
@@ -2707,7 +2711,7 @@ taxonomyService:
 subdomainSeparator: "-"
 replicas:
   website: 1
-  backend: 1
+  backend: 2
```
```suggestion
  backend: 1
```

backend/src/main/kotlin/org/loculus/backend/service/submission/S3GarbageCollectionTask.kt
```diff
@@ -0,0 +1,122 @@
...
+    private val filesDatabaseService: FilesDatabaseService,
+    private val s3Service: S3Service,
+    private val dateProvider: DateProvider,
+    private val auditLogger: AuditLogger,
+    @Value("\${${BackendSpringProperty.S3_GC_ORPHAN_RETENTION_PERIOD_MINUTES}}") private val orphanRetentionPeriod: Int,
+    @Value("\${${BackendSpringProperty.S3_GC_ENABLED}}") private val enabled: Boolean,
+) {
+
+    /**
+     * Runs once daily by default (with an initial delay of 15 minutes) and deletes S3 objects older than
```
anna-parker:
```suggestion
     * Runs once daily by default (with an initial delay of S3_GC_INITIAL_DELAY_MINUTES minutes) and deletes S3 objects older than
```

backend/src/main/kotlin/org/loculus/backend/service/submission/S3GarbageCollectionTask.kt
```diff
@@ -0,0 +1,122 @@
...
+     * `loculus.s3.gc-gc-orphan-retention-period-minutes` that are not referenced in submitted_data or processed_data.
```
anna-parker:
```suggestion
     * `loculus.s3.gc-gc-orphan-retention-period-minutes` that are not referenced in submitted_data, archive_of_submitted_data or processed_data.
```

PR #6687, opened 2026-06-16
feat!(backend): add CrossRef `writeEnabled` flag, fail crossRef.isActive check if empty strings are passed

### LINKED ISSUE #6643: SeqSet: add mint flag
Currently there is no way to limit the permissions of the crossref user to only query for existing citations and not query and mint. To enable easier testing of crossref features we would like to allow querying on staging without mint - this would require adding a flag in the code and only allowing minting in certain cases.

tombch:
resolves #6643

## BREAKING CHANGES

To enable DOI creation on PPX the `writeEnabled` flag now needs to be set to true (default is false), additionally we can now give staging access to the crossRef credentials and set `writeEnabled: false` for staging.

## Summary

- Adds a `crossref.writeEnabled` flag, defaulting to `false`, which when set to `true` enables Crossref write operations (in this case, generating DOIs).
- Alters the `createSeqSetDOI` function to require the Crossref service is both active and write enabled **before** creating any entries in the database - previously, it would create an entry in the database, and then silently skip posting to Crossref if not active.
- Updates the tests to set Crossref service is active, mocking required functions.
- Fixes Crossref `isActive` value, by requiring non-blank values rather than just non-null. This then flagged an issue in the GitHub CI where empty secrets were previously being passed into the test, and the service would still be marked as active.
- Removes some redundant `.toString()` conversions being flagged when linting the backend.

### Screenshot
image

image

### PR Checklist

🚀 Preview: https://crossref-write-enabled.loculus.org


tombch:
@claude review

kubernetes/loculus/values.yaml
```diff
@@ -13,7 +13,11 @@ seqSets:
   enabled: true
   crossRef:
     DOIPrefix: "placeholder"
-    endpoint: "https://test.crossref.org"
+    endpoint: "dummy"
```
anna-parker: I might revert these changes as we are not talking to crossref dev so even if the crossRef service is active we cant really do anything with it

anna-parker (approved):

anna-parker:
side note: before merging lets prepare a PR for PPX with this config change and adding the secrets to staging

kubernetes/loculus/values.yaml
tombch: Ah so just remove the crossref dummy variables altogether?

anna-parker: we can revert back to `endpoint: "https://test.crossref.org"` and remove the others added here, but no strong feelings

Issue #5829, opened 2026-01-11
Support extra files on edit page


chaoran-chen:
At the moment, during the review phase, the edit page does not support extra files. If an entry is edited, all previously attached files are dropped. This should be fixed: it should be possible to (1) edit metadata / sequences without uploading new files as well as (2) upload new files/exchange files.

tombch:
This has been resolved in #6694 !

PR #6694, opened 2026-06-16
feat(website): Persist previous file uploads when revising submissions

### LINKED ISSUE #5819: Revise single entry reusing existing files
While revising a single entry, it should be possible to reuse files that had been attached to the previous version.

tombch:
resolves #5819 and #5829

### Summary

- Displays existing file uploads in the website when users want to either edit an unreleased sequence entry or revise a released one.
- Adds individual 'discard file' buttons in the form file upload case.
- Adds an 'Add additional files' button in the form file upload case, appearing when a user already has an existing folder of files. This opens the users file browser and allows them to select individual additional files, and merges them into the file upload state. If the user attempts to add additional files which would replace existing ones, a confirmation dialog appears to confirm this.
-  Adds a confirmation dialog to the 'discard all files' button.

#### Additional changes
- Adds website tests for new and pre-existing functionality in the folder upload component.
- Adds an `outline-neutral` button variant.
- Refactors the `FolderUploadComponent` to support providing an existing `formSubmissionId`, and uses this instead of the `dummySubmissionId` if provided. It is a somewhat awkward prop, as its only used in the `inputMode == form` case, but I added this to remove the requirement of 're-keying' the file mapping with the actual form submission id in the edit/revise case. 

### Screenshot

image

### PR Checklist

🚀 Preview: https://single-submission-revise.loculus.org


website/src/components/common/buttonStyles.ts
```diff
@@ -33,11 +33,13 @@ const circleSizeClasses: Record<ButtonSize, string> = {
...
-    primary: 'bg-[var(--color-main)] text-white border-transparent hover:bg-primary-700 no-underline!',
-    ghost: 'bg-transparent border-transparent hover:bg-base-200 no-underline!',
-    outline: 'bg-white border-primary-600 text-primary-600 hover:bg-primary-600 hover:text-white no-underline!',
-    unstyled: '',
+    'neutral': 'bg-base-200 text-base-content border-base-300 hover:bg-base-300 no-underline!',
+    'primary': 'bg-[var(--color-main)] text-white border-transparent hover:bg-primary-700 no-underline!',
+    'ghost': 'bg-transparent border-transparent hover:bg-base-200 no-underline!',
+    'outline': 'bg-white border-primary-600 text-primary-600 hover:bg-primary-600 hover:text-white no-underline!',
+    // eslint-disable-next-line @typescript-eslint/naming-convention
+    'outline-neutral': 'bg-white border-base-300 text-base-content hover:bg-base-200 no-underline!',
```
anna-parker, line 41: does eslint want CamelCase?

website/src/components/Submission/FileUpload/FolderUploadComponent.tsx
anna-parker, line 487: could we also add a discard button to each of these files individually?

anna-parker: answering my own question: I guess this would be bulk upload as it is not the form, so this will be in a later PR

website/src/components/Submission/FileUpload/FolderUploadComponent.tsx
```diff
@@ -385,13 +488,48 @@ export const FolderUploadComponent: FC<FolderUploadComponentProps> = ({
...
 
-            <Button
-                onClick={() => setFileUploadState(undefined)}
-                data-testid={`discard_${fileCategory.name}`}
-                className='text-xs break-words text-gray-700 py-1.5 px-4 border border-gray-300 rounded-md hover:bg-gray-50'
-            >
-                Discard files
-            </Button>
+            <div className={`grid gap-2 w-full ${inputMode === 'form' ? 'grid-cols-2' : 'grid-cols-1'}`}>
+                {inputMode === 'form' && (
```
anna-parker, line 494: Is there a reason to restrict this to forms though?

anna-parker (approved):
looks great! I played around with this a bit and sees to work well, thanks Tom!

website/src/components/common/buttonStyles.ts
tombch: Yeah its part of our eslint config to require camel case, but I think in this case using dashes makes more sense

website/src/components/Submission/FileUpload/FolderUploadComponent.tsx
tombch: No, I guess its mainly the implementation for bulk uploads would potentially be quite different so I thought its best to add in the next PR for bulk uploads

corneliusroemer:
I think this would be nice to test more thoroughly in integration tests, at least the happy path. I'll make an issue

PR #6696, opened 2026-06-16
feat(website): time-scale, stepped group submission graph with markers

### LINKED ISSUE #6701: Group cumulative submission graphs show nothing when only a single release was made
Reported by @rneher here https://loculus.slack.com/archives/C061ZQQM3N1/p1781637766434859

The submission graphs on the group pages are really neat, but they don't quite work when the group has made only a single transmission, like [this group from Canada](https://pathoplexus.org/group/2344). One would probably need to

- show markers on the line for every submission
- special case the one datapoint only case or at least have a minimal x-axis width

I might also be better to plot these lines in a 'stair-case' manner, rather than just linearly connecting counts.

Image

theosanderson-agent:
resolves #6701

## Summary

The "Cumulative submissions over time" chart on group pages (`CumulativeSubmissionsChart`) didn't render usefully in a couple of cases:

- **A group with a single submission showed a blank chart.** The line had hidden points (`pointRadius: 0`) and a single data point draws no line segment, so nothing was visible.
- **The x-axis wasn't proportional to time.** It used a Chart.js *category* scale, so submission dates were spaced evenly regardless of the real gaps between them — a one-day gap and a six-month gap looked identical.
- The line used smoothed (`tension`) interpolation, which misrepresents a cumulative count that only changes on discrete submission dates.

This PR reworks the chart:

1. **Time scale** — the x-axis is now a Chart.js `time` scale (via the already-present `chartjs-adapter-date-fns`), so the line is proportional to elapsed time.
2. **Per-series points** — each organism's series is built as `{x: timestamp, y: cumulative}` points at its own submission dates, instead of a shared category-label array. Markers therefore land on real submission dates.
3. **Markers on every submission** — `pointRadius: 3` (with a larger hover radius) so individual submissions are visible, including the single-submission case.
4. **Step ('before') line** — the cumulative total now holds flat and jumps up exactly on each release date (a staircase), which is the faithful shape for a cumulative count.
5. **Single-date padding** — when every submission shares one date, the time axis would collapse to zero width, so the axis is padded ±15 days around the lone point.
6. **Date parsing** — metadata date strings (`2024`, `2024-03`, `2024-03-15`) are parsed to **local** midnight, avoiding a timezone off-by-one in the rendered/tooltip date.

## Before / after

Identical data rendered with the old config (top) and the new config (bottom). Left = a group with a single submission (the Canada-style case); right = several submissions with uneven time gaps (note Feb→Jul is ~5.5 months).

| | |
| --- | --- |
| **Before** | before |
| **After** | after |

In *Before*, the single-submission chart is blank and the multi-submission line spaces all four dates evenly. In *After*, the single submission shows a clear marker on a padded axis, and the staircase is time-proportional — the three early submissions cluster at the left and the long gap shows as a flat plateau.

> These are faithful renders of the exact Chart.js dataset/scale options from each version (the component just feeds them real LAPIS data).

## Changes

- `website/src/components/User/CumulativeSubmissionsChart.tsx` — only file changed. Swaps `CategoryScale` for `TimeScale` + date-fns adapter, restructures the data transform to per-series `{x, y}` points with a local-midnight date parser, switches the line to `stepped: 'before'` with visible markers, and pads the x-axis for the single-date case.

## Verification

- `npm run check-types` — 0 errors
- `npx eslint` on the changed file — clean
- `npm run format` / prettier — clean
- Rendered the exact before/after Chart.js config in an isolated browser harness to confirm the single-point, padding, staircase direction, and time-proportional spacing behave as intended (screenshots above).

## Notes / possible follow-ups

- The chart only appears when `dateFieldForGroupGraph` is configured for the instance, so behaviour is unchanged where it's disabled.
- With many submission dates the markers could look busy; left as-is since the request was explicitly to show a marker per submission.

### DIFF
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -54,7 +54,7 @@ ingestLimitSeconds: 1800
 getSubmissionListLimitSeconds: 600
 preprocessingTimeout: 600
 accessionPrefix: "LOC_"
-dateFieldForGroupGraph: null
+dateFieldForGroupGraph: date # TEMP(pr-6696): force-enabled to capture preview screenshots — REVERT before merge
 zstdCompressionLevel: 10
 pipelineVersionUpgradeCheckIntervalSeconds: 10
 dataUseTerms:
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -54,7 +54,7 @@ ingestLimitSeconds: 1800
 getSubmissionListLimitSeconds: 600
 preprocessingTimeout: 600
 accessionPrefix: "LOC_"
-dateFieldForGroupGraph: date # TEMP(pr-6696): force-enabled to capture preview screenshots — REVERT before merge
+dateFieldForGroupGraph: null
 zstdCompressionLevel: 10
 pipelineVersionUpgradeCheckIntervalSeconds: 10
 dataUseTerms:
--- a/website/src/components/User/CumulativeSubmissionsChart.tsx
+++ b/website/src/components/User/CumulativeSubmissionsChart.tsx
@@ -1,20 +1,31 @@
-import {
-    Chart as ChartJS,
-    CategoryScale,
-    LinearScale,
-    PointElement,
-    LineElement,
-    Title,
-    Tooltip,
-    Legend,
-} from 'chart.js';
+import { Chart as ChartJS, TimeScale, LinearScale, PointElement, LineElement, Title, Tooltip, Legend } from 'chart.js';
+import 'chartjs-adapter-date-fns';
 import { type FC, useMemo } from 'react';
 import { Line } from 'react-chartjs-2';
 
 import type { Organism } from '../../config.ts';
 import { Spinner } from '../common/Spinner';
 
-ChartJS.register(CategoryScale, LinearScale, PointElement, LineElement, Title, Tooltip, Legend);
+ChartJS.register(TimeScale, LinearScale, PointElement, LineElement, Title, Tooltip, Legend);
+
+// Padding applied to the x-axis when every submission falls on a single date,
+// so the lone point isn't rendered on a zero-width time axis.
+const SINGLE_POINT_PADDING_MS = 15 * 24 * 60 * 60 * 1000;
+
+// Parse a metadata date string ('2024', '2024-03' or '2024-03-15') into a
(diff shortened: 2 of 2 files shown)


kubernetes/loculus/values.yaml
```diff
@@ -54,7 +54,7 @@ ingestLimitSeconds: 1800
 getSubmissionListLimitSeconds: 600
 preprocessingTimeout: 600
 accessionPrefix: "LOC_"
-dateFieldForGroupGraph: null
+dateFieldForGroupGraph: sampleCollectionDate # TEMP(pr-6696): force-enabled to capture preview screenshots — REVERT before merge
```
theosanderson-agent: ⚠️ **Revert before merge.** This `dateFieldForGroupGraph` value was force-enabled **only** to make the group-page graph render on this branch's preview deployment so we could capture real in-app screenshots. It is not part of the fix (the actual change is the `CumulativeSubmissionsChart.tsx` rework) — the graph should stay config-gated/off by default. Revert/drop commit `5b35272` before merging.

kubernetes/loculus/values.yaml
```diff
@@ -54,7 +54,7 @@ ingestLimitSeconds: 1800
 getSubmissionListLimitSeconds: 600
 preprocessingTimeout: 600
 accessionPrefix: "LOC_"
-dateFieldForGroupGraph: null
+dateFieldForGroupGraph: date # TEMP(pr-6696): force-enabled to capture preview screenshots — REVERT before merge
```
theosanderson-agent: ⚠️ **Revert before merge.** This `dateFieldForGroupGraph` value was force-enabled **only** to make the group-page graph render on this branch's preview deployment so we could capture real in-app screenshots (set to `date`, the dummy organism's date field). It is not part of the fix (the actual change is the `CumulativeSubmissionsChart.tsx` rework) — the graph should stay config-gated/off by default. Revert/drop this commit before merging.

theosanderson:
@theosanderson-agent make these graphs always have only integers on the Y axis labels

theosanderson-agent:
Done — the y-axis now only shows integer labels. Set `ticks.precision: 0` on the linear y-scale so Chart.js never labels the cumulative-submission axis with fractional values (which it could do when the max count is small, e.g. ticks at 0.5/1.5). Pushed to `fix-group-submission-graph`.

rneher (approved):
this LGTM (pending the removal of the preview hacks). The changes are very localized and only affect a front end component.

PR #6700, opened 2026-06-17
feat(website): refresh typography & layout for docs/about/news pages

theosanderson-agent:
see https://preview-design-refresh-overlay.pathoplexus.org/ / https://loculus.slack.com/archives/C061ZQQM3N1/p1780659618788189

## Summary

Refreshes the reading experience on Loculus's long-form pages — **docs**, **about** and **news** — by reworking the shared `.mdContainer` prose styling and tidying the docs/about sidebar layout. Also adds a small, optional hook so a downstream instance can extend the prose styling without forking the shared stylesheet.

These are generic, deployment-agnostic changes, so every Loculus-based instance gets the improved typography. They're the upstream foundation for the Pathoplexus design refresh, which layers its own content and homepage FAQ styling on top via the overlay (companion PR: pathoplexus/pathoplexus#1042).

## What changed

**Prose typography — `website/src/styles/mdcontainer.css`**
- Larger, tighter headings (h1 `text-3xl`, h2 `text-2xl`, …) with more deliberate vertical rhythm.
- Comfortable `line-height: 1.6` and a ~`65ch` measure on flowing text (paragraphs, lists, headings, blockquotes), while tables, images and code blocks keep the full content width.
- News pages (`MdLayout` → `.mdContainerItself`) opt out of the per-element measure so prose still wraps around right-floated figures.
- Softer, underlined links; styled inline `code` pills (with a tinted variant inside info-box asides); refreshed blockquote and `.astro-code` block styling.

**Sidebar layout — `website/src/layouts/AboutLayout.astro` & `DocLayout.astro`**
- Wider sidebar (`w-72`) with a vertical divider between navigation and content.
- Content column constrained to a readable measure (`max-w-2xl`).

**Prose styling hook — `website/src/styles/extraStyles.css` (new)**
- Intentionally **empty** in stock Loculus. `mdcontainer.css` imports it, so a downstream overlay can replace this single file to add instance-specific prose styling (e.g. a homepage FAQ card grid) without forking the shared stylesheet. The Pathoplexus PR uses it to supply its `.faqCards` styling.

## Notes
- `npx prettier --check` passes on the changed files. A full `astro build` couldn't be run in my environment (a pre-existing, unrelated broken `flowbite-react` install) — CI will validate the build.

### DIFF
--- a/website/src/styles/mdcontainer.css
+++ b/website/src/styles/mdcontainer.css
@@ -36,6 +36,12 @@
  * figures. Capping individual elements to a measure leaves a right-floated
  * image stranded beyond the text, so let the (already narrow) container width
  * set the measure there and have text wrap naturally around the float.
+ *
+ * `.mdContainerWrap` is a modifier on top of `.mdContainer` and must always be
+ * applied together with it (e.g. `class="mdContainer mdContainerWrap"`). These
+ * rules override the `.mdContainer` ones above via equal specificity, so they
+ * rely on source order: keep `.mdContainerWrap` declared *after* `.mdContainer`
+ * in this file or the overrides will silently stop winning.
  */
 .mdContainerWrap h1,
 .mdContainerWrap h2,
--- a/website/src/styles/mdcontainer.css
+++ b/website/src/styles/mdcontainer.css
@@ -1,47 +1,103 @@
+@import './extraStyles.css';
 @reference './base.css';
 
 .mdContainerItself {
-    @apply max-w-3xl mx-auto break-normal w-full;
+    @apply max-w-2xl mx-auto break-normal w-full;
+}
+
+/*
+ * Base reading typography for docs / about / news prose. Set here (rather than
+ * only on the layout wrappers) so news pages, which render through MdLayout
+ * without the `text-gray-700 leading-relaxed` classes, share the same colour,
+ * size and rhythm as the docs and about pages.
+ */
+.mdContainer {
+    @apply text-gray-800;
+    line-height: 1.6;
+    text-rendering: optimizeLegibility;
+}
+
+/*
+ * Constrain flowing text to a comfortable measure (~70 characters) while
+ * letting tables, images and code blocks use the full content width.
+ */
+.mdContainer h1,
+.mdContainer h2,
+.mdContainer h3,
+.mdContainer h4,
+.mdContainer p,
+.mdContainer ul,
+.mdContainer ol,
+.mdContainer blockquote {
+    max-width: 65ch;
+}
+
+/*
+ * News pages (MdLayout) render through `.mdContainerItself` and use floated
+ * figures. Capping individual elements to a measure leaves a right-floated
+ * image stranded beyond the text, so let the (already narrow) container width
+ * set the measure there and have text wrap naturally around the float.
+ */
+.mdContainerItself h1,
+.mdContainerItself h2,
+.mdContainerItself h3,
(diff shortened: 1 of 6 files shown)


theosanderson-agent:
Thanks for the review. Addressed in 42a312a:

1. **`text-gray-700` vs `text-gray-800` conflict** — fixed by dropping `text-gray-700 leading-relaxed` from the About/Doc content wrappers, so `.mdContainer` solely owns the body colour (gray-800) and line-height (1.6). This also matches `MdLayout`, which already relies on `.mdContainer` for these.
2. **`text-rendering: optimizeLegibility`** — removed; negligible benefit for the sans body text and not worth the perf footgun.
3. **Redundant `.mdContainer ul` block** — keeping this one. The per-depth nested rules (`> ul > li > ul`, etc.) only override `list-style-type` and `margin`, **not** `padding-left`. So the generic `.mdContainer ul { padding-left: 0 }` is the only rule zeroing padding on level-2/3 nested lists; removing it reintroduces the browser's default ~40px `padding-inline-start` on nested `ul`s and shifts their indentation. (It's also pre-existing behaviour.)

website/src/styles/mdcontainer.css
```diff
@@ -1,47 +1,102 @@
...
+.mdContainer h4,
+.mdContainer p,
+.mdContainer ul,
+.mdContainer ol,
+.mdContainer blockquote {
+    max-width: 65ch;
+}
+
+/*
+ * News pages (MdLayout) render through `.mdContainerItself` and use floated
```
anna-parker: @tombch and I realized that `mdContainerItself` has to be applied after `mdContainer` for the changes to take affect, this should be probably documented

website/src/styles/mdcontainer.css
```diff
@@ -1,47 +1,102 @@
...
+ * figures. Capping individual elements to a measure leaves a right-floated
+ * image stranded beyond the text, so let the (already narrow) container width
+ * set the measure there and have text wrap naturally around the float.
+ */
+.mdContainerItself h1,
```
anna-parker: `mdContainerItself` is a very undescriptive name and is only used in Loculus, we could rename to `mdContainerFlow` or `mdContainerWrap` as its used to wrap text around images in the docs/news

website/src/styles/mdcontainer.css
theosanderson: ~This is a pre-existing class that works exactly the same after this PR as before~

website/src/styles/mdcontainer.css
theosanderson: Again, this is as-currently the case, so I wouldn't expect it to block us changing the styles, but have actioned :)

website/src/styles/mdcontainer.css
theosanderson: Well, ok, maybe the order-requirement is changed, fair enough, will do, apologies!

theosanderson: Done in https://github.com/loculus-project/loculus/pull/6700/commits/d16dd75a79cdbda40a1b8347bb307feb005a5386 - good catch thanks!!

website/src/styles/mdcontainer.css
theosanderson: (and again, I take it back given that behaviour has changed a bit and order is important now :) ) - addressed

tombch (approved):
Looks good to me!

PR #6705, opened 2026-06-17
feat(backend): run scheduled tasks with ShedLock

### LINKED ISSUE #6704: Make scheduled tasks not scale in frequency with number of backend replicas
We have scheduled tasks in backend - ideally frequency shouldn't scale with number of backend replicas. We should make it so we ensure only one replica tries to run scheduled tasks.

theosanderson-agent:
## Summary
Implementation of [ShedLock](https://github.com/lukas-krecan/ShedLock) to close #6704.

Today every `@Scheduled` task runs independently on **each** backend replica, so the effective execution frequency scales with the replica count (e.g. with 3 replicas the cleanup task runs ~3× as often, and the CrossRef citation fetch hits the external service 3×). ShedLock fixes this by having replicas compete for a shared, named lock before running a task — only the winner executes that tick; the rest skip it. The result is "run once per interval, regardless of replica count".

## How it works

ShedLock stores locks in a small `shedlock` table in the **existing Postgres database** (no new infrastructure). When a scheduled method fires:

1. The replica attempts to `INSERT`/`UPDATE` a row keyed by the task name, setting `lock_until`.
2. If it wins the row lock, it runs the task and releases the lock when done.
3. Other replicas, seeing an unexpired lock, skip that execution.

`lockAtMostFor` is a safety net: if the holding replica crashes mid-task, the lock auto-expires after that duration so another replica can take over. We use the **database clock** (`usingDbTime()`) for all lock timing, so clock drift between replicas can't cause double-runs.

## Changes

- **`build.gradle`** — add `net.javacrumbs.shedlock:shedlock-spring` and `shedlock-provider-jdbc-template` (7.7.0); `gradle.lockfile` regenerated (STRICT lock mode).
- **`BackendSpringConfig.kt`** — `@EnableSchedulerLock(defaultLockAtMostFor = "PT30M")` and a `JdbcTemplateLockProvider` bean backed by the existing `DataSource` with `usingDbTime()`.
- **Migration `V1.31__add_shedlock_table.sql`** — creates the `shedlock` table.
- **All four scheduled tasks** annotated with `@SchedulerLock`, each `lockAtMostFor` sized above its expected runtime:

  | Task | Interval | `lockAtMostFor` |
  |------|----------|-----------------|
  | `CleanUpStaleSequencesInProcessingTask` | ~60s | `PT5M` |
  | `UseNewerProcessingPipelineVersionTask` | ~10s | `PT10M` |
  | `CleanUpAuxTableTask` | 1h | `PT15M` |
  | `SeqSetCrossRefCitationsTask` | 6h | `PT1H` |

Note the `org.jetbrains.kotlin.plugin.spring` plugin auto-opens Spring components, so the `@SchedulerLock` methods are proxyable without extra `open` modifiers.

## Testing

- `./gradlew compileKotlin ktlintFormat` — pass.
- `./gradlew dependencies --write-locks` — lockfile updated.
- `BackendApplicationTest` (context load with mocked `DataSource`) — pass; the `lockProvider` bean wires up without touching the DB at construction.
- A DB-backed `EndpointTest` against the Postgres test container — pass; Flyway log confirms `Migrating schema "public" to version "1.31 - add shedlock table"` and `Successfully applied 32 migrations`.


anna-parker (dismissed):
lgtm!

I confirmed tasks are running on the preview!

theosanderson:
OK, I will let you/someone else merge as I haven't looked into this enough to take responsibility in the way that merging would imply (thanks a lot for checking!)

anna-parker:
@corneliusroemer and I took a closer look and discovered that the behavior is slightly different than I expected from quickly reading the docs. 

1. Shedlock prevents instances running a task in parallel. It does not check that a task is executed only as many times as defined by the task schedule frequency.

If a second instance tries to start a task while the first instance is running the same task (i.e. that task is locked in the table) the second task run will be skipped. However, if the second instance tried to run the task AFTER the initial instance has finished running (and removed the lock) the second task would still run. As backends will not all start in unison in theory we could get to a state where a task is still run multiple times during its defined frequency period. This will definitely prevent peaks of load when multiple backends start a task in unison but not fully prevent multiple executions of a task.

I believe that a way to fix this would be to set the `lockAtMostFor` and `lockAtLeastFor` parameters to the task frequency. However, I am wondering if it makes sense to just add code for this instead of using a library.

2. `lockAtMostFor` should be greater than the time the task is expected to run for so that another task doesnt try to start while the first task is still running.

theosanderson:
Yes my vague expectation was what you say. Avoiding parallel runs is anyway a good start.  I was thinking we could maybe run on a timetable based on times and then discard the runs that are stopped by locks (possibly)

anna-parker:
alternative suggestion that should handle these edge cases with self rolled code (quite simple as only covers what we need and leads to no external dependencies): https://github.com/loculus-project/loculus/pull/6711

anna-parker:
From discussing with @theosanderson we should increase the `lockAtMostFor` to at least the task run frequency for all tasks (for the GC task this can be higher), and probably set `lockAtLeastFor` to the task frequency (slightly less is ok if we are ok with potentially higher task run frequency).

This will require some test changes

theosanderson:
yeah I guess I would phrase it differently, I would say we set `lockAtLeastFor` to the desired frequency. We set `lockAtMostFor` to the highest interval we would be prepared to tolerate in the rare event that a job died.

theosanderson:
@codex review

backend/src/main/kotlin/org/loculus/backend/service/submission/UseNewerProcessingPipelineVersionTask.kt
```diff
@@ -18,6 +19,15 @@ class UseNewerProcessingPipelineVersionTask(private val submissionDatabaseServic
...
         timeUnit = TimeUnit.SECONDS,
     )
+    @SchedulerLock(
+        name = "useNewerProcessingPipelineVersion",
+        // Defaults to the configured check interval so an operator who tunes `interval-seconds` is
+        // honored, not overridden. The Helm chart sets `atMost` to 5x the interval; PT1M is the
+        // non-Helm fallback. Both overridable via `loculus.locks.*` (tests use PT0S).
+        lockAtLeastFor = "\${loculus.locks.useNewerProcessingPipelineVersion.atLeast:" +
+            "PT\${${BackendSpringProperty.PIPELINE_VERSION_UPGRADE_CHECK_INTERVAL_SECONDS}}S}",
+        lockAtMostFor = "\${loculus.locks.useNewerProcessingPipelineVersion.atMost:PT1M}",
```
corneliusroemer, line 29: This hard coding is problematic when the task starts taking longer on big instances

corneliusroemer: Ugh this is actually configurable but bypassing our explicit config variables

theosanderson: yeah agree it could be confusing to have this set in two ways. Would be happy to discard one

theosanderson:
Closing as Anya and Cornelius prefer a hand-rolled version: https://github.com/loculus-project/loculus/pull/6711

PR #6706, opened 2026-06-17
fix(website): Disable underline for styled button variants

tombch:
Currently, anchor elements which are styled as buttons show an underline on hover, contrary to non-anchor buttons. This PR explicitly disables hover on underline for the styled button variants, removing this conflict.

This also fixes a small bug with the `DownloadButton` used by the `DownloadDialog` where it would briefly flash with an underline on hover.

### Screenshot

Before:

https://github.com/user-attachments/assets/640e0893-6b49-4a93-a3e4-0d800af478bd

After:

https://github.com/user-attachments/assets/82d661e4-8c63-459f-b92e-1c84455d106f

### PR Checklist

### DIFF
--- a/website/src/components/common/Button.tsx
+++ b/website/src/components/common/Button.tsx
@@ -39,15 +39,10 @@ export const Button = forwardRef<HTMLButtonElement | HTMLAnchorElement, ButtonPr
 
     if (props.as === 'a') {
         const { as: _as, variant, size, circle, className, ...rest } = props;
-        // Disable underline on hover for anchor buttons unless the user explicitly added it to the className.
-        // ! is required because the global a:hover underline rule is unlayered and would otherwise win over this utility.
-        const anchorClassName = className?.includes('underline')
-            ? className
-            : [className, 'hover:no-underline!'].filter(Boolean).join(' ') || undefined;
         return (
             <a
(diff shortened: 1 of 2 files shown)


website/src/components/common/Button.tsx
```diff
@@ -39,10 +39,15 @@ export const Button = forwardRef<HTMLButtonElement | HTMLAnchorElement, ButtonPr
 
     if (props.as === 'a') {
         const { as: _as, variant, size, circle, className, ...rest } = props;
+        // Disable underline on hover for anchor buttons unless the user explicitly added it to the className.
+        // ! is required because the global a:hover underline rule is unlayered and would otherwise win over this utility.
+        const anchorClassName = className?.includes('underline')
```
theosanderson: This feels a bit complicated to me - IMO for the really button looking ones we can apply ``hover:no-underline!` - and otherwise not? or something?

tombch: Yeah I've updated this now to just set a hard no-underline rule for the styled variants. I guess this would work as a straightforward solution? I suppose we'll be wanting to move more buttons into variants anyway, and for custom styling of buttons it still works as normal

theosanderson: Great!

theosanderson (approved):
lgtm!

PR #6711, opened 2026-06-17
feat(backend): ensure a task is only executed once regardless of how many backend instances are run

### LINKED ISSUE #6704: Make scheduled tasks not scale in frequency with number of backend replicas

anna-parker:
resolves #6704

Alternative to https://github.com/loculus-project/loculus/pull/6705 (shedLock seems quite easy to implement so one less dependency that we fully understand)

### Background

Today every `@Scheduled` task runs independently on **each** backend replica, so the effective execution frequency scales with the replica count (e.g. with 3 replicas the cleanup task runs ~3× as often, and the CrossRef citation fetch hits the external service 3×).

### Overview

This PR adds a new table called `task_lock` with the columns `task_name`, `started_at` and `locked_util`. When a task attempts to start it will try to get a lock on this table for that task, this is only possible if there is no row for `task_name` or `locked_util` is less than NOW. When it gets a lock it populates the row with `task_name`, `started_at=NOW` and `locked_until=NOW + maxLockTime`

If a task finishes it will delete the row if NOW is past `locked_until` or update `lock_until` with `started_at + minLockTime`

`minLockTime` and `maxLockTime` are set to be a factor of the desired `frequency` a task runs at where the `frequency` needs to be set as higher than the anticipated run time of a task. The task can run up to `maxLockTime` without fear of parallel executions, `maxLockTime` is by default 5*`frequency`. `minLockTime` is set to be 0.9 of the expected `frequency` to prevent subsequent tasks being blocked from starting due to a previous lock being delayed in its removal due to clock skew etc. 

Note we need to use the DB time for the `task_lock` and now comparisons.

### Changes in behavior
- Now only one backend instance will run each task (was previously multiple on PPX prod) -> this is intended. 
- On previews the db will always be deleted and tasks will start on backend start up as configured (after initialDelay) -> same behavior as currently. 
- On persistent instances when the backend restarts e.g. after a rollout the task will not only get run after initialStart up but after the defined task run frequency -> this is a change in behavior, which I think is ok.

### PR Checklist
Configured 2 backends on the preview and checked the grafana logs, I can confirm the newer pipeline version check is now only run once every 10seconds (as desired) and we log that the other task is stopped by the lock as well:
image
Screenshot 2026-06-24 at 11 42 35

🚀 Preview: https://shedlock-alternative.loculus.org


anna-parker:
@claude review

theosanderson:
@codex review

backend/src/main/kotlin/org/loculus/backend/service/scheduler/TaskLockService.kt
```diff
@@ -0,0 +1,115 @@
...
+        maxLockFactor: Double = 5.0,
+        frequencyIntervalSeconds: Long,
+    ): TaskLockService = TaskLockService(
+        minLockFactor = minLockFactor,
+        maxLockFactor = maxLockFactor,
+        frequencyIntervalSeconds = frequencyIntervalSeconds,
+    )
+}
+
+class TaskLockService(
```
corneliusroemer, line 23: High level comment to give mental model

```suggestion
/**
 * Lock can be acquired when current time > `locked_until` (or no row for this task exists)
 * When acquiring, `locked_until` is set to `currentTime + maxLockFactor * frequencyIntervalSeconds`
 * When releasing, `locked_until` is reduced to `started_at + minLockFactor * frequencyIntervalSeconds`
 * maxLockFactor prevents concurrent execution even if tasks run longer than the interval
 * minLockFactor prevents multiple backends from starting new tasks before the interval is almost over
 */
class TaskLockService(
```

backend/src/main/kotlin/org/loculus/backend/service/submission/CleanUpAuxTableTask.kt
```diff
@@ -3,38 +3,54 @@ package org.loculus.backend.service.maintenance
...
-        val thresholdInstant = now.minus(
-            hourCutoff,
-            DateTimeUnit.HOUR,
-            DateProvider.timeZone,
-        ).toLocalDateTime(DateProvider.timeZone)
-        val deletedCount = uploadDatabaseService.deleteAuxTableEntriesOlderThan(thresholdInstant)
+        val taskLockService = taskLockServiceFactory.create(
+            frequencyIntervalSeconds = TimeUnit.HOURS.toSeconds(runEveryHours),
+        )
+        if (!taskLockService.acquireLock("clean-up-aux-table")) return
```
corneliusroemer: `clean-up-aux-table` should be a constant not magic string

anna-parker: done :-)

backend/src/main/kotlin/org/loculus/backend/service/scheduler/TaskLockService.kt
```diff
@@ -0,0 +1,115 @@
...
+     *
+     * @param taskName unique name identifying the task.
+     * @return true if the lock was released, false if the lock was not found or is still held.
+     */
+    fun releaseLock(taskName: String) = transaction {
+        exec(
+            """
+        WITH deleted AS (
+            DELETE FROM task_lock
+            WHERE task_name = ?
```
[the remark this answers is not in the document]
theosanderson: (this seems important to me - I'd indeed expect a lock to have a random UUID or something)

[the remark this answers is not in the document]
anna-parker: Im not sure how important this is as this only affects the case where the run exceeds the `maxDuration` causing tasks to run in parallel - in which case we have bigger issues, but Im happy to add if you think this is a big risk!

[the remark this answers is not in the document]
theosanderson: Idk, it seems not great to me, but looking into it I'm not sure ShedLock is better so I can tolerate stuff where that's the case :)

kubernetes/loculus/values.yaml
```diff
@@ -2707,7 +2707,7 @@ taxonomyService:
 subdomainSeparator: "-"
 replicas:
   website: 1
-  backend: 1
+  backend: 2
```
anna-parker: revert before merging

corneliusroemer (approved):
Thanks this is great!

kubernetes/loculus/values.yaml
```suggestion
  backend: 1
```

Issue #5110, opened 2025-09-26
Don't re-call `get-released-data` unless new data has been added to the organism and preprocessing pipeline version in question


theosanderson:
We currently use too much database compute resources. One reason is that while preprocessing is happening for one organism there are constant writes to the processed_data table. This constantly invalidates our cache which is based on the time that that table was updated but the cache should instead remain valid in many cases as:

- sometimes a different organism is having data processed
- even if the organism in question is having data processed the released data only changes at the point at which the current_pipeline_version changes once the entire dataset has been reprocessed

Our processed data is a write-only database so we could instead use as our Etag equivalent: `the highest values of the processedAt [do we have this?] timestamp in that organism for the current pipeline version`.

PR #6716, opened 2026-06-18
fix(backend): scope current pipeline etag tracker by organism

### LINKED ISSUE #6698: Add organism scoped trigger for current_preprocessing_pipeline to ensure etag isnt changed for other organisms on pipeline version update
Look into code added in https://github.com/loculus-project/loculus/pull/6597 and the slack thread: https://loculus.slack.com/archives/C05G172HL6L/p1780330026392479

clonder:
Fixes released-data ETag invalidation for `current_processing_pipeline`.

Pipeline changes for one organism no longer invalidate released-data cache checks for other organisms.

Fixes #6698 and related to #5110

### DIFF
--- a/backend/src/main/resources/db/migration/V1.31__scope_current_processing_pipeline_tracker_by_organism.sql
+++ b/backend/src/main/resources/db/migration/V1.31__scope_current_processing_pipeline_tracker_by_organism.sql
@@ -1,0 +1,35 @@
+-- Scope current_processing_pipeline tracker updates by organism.
+-- current_processing_pipeline has one row per organism, so changing the current
+-- pipeline for one organism should not invalidate released-data ETags for all
+-- organisms.
+
+
+DROP TRIGGER IF EXISTS update_tracker_trigger ON current_processing_pipeline;
+
+CREATE OR REPLACE FUNCTION update_current_processing_pipeline_tracker()
+RETURNS TRIGGER AS $$
+BEGIN
+    INSERT INTO table_update_tracker (table_name, organism, pipeline_version, last_time_updated)
+    SELECT TG_TABLE_NAME, cr.organism, NULL, timezone('UTC', CURRENT_TIMESTAMP)
+    FROM changed_rows cr
+    GROUP BY cr.organism
+    ON CONFLICT (table_name, organism, pipeline_version)
+    DO UPDATE SET last_time_updated = timezone('UTC', CURRENT_TIMESTAMP);
+    RETURN NULL;
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER update_tracker_trigger_ins
+AFTER INSERT ON current_processing_pipeline
+REFERENCING NEW TABLE AS changed_rows
+FOR EACH STATEMENT EXECUTE FUNCTION update_current_processing_pipeline_tracker();
+
+CREATE TRIGGER update_tracker_trigger_upd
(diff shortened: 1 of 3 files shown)


theosanderson:
@codex review

anna-parker (approved):
Looks great! Sorry for the delay in reviewing!

This isnt really necessary as we have tests and we have confirmed the tracker is working. But if you would like to get a feeling for how reprocessing works you could test this out on a preview and look into the grafana logs

- add the `preview` label to the PR
- Id recommend setting the preview db to persistent by adding `developmentDatabasePersistence: true` to your `values.yaml`
- After all sequences are ingested on the preview doing a pipeline version bump for one organism e.g. ebola
- Then you can confirm that there is no change in the etag of other organisms (you can see this best in the silo-import logs), they will show that if the etag changed or not:

Image

On Wednesday we can set up your kubeconfig so that you can portforward to preview databases as well :-)

backend/src/main/resources/db/migration/V1.31__scope_current_processing_pipeline_tracker_by_organism.sql
```diff
@@ -0,0 +1,35 @@
...
+
+CREATE TRIGGER update_tracker_trigger_upd
+AFTER UPDATE ON current_processing_pipeline
+REFERENCING NEW TABLE AS changed_rows
+FOR EACH STATEMENT EXECUTE FUNCTION update_current_processing_pipeline_tracker();
+
+CREATE TRIGGER update_tracker_trigger_del
+AFTER DELETE ON current_processing_pipeline
+REFERENCING OLD TABLE AS changed_rows
+FOR EACH STATEMENT EXECUTE FUNCTION update_current_processing_pipeline_tracker();
```
corneliusroemer, line 35: Previously this included `TRUNCATE` - maybe the prior PR changing how things are done also dropped `TRUNCATE`.

This would have been more succinct and more complete:

```
CREATE TRIGGER update_tracker_trigger AFTER INSERT OR DELETE OR UPDATE OR TRUNCATE ON public.current_processing_pipeline FOR EACH STATEMENT EXECUTE FUNCTION public.update_table_tracker();
```

Truncate doesn't happen in usual operations but an admin might do it.

clonder: good catch about `TRUNCATE`, thanks! i will add it!

wouldn't the suggested trigger use `update_table_tracker()` though, and therefore write a global tracker row again? my understanding was that we need the per-event triggers here so we can use transition tables and only update the affected organism rows, but I think we can't use transition tables with a combined `INSERT OR UPDATE OR DELETE`, which is why it's split. 

but maybe I'm wrong here let me know if I'm missing something :)

theosanderson: Haven't looked into the details of the specific PR here, but my recent changes to the triggers in previous PRs dropped TRUNCATE support, because the way that they worked did not support it (transition tables as Stacy mentions) and we don't actually use it - so if that's the case here I wouldn't worry

anna-parker: > but I think we can't use transition tables with a combined INSERT OR UPDATE OR DELETE, which is why it's split.

this is true (I also tried it lol)

PR #6714, opened 2026-06-18
feat!(dev): Add test crossref for previews, remove `crossref.hostUrl` config parameter

tombch:
### BREAKING CHANGES
- This PR removes the `crossref.hostUrl` parameter from the kubernetes `values.yaml` and `values.schema.json`. Any existing Loculus configurations that use the `crossref` configuration will need to also remove this parameter.

Note for maintainers, when applying this change kubernetes/argoCD might have a problem applying the changes due to the crossref secret now being owned by sealed secrets, in this case delete the old secret using: `kubectl delete secret crossref -n <namespace>`

### Summary

- Removes the partially filled in dummy Crossref values from the `values.yaml`. 
- Adds required values and secrets to authenticate with `test.crossref.org` on previews.
- Removes the `crossref.hostUrl` config parameter - this was used when constructing the DOIs `resource`, the URL that the DOI resolves to. We can instead use the `backendConfig.websiteUrl`, enabling correct test DOI urls even for previews.

### Testing

- A test SeqSet was created and a test DOI generated on a preview: https://crossref-test-on-previews.loculus.org/seqsets/LOC_SS_1.2

image

- The DOI link does not work (as expected for a test), but we can view it within Crossref test:

image

- In Crossref test, we can also view the content of the DOI, and confirm it has a correct resource tag in the XML:
```
...
        <doi_data>
          <doi>10.62599/LOC_SS_1.2</doi>
          <resource>
            https://crossref-test-on-previews.loculus.org/seqsets/LOC_SS_1.2
          </resource>
        </doi_data>
...
```

### PR Checklist
- [ ] All necessary documentation has been adapted.
- [ ] The implemented feature is covered by appropriate, automated tests.
- [ ] Any manual testing that has been done is documented (i.e. what exactly was tested?)

🚀 Preview: https://crossref-test-on-previews.loculus.org

PR #6718, opened 2026-06-19
feat(config): add back crossRef defaults to the values.yaml config

anna-parker:
In https://github.com/loculus-project/loculus/pull/6714 we removed the crossref placeholders, this is causing errors for users of loculus due to 
```
Error: values don't meet the specifications of the schema(s) in the following chart(s):
loculus:
- seqSets.crossRef: Invalid type. Expected: object, given: null
```

this adds back the defaults, they will be overwritten by previews/instances with actual crossref fields

### PR Checklist

### DIFF
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -11,7 +11,7 @@ localHost: localhost
 robotsNoindexHeader: false
 seqSets:
   enabled: true
-  crossRef: 
+  crossRef:
     DOIPrefix: "placeholder"
     endpoint: "https://test.crossref.org"
   fieldsToDisplay:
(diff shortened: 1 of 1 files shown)


tombch:
Could this be fixed by updating the `crossref` type in `values.schema.json` to be `["object", "null"]` ?

tombch:
I just tested locally and updating the `values.schema.json` to have:

```
...
        "crossRef": {
          "groups": ["general"],
          "type": ["object", "null"],
          "additionalProperties": false,
          "description": "Configuration for CrossRef integration. Set to `null` to disable CrossRef integration (you can still use SeqSets without CrossRef DOIs).",
...
```

enabled setting and successfully linting the `values.yaml` when it had:

```
...
seqSets:
  enabled: true
  crossRef: null
...
```

anna-parker:
@tombch could you make a PR with this alternative?

PR #6723, opened 2026-06-20
feat(everything): POC claude - Cross organism queries: One SILO approach

anna-parker:
🤖 Generated with Claude Code (https://claude.com/claude-code)

This PR implements the "1 SILO" architecture: instead of one LAPIS/SILO instance per organism, all organisms share a single unified LAPIS/SILO instance. All LAPIS queries are routed through the backend, which acts as a proxy. This enables a new cross-organism search page.

### Metadata structure changes

(Note this accounts for the most LOC: `+2,111 -1,627` in the values.yaml and approx 1000 LOC in the backend config test files)

  - Introduced a `sharedMetadata` top-level key in values.yaml for metadata fields shared across all organisms (collection date, geo-location, display name, NCBI release date,
  etc.) and an `organismSpecificMetadata` field for organism specific fields (segment specific fields like INSDC accession, nextclade metadata)
  - The unified SILO database config (_siloDatabaseConfig.tpl) merges all organisms into one schema:
    - An organism string field is added as the first metadata entry (indexed) so SILO can filter by organism.
    - Nucleotide and gene names are organism-prefixed: `{organism}_{segment}-{reference}` or `{organism}_{gene}-{reference}`
    - All metadata fields from all organisms are merged (deduplication by name, first occurrence wins). perSegment fields are expanded per multi-segment organism.

### New backend endpoints

  - LapisProxyController (/query/{endpoint}): proxies all LAPIS/SILO queries from clients through the backend. POST endpoints pass the body (with organism as a filter field)
directly to LAPIS. GET sequence-download endpoints (`/query/unalignedNucleotideSequences`, `/query/alignedNucleotideSequences`, `/query/alignedAminoAcidSequences`) take `organism`,
`segment`, and `reference` as query params and remap them to the correct unified segment/gene name (e.g. `{organism}_{segment}-{reference}`).
  - LapisProxyService: wraps Java's HttpClient to forward POST and GET requests to the internal LAPIS URL, streaming the response body.
  - UnifiedReleasedDataController (/get-released-data): extended to accept an optional organism query param. When omitted, it streams released data for all organisms in a single NDJSON response, padding each row with null values for columns belonging to other organisms (required by SILO, which expects every defined column in every row). Sequence keys are organism-prefixed to match the unified SILO schema.
   - The LAPIS ingress is removed — LAPIS is no longer externally exposed. A single loculus-lapis-service replaces the per-organism services.

### Cross-organism search
  
  - New page at `/search` (website/src/pages/search/index.astro): computes the intersection of metadata fields present across all organism schemas, builds a synthetic
`crossOrganismSchema` with an organism field prepended, and renders the standard SearchFullUI against the unified LAPIS endpoint. Mutation search and sequence downloads are currently disabled.
  - Website code is refactored to call the backend with the appropriate parameters

### What this PR does not handle:
- Metadata name conflicts in metadata of a different type in organismSpecificFields
- Lineage searches (not too much to add actually I just wanted to simplify)
- Potentially organism specific metadata could be further split into organism sample and organism segment-reference specific metadata (i.e. sampling procedure vs. nextclade alignment details)
- Unaligned sequence downloads on the cross organism query page
- Mutation search is broken as it now requires a `{organism}_{segment}-{reference}` prefix, this should be removed via the webpage or the backend
- Create a new `INSDC accession(s)` metadata field that includes a list of all INSDC accessions for segmented organisms so this can be searched for on the cross organism page.

## Limitations:
- One SILO is great in many ways BUT our current approach to updating SILO is just not scalable. Now when there is any new data ALL data must be streamed again to SILO. Switching to one SILO will only be possible when SILO supports incremental uploads with deletions.
- The backend currently parses and modifies all requests to LAPIS but does not modify/parse the response. Modifying requests is feasible. However, as soon as LAPIS/SILO has different sequence and metadata names than the ones we expose to users we would need to modify these names in the response we send to users. This is definitely possible but another level of complexity, especially if the user for example requested compressed mutations. Potentially adding an option in LAPIS to map the exported gene/segment names would be a better alternative than mapping in the backend.

#### Alternatives until SILO supports incremental uploads:
- Continue with one SILO per organism and add new SILO with all shared metadata and unaligned sequences (this does mean we duplicate the metadata and unaligned sequences across LAPIS/SILO instances yet again which is inefficient), this can then be easily switched out later if LAPIS/SILO is not exposed via the backend. Requires: new metadata structure, proxying of endpoints in the backend


backend/src/main/kotlin/org/loculus/backend/model/ReleasedDataModel.kt
```diff
@@ -63,6 +63,64 @@ open class ReleasedDataModel(
...
+                    aminoAcidInsertions = allAminoAcidKeys.associateWith {
+                        prefixedAminoAcidInsertions[it]
+                            ?: emptyList()
+                    },
+                )
+            }
+        }
+    }
+
     @Transactional(readOnly = true)
```
anna-parker, line 125: this should be deleted now, this is a sloppy refactor

backend/src/main/kotlin/org/loculus/backend/controller/UnifiedReleasedDataController.kt
```diff
@@ -0,0 +1,97 @@
+package org.loculus.backend.controller
```
anna-parker, line 1: Im not sure how efficient SILO is in loading all data in, in theory we could already sort data in the get-released-data endpoint by organism and reference to improve SILO's binary encoding

anna-parker:
This made me start thinking of multi-pathogen (multi-seg/multi-ref) and how tedious the whole mutation querying is as the mutations need a unique prefix which we create currently as `{seg}-{ref}`, or `{gene}-{ref}`, the website needs to add ref to the prefix when calling lapis and remove when reading it in. Ideally, the backend would do this but ref needs to be passed with each mutation. For query responses we could easily add a the reference instead of the prefix, allowing easy display on the website. For querying we would either continue to modify the mutation prefix or potentially allow a json structure of `{segment:seg, reference:ref1, mutation:12T}`

PR #6724, opened 2026-06-22
refactor(website): Split fetching of backend and lapis sequence details in `findOrganismAndData`

tombch:
### Summary

- The `findOrganismAndData` function is used in the website to retrieve sequence details when viewing sequences both in the search UI and on an individual sequence details page for an accession. Previously, this function called `getSequenceDetailsTableData` for each organism, which would query LAPIS (for sequence details) and the backend (for data use terms history). The backend calls were duplicated across organisms when only a single call needs to be made.
- This PR refactors the `findOrganismAndData` and `getSequenceDetailsTableData` to keep the fan-out approach for LAPIS queries, but have a single call to all required backend endpoints. It does this by splitting the LAPIS and backend data fetching into separate `getLapisSequenceDetails` and `getBackendSequenceDetails` functions. 
- With the `getBackendSequenceDetails` now called once rather than per-organism, we add the sequence citation fetching to the `getBackendSequenceDetails`, to remove some duplicated code. 

### PR Checklist

### DIFF
--- a/website/src/pages/seq/[accessionVersion]/details.json.ts
+++ b/website/src/pages/seq/[accessionVersion]/details.json.ts
@@ -2,24 +2,15 @@ import { type APIRoute } from 'astro';
 
 import { findOrganismAndData } from './findOrganismAndData';
 import { SequenceDetailsTableResultType } from './getSequenceDetailsTableData';
-import { getRuntimeConfig, getSchema, seqSetsAreEnabled } from '../../../config';
+import { getRuntimeConfig, getSchema } from '../../../config';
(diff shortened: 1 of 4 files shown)


corneliusroemer:
What about a backend endpoint that returns the organism so we don't need to fan out to all the LAPISes? Single lookup in sequence entries table.

theosanderson:
> What about a backend endpoint that returns the organism so we don't need to fan out to all the LAPISes? Single lookup in sequence entries table.

Tom has suggested this in the past but I've argued for not investing in this til we decide the new API, which we're working on atm

PR #6725, opened 2026-06-22
fix(backend): batch processed file ownership validation

### LINKED ISSUE #3951: Improve performance of file validation for `/submit-processed-data`
The `throwIfFilesDoNotExistOrBelongToWrongGroup()` function implemented in #3947 is called for each processed sequence entry and is inefficient as it has to make two database queries to fetch the group ID of the sequence entry and the group IDs of the associated files. We should improve that; one solution would be to cache the group IDs of sequence entries and/or files in the backend.

clonder:
Resolves #3951

Batches file ownership validation in `/submit-processed-data` so the backend checks file groups once per stream chunk instead of once per sequence entry.

This keeps the existing safety check that processed data cannot reference files from another group, while reducing many small DB lookups during large preprocessing batches.


backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -329,6 +324,42 @@ class SubmissionDatabaseService(
...
+    ): Map<AccessionVersion, Set<FileId>> {
+        val filesToValidate = mutableMapOf<AccessionVersion, Set<FileId>>()
+        submittedProcessedDataBatch.forEach { submittedProcessedData ->
+            submittedProcessedData.data.files?.let { fileMapping ->
+                // Validate file names, categories, upload completion, and existence before ownership checks.
+                fileMappingPreconditionValidator
+                    .validateFilenameCharacters(fileMapping)
+                    .validateFilenamesAreUnique(fileMapping)
+                    .validateCategoriesMatchOutputSchema(fileMapping, organism)
+                    .validateMultipartUploads(fileMapping.fileIds)
```
anna-parker: `validateMultipartUploads` also calls the db (`FilesTable`), so this will still be done once per entry

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -329,6 +324,42 @@ class SubmissionDatabaseService(
...
+                    .validateFilesExist(fileMapping.fileIds)
```
anna-parker: this also calls the `FilesTable` twice actually for all fileIds

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -477,7 +508,6 @@ class SubmissionDatabaseService(
         organism: Organism,
     ) = try {
         throwIfIsSubmissionForWrongOrganism(submittedProcessedData, organism)
-        validateFilesBelongToSubmittingGroup(submittedProcessedData)
```
anna-parker, line 480: why is this removed here?

anna-parker: ah I see this was moved after validateFileMappingsAndCollectFileIds - which makes a lot more sense!

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -276,29 +276,24 @@ class SubmissionDatabaseService(
...
+        // Process the NDJSON stream in chunks so DB lookups are batched without buffering the whole request.
+        reader.lineSequence().chunked(streamBatchSize).forEach { lines ->
+            val submittedProcessedDataBatch = lines.map { parseSubmittedProcessedDataLine(it) }
+
+            val filesToValidate = validateFileMappingsAndCollectFileIds(
+                submittedProcessedDataBatch,
+                organism,
+                processedFiles,
+            )
+            validateFilesBelongToSubmittingGroups(filesToValidate)
```
anna-parker: I think the other file validations that call the db: `validateMultipartUploads` and `validateFilesExist` can be added here, so that they only call the db once per batch (they take a set of fileIds as input so I dont see an issue moving them unless I overlooked sth)

anna-parker:
@claude review

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
anna-parker: this is a preexisting issue but Im now concerned the stream is not being closed correctly, I believe ServletInputStream (the inputStream) handles closing itsself but Im not sure this BufferedReader class does... probably this isnt an issue as weve been doing it but might be good to check

anna-parker: https://stackoverflow.com/questions/30862577/does-closing-a-stream-close-the-bufferedreader-source

anna-parker: doesnt seem to close the stream on its own? https://docs.oracle.com/javase/8/docs/api/java/io/BufferedReader.html

anna-parker: @claude what are your thoughts? do we need to close this stream explicitly?

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
clonder: Moved validateMultipartUploads and validateFilesExist out of the per-entry loop.

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
clonder: added `use {}` here so the lifecycle is explicit

backend/src/test/kotlin/org/loculus/backend/controller/submission/SubmitProcessedDataEndpointTest.kt
```diff
@@ -425,6 +427,58 @@ class SubmitProcessedDataEndpointTest(
...
+                ),
+            ),
+            PreparedProcessedData.withFiles(
+                accessions[1],
+                mapOf(
+                    "myFileCategory" to listOf(
+                        FileIdAndName(secondFileIdAndUrl.fileId, "foo-2.txt"),
+                    ),
+                ),
+            ).copy(errors = PreparedProcessedData.withErrors(accessions[1]).errors),
```
anna-parker: The change that now also errored files are still be checked for ownership isnt too relevant IMO so I would probably remove it from the tests
```suggestion
            ),
```

anna-parker: (think its important we test the case where they dont have other errors) - but you could also add both

anna-parker (approved):
looks great!

PR #5760, opened 2025-12-18
feat(website): add diff view for comparing sequence versions

corneliusroemer:
Based on quick vibe code session with @maverbiest

First attempt at offering some diffing between versions - we have nothing so far. This is better than nothing - clearly has room for improvement.

Try it out:
- Sequences with 2 versions show diff automatically: https://diff-versions.loculus.org/seq/LOC_0002GVP.2/versions?compare=1%2C2
- Sequences with more than 2 allow selecting 2 to diff: https://diff-versions.loculus.org/seq/LOC_0002H0D.4/versions?compare=2%2C4

image

## Summary
Add interactive diff view to compare metadata changes between sequence versions on the `/seq/{accession}/versions` page.

### Features
- **Smart defaults**: Auto-compares when only 2 versions exist, shows checkboxes for 3+ versions
- **URL persistence**: Shareable links with `?compare=1,3` format
- **Toggle view**: Show all fields or only changed fields
- **Noisy field handling**: Timestamps and version info displayed greyed out at bottom
- **Visual design**: Changed fields highlighted with amber background, grouped by metadata headers

### Components
- `VersionsWithDiff.tsx` - Main React component with version selection and data fetching
- `DiffTable.tsx` - Table display with field comparisons
- `compareVersions.ts` - Logic to identify changed/unchanged/noisy fields
- `types.ts` - TypeScript types for comparison results

### Test Plan
- [x] Type checking passes (`npm run check-types`)
- [x] Unit tests pass (`CI=1 npm run test`)
- [x] Code formatted (`npm run format`)
- [ ] Manual test: Browse to a sequence with 2 versions on preview, verify auto-compare
- [x] Manual test: Browse to a sequence with 3+ versions, select 2, verify comparison
- [ ] Manual test: Toggle "Show all fields", verify unchanged fields appear
- [x] Manual test: Verify URL params persist and are shareable
- [ ] Manual test: Verify noisy fields (timestamps) are greyed out

🤖 Generated with [Claude Code](https://claude.com/claude-code)

🚀 Preview: Add `preview` label to enable

PR #6726, opened 2026-06-22
feat(website): add version diff table on "All versions" page

### LINKED ISSUE #2803: Diff view of Loculus versions
It would be great if people could easily check what changed between Loculus versions.

This could be a new REST endpoint `/diff` that takes two versions and outputs only changes. This could then be surfaced on the website in some way.

We could eventually even offer _blame_, which could be something cool to surface on every sequence details page, i.e. annotate fields with the version since which the field has been what it is now.

corneliusroemer:
Add version diff table to `/seq/[ACCESSION]/versions` to allow quick identification of what changed between versions.

Preview ready for testing (until it gets wiped): https://diff-versions-v2.loculus.org/seq/LOC_0002U7N.3/versions?compare=2%2C3

Based on https://github.com/loculus-project/loculus/pull/5760
Resolves #2803

Key features:
- Users can toggle between hiding unchanged fields (on by default) and showing all fields. Changed fields have yellow background to set them apart. Fields that are always going to be different (release time etc) are greyed out.
- Users can toggle hiding shared substitutions/indels (on by default) 
- Fields are ordered and grouped under same headers same as on sequence details page
- URL encodes compared versions to allow sharing. By default the most recent 2 versions are selected

Caveat: There are edge cases where a sequence can change and this is not visible in this view as `/details.json` does not include all information required to fully reconstruct a sequence. E.g. the position of Ns could change and this would not necessarily be visible (unless it causes differences in substitutions/indels). This could be worked around by fetching sequences separately from `/details.json` but I decided to keep it simple for now

### Additional changes

- Fix bug in `withQueryProvider` function to not recreate query client on each React render through lazy initialization
- Make some sequence details page functions reusable to avoid duplication of header grouping/ordering
- Fix mutation badge component to only show horizontal line below segment/gene name when that name is not empty

### Tests

I extended existing integration tests that revise sequences to now also check basic functionality of the diff component. 

### Screenshot/cast

Google Chrome 2026-06-22 16 32 03

Google Chrome 2026-06-22 16 32 14

https://github.com/user-attachments/assets/459dc8c2-ffa6-4b16-b081-feaf95d46743

🚀 Preview: https://diff-versions-v2.loculus.org


website/src/pages/seq/[accessionVersion]/versions.astro
Comment (anna-parker) [inline website/src/pages/seq/[accessionVersion]/versions.astro:1]: I wonder if this page should be called `Compare Versions` instead of `All Versions` - I am not sure I would have found it as a normal user

website/src/components/common/withQueryProvider.tsx
```diff
@@ -1,16 +1,21 @@
...
-        const queryClient = new QueryClient({
-            defaultOptions: {
-                queries: {
-                    refetchOnWindowFocus: false,
-                },
-            },
-        });
+        // Create the client lazily once and keep it stable across re-renders, otherwise a new
+        // QueryClient (and thus an empty cache) would be created on every render.
+        const [queryClient] = useState(
```
anna-parker, line 9: this worries me wont this mean if the data updates in between renders, for example there is a new version or if external metadata is added the page wont reload with the changes? it seems like the standard would be to set a staleTime of e.g. 60seconds and after that if the user comes back to a tab it will be refreshed: `staleTime: 60_000,`

website/src/components/VersionDiff/VersionsWithDiff.tsx
```diff
@@ -0,0 +1,122 @@
...
+                                        </p>
+                                    </div>
+                                </div>
+                            </div>
+                        </li>
+                    ))}
+                </ul>
+            </div>
+
+            {(comparison !== undefined || selectedPair !== null) && (
```
anna-parker, line 73: shouldnt this be?
```suggestion
            {(comparison !== undefined && selectedPair !== null) && (
```

website/src/components/common/withQueryProvider.tsx
anna-parker, line 1: I wonder if this change should be in a separate PR as it affects other files that use the QueryProvider

website/src/components/common/withQueryProvider.tsx
corneliusroemer: This is just a bug fix for situations where the query client was called from react components that rerendered. Which isn't that many places but affected the feature here.

website/src/components/VersionDiff/VersionsWithDiff.tsx
corneliusroemer: No, comparison is the current loaded comparison and selected is the future currently loading one. We keep showing the current one greyed out to not have the scroll state flicker about

website/src/components/common/withQueryProvider.tsx
corneliusroemer: It's just a small bug fix, you can make a separate PR for it if you want but it just complicates things

website/src/components/VersionDiff/VersionsWithDiff.tsx
corneliusroemer: codex: Changing it to && would hide the whole comparison panel during the initial load, because selectedPair !== null is true but comparison is still undefined, so the spinner would not render.

One could make it ` {selectedPair !== null && (` to hide the comparison if one selects only 1

anna-parker: ah... Im fine with keeping it this way then, thanks for the explanation!

website/src/components/VersionDiff/VersionsWithDiff.tsx
```diff
@@ -0,0 +1,122 @@
...
+        <div>
+            <div className='mb-6'>
+                <h2 className='text-xl font-semibold mb-4'>Version History</h2>
+                {showCheckboxes && <p className='text-sm text-gray-600 mb-3'>Select two versions to compare</p>}
+                <ul className='p-3'>
+                    {versions.map((version) => (
+                        <li key={version.version} className='mb-4'>
+                            <div className='flex items-start gap-3'>
+                                {showCheckboxes && (
+                                    <Checkbox
```
theosanderson, line 41: I think eventually we should change this to two columns of radio buttons - can make an issue some time - but this is fine for now

theosanderson: (wikipedia is a good example)

website/src/components/common/withQueryProvider.tsx
anna-parker: I just confirmed that this is only relevant when the component is created with `client:only='react'`, which is only applied here and in `ErrorFeedback`, so this doesnt seem to be a problem. I also looked that when the queryClient is used `refetchOnWindowFocus` can still be set to true and a staleTime added (which you do actually) - so I will resolve this for now, thanks @tombch for looking explaining this to me!

integration-tests/tests/specs/features/sequence-version-banners.spec.ts
```diff
@@ -21,6 +22,19 @@ const TEST_SEQUENCE =
     'CAATATCCAACTTCCTGGCAATCAGTTGGACACATGATGGTGATCTTCCGTTTGATGAGAACAAACTTTTTAATCAAGTTCCT' +
     'ACTAATACATCAGGGGATGCACATGG';
 
+// TEST_SEQUENCE with one point mutation (SNP) and one in-frame (3-nt) deletion, used to
```
anna-parker, line 25: The tests do not appear to check that these sequence mutation differences are shown on the comparison page?

website/src/components/VersionDiff/mutationDiff.ts
```diff
@@ -0,0 +1,73 @@
...
+        const onlyInA = mutations.filter((m) => !bKeys.has(badgeKey(m)));
+        if (onlyInA.length > 0) {
+            result.push({ segment, mutations: onlyInA });
+        }
+    }
+    return result;
+}
+
+/** Returns the segments/strings of `a` that are not present (per segment) in `b`. */
+function subtractSegmentedStrings(
```
anna-parker, line 26: these are insertions and deletions right? maybe we should add to doc string

anna-parker (approved):
looks great! I revised metadata and sequences across all organisms (includings EVs and multi-seg, multi-ref) and this works great!

website/src/components/VersionDiff/mutationDiff.ts
corneliusroemer: In practice indels yes but this might depend on config so maybe better to leave it out to not become stale

website/src/components/VersionDiff/VersionsWithDiff.tsx
corneliusroemer: Good idea and example! Made an issue: https://github.com/loculus-project/loculus/issues/6735

integration-tests/tests/specs/features/sequence-version-banners.spec.ts
corneliusroemer: Good idea, I now test for presence/absence of these mutations

website/src/pages/seq/[accessionVersion]/versions.astro
Comment (corneliusroemer) [reply website/src/pages/seq/[accessionVersion]/versions.astro:1]: Changed to `Compare versions`

corneliusroemer:
Renamed `All versions` -> `Compare versions` and also test for mutation/indel diff in integration test

PR #6731, opened 2026-06-24
feat(prepro): Pass files through nextclade pipeline

maverbiest:
Partly resolves https://github.com/loculus-project/loculus/issues/6758

Currently, user-submitted data are not read in by the nextclade preprocessing pipeline, causing them to be dropped form the submission process. Since we're now working to implement file sharing on Loculus, we need to make it so user-submitted files make it through the preprocessing pipeline as well.

This PR makes it so user-submitted files are forwarded through the nextclade pipeline without any processing or checking of file contents. Future PRs will add functionality to, for example, check for host sequences in submitted raw reads.

## Implementation

- `parse_ndjson` now also parses the file related information sent to preprocessing by the backend
- `UnprocessedData` and `UnprocessedAfterNextclade` both get a `files` attribute
- test factories in `factory_methods.py` can now be given files to add to test objects, also added a test case in `test_nextclade_preprocessing.py` that carries file information

## Manual testing

I created a preview to test whether files now make it through the nextclade pipeline. When I submit sequences to the preview with attached raw reads, this file now appears in the submission review page:

grafik

And also on the sequence details page after the sequence is released:

grafik

## Open questions

One thing I ran in to when implementing this is that you need to add file categories two times in the config: one time under `submissionDataTypes` (file categories that users are allowed to submit) and then again under a top-level `files` field (files accepted as outputs of prepro pipelines):

```
defaultOrganismConfig: &defaultOrganismConfig
  schema: &schema
    submissionDataTypes: &defaultSubmissionDataTypes
      consensusSequences: true
      maxSequencesPerEntry: 1
      files:
        enabled: true
        categories:
          - name: raw_reads
            displayName: Raw reads
    ...
    files:
      - name: annotations
        displayName: Annotations
      - name: raw_reads
        displayName: Raw reads
```

Would it be nicer to always allow file categories listed under `submissionDataTypes.files` to be output but prepro? Or will we ever have cases where users submit one thing, prepro processes it, and then outputs another filetype? 

Probably safest to keep as-is for now but just wanted to flag since not doing this properly got me into a weird state where submissions stay in 'processing' indefinitely but never error because the backend doesn't accept the preprocessing output (it only logs errors in the backend).

### PR Checklist

🚀 Preview: https://pass-files-through.loculus.org


preprocessing/nextclade/src/loculus_preprocessing/backend.py
```diff
@@ -94,14 +94,13 @@ def parse_ndjson(ndjson_data: str) -> Sequence[UnprocessedEntry]:
             key: trim_ns(value) if value else None
             for key, value in unaligned_nucleotide_sequences.items()
         }
-        user_files = json_object["data"].get("files")
```
anna-parker: Im not sure these changes are needed (the comment might also be useful)

anna-parker: sorry I reviewed the wrong diff lol

preprocessing/nextclade/tests/test_nextclade_preprocessing.py
```diff
@@ -284,6 +285,40 @@ def invalid_sequence() -> str:
...
+            "totalInsertedNucs": 0,
+            "totalSnps": 1,
+            "totalDeletedNucs": 0,
+            "length": len(consensus_sequence("single")),
+            "nonExistentField": "None",
+            "variant": True,
+        },
+        expected_errors=[],
+        expected_warnings=[],
+        expected_processed_alignment=ProcessedAlignment(
```
anna-parker, line 316: it seems like files are dropped in this test?

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
anna-parker, line 1: Im not sure the path for sequences without an alignment is currently handled, e.g. the `process_single_unaligned` seems to still drop files I believe

anna-parker:
@maverbiest regarding your open question, yes I believe it is quite likely that prepro will in future receive one file and output a different one so I think the current distinction is good

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
maverbiest: Oh thanks for flagging will look into it!

maverbiest: The files should also get passed through for the sequences without alignment!

I'm adding them to the `ProcessedData` in `processed_entry_no_alignment` here: https://github.com/loculus-project/loculus/pull/6731/changes#diff-e2cd7309bc036f2278b096358274158e03166a2d7c1fb2842c856f626da9acb0R299

preprocessing/nextclade/src/loculus_preprocessing/datatypes.py
```diff
@@ -82,6 +88,7 @@ class UnprocessedData:
     submissionId: str  # noqa: N815
     metadata: InputMetadata
     unalignedNucleotideSequences: dict[SequenceName, NucleotideSequence | None]  # noqa: N815
+    files: dict[str, list[FileIdAndName]] | None
```
tombch: Might be worth adding `FileCategory = str` as a type alias further up, then using it in place of `str` for the `files` attributes etc? Would help with reading what the mapping is to me

preprocessing/nextclade/tests/test_nextclade_preprocessing.py
maverbiest: Ah no I was just always setting both `unprocessed_entry.files` and `expected_output.files` to `files` since we're just passing the files through and not modifying them at all atm.

But agree that this was confusing to read, I now changed it so you need to define `expected_files` and `input_files` explicitly

anna-parker: yes that is much clearer!

anna-parker (approved):

preprocessing/nextclade/src/loculus_preprocessing/datatypes.py
maverbiest: Thanks! I added this now

PR #6733, opened 2026-06-24
fix(website): When in search ui, navigating to the latest version from sequences banner keeps in search ui

tombch:
### Summary

- Currently, when viewing an earlier version of a sequence in the search page, if you click the sequences banner that prompts to go to the latest version, it takes you away from the search page and onto the sequence details page. I find this behaviour a bit counter-intuitive - I would expect to go to the latest version but still within the search ui.
- This PR adds an optional hook to the `SequencesBanner` so that when in the `SeqPreviewModal`, clicking this takes you to the latest version, but still within the search ui.

### Screenshot
This is the banner in the search page:

image

Previously when clicking, it would take you to the sequence details page for the latest version:

image

now when its clicked, it keeps the user on the search page:

image

### PR Checklist


website/src/components/SequenceDetailsPage/SequencesBanner.tsx
```diff
@@ -26,12 +32,21 @@ const SequencesBanner: FC<SequencesBannerProps> = ({ sequenceEntryHistory, acces
...
                         <p>
                             The latest version is:
-                            <a
-                                href={routes.sequenceEntryDetailsPage(latestAccessionVersion)}
-                                className='font-bold underline mx-1'
-                            >
-                                {getAccessionVersionString(latestAccessionVersion)}
-                            </a>
+                            {handleNavigateToVersion ? (
+                                <Button
```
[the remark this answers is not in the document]
tombch: IMO this is fine to leave but can change if anyone thinks otherwise!

website/src/components/SequenceDetailsPage/SequencesBanner.tsx
```diff
@@ -3,13 +3,19 @@ import type { FC } from 'react';
...
     sequenceEntryHistory: SequenceEntryHistory;
     accessionVersion: string;
+    handleNavigateToVersion?: (accessionVersion: string) => void;
 };
 
-const SequencesBanner: FC<SequencesBannerProps> = ({ sequenceEntryHistory, accessionVersion }) => {
+const SequencesBanner: FC<SequencesBannerProps> = ({
+    sequenceEntryHistory,
+    accessionVersion,
+    handleNavigateToVersion,
```
theosanderson: Probably if I were writing this I would keep the name `setPreviewedSeqId` to avoid introducing a new name (but it's a matter of taste and not blocking)

theosanderson: (partly I guess as otherwise I'd expect this function to just accept a version number)

theosanderson (approved):
nice

website/src/components/SequenceDetailsPage/SequencesBanner.tsx
tombch: Yeah my original thinking was if we wanted a generic prop to reuse if we wanted other navigation in the future with the sequences banner, but now I'm re-looking at it the naming is a bit confusing so will switch to this for now!

PR #6738, opened 2026-06-24
chore(website): delete dead files (FieldProps, CitationPlot)

theosanderson-agent:
## Summary

Deletes two unreferenced files in the website, found by running [`knip`](https://knip.dev/) and then verified by hand.

- **`src/components/SearchPage/fields/FieldProps.tsx`** — exports only a `FieldProps` type, which no module imports.
- **`src/components/SeqSetCitations/CitationPlot.tsx`** — the `CitationPlot` component is referenced nowhere.

## Verification

knip on the website surfaced a large list of "unused" files/exports, but most are false positives once you account for:
- the **pathoplexus** deployment overlay (loculus is a base template whose `website/` source is overlaid downstream), and
- Astro `.mdx` frontmatter `layout:` edges, which knip does not follow.

After cross-checking every candidate against loculus **and** the pathoplexus overlay across `.tsx`/`.ts`/`.astro`/`.mdx`, only these two files are genuinely dead. Both were confirmed to have zero importers in either repo.

The remaining knip "unused export" findings are internally-used (over-exported) symbols and are handled in a separate, low-risk PR — not deletions.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

🚀 Preview: Add `preview` label to enable

PR #6740, opened 2026-06-24
chore(website): drop redundant export on internal-only symbols

theosanderson-agent:
## Summary

[`knip`](https://knip.dev/) flagged a number of "unused exports" in the website. Investigating each, these symbols are **not** dead — they are used, but **only within their own module**. The `export` keyword was therefore unnecessary. This PR removes the redundant `export` while keeping the code; there is **no behaviour change** and nothing is deleted.

| Symbol | File | Internal use |
|---|---|---|
| `jsDateToISOString`, `isoStringToJsDate`, `jsDateToTimestamp`, `timestampToJsDate` | `SearchPage/fields/DateField.tsx` | used as the date/value converters in the same file |
| `displayRevocationDialog`, `RevocationDialog` | `SequenceDetailsPage/RevokeButton.tsx` | `displayRevocationDialog` is called in-file; it renders `RevocationDialog` |
| `InnerSearchFullUI` | `SearchPage/SearchFullUI.tsx` | rendered by the exported wrapper in the same file |
| `CopyUrlButton` | `SearchPage/DownloadDialog/DownloadButton.tsx` | rendered in the same file |

## Verification

- Each symbol has **zero importers** in loculus or in the **pathoplexus** deployment overlay (checked across `.tsx`/`.ts`/`.astro`/`.mdx`, including Astro `.mdx` frontmatter `layout:` edges).
- Each symbol is still referenced internally, so removing `export` is safe.
- `npm run check-types` shows no new errors in the touched files (pre-existing errors on `main` are unrelated: a not-yet-installed `exceljs` dep and an implicit-any in `submission/template/index.ts`).
- `eslint` and `prettier` clean on all four files.

Companion PR #6738 deletes the two genuinely-dead files found in the same knip pass.

### DIFF
--- a/website/src/components/SearchPage/DownloadDialog/DownloadButton.tsx
+++ b/website/src/components/SearchPage/DownloadDialog/DownloadButton.tsx
@@ -15,7 +15,7 @@ type DownloadButtonProps = {
     onClick?: () => void;
 };
 
-export const CopyUrlButton: FC<{ url: string }> = ({ url }) => {
+const CopyUrlButton: FC<{ url: string }> = ({ url }) => {
     const [copied, setCopied] = useState(false);
 
     const handleCopy = () => {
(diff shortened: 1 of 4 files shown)


maverbiest (approved):
Looks good to me!

PR #6769, opened 2026-06-25
feat(website): add SeqSet citations section to admin dashboard

theosanderson-agent:
## Summary

Adds a **SeqSet citations** section to the admin dashboard that lists every publication citing a SeqSet in the instance. This gives super users a single place to see all external citations across all SeqSets, complementing the existing per-SeqSet (`get-seqset-citations`) and per-sequence (`get-sequence-citations`) views, which previously had no aggregate/admin-level counterpart.

It also lets super users **manually register a citation** ("curated", as opposed to "crossref"-discovered) — useful when a publication hasn't been indexed by CrossRef yet, and as a way to seed citation data for e2e testing without depending on an external CrossRef call.

## Backend

- New endpoint `GET /admin/get-all-seqset-citations` on `SeqSetCitationsController`.
  - Placing it under `/admin/*` means it is automatically gated to **super users** by the existing `SecurityConfig` matcher (`auth.requestMatchers("/admin/*").hasAuthority(SUPER_USER)`), so no new security wiring is needed.
  - It stays on `SeqSetCitationsController`, which is `@ConditionalOnProperty(ENABLE_SEQSETS)`, so the endpoint only exists when SeqSets are enabled.
- `SeqSetCitationsDatabaseService.getAllSeqSetCitations()` joins the citation source, join, and SeqSets tables and returns each citation source together with the list of SeqSets it references (accession version, name and DOI). Results are grouped per citation source and ordered by year (descending), mirroring the existing `getSequenceCitations` query shape.
- New API types `AdminSeqSetCitation` and `CitedSeqSet`.
- New `POST /admin/add-seqset-citation` and `DELETE /admin/delete-seqset-citation` endpoints (also under `/admin/*`, so also super-user gated):
  - Adding sets `origin = CURATED` on `seqset_citation_source` (the schema already had a `CROSSREF`/`CURATED` enum, but nothing ever wrote `CURATED` before this). Re-submitting the same `sourceDOI` updates its metadata and links any newly listed SeqSets without dropping existing links.
  - Deleting only works for `CURATED` citations — `CROSSREF`-discovered citations can't be removed through this admin path.
  - `AdminSeqSetCitation` now also returns `origin`, so the frontend knows which rows are deletable.

## Website

- `BackendClient.getAllSeqSetCitations(token)` calls the new endpoint (parsed with the new `adminSeqSetCitations` zod schema).
- New `AdminSeqSetCitationsTable` component (renamed from `SeqSetCitationsTable` for clarity, since it's specifically the admin aggregate view) renders, per citation: the title (linking to the source DOI), contributors, year, and the cited SeqSets (each linking to its SeqSet page, with accession version and DOI), plus a "Remove" button on `CURATED` rows.
- New `AddSeqSetCitationForm` renders a form (source DOI, title, year, contributors, cited SeqSet accession-versions) that posts to the new add endpoint.
  - It also has a "Fetch from DOI" button that calls CrossRef's public REST API (`api.crossref.org/works/{doi}`) directly from the browser to autopopulate title, year, and contributors. That endpoint sends `Access-Control-Allow-Origin: *`, so no backend proxy is needed for this lookup (unlike the institutional CrossRef deposit/cited-by API the backend already talks to, which needs credentials).
- New `AdminSeqSetCitationsSection` is a small client island that owns the citation list state so add/remove feel immediate, without a full page reload.
- `admin/dashboard.astro` fetches and renders the new section below the pipeline statistics. The fetch is only issued when `seqSetsAreEnabled()`, and the section degrades gracefully on error.

## Tests

- Backend endpoint tests in `CitationEndpointsTest`:
  - returns **401**/**403** without auth / for a non-super user (for get, add, and delete),
  - returns an empty list when there are no citations,
  - returns each citation with its cited SeqSet metadata (accession version, name, DOI) after a simulated CrossRef run,
  - adding a citation for an existing SeqSet sets `origin = CURATED` and shows up in both the admin and per-SeqSet citation views,
  - adding a citation for a non-existent SeqSet returns 404,
  - re-adding the same `sourceDOI` with an additional SeqSet updates metadata and links both SeqSets,
  - deleting a curated citation removes it; deleting a `CROSSREF`-origin citation is rejected; deleting an unknown DOI returns 404.
- New Playwright spec `admin-citations.dependent.spec.ts`: logs in as the super user, creates a SeqSet, adds a citation through the admin UI, asserts it renders, deletes it via the UI, asserts it disappears.

## Verification

- `./gradlew ktlintFormat` and the full backend test suite (`./gradlew test`, 72 test classes) pass, real Postgres via Testcontainers.
- Website `npm run check-types`, `npm run format`, and `CI=1 npm run test` pass (683 unit tests).
- Manually exercised the full flow on the preview deployment as the seeded super user, including after the review-feedback fixes: created a SeqSet, added a curated citation (via "Fetch from DOI" autofill against a real CrossRef DOI), confirmed it appears in the admin table (now above the form), the SeqSet's own citations dialog, and the cited sequence's "Cited in" section, then removed it via the admin UI and confirmed it disappears everywhere. Also confirmed the Year-validation fix rejects a blank year instead of silently submitting `0`. Screenshots below.

### Admin dashboard: citations table above the add-citation form

Admin dashboard, empty citations table above the form

### Autofilling from a DOI via CrossRef

Typing a DOI and clicking "Fetch from DOI" populates title, year, and contributors from CrossRef's public API:

DOI autofill populated the form

### Citation appears in the table (above the form), with real CrossRef metadata and a Remove button

Citation added, table above form

### Citation also appears on the SeqSet's own citations dialog, and on the cited sequence's page

SeqSet page citations dialog

Sequence page Cited in section

🚀 Preview: https://feat-admin-seqset-citatio.loculus.org


theosanderson:
@codex review

backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCitationsDatabaseService.kt
```diff
@@ -585,6 +588,121 @@ class SeqSetCitationsDatabaseService(
...
+        if (missing.isNotEmpty()) {
+            throw NotFoundException(
+                "The following SeqSets do not exist: ${missing.joinToString(", ") { it.displayAccessionVersion() }}",
+            )
+        }
+
+        val citationSourceId = SeqSetCitationSourceTable
+            .batchUpsert(listOf(request.source), SeqSetCitationSourceTable.sourceDOI) {
+                this[SeqSetCitationSourceTable.sourceDOI] = it.sourceDOI
+                this[SeqSetCitationSourceTable.origin] = CitationOrigin.CURATED
```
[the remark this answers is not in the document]
theosanderson: Let's reject manual adds for CrossRef-owned DOIs for now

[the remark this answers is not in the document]
theosanderson-agent: Fixed in 360cdf5: `addCuratedCitation` now checks for an existing `CROSSREF`-origin row with the same `sourceDOI` and rejects the request with 422 rather than silently flipping it to `CURATED` (per the discussion below — reject rather than preserve-and-merge). Added a test (`WHEN adding a citation for a DOI already discovered via CrossRef THEN returns unprocessable entity`) covering it.

backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCitationsDatabaseService.kt
```diff
@@ -585,6 +588,121 @@ class SeqSetCitationsDatabaseService(
...
+
+        SeqSetToCitationSourceTable.batchInsert(accessionVersions, ignore = true) { accessionVersion ->
+            this[SeqSetToCitationSourceTable.citationSourceId] = citationSourceId
+            this[SeqSetToCitationSourceTable.seqSetId] = accessionVersion.accession
+            this[SeqSetToCitationSourceTable.seqSetVersion] = accessionVersion.version
+        }
+
+        return AdminSeqSetCitation(
+            source = request.source,
+            seqSets = accessionVersions.map { accessionVersion ->
```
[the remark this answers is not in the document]
theosanderson-agent: Fixed in 360cdf5: after the join-table insert, `addCuratedCitation` now re-queries `SeqSetToCitationSourceTable` joined with `SeqSetsTable` for everything linked to that `citationSourceId`, instead of building the response from only the SeqSets in the current request. Extended the existing re-add test to only list the *new* SeqSet on the second call (not re-listing the first), so it actually exercises this — it would have passed before the fix too, since it happened to list both each time.

website/src/components/AdminDashboard/AddSeqSetCitationForm.tsx
```diff
@@ -0,0 +1,254 @@
...
+        // eslint-disable-next-line @typescript-eslint/no-misused-promises
+        <form className='max-w-lg mt-4 mb-8' onSubmit={handleSubmit}>
+            <h3 className='text-lg font-semibold mb-2'>Manually add a citation</h3>
+            <div className='mb-4'>
+                <label htmlFor='citation-source-doi' className='block mb-1 text-sm font-medium text-gray-900'>
+                    * Source DOI
+                </label>
+                <div className='flex gap-2'>
+                    <input
+                        id='citation-source-doi'
```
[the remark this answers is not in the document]
theosanderson-agent: Fixed in 360cdf5: added `useClientFlag` and `disabled={!isClient}` on all five raw inputs/textareas in the form. The two buttons (Add citation, Fetch from DOI) already get this for free since `Button` wraps `DisabledUntilHydrated` internally.

website/src/components/AdminDashboard/AddSeqSetCitationForm.tsx
```diff
@@ -0,0 +1,254 @@
...
+        setValidationMessage('');
+
+        const seqSetAccessionVersions = splitAccessions(seqSetAccessionsInput);
+        const parsedYear = Number(year);
+
+        if (sourceDOI.trim() === '' || title.trim() === '') {
+            setValidationMessage('Source DOI and title are required.');
+            return;
+        }
+        if (!Number.isInteger(parsedYear)) {
```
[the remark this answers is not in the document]
theosanderson-agent: Good catch, thanks. Fixed in 360cdf5 with essentially your suggested change (`year.trim() === '' || !Number.isInteger(parsedYear)`).

website/src/services/backendClient.ts
tombch, line 192: Not introduced in this PR so could be addressed separately, but I noticed the error message for the citations form doesn't include the actual API problem detail, just a generic 'Failed to make request: Request failed with code 422' message. This could be updated to something like: 

```
        } catch (e) {
            const axiosError = e as AxiosError;

            const problemDetailResult = problemDetail.safeParse(axiosError.response?.data);
            if (problemDetailResult.success) {
                return err(problemDetailResult.data);
            }

            return err({
                type: 'about:blank',
                title: 'bad response',
                status: axiosError.response?.status ?? 0,
                detail: `Failed to make request: ${axiosError.message}`,
            });
        }
```

website/src/components/AdminDashboard/AddSeqSetCitationForm.tsx
```diff
@@ -0,0 +1,261 @@
...
+            return;
+        }
+        if (year.trim() === '' || !Number.isInteger(parsedYear)) {
+            setValidationMessage('Year must be a whole number.');
+            return;
+        }
+        if (seqSetAccessionVersions.length === 0) {
+            setValidationMessage('At least one cited SeqSet accession.version is required.');
+            return;
+        }
```
tombch, line 140: The 422 error I encountered was an `Invalid accession version format 'LOC_SS_1', expected 'accession.version` issue, so could add an additional validation here?

website/src/services/backendClient.ts
theosanderson: (Would prefer to leave anything not blocking until later)

website/src/types/seqSetCitation.ts
```diff
@@ -65,3 +68,25 @@ const sequenceCitation = z.object({
 export type SequenceCitation = z.infer<typeof sequenceCitation>;
 
 export const sequenceCitations = z.array(sequenceCitation);
+
+const citedSeqSet = z.object({
```
tombch: Not blocking but in future I might try to consolidate these, we have quite a few different but similar formats for the citations now - same for the API types in `backend/src/main/kotlin/org/loculus/backend/api/SeqSetCitations.kt`

website/src/components/AdminDashboard/AddSeqSetCitationForm.tsx
```diff
@@ -0,0 +1,261 @@
...
+        setValidationMessage('');
+
+        const seqSetAccessionVersions = splitAccessions(seqSetAccessionsInput);
+        const parsedYear = Number(year);
+
+        if (sourceDOI.trim() === '' || title.trim() === '') {
+            setValidationMessage('Source DOI and title are required.');
+            return;
+        }
+        if (year.trim() === '' || !Number.isInteger(parsedYear)) {
```
tombch, line 133: Also not blocking for an admin panel but negative years can currently be added

tombch (approved):
Looks good to me!

website/src/types/seqSetCitation.ts
theosanderson: @theosanderson-agent what consolidation options are there - reply to this thread

theosanderson-agent: Agreed there is duplication here. Mapping out the family first, because it makes the options clearer — every one of these types is `{ source: CitationSource, ...how it describes the cited SeqSets }`:

| Type | Used by | Per-SeqSet payload | Extra |
|---|---|---|---|
| `SeqSetCitation` | `GET /get-seqset-citations` | none — 1-field wrapper | |
| `SequenceCitation` | `GET /get-sequence-citations` | `{seqSetAccessionVersion, sequenceAccession}` | |
| `AdminSeqSetCitation` | admin list + add | `{seqSetAccessionVersion, name, seqSetDOI}` | `origin` |
| `AddSeqSetCitationRequest` | admin add body | bare accessionVersion strings | |
| `SeqSetCitationSource` | CrossRef sync only, never serialized | `seqSetDOIs: Set<String>` | |

So `CitationSource` is already the shared core, and the divergence is entirely in the leaf "cited SeqSet" type. Options, cheapest first:

**1. Get `SeqSetCitationSource` out of the API package (free, no API change).** It is not an API type at all — it only exists on the CrossRefService → `mergeCitationSources` → `updateCitationSourcesFromCrossRef` path and is never serialized. Moving it to `service/crossref` and naming it for what it is (`CrossRefMatch` or similar) removes one of the five "similar formats" from the public surface outright. It is also the only one that keys SeqSets by DOI rather than accessionVersion, which is part of why it reads as inconsistent.

**2. Kill `SeqSetCitation` (website-only version is non-breaking).** It is `{source}` and nothing else. Two flavours:
- *(a) wire-level:* have `/get-seqset-citations` return `List<CitationSource>` directly. Deletes the Kotlin type and the zod schema, but changes the response shape from `{source: {...}}` to `{...}` — breaking for any API consumer.
- *(b) website-only:* keep the wire shape, drop the distinct TS type. `CitationList`/`CitationTable`/`DataTable` currently thread a `SeqSetCitation | SequenceCitation` union around and narrow with `'seqSets' in citation`; they could just take a structural `{ source: CitationSource; seqSets?: CitingSeqSet[] }`. Cheap, no backend change, removes the union.

**3. Unify the leaf types — the real duplication.** `CitedSeqSet {seqSetAccessionVersion, name, seqSetDOI}` and `SeqSetCitingSequence {seqSetAccessionVersion, sequenceAccession}` are the same entity from two angles. Three ways:
- *Nullable superset:* one `CitedSeqSet(accessionVersion, name?, seqSetDOI?, sequenceAccession?)`. Least code, but which fields are populated becomes endpoint-dependent and undocumentable. I would avoid this one.
- *Compose:* `SequenceCitation.seqSets: List<{seqSet: CitedSeqSet, sequenceAccession: String}>`. Honest, one extra level of nesting.
- *Widen the sequence endpoint:* serve the full `CitedSeqSet` on both, with `sequenceAccession` as the sequence-endpoint-only extra. Nice side effect: the sequence details citation list currently shows only the bare accessionVersion, and this would hand it the SeqSet name and DOI for free. Breaking (additive-plus-reshape) for `/get-sequence-citations`.

**4. Parameterise rather than duplicate.** On the zod side this is easy and worth doing regardless: `const citationOf = <T extends ZodTypeAny>(seqSet: T) => z.object({ source: citationSource, seqSets: z.array(seqSet) })`, with admin as `.extend({ origin })`. On the Kotlin side, `data class Citation<T>(source, seqSets: List<T>)` I would *not* do — springdoc renders generics as synthetic schema names (`CitationCitedSeqSet` etc.) and the resulting OpenAPI contract is uglier than the duplication it saves.

**5. One wire type for everything.** A single `SeqSetCitation {source, seqSets: List<CitedSeqSet>, origin}` served by all three read endpoints, with `sequenceAccession` optional on the leaf. One Kotlin type, one zod schema, one component prop type. Costs: breaks both public endpoints; exposes `origin` publicly (arguably a feature — "discovered via CrossRef" is reasonable to show); and the per-SeqSet endpoint pays joins for data nothing renders.

One thing I would explicitly *not* consolidate: `AddSeqSetCitationRequest`. `name`, `seqSetDOI` and `origin` are all server-derived, so a request type that is a strict subset of the response type is correct rather than redundant.

**Suggested split:** 1 + 2(b) + 4(zod) are non-breaking, mechanical, and get most of the readability win — good candidate for a follow-up PR right after this one. 3 and 5 are the actual merge but change two public response shapes, so they want to be deliberate; the natural trigger is whenever the sequence details citation list wants SeqSet names/DOIs, since that needs the `CitedSeqSet` payload there anyway.

Happy to open a PR for the non-breaking subset if that would be useful.

theosanderson: Nice catch, thanks: https://github.com/loculus-project/loculus/pull/6769/commits/50b53ac4004264745a70ca3a1a1f9a6ca32f61bd

website/src/services/backendClient.ts
theosanderson: Hopefully addressed

tombch: Yeah I tested it, looks good to me! Now shows the detailed error message:

Image

Issue #6758, opened 2026-06-24
MVP: Raw reads prepro - ensure uploaded files are passed through prepro and validated


anna-parker:
We need to:
- ensure files are passed through (https://github.com/loculus-project/loculus/pull/6731)
- ensure files are in fastq format (reject if wrong format) (https://github.com/loculus-project/loculus/pull/6773) [PR only checks extension, could contain anything]
- ensure a max of 2 fastq files are uploaded per submission (paired or not paired) (https://github.com/loculus-project/loculus/pull/6773)
- validate that files do not contain human reads (reject if bede's tool flags has human data)

- Ensure fastq file is _submittable_ to ENA: For submission to ENA fastq need to satisfy a long list of requirements: https://ena-docs.readthedocs.io/en/latest/submit/fileprep/reads.html#fastq-format - currently we only check the extension, don't look into the file at all (this PR only checks extension: https://github.com/loculus-project/loculus/pull/6773)

corneliusroemer:
Best way to validate reads would be to run ENA's readtools code through a CLI and call that in prepro ourselves: https://github.com/enasequence/readtools/issues/23

Forking and adding a CLI entrypoint should be fairly straightforward.

PR #6773, opened 2026-06-26
feat(prepro): basic validation of raw read submissions

maverbiest:
Partially resolves https://github.com/loculus-project/loculus/issues/6758

This PR adds general functionality to validate user-submitted files in the nextclade preprocessing pipeline. How files are processed is determined based on the `FileCategory` and a simple switch statement.

The PR introduces a function to validate a submission of raw sequencing reads as a first use case:

```python
def validate_raw_reads_submission(
    files: list[FileIdAndName],
) -> tuple[list[ProcessingAnnotation], list[ProcessingAnnotation]]:
```

The validation is very basic for the moment, and only considers the number of submitted files and their files extensions. This could be built out in the future to include more robust checks.

## Alternative approaches
Rather than having a switch statement on `FileCategory` in prepro to determine how different files are processed, we could define this in the `values.yaml`, similar to how we do for metadata fields. E.g.:

```yaml
    submissionDataTypes:
      files:
        enabled: true
        categories:
          - name: raw_reads
            displayName: Raw reads
            preprocessing:
              function: validate_raw_reads_submission
          ...
```

I decided against this for now to get a working version with a minimal diff, but I would be happy to implement this if people feel it's better.

### PR Checklist

🚀 Preview: https://simple-fq-validation.loculus.org

### DIFF
--- a/preprocessing/dummy/main.py
+++ b/preprocessing/dummy/main.py
@@ -129,7 +129,7 @@ def parse_ndjson(ndjson_data: str) -> list[Sequence]:
 
 
 def process(unprocessed: list[Sequence]) -> list[Sequence]:
-    with open("mock-sequences.json", "r") as f:
+    with open("mock-sequences.json") as f:
         mock_sequences = json.load(f)
     possible_lineages = ["A.1", "A.1.1", "A.2"]
 
@@ -160,7 +160,7 @@ def process(unprocessed: list[Sequence]) -> list[Sequence]:
             "nucleotideInsertions": {},
             "aminoAcidInsertions": {},
         }
-        
+
         if not disableConsensusSequences:
             data = {**data, **mock_sequences}
             data["sequenceNameToFastaId"] = {"main": submissionId}
--- a/preprocessing/nextclade/src/loculus_preprocessing/datatypes.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/datatypes.py
@@ -37,7 +37,7 @@ class FileCategory(StrEnum):
 class AnnotationSourceType(StrEnum):
     METADATA = "Metadata"
     NUCLEOTIDE_SEQUENCE = "NucleotideSequence"
-    SUBMITTED_FILE = "SubmittedFile"
+    FILE = "File"
 
 
 @dataclass(frozen=True)
--- a/preprocessing/nextclade/src/loculus_preprocessing/datatypes.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/datatypes.py
@@ -28,8 +28,9 @@ ProcessingAnnotationAlignment: Final = "alignment"
 
 
 @unique
-class SubmissionFileCategory(StrEnum):
+class FileCategory(StrEnum):
     RAW_READS = "raw_reads"
+    ANNOTATIONS = "annotations"
 
 
 @unique
@@ -94,7 +95,7 @@ class UnprocessedData:
     submissionId: str  # noqa: N815
     metadata: InputMetadata
     unalignedNucleotideSequences: dict[SequenceName, NucleotideSequence | None]  # noqa: N815
-    files: dict[SubmissionFileCategory, list[FileIdAndName]] | None
+    files: dict[FileCategory, list[FileIdAndName]] | None
 
 
 @dataclass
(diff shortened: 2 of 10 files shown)


preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -436,6 +438,27 @@ def get_output_metadata(
...
+    for category, files in file_mapping.items():
+        if not files:
+            # Backend always includes a key with empty list for enabled categories
+            continue
+        match category:
+            case FileCategory.RAW_READS:
+                rr_errors, rr_warnings = validate_raw_reads_submission(files)
+                errors.extend(rr_errors)
+                warnings.extend(rr_warnings)
+            case _:
```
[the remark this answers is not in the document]
maverbiest: Yes currently when we add a new input file category, this match statement will need to be updated to handle it. That is as intended (see discussion of alternative in the PR description)

preprocessing/dummy/main.py
maverbiest, line 1: This file caught some stray updates from running the formatter

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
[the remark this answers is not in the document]
tombch: I'm wondering if this should emit a hard exception, as (if my understanding is correct) a file category could be added to the `values.yaml`, and so will be accepted in the backend, but then if its missing in this match statement, it would pass through without validation?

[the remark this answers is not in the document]
tombch: I guess a hard exception would prevent other categories being allowed for other users of loculus which might not be a good thing

tombch (approved):
Tested it out and this looks good to me!

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -436,6 +438,27 @@ def get_output_metadata(
     return output_metadata, errors, warnings
 
 
+def process_submitted_files(
```
corneliusroemer: Maybe worth moving the file related functions (this one and validate raw reads) to a new file to not grow this 2k line monstrosity even further

preprocessing/nextclade/src/loculus_preprocessing/datatypes.py
```diff
@@ -23,15 +23,21 @@
...
+class FileCategory(StrEnum):
+    RAW_READS = "raw_reads"
+    ANNOTATIONS = "annotations"
+
+
 @unique
 class AnnotationSourceType(StrEnum):
     METADATA = "Metadata"
     NUCLEOTIDE_SEQUENCE = "NucleotideSequence"
+    SUBMITTED_FILE = "SubmittedFile"
```
anna-parker: I think this name is a bit confusing as it implies this is a file submitted by the user but annotations are created by the prepro pipeline, I think just "File" is ok

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -1948,6 +1949,39 @@ def single_metadata_annotation(
...
 
 
+def validate_raw_reads_submission(
+    files: list[FileIdAndName],
+) -> tuple[list[ProcessingAnnotation], list[ProcessingAnnotation]]:
+    errors: list[ProcessingAnnotation] = []
+    warnings: list[ProcessingAnnotation] = []
+
+    if len(files) > 2:  # noqa: PLR2004
+        message = f"Raw reads must be submitted as one or two files, got {len(files)}"
```
anna-parker: This is a message for submitters so I would actually state we want to have paired-end or single-end raw reads, and thus accept a max of 2 files.

anna-parker (approved):
some small improvements but also looks good overall!

anna-parker:
Regarding the alternative suggestion, I like it but I would hold off for now - its not required for the beta and we can make the config a bit nicer in a later step, you can create an issue for this though to track it :-)

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
maverbiest: I moved them into a new file called `file_processing_functions.py`

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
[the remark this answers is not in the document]
maverbiest: Yes you're right, it's better to have an explicit match arm for each configured category, even if we'd then decide to just pass those files through without validation.

The `case _:` arm now adds an error message so that submissions won't go through

PR #6774, opened 2026-06-29
feat!(prepro): use submissionId for displayName if specimenCollectorSampleId cannot be parsed

### LINKED ISSUE #6763: Have build_display_name fall back to submissionId if specimenCollectorSampleId cannot be parsed
Preprocessing has a function `build_display_name`

https://github.com/loculus-project/loculus/blob/352c1c7c9b4e84b1fc175e800721d6ca9796be5a/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py#L1573-L1578

We use this to incorporate an identifier field in our displayNames that is populated from either the submissions `specimenCollectorSampleId` (isolate name) or - if this is absent - the `submissionId`. If this identifier is in a format that we cannot parse, we fall back to using the `accessionVersion` instead. 

```md
# Option A: `specimencollectorSampleId` exists
specimencollectorSampleId -- succesfully parsed --> use in displayName
                 | 
            failed to parse --> use accessionVersion

# Option B: `specimencollectorSampleId` does not exist
submissionId -- succesfully parsed --> use in displayName
         | 
    failed to parse --> use accessionVersion
```

In this model, we never look at the `submissionId` if `specimencollectorSampleId` is present but we fail to parse it, instead we fall back immediately to the `accessionVersion`. This is unfortunate as the `submissionId` may actually be more informative than the `accessionVersion`. 

We could instead fall back to the `submissionId` first if we fail to parse the `specimenCollectorSampleId`, so the two flows become:
```md
# Option A: `specimencollectorSampleId` exists
specimencollectorSampleId -- succesfully parsed --> use in displayName
                 | 
            failed to parse --> submissionId -- succesfully parsed --> use in displayName
                                   | 
                              failed to parse --> use accessionVersion

# Option B: `specimencollectorSampleId` does not exist
submissionId -- succesfully parsed --> use in displayName
         | 
    failed to parse --> use accessionVersion
```

maverbiest:
resolves #6763

`build_display_name` now attempts to use the `submissionId` as the identifier field in displayName when `specimenCollectorSampleId` is present but cannot be parsed (see extended description in the issue). It also introduces the `human_readable_pattern` arg to `build_display_name` (see breaking changes below)

Additionally, this PR extracts the identifier string parsing logic into its own function.

## Breaking changes

If you configure the `build_display_name` preprocessing function with a `regex_pattern`, you must now also provide a `human_readable_pattern`. This is a human-readable representation of the format the `regex_pattern` parses and will be shown to users in a warning message if the regex parsing fails.

```yaml
        - name: displayName
          ...
          preprocessing:
            function: build_display_name
            inputs:
              ...
            args:
              ...
              regex_pattern: '^(?:[^/]+/)?[^/]+/(?P<identifier>[^/]+)/\d{4}(?:-\d{2}){0,2}$'
              human_readable_pattern: "<any>/<any>/<identifier>/<date>"
```

## Screenshots

Three different cases showing new fallback behaviour:

1. specimenCollectorSampleId 'INYO_2_2011' does not need to be parsed and is used director
2. No isolate name -> fall back to submissionId, which is 'test_WNV/Bakersfield_CA_USA/W417/2010-07-08' -> parse this and pull out the identifier 
3. No isolate name -> fall back to submissionId, which is 'test_WNV/mosquito/Larimer-CO-USA/2016-06-27/W1819' -> cannot be parsed, show new warning message with supported format for regex

grafik

## Examples of less nice displayNames

I tested this new approach of generating displayNames on all sequences currently in Pathoplexus. Overall it looks good and my opinion is that this is an improvement over the current state. However, there are instances where the new displayNames are less nice. These broadly fall into three classes from what I've seen:

1) **Redundancy in displayName (examples PP_006XBKH.2, PP_000UMND.2)**
    - old: `France/PP_006XBKH.2/2026-05-10`, new: `France/26-0090|France|Homo_sapiens|2026-05-10/2026-05-10`
    - old: `USA/PP_000UMND.2/2022-07`, new: `USA/MPXV_USA_2022_NJ0016/2022-07`
2) **Uninformative/short identifier (example PP_003HM09.1)**
    - old: `Gabon/PP_003HM09.1/2017`, new: `Gabon/3/2017`
3) **Just adding `unknown` (example PP_0021GXK.1)**
    - old: `PP_0021GXK.1`, new: `unknown/PP_0021GXK.1/unknown`

To address 1), we'd have to alter our regex or extend our set of forbidden characters to either start parsing these types of identifiers (hard, esp. if there's multiple separators in the same identifer) or to reject them and just use accessionVersion (which has the downside of rejecting good identifiers). For 2) we could consider having a minimum length for identifier fields. For 3) we could consider not adding the `unknown` placeholder, but in general I think it's nice that we now have a consistent meaning for each position in the displayName for each sequence. All this does not mean we can't upate/tweak displayName creation in the future, but for now I think this is a good starting point

### PR Checklist

### DIFF
--- a/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
@@ -1181,21 +1181,17 @@ class ProcessingFunctions:
             concatenate_field_types = replace_identifier(field_types, "string")
             input_data["IDENTIFIER"] = identifier
         else:
-            # Unable to parse specimenCollectorSampleId or submissionID, use ACCESSION_VERSION
+            # Unable to parse specimenCollectorSampleId and submissionID, use ACCESSION_VERSION
             if not insdc_ingested and regex_pattern is not None:
-                failed_source = (
-                    collector_id
-                    if isinstance(collector_id, str) and collector_id
-                    else submission_id
-                )
                 warnings.append(
                     ProcessingAnnotation.from_fields(
                         input_fields,
                         [output_field],
                         AnnotationSourceType.METADATA,
                         message=(
-                            f"identifier string '{failed_source}' could not be parsed,"
-                            " using ACCESSION_VERSION in displayName instead"
+                            f"specimencollectorSampleId '{collector_id}' and submissionId"
+                            f" '{submission_id}' could not be parsed, using ACCESSION_VERSION"
+                            f" in displayName instead"
                         ),
                     )
                 )
@@ -1580,7 +1576,7 @@ def process_phenotype_values(input: str | None, args: FunctionArgs | None) -> In
 def parse_identifier_string(
     input: ProcessedMetadataValue, insdc_ingested: bool, regex_pattern: str | None = None
 ) -> str | None:
-    """Return a IDENTIFIER string to use in the displayName or None if `input` cannot be used
+    """Return an IDENTIFIER string to use in the displayName or None if `input` cannot be used
     as an identifier.
     """
     if not isinstance(input, str):
--- a/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
@@ -1196,7 +1196,7 @@ class ProcessingFunctions:
                 warnings.append(
                     f"specimenCollectorSampleId and submissionId could not be parsed, using "
                     f"ACCESSION_VERSION in displayName instead. To include your own identifier, "
-                    f"remove whitespace and '/' characters or use format '{human_readable_pattern}'"
+                    f"remove whitespace and '/' characters or use the format '{human_readable_pattern}' and we will parse the `identifier` from the submission."
                 )
             concatenate_order = replace_identifier(order, "ACCESSION_VERSION")
             concatenate_field_types = replace_identifier(field_types, "ACCESSION_VERSION")
--- a/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
@@ -1114,7 +1114,7 @@ class ProcessingFunctions:
         return RawProcessingResult()
 
     @staticmethod
-    def build_display_name(  # noqa: C901
+    def build_display_name(
         input_data: InputMetadata,
         output_field: str,
         input_fields: list[str],
@@ -1127,12 +1127,13 @@ class ProcessingFunctions:
         args, as well as adding some additional checks and requirements:
             - submissionId and specimenCollectorSampleId must be in the input_data
             - IDENTIFIER keyword must be in args['order'] and args['type']
-            - if the IDENTIFIER is in an unrecognized format, it will be replaced with the ACCESSION_VERSION
-            - if fallback_value is not in args, { 'fallback_value': 'unknown' } is added to the args before passing
-              them on to concatenate()
-            - for sequences ingested from INSDC, we do not try to parse the IDENTIFIER field using regex. We
-              will use the Isolate Name as IDENTIFIER field if it contains no slashes or spaces (otherwise we fall back to
-              ACCESSION_VERSION)
+            - the IDENTIFIER is resolved by trying specimenCollectorSampleId first, then
+              submissionId; if neither yields a usable value it is replaced with ACCESSION_VERSION
+            - if fallback_value is not in args, { 'fallback_value': 'unknown' } is added to the args
+              before passing them on to concatenate()
+            - for sequences ingested from INSDC, we do not try to parse the IDENTIFIER field using
+              regex. We will use the Isolate Name as IDENTIFIER field if it contains no slashes or
+              spaces (otherwise we fall back to ACCESSION_VERSION)
         """
         collector_id = input_data.get("specimenCollectorSampleId", None)
         submission_id = input_data.get("submissionId", None)
@@ -1167,40 +1168,39 @@ class ProcessingFunctions:
         def replace_identifier(values, replacement):
             return [replacement if v == "IDENTIFIER" else v for v in values]
 
-        identifier: ProcessedMetadataValue = collector_id or submission_id
-        if not isinstance(identifier, str):
-            identifier = None
-        elif args["is_insdc_ingest_group"]:
-            # For INSDC ingested sequence: use ID as is unless it contains ' ' or '/'
-            # If it does: fall back to ACCESSION_VERSION
-            if " " in identifier or "/" in identifier:
-                identifier = None
-        elif "/" in identifier:
-            # For direct submissions with "/": try to extract ID field using regex
-            if regex_pattern is None:
-                identifier = None
-            else:
-                extract_result = ProcessingFunctions.extract_regex(
-                    input_data={"regex_field": identifier},
-                    output_field="IDENTIFIER",
-                    input_fields=[],
-                    args={"pattern": regex_pattern, "capture_group": "identifier"},
-                )
-                if extract_result.datum is None:
-                    # regex extraction of ID field failed, fall back to ACCESSION_VERSION
-                    warnings.append(
-                        f"identifier string '{identifier}' could not be parsed, using ACCESSION_VERSION in displayName instead"
-                    )
-                identifier = extract_result.datum
+        insdc_ingested = bool(args["is_insdc_ingest_group"])
+        pattern = str(regex_pattern) if regex_pattern is not None else None
 
+        # Try specimenCollectorSampleId first, fall back to submissionId.
+        identifier = parse_identifier_string(collector_id, insdc_ingested, pattern)
         if identifier is None:
-            # Use ACCESSION_VERSION instead of IDENTIFIER
+            identifier = parse_identifier_string(submission_id, insdc_ingested, pattern)
+
+        if identifier is not None:
+            # We were able to parse an IDENTIFIER, treat it as a string
+            concatenate_field_types = replace_identifier(field_types, "string")
+            input_data["IDENTIFIER"] = identifier
+        else:
+            # Unable to parse specimenCollectorSampleId or submissionID, use ACCESSION_VERSION
+            if not insdc_ingested and regex_pattern is not None:
+                failed_source = (
+                    collector_id
+                    if isinstance(collector_id, str) and collector_id
+                    else submission_id
+                )
+                warnings.append(
+                    ProcessingAnnotation.from_fields(
+                        input_fields,
+                        [output_field],
+                        AnnotationSourceType.METADATA,
+                        message=(
+                            f"identifier string '{failed_source}' could not be parsed,"
+                            " using ACCESSION_VERSION in displayName instead"
+                        ),
+                    )
+                )
             concatenate_order = replace_identifier(order, "ACCESSION_VERSION")
             concatenate_field_types = replace_identifier(field_types, "ACCESSION_VERSION")
-        else:
-            # Keep IDENTIFIER but treat it as string
-            concatenate_field_types = replace_identifier(field_types, "string")
-            input_data["IDENTIFIER"] = str(identifier)
 
         new_args = args.copy()
         new_args.update(
@@ -1575,6 +1575,36 @@ def process_phenotype_values(input: str | None, args: FunctionArgs | None) -> In
             ),
         )
     return InputData(datum=None)
+
+
+def parse_identifier_string(
+    input: ProcessedMetadataValue, insdc_ingested: bool, regex_pattern: str | None = None
+) -> str | None:
+    """Return a IDENTIFIER string to use in the displayName or None if `input` cannot be used
+    as an identifier.
+    """
+    if not isinstance(input, str):
+        return None
+    has_forbidden_char = " " in input or "/" in input
+
+    if insdc_ingested:
+        # For INSDC ingested sequences: use the value as-is unless it contains a ' ' or '/'
+        return None if has_forbidden_char else input
+
+    if not has_forbidden_char:
+        # Direct submission without forbidden_char: use the value as-is, no regex parsing
+        return input
+
+    # Direct submission containing forbidden_char: attempt regex extraction of identifier field
+    if regex_pattern is None:
+        return None
+    extract_result = ProcessingFunctions.extract_regex(
+        input_data={"regex_field": input},
+        output_field="IDENTIFIER",
+        input_fields=[],
+        args={"pattern": regex_pattern, "capture_group": "identifier"},
+    )
+    return None if extract_result.datum is None else str(extract_result.datum)
 
 
 def trim_ns(sequence: str) -> str:
--- a/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
@@ -1120,8 +1120,8 @@ class ProcessingFunctions:
         input_fields: list[str],
         args: FunctionArgs,
     ) -> RawProcessingResult:
-        """Builds a displayName from input_fields. The identifier field in the displayName is based on
-        specimenCollectorSampleId or - if it is not set - submissionId.
+        """Builds a displayName from input_fields. The identifier field in the displayName is based
+        on specimenCollectorSampleId or - if it is not set - submissionId (direct submissions only).
 
         This method wraps ProcessingFunctions.concatenate(). Thus, it has the same required input
         args, as well as adding some additional checks and requirements:
@@ -1138,8 +1138,6 @@ class ProcessingFunctions:
         collector_id = input_data.get("specimenCollectorSampleId", None)
         submission_id = input_data.get("submissionId", None)
         warnings: list[str] = []
-        if submission_id is None:
-            return raw_internal_error("'submissionId' must not be None for build_display_name().")
 
         order = args.get("order")
         field_types = args.get("type")
@@ -1163,9 +1161,6 @@ class ProcessingFunctions:
         concatenate_order = order.copy()
         concatenate_field_types = field_types.copy()
 
-        def replace_identifier(values, replacement):
-            return [replacement if v == "IDENTIFIER" else v for v in values]
-
         insdc_ingested = bool(args["is_insdc_ingest_group"])
 
         # Try to parse the specimenCollectorSampleId first
@@ -1173,6 +1168,9 @@ class ProcessingFunctions:
         if identifier is None and not insdc_ingested:
             # For direct submissions only: try to parse the submissionId
             identifier = parse_identifier_string(submission_id, insdc_ingested, regex_pattern)
+
+        def replace_identifier(values, replacement):
+            return [replacement if v == "IDENTIFIER" else v for v in values]
 
         if identifier is not None:
             # We were able to parse an IDENTIFIER, treat it as a string
--- a/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
@@ -1134,6 +1134,9 @@ class ProcessingFunctions:
             - for sequences ingested from INSDC, we do not try to parse the IDENTIFIER field using
               regex. We will use the Isolate Name as IDENTIFIER field if it contains no slashes or
               spaces (otherwise we fall back to ACCESSION_VERSION)
+            - if regex_pattern is provided, human_readable_pattern must also be provided. It is a
+              submitter-friendly rendering of the pattern to show in error messages
+              (e.g. '<any>/<any>/<identifier>/<date>')
         """
         collector_id = input_data.get("specimenCollectorSampleId", None)
         submission_id = input_data.get("submissionId", None)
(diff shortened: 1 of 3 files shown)


maverbiest:
@claude review

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -1658,47 +1661,35 @@ def build_display_name(  # noqa: C901
...
-                            [output_field],
-                            AnnotationSourceType.METADATA,
-                            message=(
-                                f"identifier string '{identifier}' could not be parsed, using ACCESSION_VERSION in displayName instead"
-                            ),
-                        )
-                    )
-                identifier = extract_result.datum
+        insdc_ingested = bool(args["is_insdc_ingest_group"])
+        pattern = str(regex_pattern) if regex_pattern is not None else None
```
anna-parker: this shouldnt be needed you should be able to just use regex_pattern - it is of type string| None

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -1658,47 +1661,35 @@ def build_display_name(  # noqa: C901
         def replace_identifier(values, replacement):
```
anna-parker: maybe we could move this helper function to where it is used

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -1570,40 +1570,43 @@ def assign_custom_lineage(  # noqa: C901
...
+            - if fallback_value is not in args, { 'fallback_value': 'unknown' } is added to the args
+              before passing them on to concatenate()
+            - for sequences ingested from INSDC, we do not try to parse the IDENTIFIER field using
+              regex. We will use the Isolate Name as IDENTIFIER field if it contains no slashes or
+              spaces (otherwise we fall back to ACCESSION_VERSION)
         """
         collector_id = input_data.get("specimenCollectorSampleId", None)
         submission_id = input_data.get("submissionId", None)
         warnings: list[ProcessingAnnotation] = []
         if submission_id is None:
```
anna-parker: tbf we can actually probably drop this check, the build_display_name uses "ACCESSION_VERSION" as the ultimate fallback value - if anything we could throw an error if "ACCESSION_VERSION" isnt passed by this would be an internal preprocessing error which is handled.

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -1658,47 +1661,35 @@ def build_display_name(  # noqa: C901
...
+            if not insdc_ingested and regex_pattern is not None:
+                warnings.append(
+                    ProcessingAnnotation.from_fields(
+                        input_fields,
+                        [output_field],
+                        AnnotationSourceType.METADATA,
+                        message=(
+                            f"specimencollectorSampleId '{collector_id}' and submissionId"
+                            f" '{submission_id}' could not be parsed, using ACCESSION_VERSION"
+                            f" in displayName instead"
```
corneliusroemer: As this shows to users it should be clear for users what the problem is and how to fix it - direct submitters can then potentially fix it. Ideally:
- why it didn't parse
- what format is legal, maybe link to some docs

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -2173,6 +2164,36 @@ def process_phenotype_values(input: str | None, args: FunctionArgs | None) -> In
...
 
+def parse_identifier_string(
+    input: ProcessedMetadataValue, insdc_ingested: bool, regex_pattern: str | None = None
+) -> str | None:
+    """Return an IDENTIFIER string to use in the displayName or None if `input` cannot be used
+    as an identifier.
+    """
+    if not isinstance(input, str):
+        return None
+    has_forbidden_char = " " in input or "/" in input
```
corneliusroemer: I don't remember: Might this contain whitespace other than space or do we sanitize this in backend already? Things like tab, newline

corneliusroemer: Should we also forbid `|` ? Why only ` ` and `/`?

Sometimes `|` is used causing things like `Argentina/Argentina|XYZ|2026/2026`

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -2173,6 +2164,36 @@ def process_phenotype_values(input: str | None, args: FunctionArgs | None) -> In
...
+
+    if insdc_ingested:
+        # For INSDC ingested sequences: use the value as-is unless it contains a ' ' or '/'
```
corneliusroemer: Why don't we parse INSDC ingested? Was this found to be problematic?

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
theosanderson: Unsure specifically why we have an issue with ` `. Our aim is not to avoid ever having redundancy, which I guess is why `|` is allowed, but yeah one can argue that the fact that this will mess up FASTA parsing is an argument to at least sanitise it downstream (and maybe same with `/` (and ` `)  instead of banning it`).

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
anna-parker: I think the order of fields was just so all over the place that we decided to just not try to extract the submissionId using regex

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
anna-parker: we dont want to have " " in the displayName, but I guess that can always be replaced with "_" in a str.replace() step.

I think @maverbiest tried parsing "|" as well at some point, I cant recall why it wasnt added (maybe bad results), I also would have thought it should be included

anna-parker: good point! LAPIS allows us to download tsv with `TSV-ESCAPED` but in my understanding the backend doesnt do this!

maverbiest: `submissionId` goes through `getValueAndValidateNoWhitespace` in the backend, which should guard against all whitespace:

https://github.com/loculus-project/loculus/blob/36b77c971aeec6b0e9611ad65f77db93caafc428/backend/src/main/kotlin/org/loculus/backend/utils/MetadataEntry.kt#L99-L104

But I don't think the `specimenCollectorSampleId` goes through this. Might be safer to check for all whitespace (although I haven't seen this in practice yet)

maverbiest: Indeed we didn't want " " in the displayName. My recollection if that we only check for '/' because that's the separator we use in the displayName ourselves. We didn't include other separators because we couldn't decide on where to draw the line (`|`, `-`, `_`, ..., also what do we do if there's multpile types?)

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
maverbiest: This parsing was intended as a service to direct submitters so they wouldn't have to reformat their identifiers. E.g., if someone already formatted their identifiers to submit to a different DB that expects a certain format, we could try to parse out the relevant field to use

(also INSDC was a mess)

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
maverbiest: `regex_pattern` is of type `ArgValue` which is very broad:

https://github.com/loculus-project/loculus/blob/36b77c971aeec6b0e9611ad65f77db93caafc428/preprocessing/nextclade/src/loculus_preprocessing/datatypes.py#L17-L19

So we do need to narrow to str or None first

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
maverbiest: I extended the message to say submitters can remove whitespace and '/' so we can incorporate the identifier in the displayName.

I decided against showing the regex pattern in the warning message as that may not be useful to most users, and I don't think we have any docs in Loculus for this (we do in PPX). So hopefully the updated warning is clear enough now

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -1153,54 +1152,42 @@ def build_display_name(  # noqa: C901
...
 
         concatenate_order = order.copy()
         concatenate_field_types = field_types.copy()
 
+        insdc_ingested = bool(args["is_insdc_ingest_group"])
+
+        # Try to parse the specimenCollectorSampleId first
+        identifier = parse_identifier_string(collector_id, insdc_ingested, regex_pattern)
+        if identifier is None and not insdc_ingested:
+            # For direct submissions only: try to parse the submissionId
```
anna-parker, line 1181: maybe add an explanation that the submissionId is just the concatenated INSDC nuccore accession for sequences ingested from the INSDC

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
anna-parker: lets add a comment that the of INSDC isolateNames is not good enough to parse using regex

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
anna-parker: c.isspace() should do the trick

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -1153,54 +1152,42 @@ def build_display_name(  # noqa: C901
...
+            # We were able to parse an IDENTIFIER, treat it as a string
+            concatenate_field_types = replace_identifier(field_types, "string")
+            input_data["IDENTIFIER"] = identifier
+        else:
+            # Unable to parse specimenCollectorSampleId and submissionID, use ACCESSION_VERSION
+            if not insdc_ingested and regex_pattern is not None:
+                warnings.append(
+                    f"specimenCollectorSampleId '{collector_id}' and submissionId"
+                    f" '{submission_id}' could not be parsed, using ACCESSION_VERSION"
+                    f" in displayName instead. Alternatively, you may edit the"
```
anna-parker: this isnt fully true you can also edit the specimenCollectorSampleId to fit the expected regex format, I think we should add this to the warning, we could offer a "human readable" example: "`{geoLocCountry}/{identifier}/{sampleCollectionDate}`" e.g. `Colombia/iso34-500/2012-05`

preprocessing/nextclade/tests/test_metadata_processing_functions.py
```diff
@@ -1174,238 +1176,256 @@ def test_parse_date_into_range() -> None:
...
+        input_fields=["date", "country"],
+        concatenate_args={
+            "ACCESSION_VERSION": "accession.1",
+            "order": ["date", "country"],
+            "type": ["dateRangeString", "string"],
+        },
+        expected="2021-01-01_TO_2021-12-31/USA",
+    ),
+    ConcatenateCase(
+        name="empty_int_field_skipped",
```
anna-parker: both country and someInt are skipped (or even dropped)
```suggestion
        name="empty_fields_at_start_dropped",
```

preprocessing/nextclade/tests/test_metadata_processing_functions.py
```diff
@@ -1174,238 +1176,256 @@ def test_parse_date_into_range() -> None:
...
+        input_fields=["geoLocCountry", "sampleCollectionDate"],
+        concatenate_args={
+            "ACCESSION_VERSION": "accession.1",
+            "order": ["someInt", "geoLocCountry", "ACCESSION_VERSION", "sampleCollectionDate"],
+            "type": ["integer", "string", "ACCESSION_VERSION", "date"],
+        },
+        expected="accession.1/2025-01-01",
+    ),
+    ConcatenateCase(
+        name="present_int_field_and_empty_string_field_with_no_fallback",
```
anna-parker:
```suggestion
        name="empty_field_in_middle_with_no_fallback_results_is_not_dropped",
```

preprocessing/nextclade/tests/test_metadata_processing_functions.py
```diff
@@ -1174,238 +1176,256 @@ def test_parse_date_into_range() -> None:
...
+        input_fields=["geoLocCountry", "sampleCollectionDate"],
+        concatenate_args={
+            "ACCESSION_VERSION": "accession.1",
+            "order": ["someInt", "geoLocCountry", "ACCESSION_VERSION", "sampleCollectionDate"],
+            "type": ["integer", "string", "ACCESSION_VERSION", "date"],
+        },
+        expected="0//accession.1/2025-01-01",
+    ),
+    ConcatenateCase(
+        name="empty_string_field_uses_fallback_value",
```
anna-parker:
```suggestion
        name="empty_field_uses_fallback_value",
```

anna-parker: this should still work for int I think?

preprocessing/nextclade/tests/test_metadata_processing_functions.py
```diff
@@ -1174,238 +1176,256 @@ def test_parse_date_into_range() -> None:
...
+        concatenate_args={
+            "ACCESSION_VERSION": "accession.1",
+            "order": ["someInt", "geoLocCountry", "ACCESSION_VERSION", "sampleCollectionDate"],
+            "type": ["integer", "string", "ACCESSION_VERSION", "date"],
+            "fallback_value": "unknown",
+        },
+        expected="0/unknown/accession.1/2025-01-01",
+    ),
+    ConcatenateCase(
+        name="accession_version_omitted_from_order",
```
anna-parker:
```suggestion
        name="concatenate_only_uses_fields_in_list",
```

preprocessing/nextclade/tests/test_metadata_processing_functions.py
```diff
@@ -1174,238 +1176,256 @@ def test_parse_date_into_range() -> None:
...
+        input_fields=["geoLocCountry", "sampleCollectionDate"],
+        concatenate_args={
+            "ACCESSION_VERSION": "accession.1",
+            "order": ["someInt", "geoLocCountry", "sampleCollectionDate"],
+            "type": ["integer", "string", "date"],
+            "fallback_value": "unknown",
+        },
+        expected="0/unknown/2025-01-01",
+    ),
+    ConcatenateCase(
```
anna-parker: I would actually drop this and in the top test with fallback value ensure it can be added for int, date and string all in one test

preprocessing/nextclade/tests/test_metadata_processing_functions.py
```diff
@@ -1174,238 +1176,256 @@ def test_parse_date_into_range() -> None:
...
+    "specimenCollectorSampleId",
+    "submissionId",
+    "sampleCollectionDate",
+]
+base_args: FunctionArgs = {
+    "ACCESSION_VERSION": "accession.1",
+    "is_insdc_ingest_group": False,
+    "order": ["nextclade.clade", "geoLocCountry", "IDENTIFIER", "sampleCollectionDate"],
+    "type": ["string", "string", "IDENTIFIER", "string"],
+    "regex_pattern": r"^[^\/][^/]*/[^/]+/(?P<identifier>[^/]+)/\d{4}(?:-\d{2}){0,2}$",
```
anna-parker: this is confusing - it doesnt use the same regex we use in values.yaml. I would use the same regex as used there

preprocessing/nextclade/tests/test_metadata_processing_functions.py
```diff
@@ -1174,238 +1176,256 @@ def test_parse_date_into_range() -> None:
...
+        specimen_collector_id="hDENV1/myExtractedSample/2025",
+        submission_id="mySample",
+        geo_loc_country="Switzerland",
+        expected_regular="DENV-1/Switzerland/mySample/2025",
+        expected_insdc="DENV-1/Switzerland/accession.1/2025",  # INSDC skips submissionId
+        expected_prefix="hYF/Switzerland/mySample/2025",
+    ),
+    DisplayNameCase(
+        name="both_ids_unparseable_empty_country_uses_accession_version",
+        specimen_collector_id="hDENV1/myExtractedSample/2025",
```
anna-parker: I would use an example that is also unparsable given the regex that we have

preprocessing/nextclade/tests/test_metadata_processing_functions.py
```diff
@@ -1174,238 +1176,256 @@ def test_parse_date_into_range() -> None:
...
+        submission_id="hDENV1/myExtractedSample/2025",
+        geo_loc_country="",
+        expected_regular="DENV-1/unknown/accession.1/2025",
+        expected_insdc="DENV-1/unknown/accession.1/2025",
+        expected_prefix="hYF/unknown/accession.1/2025",
+        warning_regular=unparseable_identifier_warning,
+        warning_prefix=unparseable_identifier_warning,
+    ),
+    DisplayNameCase(
+        name="fallback_value_replaces_unknown_country_when_ids_unparseable",
```
anna-parker: this name is misleading the fallback values should be replaced for country even when the ids are parsable these are unrelated

anna-parker (commented):
much easier to read but I made a couple of comments to further improve test readability but this is a significant improvement, thanks!

preprocessing/nextclade/tests/test_metadata_processing_functions.py
maverbiest: You're right these last two test cases are totally nonsensical, guess I didn't look close enough at the LLM generated test cases when reviewing. Thanks for flagging!

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
maverbiest: I agree that showing the human readable format would be much nicer, but the regex could be configured differently for different organsims (even though we don't have this at the moment). Won't that be an issue?

anna-parker: I would make this configurable so that it comes in from the config

maverbiest: That might be the cleanest solution actually, passing a human-readable version to use in the message through a `supported_format` arg or something

maverbiest: I added a `human_readable_pattern` argument to `build_display_name` that should be a human-readable representation of what the `regex_pattern` targets. It gets used in the warning message to users and is required if `regex_pattern` is provided.

anna-parker: this PR should now be marked as breaking

maverbiest: True! I marked it as breaking and updated the PR description

anna-parker:
@maverbiest has this been tested on staging to see the changes that will occur? might be good to do before merging - also to prepare a PR with the new config field that will be required or at least have in the in the PR description :-)

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -1153,54 +1155,51 @@ def build_display_name(  # noqa: C901
...
+            # We were able to parse an IDENTIFIER, treat it as a string
+            concatenate_field_types = replace_identifier(field_types, "string")
+            input_data["IDENTIFIER"] = identifier
+        else:
+            # Unable to parse specimenCollectorSampleId and submissionID, use ACCESSION_VERSION
+            if not insdc_ingested and regex_pattern is not None:
+                warnings.append(
+                    f"specimenCollectorSampleId and submissionId could not be parsed, using "
+                    f"ACCESSION_VERSION in displayName instead. To include your own identifier, "
+                    f"remove whitespace and '/' characters or use format '{human_readable_pattern}'"
```
anna-parker:
```suggestion
                    f"remove whitespace and '/' characters or use the format '{human_readable_pattern}' and we will parse the `identifier` from the submission."
```

preprocessing/nextclade/tests/test_metadata_processing_functions.py
```diff
@@ -1174,238 +1176,264 @@ def test_parse_date_into_range() -> None:
...
+        sample_collection_date="2025",
+        expected_regular="DENV-1/Switzerland/accession.1/2025",
+        expected_insdc="DENV-1/Switzerland/accession.1/2025",  # INSDC never uses regex
+        expected_prefix="hYF/Switzerland/accession.1/2025",
+        warning_regular=unparseable_identifier_warning,
+        warning_insdc=None,  # warning is only emitted for direct (non-INSDC) submissions
+        warning_prefix=unparseable_identifier_warning,
+    ),
+    DisplayNameCase(
+        name="empty_country_uses_default_fallback",
```
anna-parker: would probably be good to check in this test that the fallback can be used for dates as well

anna-parker (approved):
looks great - lets still test on staging though before merging - also this will probably require reprocessing of all data so we need to add that to the PR description :-)

maverbiest:
@anna-parker thank you! Yes you're right I'll hold off on merging until this new implementation is tested on staging data

Issue #1548, opened 2024-04-02
Remove last vestiges of MUI


theosanderson:
We now have not much MUI left and it should be relatively easy to remove. We should remove it when we have time, in part to avoid people being tempted to use it more in future. (MUI plays badly with Astro)

PR #6780, opened 2026-06-29
refactor(website): unify modals onto BaseDialog, drop MUI modal and react-confirm-alert

### LINKED ISSUE #6744: Confirmation dialogs/modals should make background greyed out items inert for keyboard navigation
I noticed that our confirmation dialogs are hard to use with keyboard navigation as tabbing goes through buttons in the greyed out background first.

We should somehow make the modal be in the foreground and everything else inert.

Image

Also the focus outline of blue buttons should be a color other than blue - it's hardly noticeable right now that the buttons are focused.

corneliusroemer:
Relates to #1548 (removing a MUI component)
Fixes https://github.com/loculus-project/loculus/issues/6744 (make background behind modal inert to keyboard navigation)

## Summary

The website had **four** different modal/dialog mechanisms and **three** dependencies powering them:

| Mechanism | Used by | Dismissible on backdrop click? |
|---|---|---|
| `BaseDialog` (Headless UI) | ~10 call sites | yes |
| `Modal` (`@mui/material/Modal`) | SeqSet create form | yes |
| `confirmAlert` (`react-confirm-alert`) + `ModalBox` | confirmation dialogs | configurable |
| hand-rolled native `<dialog>` + `ModalBox` | edit data use terms | no |

This PR consolidates everything onto a **single modal component, `BaseDialog`**, and removes the two extra dependencies (`@mui/material/Modal` usage and `react-confirm-alert`).

## Testing

Clicked through the various changed modals. All seem to behave well. The keyboard navigation works great with background inert.

<details>
<summary>Detailed Claude generated description</summary>

## What changed

**One unified modal: `BaseDialog`**
- Added a `dismissible` prop (default `true`). When `false`, backdrop clicks and Escape no longer close the dialog — it can only be closed via the close button or an explicit action. This folds the old "sticky `Modal`" behaviour into a single prop, and the component documents when to use each mode:
  - **default (dismissible)** — lightweight, informational modals where an accidental close is harmless (details, citations).
  - **`dismissible={false}`** — confirmations and involved modals (forms) where the user might lose significant state to an accidental backdrop click.

**Migrated every other mechanism onto it**
- The MUI-based `Modal` (only used by the SeqSet create form) → `<BaseDialog dismissible={false}>`; deleted `common/Modal.tsx`.
- `EditDataUseTermsButton` (hand-rolled native `<dialog>` + `ModalBox`) → `<BaseDialog dismissible={false}>`.
- `RevokeButton` (`confirmAlert` + `ModalBox`, has a reason input) → declarative `<BaseDialog dismissible={false}>` with local state.

**Confirmations are now declarative**
- Replaced the imperative `displayConfirmationDialog` / `react-confirm-alert` with a `useConfirmDialog()` hook built on `BaseDialog`. It returns `{ confirm, confirmDialog }`: render `confirmDialog` once and call `confirm({ dialogText, onConfirmation, ... })` from handlers. One hook instance per component covers any number of confirm actions (only one shows at a time).
- Migrated all 11 call sites across `ReviewPage`, `GroupPage`, `EditPage`, `DataUploadForm`, `SeqSetItem`, `SeqSetItemActions`.

**Cleanup**
- Deleted now-unused `common/Modal.tsx` and `common/ModalBox.tsx`.
- Removed the `react-confirm-alert` dependency and its `.react-confirm-alert-overlay` CSS.

**Scroll fix**
- When `title=''`, `BaseDialog` previously rendered an empty `<h3>` (`leading-6` + `mb-4`), adding ~2.5rem of dead vertical space that could push the panel past the viewport and cause a small spurious scroll. It now uses `sr-only` instead of `mb-4` when there is no title, so it takes no layout space while keeping the `DialogTitle` element for accessibility.

> Note: `@mui/material` itself is **not** fully removed — `Pagination` still uses it (`SearchPagination`, `SeqSetList`, `SeqSetItem`). Dropping MUI entirely is a separate follow-up.

## Behaviour change worth flagging

The generic confirmation dialog is now **not dismissible by backdrop click** (it was `closeOnClickOutside: true` before), matching the principle that confirmations shouldn't be lost accidentally. Users still cancel via the Cancel or close (✕) button.

## Testing

- `npm run check-types`: clean (in `website/src`).
- `eslint` on `src`: clean.
- `prettier`: clean.
- Unit tests: 683 passed.
- Manually verified against the `main` preview backend on localhost (SeqSet create modal, revoke flow, confirmations) — including that the scroll glitch is gone.
- Playwright integration tests drive these dialogs via button text / dialog text / input placeholder (`Confirm`, `Discard`, `Approve`, `Continue under Open terms`, `Revoke this sequence`, "Are you sure you want to revoke this sequence?", "Enter reason for revocation"), all of which are preserved.

</details>


website/src/components/ReviewPage/ReviewPage.tsx
```diff
@@ -408,6 +409,7 @@ const InnerReviewPage: FC<ReviewPageProps> = ({
             </div>
             {reviewCards}
             {pagination}
+            {confirmDialog}
```
theosanderson: I do think the original pattern without this is nicer - I'm guessing this might be quite achievable whilst still consolidating

corneliusroemer: It's quite normal react to have to say where to put something in the DOM, there's a hacky/implicit way one can do suure, see incoming commit

theosanderson: I'm not against being explicit - but I think one can do it something like this https://github.com/loculus-project/loculus/blob/6c6d2f55d2395c329abe7cfa042b269ce0ee3704/website/src/layouts/BaseLayout.astro#L78

theosanderson: (but also happy with whatever's in that commit if it works! main thing was avoiding having to put this in every location)

corneliusroemer: While we can hack it so we don't need the declarative {confirmDialog} there are accessibility/UX reasons to prefer the declarative, e.g. keyboard navigation works properly.

If it weren't for that I'd be happy with the hack but this makes me prefer the declarative solution.

See Claude convo:

```

❯ interesting, now after your changes the background is no longer inert for confirmation dialogs - why is that?

Good catch — that's a real regression, and it's a direct consequence of the imperative (createRoot) approach we just reverted to.

Why: The "background goes inert" behavior isn't something our code does — it's provided by Headless UI's Dialog. When a Dialog opens, Headless UI walks the DOM and applies inert to everything outside the dialog, plus traps focus and locks scroll. Crucially, it coordinates all of that through React context (its MainTreeProvider / inert-others machinery) — it needs the Dialog to be part of the same React tree as the rest of the app so it can identify "the rest of the app" as the thing to make inert.

- Our other BaseDialog usages render inside the app's React tree → Headless UI establishes the main-tree relationship correctly → background is inert. ✅
- The confirmation dialog is now mounted via a standalone createRoot() into a detached <div> on document.body. That root is its own isolated React tree with a fresh, empty context — Headless UI has no link back to the app tree, so it can't determine/mark the app content as inert. ❌

So mounting detached gives you the dialog visually, but you lose the inert/focus-coordination that depends on tree/context membership.

Notably, the old react-confirm-alert + ModalBox version was also never truly inert (ModalBox is a plain div with no focus trap; the overlay only blocked clicks visually). So this isn't a regression vs. main — it's a regression vs. the intermediate declarative hook version, where the dialog rendered in-tree and inerted correctly.

This is actually the strong argument for the declarative approach that I think outweighs the call-site ergonomics: rendering the dialog in the React tree (the useConfirmDialog() hook with {confirmDialog}) isn't just stylistically different — it's the version that gets inert background, focus trapping, and scroll lock right, because that's how Headless UI is designed to be used.

My recommendation: revert the confirmation dialog back to the declarative useConfirmDialog() hook. I'd then note to Theo that the imperative version breaks inert/focus-trapping, which is why in-tree rendering is the better call here.

Want me to switch it back to the hook? (I can keep the call sites as clean as possible — one {confirmDialog} per component — and reply on the PR explaining the inertness trade-off.)
```

theosanderson: I think the method I proposed will address the concern you have https://github.com/loculus-project/loculus/pull/6780#discussion_r3491330058

theosanderson: I'll have a look at some point if you don't have bandwidth, let me know

corneliusroemer: I'm not sure this is an Astro island issue? Don't think it's worth spending time hacking around this to avoid a single line in 5 files?

Headless' dialog just has this pattern of needing to put the element there explicitly. we do this in all the other places we use `BaseDialog`

theosanderson: It's about avoiding having to remember to add both things and avoiding any situation where you end up not having the actual element due to an early return, for example. I don't think you're understanding the pattern I'm suggesting (which is also a very common pattern). It's not about it being an Astro island issue. I'll make a PR to this one.

corneliusroemer: Yeah early return case would be annoying.

I didn't get it before, now after #6786 I think I do - you want to put a container in base layout - that might work, not sure about the inert/keyboard handling in that case.

theosanderson: nearly there with something

theosanderson: https://github.com/loculus-project/loculus/pull/6789

had forgotten that the islanded react does make it more complicated, but I do think this is a good pattern - have tested locally

theosanderson: working on the tests (false positive) now

website/src/components/Submission/FileUpload/FolderUploadComponent.tsx
```diff
@@ -393,146 +393,154 @@ export const FolderUploadComponent: FC<FolderUploadComponentProps> = ({
         } else void addAdditionalFiles();
     };
 
-    return fileUploadState === undefined || fileUploadState.type === 'awaitingUrls' ? (
```
tombch, line 396: I think the large diff in this file is not really needed? Its only a few lines that actually required changing right

(otherwise looks good to me)

PR #6788, opened 2026-06-29
chore(website): fix false positive Astro warning about unused local variable

corneliusroemer:
## Summary

This PR fixes false positive Astro lint:

```
16:10:33 [check] Getting diagnostics for Astro files in /Users/cr/code/loculus-main/website...
src/pages/seqsets/[seqSetId].astro:20:19 - warning ts(6133): 'latestVersion' is declared but its value is never read.

20             const latestVersion = Math.max(...versions.map((s) => s.seqSetVersion));
                     ~~~~~~~~~~~~~
```


theosanderson (approved):
I guess one could also have suppressed the warning with a comment to maintain readability - but fine either way

PR #6789, opened 2026-06-29
refactor(website): make confirmation dialog a global singleton

theosanderson:
PR to Cornelius's PR https://github.com/loculus-project/loculus/pull/6780

Replace the useConfirmDialog() hook (which returned { confirm, confirmDialog } and required every call site to both call confirm() and render {confirmDialog}) with a module-level store and a single ConfirmDialogContainer host mounted once in BaseLayout. Call sites now just import and call confirm() — no element to render, and no risk of an early return omitting the dialog.

A module-level store rather than React context is required because each Astro client:* island is an isolated React tree that context cannot cross. The host renders in-tree (not a detached createRoot), so Headless UI still makes the background inert and traps focus.

🚀 Preview: https://confirmation.loculus.org

### DIFF
--- a/website/src/components/ConfirmationDialog.tsx
+++ b/website/src/components/ConfirmationDialog.tsx
@@ -1,32 +1,60 @@
-import { type ReactNode, useCallback, useState } from 'react';
+import { type ReactNode, useSyncExternalStore } from 'react';
 
 import { BaseDialog } from './common/BaseDialog';
 import { Button } from './common/Button';
 
-type ConfirmOptions = {
+export type ConfirmOptions = {
     dialogText: ReactNode;
     confirmButtonText?: string;
     closeButtonText?: string;
     onConfirmation: () => Promise<void> | void;
 };
 
-/**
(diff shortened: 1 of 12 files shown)


corneliusroemer:
Very nice, tested locally and works 🎉 

Just the tests now are unhappy:

```

Error: TestingLibraryElementError: Unable to find role="button" and name "Continue under Open terms"

Ignored nodes: comments, script, 
```

theosanderson:
yep working on it, thanks!

theosanderson:
annoying amount of churn to supply this for the tests - I tried a global pattern where one wouldn't have needed changes for each test but that broke the hydrated button test. I think that possibly indicates that that test is overly fragile. Anyway, I'd be very keen to keep this single-function pattern one way or another.

corneliusroemer:
Not so worried about the test churn. The global state outside React isn't great but if you want the DX of single call I'm fine with it.

theosanderson:
This is the pattern we already have for toasts, and I do think it's significantly safer. Thanks for tolerating!

PR #6790, opened 2026-06-29
feat(prepro): make tests more readable

anna-parker:
While reviewing https://github.com/loculus-project/loculus/pull/6774 I had a lot of trouble reading the tests, I think having clear test cases with descriptive names of what the test is checking makes this a lot easier to parse (or grok as @corneliusroemer would say). I got claude to reformat these tests like the other prepro test `Case` objects.

### PR Checklist

### DIFF
--- a/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
@@ -1638,7 +1638,10 @@ class ProcessingFunctions:
             )
 
         regex_pattern = args.get("regex_pattern")
-        if regex_pattern is not None and "<identifier>" not in str(regex_pattern):
+        if (
+            regex_pattern is not None
+            and "identifier" not in re.compile(str(regex_pattern)).groupindex
+        ):
             return ProcessingResult(
                 datum=None,
                 warnings=warnings,
--- a/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
@@ -1591,7 +1591,7 @@ class ProcessingFunctions:
         input_fields: list[str],
         args: FunctionArgs,
     ) -> ProcessingResult:
-        """Wraps concatenate() to resolve the IDENTIFIER slot in args['order']/args['type'].
+        """Create a display name by wrapping concatenate() and resolving the IDENTIFIER slot.
 
         The IDENTIFIER slot is filled by the best available sample identifier:
           1. specimenCollectorSampleId — used as-is if it contains no slashes or spaces;
@@ -2145,7 +2145,7 @@ DISPLAY_NAME_SEPARATORS = [" ", "/"]
 
 
 def has_display_name_separators(input: str) -> bool:
-    """Check if the input string contains any characters that are not allowed in a displayName."""
+    """Check if the input string contains any characters that are used as displayName separators."""
     return any(char in input for char in DISPLAY_NAME_SEPARATORS)
 
 
--- a/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
@@ -2149,13 +2149,11 @@ def has_display_name_separators(input: str) -> bool:
     return any(char in input for char in DISPLAY_NAME_SEPARATORS)
 
 
-def parse_identifier_string(
-    input: str, regex_pattern: str
-) -> str | None:
+def parse_identifier_string(input: str, regex_pattern: str) -> str | None:
     """Extract a usable display-name identifier from `input`, or return None.
 
-    Returns input as-is if it contains no displayName separators (space, slash, …).
-    If it contains separators, attempts to extract a usable identifier 
+    Returns input as-is if it contains no DISPLAY_NAME_SEPARATORS.
+    If it contains separators, attempts to extract a usable identifier
     using the provided regex_pattern.
     """
     if not has_display_name_separators(input):
--- a/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
@@ -1591,21 +1591,22 @@ class ProcessingFunctions:
         input_fields: list[str],
         args: FunctionArgs,
     ) -> ProcessingResult:
-        """Builds a displayName from input_fields. The identifier field in the displayName is
-        based on specimenCollectorSampleId; if that cannot be parsed we fall back to submissionId,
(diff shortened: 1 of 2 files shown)


preprocessing/nextclade/tests/test_metadata_processing_functions.py
```diff
@@ -1176,239 +1178,251 @@ def test_parse_date_into_range() -> None:
...
-    )
 
-    args["fallback_value"] = "unknown"
-    res_fallback = ProcessingFunctions.concatenate(
-        input_data,
-        output_field,
-        input_fields,
-        args,
-    )
+_CONCATENATE_CASES = [
```
maverbiest:
```suggestion
concatenate_case_definitions = [
```

Maybe this? To be more consistent with `test_case_definitions` etc. in the same file

preprocessing/nextclade/tests/test_metadata_processing_functions.py
```diff
@@ -1176,239 +1178,251 @@ def test_parse_date_into_range() -> None:
...
-    res_insdc = ProcessingFunctions.build_display_name(
-        input_data, output_field, input_fields(), args_insdc()
-    )
-    res_prefix = ProcessingFunctions.build_display_name(
-        input_data, output_field, input_fields(), args_with_prefix()
-    )
-    assert res.datum == "DENV-1/Switzerland/mySample/2025"
-    assert res_insdc.datum == "DENV-1/Switzerland/mySample/2025"
-    assert res_prefix.datum == "hYF/Switzerland/mySample/2025"
+_DISPLAY_NAME_CASES = [
```
maverbiest, line 1318:
```suggestion
display_name_case_definitions = [
```

preprocessing/nextclade/tests/test_metadata_processing_functions.py
```diff
@@ -1176,239 +1178,251 @@ def test_parse_date_into_range() -> None:
...
-
-    def input_fields():
-        return [
-            "nextclade.clade",
-            "geoLocCountry",
-            "specimenCollectorSampleId",
-            "submissionId",
-            "sampleCollectionDate",
-        ]
+_DISPLAY_NAME_REGEX = r"^[^\/][^/]*/[^/]+/(?P<identifier>[^/]+)/\d{4}(?:-\d{2}){0,2}$"
```
maverbiest, line 1274: Not blocking, but idk why claude likes this `_UPPERCASE_WITH_LEADING_UNDERSCORE` so much. I'm not the biggest fan and it's also inconsistent with other variable definitions in the same file e.g., https://github.com/loculus-project/loculus/blob/36b77c971aeec6b0e9611ad65f77db93caafc428/preprocessing/nextclade/tests/test_metadata_processing_functions.py#L656

Could we refactor these?

preprocessing/nextclade/tests/test_metadata_processing_functions.py
```diff
@@ -1176,239 +1178,251 @@ def test_parse_date_into_range() -> None:
...
-    }
-
-    res_no_fallback_no_int = ProcessingFunctions.concatenate(
-        input_data,
-        output_field,
-        input_fields,
-        args,
-    )
+@dataclass
+class ConcatenateCase:
```
maverbiest, line 1182: Would this class definition (and the `DisplayNameCase` definition) fit better in `factory_methods.py`, next to the definition of `Case`?

maverbiest (approved):
Thanks for this, these tests had really turned into a mess.

I left some small comments but overall I like this approach much more!

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -1620,61 +1618,43 @@ def build_display_name(
...
-
         insdc_ingested = bool(args["is_insdc_ingest_group"])
-        pattern = str(regex_pattern) if regex_pattern is not None else None
-
-        # Try specimenCollectorSampleId first, fall back to submissionId.
-        identifier = parse_identifier_string(collector_id, insdc_ingested, pattern)
-        if identifier is None:
-            identifier = parse_identifier_string(submission_id, insdc_ingested, pattern)
+        if insdc_ingested:
+            # For INSDC ingested sequences the submissionId is the INSDC accession
```
anna-parker, line 1643: @maverbiest I realized now that the logic of the PR should maybe be different? I thought we did not want to fallback to the INSDC accession for ingested sequences

preprocessing/nextclade/tests/test_metadata_processing_functions.py
anna-parker: as we only use it in this file and it is very specific to this function I dont think it is necessary

preprocessing/nextclade/tests/test_metadata_processing_functions.py
anna-parker: tbf capitalising constants is actually pretty common and considered good practice I think, the leading underscore is weird though, I will remove it

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
maverbiest: Oh hmm I don't remember exactly what we decided on this. I'll check on Slack to see what people's thoughts are

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -1620,61 +1619,44 @@ def build_display_name(
...
+            # - do not use it in the displayName;
+            # use the collector_id as-is unless it already contains displayName separators
+            identifier = (
+                None if has_display_name_separators(str(collector_id)) else str(collector_id)
+            )
+        else:
+            identifier = parse_identifier_string(str(collector_id), str(regex_pattern))
```
maverbiest, line 1650: This is why the unit test is failing, I think: if `regex_pattern` is `None`, `str(None) -> "None"`

We need to narrow to `str | None` from `ArgValue` and then check for `None` before passing

maverbiest (commented):
Since we're now anyway doing a big refactor of the prepro code in https://github.com/loculus-project/loculus/pull/6537, could we maybe just keep the changes to tests in this PR?

We can then see how we want to update to logic after rebasing the `display-name-fallback` branch

preprocessing/nextclade/tests/test_metadata_processing_functions.py
maverbiest: Yeah fair, I don't mind the capitalization either really it's mostly the leading underscore

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
maverbiest: (might be wrong though, need to take a closer look)

anna-parker:
@maverbiest, yes! you can cherry-pick the test changes to your PR if you would like :-)

PR #6807, opened 2026-06-30
feat(observability): expose Prometheus metrics and wire ServiceMonitor

### LINKED ISSUE #4807: Expose and scrape backend metrics via spring prometheus
Learning about Spring Boot, we can get some backend metrics very easily by configuring the actuator to expose some stats. Would be great to have better insights into number of requests, request latencies etc.

Not high priority but useful in the long run and should be fun.

clonder:
- add micrometer-registry-prometheus dependency to backend so /actuator/prometheus endpoint exists
- enable the endpoint via backendExtraArgs in values.yaml
- add app/component labels to loculus-backend-service so the ServiceMonitor selector can find it
- add ServiceMonitor for Prometheus Operator to scrape the backend every 15s (requires release=kube-prometheus-stack label to be picked up by the existing kube-prometheus-stack instance)

🚀 Preview: https://feat-observability-stack.loculus.org


corneliusroemer:
I think this relates to issue:
- https://github.com/loculus-project/loculus/issues/4807

kubernetes/loculus/values.yaml
```diff
@@ -2709,6 +2709,11 @@ taxonomyService:
   dbVersion: "ncbi_taxonomy_2026-05-01"
   taxonomyDbPath: "/data/taxonomy.sqlite" # this is where the init container mounts the database
 subdomainSeparator: "-"
+enableServiceMonitor: false
+backendExtraArgs:
+  - "--management.endpoint.prometheus.enabled=true"
+  - "--management.endpoints.web.exposure.include=health,prometheus"
```
corneliusroemer: Does it make sense to have these args present when serviceMonitor is off? If not, we could add those args programmatically depending on `enableServiceMonitor` flag in the backend service file.

corneliusroemer:
Nice, so easy to add!

Google Chrome 2026-06-30 19 43 50

[URL for above graph](https://grafana.loculus.org/explore?schemaVersion=1&panes=%7B%2282l%22:%7B%22datasource%22:%22decyzhq5ockqoe%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22rate%28http_server_requests_seconds_count%7Bnamespace%3D%5C%22prev-feat-observability-stack%5C%22%7D%5B1m%5D%29%22,%22range%22:true,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22decyzhq5ockqoe%22%7D,%22editorMode%22:%22builder%22,%22legendFormat%22:%22__auto%22,%22useBackend%22:false,%22disableTextWrap%22:false,%22fullMetaSearch%22:false,%22includeNullMetadata%22:false%7D%5D,%22range%22:%7B%22from%22:%22now-6h%22,%22to%22:%22now%22%7D%7D%7D&orgId=1)

kubernetes/loculus/values.yaml
clonder: Good point! 
Moved the args under the condition

fhennig:
I had a look at the code briefly, looks generally good to me! Also useful! Some docs about this could maybe be useful, a section like "Observability" somewhere? But can also be in a different PR

clonder:
> I had a look at the code briefly, looks generally good to me! Also useful! Some docs about this could maybe be useful, a section like "Observability" somewhere? But can also be in a different PR

Thank you :) The docs are [here](https://app.nuclino.com/Loculus/Development/Loculus-Grafana-dashboard-a5bb25f7-94d2-4240-8e00-08ec3bd3f5fe) !

kubernetes/loculus/values.schema.json
```diff
@@ -1599,6 +1599,12 @@
       "default": false,
       "description": "If true, disables the backend API deployment."
     },
+    "enableServiceMonitor": {
+      "groups": ["services"],
+      "type": "boolean",
+      "default": false,
+      "description": "If true, deploys a Prometheus Operator ServiceMonitor for the backend. Requires kube-prometheus-stack CRDs to be installed."
```
chaoran-chen: How should one install `kube-prometheus-stack CRDs`? I think some additional information would be very useful for admins to know how to activate this (but if this needs a longer explanation and generalization, I'm also happy with moving this to a new issue)

chaoran-chen (commented):
Thanks very much @clonder for this PR! Apart from the specific point raised in the code, I've two general comments

- I'm wondering whether functions like `timeWritePhase` should be extracted and stored separately and whether it's possible to work more systematically with higher-order functions. At the moment, the measurement code seems rather mixed with the main logic which can make the code more difficult to maintain later.
- This is already pre-existing problem but maybe good to keep in mind when developing this further: The metrics endpoint, when activated, are public which may not be safe. It can give an attacker useful information about the state of the server.  For now, when this is only for test instances for performance measurements, that's fine, but if we want this to be used for measuring a production instance, this should be changed by only allowing e.g. cluster internal access or another authentication measure.

I think both comments are not blocking, so I'm happy to approve if you think the PR is ready. (The PR is a bit out-of-date and should be synchronized with main again.)

(@claude, can you please also review? (Do you listen to me?))

clonder:
@claude review

backend/src/main/kotlin/org/loculus/backend/model/SubmitModel.kt
```diff
@@ -35,6 +36,17 @@ const val FASTA_IDS_SEPARATOR = " "
...
+private const val REVISE_ENDPOINT = "revise"
+private const val VALIDATE_UPLOAD_PHASE = "validate-upload"
+private const val COPY_TO_AUX_TABLE_PHASE = "copy-to-aux-table"
+private const val LOAD_METADATA_SUBMISSION_IDS_PHASE = "load-metadata-submission-ids"
+private const val VALIDATE_CONSENSUS_SEQUENCES_PHASE = "validate-consensus-sequences"
+private const val ASSOCIATE_REVISED_DATA_PHASE = "associate-revised-data"
+private const val VALIDATE_FILE_MAPPING_PHASE = "validate-file-mapping"
+private const val GENERATE_ACCESSIONS_PHASE = "generate-accessions"
+private const val INSERT_SEQUENCE_ENTRIES_PHASE = "insert-sequence-entries"
+private const val CLEANUP_UPLOAD_DATA_PHASE = "cleanup-upload-data"
```
fhennig: nit: I'd probably move these into either `SubmissionMetrics.kt` or into a dedicated "metrics constants" file.

backend/src/main/kotlin/org/loculus/backend/log/Logging.kt
```diff
@@ -33,25 +45,161 @@ class OrganismMdcInterceptor : HandlerInterceptor {
...
+                    measurement.asynchronousRequestCompleted()
+                }
+            },
+        )
+    }
+
+    private fun currentThreadCpuTime() =
+        if (threadBean.isCurrentThreadCpuTimeSupported) threadBean.currentThreadCpuTime else -1L
+
+    private inner class RequestMeasurement(
```
fhennig: Hmm, why is this class thrown in here with `Logging.kt`? IMO maybe a new module `metrics` alongside `log` would be good where all the metrics related stuff could go. But not blocking.

backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
```diff
@@ -639,38 +692,73 @@ open class SubmissionController(
...
+            return
+        }
+
+        submissionMetrics.recordPollingRequest(
+            endpoint = endpoint,
+            organism = organism.name,
+            status = status.value().toString(),
+            duration = Duration.ofNanos(System.nanoTime() - requestStartNanos),
+        )
+    }
```
fhennig: This function as well as the consts at the top of the file (`POLLING_ENDPOINTS`) maybe also in a dedicated file? Or in `SubmissionMetrics.kt`? I think it could be simplified.

fhennig (commented):
Yes, looks useful, great!

I found your code to be quite readable and it all made sense to me; I'd only say I think it'd be good to have more cohesion for the metrics related code and consts; instead of spreading the phase name consts across multiple files, I'd put them all in one file. Maybe it would make sense to have a module where the consts and functions could live, like `log` - maybe `metrics`?

fhennig (approved):
LGTM!

PR #6808, opened 2026-06-30
feat(deposition): allow automatic revision of author fields in assemblies

anna-parker:
See https://loculus.slack.com/archives/C0757PTR607/p1782745074666639 - we have been informed revision of fields in the manifest.tsv is now possible via the cli.

### PR Checklist

### DIFF
--- a/ena-submission/src/ena_deposition/create_assembly.py
+++ b/ena-submission/src/ena_deposition/create_assembly.py
@@ -407,7 +407,11 @@ def can_be_revised(config: Config, db_engine: Engine, submission_row: Submission
                 update_type="revision",
             )
             return False
-
+    if config.allow_retry_for_manifest_changes:
+        logger.debug(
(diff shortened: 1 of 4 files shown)


ena-submission/src/ena_deposition/create_assembly.py
```diff
@@ -407,7 +407,11 @@ def can_be_revised(config: Config, db_engine: Engine, submission_row: Submission
                 update_type="revision",
             )
             return False
-
+    if config.allow_retry_for_manifest_changes:
```
corneliusroemer: I'd move this into the `if fields_changed` branch - early return for True is not the right pattern here where we're validating a bunch of criteria. one can accidentally add something after and that gets wrongly short-circuited

corneliusroemer (approved):
Thanks! Good for now, looking into it while we can deploy this as first stop-gap.

PR #6813, opened 2026-06-30
feat(backend): keep only latest preprocessing pipeline version

corneliusroemer:
## Summary

The pipeline-version garbage collector (`UseNewerProcessingPipelineVersionTask`) runs after the backend bumps an organism to a newer preprocessing pipeline version, deleting outdated preprocessed data. Previously it retained the **two most recent** pipeline versions by passing `latestVersion - 1` as the earliest version to keep.

This PR changes it to keep **only the latest** version (passing `latestVersion`), so older preprocessed data is no longer retained unnecessarily — keeping only what's needed.

## Changes

- **`UseNewerProcessingPipelineVersionTask.kt`**: pass `latestVersion` instead of `latestVersion - 1` to `cleanUpOutdatedPreprocessingData`. This is the only behavioral change; the deletion is still scoped per-organism and only triggers when a version upgrade actually happens.
- **`UseNewerProcessingPipelineVersionTaskTest.kt`**: updated the existing GC test assertions to expect only the latest version surviving (`[2L]` after the v2 bump, `[3L]` after v3), with clarified comments. The per-organism scoping assertions (`OTHER_ORGANISM` untouched) are unchanged.
- **`backend/AGENTS.md`**: clarified test instructions to check for Docker first and prefer the Docker route, falling back to `USE_NONDOCKER_INFRA=true` only when Docker is absent or its run fails.

## Testing

`./gradlew test --tests "...UseNewerProcessingPipelineVersionTaskTest"` passes.

### DIFF
--- a/backend/AGENTS.md
+++ b/backend/AGENTS.md
@@ -1,10 +1,10 @@
 Kotlin dependent packages have already been installed for you.
 
-Run tests like this (if you have Docker set up properly):
+First check whether Docker is present (e.g. `docker info`). If it is, prefer running tests through Docker:
 
 ./gradlew test --console=plain
 
-If that doesn't work due to Docker issues because you're running inside a cloud environment, try this:
+Only fall back to the non-Docker route if Docker is not present or the Docker-based run fails (e.g. when running inside a cloud environment):
 
 USE_NONDOCKER_INFRA=true ./gradlew test --console=plain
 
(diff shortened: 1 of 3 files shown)


theosanderson:
At first glance I'm not sure about this. I can imagine various race conditions that might result (not necessarily with current code, unsure, but with possible code) and I'm not sure DB size is such a big issue

PR #6820, opened 2026-06-30
fix(backend,website): remove broken, unused `/get-user-cited-by-seqset` endpoint

### LINKED ISSUE #6497: Inconsistent use of the term 'Citation' in backend
When reviewing https://github.com/loculus-project/loculus/pull/6304, I noticed that backend uses the term `Citation` to mean two different things:

1. A sequence being part of a seqset (e.g., `getUserCitedBySeqSet()` in `SeqSetCitationsDatabseService.kt`)
2. An external source (like a peer-reviewed publication indexed in Crossref) referencing a seqset (e.g., `getSeqSetCitedByPublication()` in `SeqSetCitationsDatabseService.kt`)

This muddles the meaning of the term and makes the codes harder to reason about while reading it ('what type of citation is this function I'm looking at referring to?'). To me, the obvious correct usage of the term is 2. We should refactor the backend code to update functionality relating to 1 to use the term 'Inclusion', or similar. For example, `getUserCitedBySeqSet()` could become `getUserSequencesIncludedInSeqSets()`.

### LINKED ISSUE #6819: Remove completely broken `getApprovedUserAccessionVersions` function and only caller `GET /get-user-cited-by-seqset` (unused by website)
Claude identified a bug in that the `GET /get-user-cited-by-seqset` is completely broken, it always returns HTTP 500 no matter how called because it calls `getApprovedUserAccessionVersions` which has an SQL bug.

Given both functions have been around 19 months or more without being used and they are broken and unused (otherwise we would have noticed they are broken) I propose we remove both.

It's quite interesting how this was missed by tests. The endpoint mocked that function - and the function itself was untested.

Here's Claude's writeup (don't want to make you read it but if you're curious it's got some fun insights):

<details>

### Summary

`getApprovedUserAccessionVersions()` (`backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt:1251-1263`)
passes a row-filter expression to `.groupBy()` instead of `.where()`/`.andWhere()`:

```kotlin
fun getApprovedUserAccessionVersions(authenticatedUser: AuthenticatedUser): List<AccessionVersion> =
    SequenceEntriesView.select(
        SequenceEntriesView.accessionColumn,
        SequenceEntriesView.versionColumn,
    )
        .where(SequenceEntriesView.statusIs(APPROVED_FOR_RELEASE))
        .groupBy(getGroupCondition(null, authenticatedUser))   // <- should be .andWhere(...)
        .map { ... }
```

`getGroupCondition(...)` returns an `Op<Boolean>` (either `Op.TRUE` for superusers or `group_id IN (...)` for regular
users) — neither makes `accession`/`version` functionally dependent on the `GROUP BY` clause, so Postgres rejects the
query outright. This is the only caller of this method
(`backend/src/main/kotlin/org/loculus/backend/controller/SeqSetCitationsController.kt:104-107`,
`getUserCitedBySeqSet`), so `GET /get-user-cited-by-seqset` always 500s, unconditionally, for every caller.

### PoC

```
$ TOKEN=$(curl -s -X POST https://authentication-main.loculus.org/realms/loculus/protocol/openid-connect/token \
    -d 'username=testuser&password=testuser&grant_type=password&client_id=backend-client' | jq -r .access_token)

$ curl -s https://backend-main.loculus.org/get-user-cited-by-seqset -H "Authorization: Bearer $TOKEN" | jq '.'

{
  "type": "about:blank",
  "title": "Internal Server Error",
  "status": 500,
  "detail": "org.postgresql.util.PSQLException: ERROR: column \"sequence_entries_view.accession\" must appear in the GROUP BY clause or be used in an aggregate function\n  Position: 8",
  "instance": "/get-user-cited-by-seqset"
}
```

Reproduced live against `main.loculus.org` on 2026-06-30. Also independently reproduced offline against a clean local
Postgres with the equivalent `sequence_entries`/`sequence_entries_view` schema — both the superuser branch
(`GROUP BY TRUE`) and the regular-user branch (`GROUP BY (group_id IN (...))`) fail with the same class of error, so
this isn't an environment quirk; the generated SQL is unconditionally invalid. (Also confirmed from Exposed's `Query.kt`
source that `.groupBy()` does nothing but append the literal expression to the clause — no ORM-side rescue is possible.)

Script: `poc_broken_query_get_user_cited_by_seqset.sh` (same directory as this issue).

### Why CI didn't catch it

`CitationEndpointsTest.kt:38-39` replaces the entire `SubmissionDatabaseService` with a MockK mock for this test class:

```kotlin
@MockkBean
lateinit var submissionDatabaseService: SubmissionDatabaseService
```

and `setup()` (lines 49-53) stubs out exactly the broken method:

```kotlin
every {
    submissionDatabaseService.getApprovedUserAccessionVersions(match { it.username == DEFAULT_USER_NAME })
} returns listOf(AccessionVersion(MOCK_SEQ_ACCESSION, MOCK_SEQ_VERSION))
```

So the real `.groupBy(Op<Boolean>)` query never executes against a database in CI — the test only exercises the
citation-aggregation/response-shape/auth logic *downstream* of that call. The two existing test cases
(`CitationEndpointsTest.kt:71-79`, `82-98`) assert `200 OK`, which is true only because the one thing that's actually
broken is faked out before it can run. This is a controller test that looks like an integration test (real Spring
context, real HTTP via MockMvc) but unit-tests around its most important dependency.

### Blast radius / who actually sees this

`getApprovedUserAccessionVersions` has exactly one caller in the entire codebase
(`grep -rn getApprovedUserAccessionVersions backend/src` → `SeqSetCitationsController.kt:106` and the test mock above,
nothing else), and that caller backs exactly one endpoint, `GET /get-user-cited-by-seqset`.

- **Website: not reachable today.** A client method exists (`website/src/services/seqSetCitationClient.ts:26-28`,
  `seqSetCitationApi.ts:25-28`) but it is never called from any page or component — grepped every `.tsx`/`.astro` file
  in `website/src`, zero callers. The "Cited by N" text users see on SeqSet pages comes from a *different*, working
  endpoint (`get-seqset-citations`/`get-sequence-citations`). So no regular website user hits this in normal use.
- **Who can actually trigger it**: any authenticated user/integrator who calls the endpoint directly — via
  swagger-ui's "Try it out", a script, or a third-party tool built against the public API
  (`/swagger-ui/index.html`, requires only a valid bearer token, no special role). They'd get a bare `500` plus,
  because of the separate catch-all exception handler (`ExceptionHandler.kt:24-33`, see related finding), the raw
  Postgres error text including internal table/column names.
- **Likely intent**: the dead frontend client code suggests a "your sequences cited by SeqSets" UI feature was planned
  or removed; right now it has no live UI surface, so real-world impact today is limited to API consumers, not the
  general website audience.

### History

- **Introduced 2024-11-11, PR [#3193](https://github.com/loculus-project/loculus/pull/3193) "refactor getSequences usage
  in SeqSetCitationsController".** Per the PR's own description, this was a pure refactor in preparation for an unrelated
  feature (issue #1000, returning organism-scoped metadata from `getSequences`) — splitting out the organism-less usage
  into a new function. The author wrote a fresh `getApprovedUserAccessionVersions()` to replace a previously-working call
  to `getSequences(...)`, and introduced the `.groupBy()`/`.where()` mix-up in that rewrite. The PR's own checklist left
  **"The implemented feature is covered by an appropriate test" unchecked**, and the single review was an "LGTM, thanks :)"
  approval with one unrelated style nit — nothing caught the bug. The accompanying test diff just updated the existing
  mock (`getSequences` → `getApprovedUserAccessionVersions`), continuing a mocking pattern that was already in place
  before this PR, not introducing it. The bug has been live, unnoticed, ever since — about 19 months at the time of this
  review.
- **2026-05-26, issue [#6497](https://github.com/loculus-project/loculus/issues/6497) "Inconsistent use of the term
  'Citation' in backend".** An agent-assisted deep dive into this exact code path (as part of an unrelated terminology
  cleanup) documented that the website's "Cited by" panel on `seqsets/index.astro` was "currently hard-coded to empty
  data" — i.e. even that review didn't surface that the underlying endpoint is completely broken, only that the panel
  wasn't really wired up. Proposes renaming `getUserCitedBySeqSet`/`get-user-cited-by-seqset` to
  `getUserSequencesIncludedInSeqSets`/`get-user-sequences-included-in-seqsets` to fix a separate terminology confusion
  (this endpoint means "sequence included in a SeqSet", not "SeqSet cited by a publication" — the latter is the newer,
  unrelated `get-seqset-citations`/`get-sequence-citations` feature). Still open as of this review.
- **2026-06-10 to 2026-06-24, a cluster of active citation-feature work**: PR
  [#6304](https://github.com/loculus-project/loculus/pull/6304) "SeqSet citation tracking" (adds real Crossref-backed
  citation data, replaces `get-seqset-cited-by-publication` with `get-seqset-citations`), PR
  [#6491](https://github.com/loculus-project/loculus/pull/6491) "Sequence citations" (adds `get-sequence-citations`,
  a per-sequence citation list on the sequence details page), and PR
  [#6738](https://github.com/loculus-project/loculus/pull/6738) "delete dead files (FieldProps, CitationPlot)" — the
  last of which removed `CitationPlot.tsx`, the very component issue #6497 had flagged as the (non-functional)
  "Cited by" panel UI. None of this recent work touched `SubmissionDatabaseService.kt` or
  `getApprovedUserAccessionVersions`/`get-user-cited-by-seqset` at all. **Net effect: the team has been actively building
  out citation features for the last several weeks, immediately adjacent to this endpoint, including removing its last
  remaining (non-functional) UI placeholder — but the endpoint itself was never touched, fixed, or removed.** It's now an
  orphan: old, broken since birth, never exercised by real traffic, and sitting untouched in the middle of an otherwise
  actively-developed feature area. Worth deciding explicitly whether to fix it (per the rename proposal in #6497) or
  delete it outright, rather than leaving it to keep failing silently for whoever next stumbles into it via the API.

### Is this the only orphan, and is this masking pattern widespread? (follow-up investigation)

**Orphan check — yes, this is the only dead SeqSet/citation endpoint.** Checked every endpoint in
`SeqSetCitationsController.kt` against actual website callers (not just client-method existence, but real usage in
`.tsx`/`.astro`/server-side `.call(...)` code): `get-seqset`, `get-seqset-records`, `create-seqset`, `update-seqset`,
`delete-seqset`, `create-seqset-doi`, `get-seqsets-of-user`, `get-seqset-citations`, `get-sequence-citations`, and
`get-author` all have confirmed real callers. `get-user-cited-by-seqset` is the only one with zero.

**Masking pattern check — this is a narrow, structurally-explained blind spot, not a recurring "broken windows" pattern.**
Grepped every `@MockkBean` in the backend test suite (10 files). Almost all mock genuinely external dependencies
(`KeycloakAdapter`, `CrossRefService`, `S3Service`) — legitimate, not a smell. `AccessionPreconditionValidator` is mocked
in two SeqSet-citation test classes, but that class has real coverage elsewhere (it's `@Autowired`, not mocked, in all 16
`SubmissionController` endpoint tests — `ApproveProcessedDataEndpointTest`, `RevokeEndpointTest`,
`SubmitEditedSequenceEntryVersionEndpointTest`, etc. — which exercise it against real Testcontainers Postgres), so
mocking it here just avoids redundant re-testing, not hiding untested logic.

`SubmissionDatabaseService` itself is the one case worth flagging structurally: **`CitationEndpointsTest.kt` is the only
test file in the entire backend that mocks it wholesale.** Every one of the 16 `SubmissionController` endpoint test
classes autowires the real service against real Postgres. But `SeqSetCitationsController` calls exactly one
`SubmissionDatabaseService` method — `getApprovedUserAccessionVersions` — so the mock in `CitationEndpointsTest` happens
to mask exactly, and only, the one function with no other caller and thus no other test surface. It's not a wide blind
spot; it's a single coincidental seam where "the only caller" and "the only over-mock" are the same place. No similar
`.groupBy()` misuse was found elsewhere in the codebase either (checked every `Query.groupBy()` call site; the one other
hit, `SeqSetCitationsDatabaseService.kt:462`, is correctly paired with `MAX(seqSetVersion) GROUP BY seqSetId`).

### Suggested fix

Change `.groupBy(getGroupCondition(null, authenticatedUser))` to `.andWhere { getGroupCondition(null, authenticatedUser) }`
(or fold it into the existing `.where(...)`), then **delete the `SubmissionDatabaseService` mock from
`CitationEndpointsTest`** (or add a separate test that exercises the real service against the Testcontainers Postgres
instance) so this class of bug can't hide behind a mock again.

```

corneliusroemer:
Resolves #6819
Resolves #6497

The endpoint is broken and unused for more than a year. Removing it makes it easier to understand the codebase. Even if bug is fixed we wouldn't use it.

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/api/SeqSetCitations.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/api/SeqSetCitations.kt
@@ -117,21 +117,6 @@ data class SequenceCitation(val source: CitationSource, val seqSets: List<SeqSet
 
 data class ResponseSeqSet(val seqSetId: String, val seqSetVersion: Long)
 
-data class CitedBy(
-    @Schema(
-        description = "The years in which the SeqSet or sequence was cited.",
-        type = "array",
-        example = "[2000, 2001, 2002]",
-    )
-    val years: MutableList<Long>,
-    @Schema(
-        description = "The number of citations per year.",
-        type = "array",
-        example = "[1, 2, 3]",
-    )
-    val citations: MutableList<Long>,
-)
-
 data class AuthorProfile(
     val username: String,
     val firstName: String,
--- a/backend/src/main/kotlin/org/loculus/backend/controller/SeqSetCitationsController.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/controller/SeqSetCitationsController.kt
@@ -4,7 +4,6 @@ import io.swagger.v3.oas.annotations.Operation
 import io.swagger.v3.oas.annotations.security.SecurityRequirement
 import org.loculus.backend.api.AccessionVersion
 import org.loculus.backend.api.AuthorProfile
-import org.loculus.backend.api.CitedBy
 import org.loculus.backend.api.ResponseSeqSet
 import org.loculus.backend.api.SeqSet
 import org.loculus.backend.api.SeqSetCitation
@@ -19,7 +18,6 @@ import org.loculus.backend.config.BackendSpringProperty
 import org.loculus.backend.config.ENABLE_SEQSETS_TRUE_VALUE
 import org.loculus.backend.service.KeycloakAdapter
 import org.loculus.backend.service.seqsetcitations.SeqSetCitationsDatabaseService
-import org.loculus.backend.service.submission.SubmissionDatabaseService
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
 import org.springframework.validation.annotation.Validated
 import org.springframework.web.bind.annotation.DeleteMapping
@@ -39,7 +37,6 @@ import org.springframework.web.bind.annotation.RestController
 )
 class SeqSetCitationsController(
     private val seqSetCitationsService: SeqSetCitationsDatabaseService,
-    private val submissionDatabaseService: SubmissionDatabaseService,
     private val keycloakAdapter: KeycloakAdapter,
 ) {
     @Operation(description = "Get a SeqSet")
@@ -98,13 +95,6 @@ class SeqSetCitationsController(
         @RequestParam seqSetId: String,
         @RequestParam version: Long,
     ): ResponseSeqSet = seqSetCitationsService.createSeqSetDOI(authenticatedUser, seqSetId, version)
-
-    @Operation(description = "Get count of user sequences cited by SeqSets")
-    @GetMapping("/get-user-cited-by-seqset")
-    fun getUserCitedBySeqSet(@HiddenParam authenticatedUser: AuthenticatedUser): CitedBy =
-        seqSetCitationsService.getUserCitedBySeqSet(
-            submissionDatabaseService.getApprovedUserAccessionVersions(authenticatedUser),
-        )
 
     @Operation(description = "Get citations for a SeqSet from publications or other sources")
     @GetMapping("/get-seqset-citations")
--- a/backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCitationsDatabaseService.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCitationsDatabaseService.kt
@@ -6,14 +6,12 @@ import kotlinx.datetime.toJavaLocalDateTime
 import kotlinx.datetime.toLocalDateTime
 import mu.KotlinLogging
 import org.jetbrains.exposed.sql.Database
-import org.jetbrains.exposed.sql.JoinType
 import org.jetbrains.exposed.sql.LikePattern
 import org.jetbrains.exposed.sql.Op
 import org.jetbrains.exposed.sql.SortOrder
 import org.jetbrains.exposed.sql.SqlExpressionBuilder.eq
 import org.jetbrains.exposed.sql.SqlExpressionBuilder.inList
 import org.jetbrains.exposed.sql.SqlExpressionBuilder.like
-import org.jetbrains.exposed.sql.alias
 import org.jetbrains.exposed.sql.and
 import org.jetbrains.exposed.sql.andWhere
 import org.jetbrains.exposed.sql.batchInsert
@@ -29,7 +27,6 @@ import org.loculus.backend.api.AccessionVersion
 import org.loculus.backend.api.AuthorProfile
 import org.loculus.backend.api.CitationOrigin
 import org.loculus.backend.api.CitationSource
-import org.loculus.backend.api.CitedBy
 import org.loculus.backend.api.ResponseSeqSet
 import org.loculus.backend.api.SeqSet
 import org.loculus.backend.api.SeqSetCitation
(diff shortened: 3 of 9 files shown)


theosanderson (approved):

maverbiest (approved):
Looks good, thanks for investigating

corneliusroemer:
Thanks for the quick reviews @theosanderson and @maverbiest!

@maverbiest does this solve the inconsistency you reported in https://github.com/loculus-project/loculus/issues/6497? You only listed one for point 1 and this is removed now :) So maybe we can also close this issue then!

I'm closing but feel free to reopen if my reading is wrong and there's still inconsistency.

maverbiest:
Hey @corneliusroemer sorry yes you were correct in closing the issue, thanks. This resolves it!

PR #6846, opened 2026-07-06
feat(backend, config): configure S3 to use IAM roles

anna-parker:
Access is scoped not cluster-wide. Two things enforce the scoping, one on each side:

  AWS side (the actual security boundary): the IAM role's trust policy restricts who can assume it via a condition on
  the OIDC token's sub claim, tied to your cluster's specific OIDC provider and one exact namespace:serviceaccount-name
  pair:
```
  {
    "Version": "2012-10-17",
    "Statement": [
      {
        "Effect": "Allow",
        "Principal": {
          "Federated": "arn:aws:iam::<account-id>:oidc-provider/oidc.eks.<region>.amazonaws.com/id/<oidc-id>"
        },
        "Action": "sts:AssumeRoleWithWebIdentity",
        "Condition": {
          "StringEquals": {
            "oidc.eks.<region>.amazonaws.com/id/<oidc-id>:sub": "system:serviceaccount:<namespace>:loculus-backend",
            "oidc.eks.<region>.amazonaws.com/id/<oidc-id>:aud": "sts.amazonaws.com"
          }
        }
      }
    ]
  }
```
  Only a pod whose ServiceAccount token has that exact sub claim can exchange it for this role's credentials — nothing
  else in the cluster can, regardless of what IAM policy is attached.

  K8s side (what I already set up): the eks.amazonaws.com/role-arn annotation lives on the standalone loculus-backend
  ServiceAccount, and serviceAccountName: loculus-backend is only set on the backend Deployment's pod spec — every other
  workload in the chart (website, preprocessing, ingest, etc.) still runs under default, so the EKS Pod Identity
  webhook never injects the web-identity token into them, and they have no way to assume the role. You can confirm this
  yourself: helm template ... | grep -B5 serviceAccountName shows it appears exactly once, on the backend Deployment.

  One caveat worth knowing: this ServiceAccount is namespace-scoped, so any pod deployed into the same namespace that
  explicitly sets serviceAccountName: loculus-backend in its own spec would also get these credentials — a pod can't
  reference a ServiceAccount from a different namespace, but within the namespace, referencing the SA name is all it
  takes (same trust model as any k8s Secret). If that namespace is exclusively used by this Helm release, that's a
  non-issue in practice.

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/config/BackendSpringConfig.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/config/BackendSpringConfig.kt
@@ -104,13 +104,19 @@ class BackendSpringConfig {
         @Value("\${${BackendSpringProperty.S3_BUCKET_INTERNAL_ENDPOINT}:#{null}}") internalEndpoint: String? = null,
         @Value("\${${BackendSpringProperty.S3_BUCKET_REGION}}") region: String? = null,
         @Value("\${${BackendSpringProperty.S3_BUCKET_BUCKET}}") bucket: String? = null,
-        @Value("\${${BackendSpringProperty.S3_BUCKET_ACCESS_KEY}}") accessKey: String? = null,
-        @Value("\${${BackendSpringProperty.S3_BUCKET_SECRET_KEY}}") secretKey: String? = null,
+        @Value("\${${BackendSpringProperty.S3_BUCKET_ACCESS_KEY}:#{null}}") accessKey: String? = null,
+        @Value("\${${BackendSpringProperty.S3_BUCKET_SECRET_KEY}:#{null}}") secretKey: String? = null,
     ): S3Config {
         if (!enabled) {
             return S3Config(false, null)
         }
-        if (endpoint != null && bucket != null && accessKey != null && secretKey != null) {
+        if ((accessKey == null) != (secretKey == null)) {
+            throw IllegalStateException(
+                "S3 bucket access-key and secret-key must either both be set (static credentials) " +
+                    "or both be omitted (to use IAM role credentials instead, e.g. IRSA on EKS).",
+            )
+        }
+        if (endpoint != null && bucket != null) {
             return S3Config(true, S3BucketConfig(endpoint, internalEndpoint, region, bucket, accessKey, secretKey))
         }
         throw IllegalStateException("S3 bucket configurations are incomplete.")
--- a/backend/src/main/kotlin/org/loculus/backend/config/S3Config.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/config/S3Config.kt
@@ -7,6 +7,9 @@ data class S3BucketConfig(
     val internalEndpoint: String?,
     val region: String?,
     val bucket: String,
-    val accessKey: String,
-    val secretKey: String,
+    // accessKey/secretKey are only set when using static credentials (e.g. MinIO). When both
+    // are null, S3Service falls back to the AWS SDK's default credential chain, e.g. IAM role
+    // credentials from IRSA on EKS.
+    val accessKey: String?,
+    val secretKey: String?,
 )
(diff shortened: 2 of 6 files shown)


anna-parker:
steps by claude (need to be confirmed)

## Step 1 — Confirm the cluster has an OIDC provider (prerequisite for IRSA)

```
  aws eks describe-cluster --name <cluster-name> --query "cluster.identity.oidc.issuer" --output text
  # -> https://oidc.eks.<region>.amazonaws.com/id/<oidc-id>
```

  If no OIDC provider is associated yet with IAM:

```
  eksctl utils associate-iam-oidc-provider --cluster <cluster-name> --approve
```
## Step 2 — Create the IAM policy
```
  {
    "Version": "2012-10-17",
    "Statement": [
      {
        "Sid": "LoculusBackendFileObjects",
        "Effect": "Allow",
        "Action": ["s3:GetObject", "s3:PutObject", "s3:DeleteObject", "s3:PutObjectTagging"],
        "Resource": "arn:aws:s3:::<bucket-name>/files/*"
      }
    ]
  }
```

```
  aws iam create-policy \
    --policy-name loculus-backend-s3 \
    --policy-document file://loculus-backend-s3-policy.json
```

## Step 3 — Create the role + attach the policy, scoped to exactly the backend's ServiceAccount

  Easiest is eksctl, using --role-only so it creates just the IAM role (trust policy included) without touching
  Kubernetes — the Helm chart already owns creating the loculus-backend ServiceAccount:

```
  eksctl create iamserviceaccount \
    --cluster <cluster-name> \
    --namespace <namespace> \
    --name loculus-backend \
    --attach-policy-arn arn:aws:iam::<account-id>:policy/loculus-backend-s3 \
    --role-only \
    --role-name loculus-backend-s3 \
    --approve
```
  Without eksctl, the equivalent is aws iam create-role with a trust policy like:
```
  {
    "Version": "2012-10-17",
    "Statement": [{
      "Effect": "Allow",
      "Principal": {
        "Federated": "arn:aws:iam::<account-id>:oidc-provider/oidc.eks.<region>.amazonaws.com/id/<oidc-id>"
      },
      "Action": "sts:AssumeRoleWithWebIdentity",
      "Condition": {
        "StringEquals": {
          "oidc.eks.<region>.amazonaws.com/id/<oidc-id>:sub": "system:serviceaccount:<namespace>:loculus-backend",
          "oidc.eks.<region>.amazonaws.com/id/<oidc-id>:aud": "sts.amazonaws.com"
        }
      }
    }]
  }
```
  then 
```
aws iam create-role --role-name loculus-backend-s3 --assume-role-policy-document file://trust-policy.json
```
  followed by aws iam attach-role-policy. The <namespace> and SA name (loculus-backend) here must exactly match where
  you deploy the chart — that's the entire access boundary, so get it right.

## Step 4 — Get the role ARN and wire it into your environment's Helm values

```
  aws iam get-role --role-name loculus-backend-s3 --query Role.Arn --output text
```
```
  runDevelopmentS3: false
  s3:
    credentialsSecretName: "" # disables static creds, forces DefaultCredentialsProvider
    serviceAccountAnnotations:
      eks.amazonaws.com/role-arn: arn:aws:iam::<account-id>:role/loculus-backend-s3
    bucket:
      endpoint: https://s3.<region>.amazonaws.com
      region: <region>
      bucket: <bucket-name>
```
## Step 5 — Deploy and verify
  
```
  helm upgrade --install loculus kubernetes/loculus -n <namespace> -f values.yaml -f <your-env-values>.yaml
  kubectl -n <namespace> get sa loculus-backend -o yaml          # annotation present?
  kubectl -n <namespace> exec deploy/loculus-backend -- env | grep AWS_ROLE_ARN   # webhook injected it?
```
  Then exercise the actual flow — call /files/request-upload, PUT to the returned presigned URL, then /files/get/... to
  confirm the download redirect works — to confirm the role's permissions are sufficient end-to-end, not just that the
  pod started.

## CORS
  One thing this doesn't cover, separate from IAM: if the website PUTs directly to presigned URLs from the browser, the
  bucket needs a CORS configuration allowing PUT/GET/HEAD from your website's origin — that's an S3 bucket setting, not
  an IAM concern, easy to forget and would show up as a browser-side CORS error rather than an AWS permissions error.

## Rotation
https://loculus.slack.com/archives/C0B95CH8BEG/p1783370769006269?thread_ts=1783369898.868879&cid=C0B95CH8BEG

PR #6849, opened 2026-07-07
fix(prepro): Pass noInput from config to prepro, use to construct error message

### LINKED ISSUE #6793: When not providing sampleCollectionDate field, error message for user says "Please provide input metadata field(s): `ncbiReleaseDate`"
When not providing sampleCollectionDate field, error message for user says "Please provide input metadata field(s): `ncbiReleaseDate`"

I don't think that makes sense. This seems to be a wrongly constructed error message in prepro

Image

maverbiest:
resolves #6793

## Background 

When a required metadata field is not provided, the submission is rejected and an error message is shown. If the required field has other input_fields besides the output_field itself, those are included in the error message to tell users what inputs they need to provide. 

This can lead to an unhelpful & misleading error message in cases where one of the input_fields is actually not a field that can be provided by users. See below, where the user is asked to provide `ncbiReleaseDate`, which is not a field they can input:

grafik

## Solution

This PR threads the `noInput` field from the confiug through to preprocessing, where this information is used to construct an error message that only asks users to provide data they can actually specify. I also added a test that covers this behaviour.

image

### PR Checklist

🚀 Preview: https://fix-prepro-error-messages.loculus.org

### DIFF
--- a/preprocessing/nextclade/src/loculus_preprocessing/config.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/config.py
@@ -24,6 +24,9 @@ logger = logging.getLogger(__name__)
 CLI_TYPES = [str, int, float, bool]
 
 METADATA_DEPENDENCY_PREFIX = "processed."
+NEXTCLADE_PREFIX = "nextclade."
+ASSIGNED_REFERENCE_PREFIX = "ASSIGNED_REFERENCE"
+INTERNAL_INPUT_PREFIXES = (NEXTCLADE_PREFIX, ASSIGNED_REFERENCE_PREFIX)
 
 
 class EmblInfoMetadataPropertyNames(BaseModel):
@@ -186,10 +189,11 @@ class Config(BaseModel):
             raise Exception
         return datasets[0]
 
-    def is_submittable(self, field: str) -> bool:
-        if (spec := self.processing_spec.get(field)) is None:
-            return False
-        return not spec.no_input
+    def is_user_input(self, field: str) -> bool:
+        if (spec := self.processing_spec.get(field)) is not None:
+            return not spec.no_input
+        # fields without a spec may be `extraInputFields` and therefore still user input
+        return not field.startswith(INTERNAL_INPUT_PREFIXES)
 
 
 def set_sequence_name(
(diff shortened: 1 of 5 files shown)


maverbiest:
@claude review

maverbiest:
> Possible bug: output field can still appear in its own "please provide" list

This was intentional and I disagree with @claude 's feedback here. What if we have a situation where field A depends on inputs A and B (both are submittable by users). If the user provides neither input, the message becomes:

```
Metadata field `A` is required. Please provide input metadata field(s): `B`
```

This makes it seem like the user can fix the error by only providing `B`, which is incorrect as they must also supply `A`. I'm happy with a little bit of redundancy here as I feel it's a much clearer message:


```
Metadata field `A` is required. Please provide input metadata field(s): `A`, `B`
```

preprocessing/nextclade/src/loculus_preprocessing/config.py
```diff
@@ -185,6 +186,11 @@ def get_dataset_by_name(self, name: str) -> NextcladeSequenceAndDataset:
             raise Exception
         return datasets[0]
 
+    def is_submittable(self, field: str) -> bool:
```
anna-parker: I wonder if `is_user_input` would be a better name

preprocessing/nextclade/src/loculus_preprocessing/config.py
```diff
@@ -185,6 +186,11 @@ def get_dataset_by_name(self, name: str) -> NextcladeSequenceAndDataset:
...
+        if (spec := self.processing_spec.get(field)) is None:
```
anna-parker: ah I think this is an issue `extraInputFields` arent sent to prepro and are thus dropped from here which we dont want and leads to wrong error messages when the input should come from that field:

Image

maverbiest: Okay I think I fixed this with my latest commit. I'll check it on the preview

maverbiest: grafik

anna-parker (approved):
thanks!

PR #6851, opened 2026-07-07
fix(website): restore toasts in astro dev after v6 upgrade

### LINKED ISSUE #6850: Toasts no longer display in `astro dev` after v6 upgrade
I just noticed the toasts have disappeared from the website when running the astro development server (`npm run dev`). Running from a build of the website (`npm run build` && `npm run preview`) still works just fine!

Haven't been able to find a fix for this yet, but I have traced the issue back to beginning after the Astro v6 upgrade (#6779).

theosanderson-agent:
Fixes #6850.

## Summary

Toasts stopped rendering under `astro dev` — but not in a production build (`npm run build` && `npm run preview`) — since the Astro v6 upgrade (#6779). This restores them.

## Root cause

A react-toastify **module-instance split** in Vite's dev server.

`ToastContainer` was imported from `react-toastify` and used directly as a `client:load` island in `BaseLayout.astro`. In `astro dev`:

- A bare framework import referenced **directly by a client directive in an `.astro` file** is served as **raw source**: `/node_modules/react-toastify/dist/index.mjs`.
- The `toast()` callers live in `.tsx` islands, whose imports go through **Vite's dependency optimizer**: `/node_modules/.vite/deps/react-toastify.js`.

These are two different module instances. react-toastify keeps its toast queue in module-level state, so `toast()` pushes into the optimized-dep instance's store while `ToastContainer` (raw-source instance) subscribes to a *different* store — nothing renders.

The production build bundles everything into a single chunk, so the store is shared and toasts work. That's exactly why only dev was affected.

I confirmed this in the running dev server by inspecting the island hydration scripts:

```
# before the fix
ToastContainer island  -> component-url="/node_modules/react-toastify/dist/index.mjs"   (raw source)
toast() caller (.tsx)  -> import { toast } from "/node_modules/.vite/deps/react-toastify.js?v=..."   (optimized)
```

## Fix

Wrap `ToastContainer` in a local `AppToastContainer.tsx` component and use that as the island. Its import now goes through the same optimized dependency as the `toast()` callers, so they share one store. The react-toastify CSS import moves into the wrapper alongside it.

```
# after the fix — both resolve to the same instance
AppToastContainer.tsx  -> import { ToastContainer } from "/node_modules/.vite/deps/react-toastify.js?v=..."
toast() caller (.tsx)  -> import { toast }          from "/node_modules/.vite/deps/react-toastify.js?v=..."
```

## Testing

Reproduced and verified end-to-end in `astro dev` with a headless browser on a minimal two-island page (a button calling `toast()` + the container):

- **Direct `ToastContainer` island** (the previous pattern): clicking the button rendered **no** toast (`.Toastify__toast` absent).
- **Wrapped `AppToastContainer` island** (this fix): the toast rendered with its text.

Also verified against a real `BaseLayout` page that the toast island now resolves to the optimized dependency, and that `astro check` + `tsc` pass with 0 errors.

### DIFF
--- a/website/src/components/common/AppToastContainer.tsx
+++ b/website/src/components/common/AppToastContainer.tsx
@@ -5,17 +5,8 @@ import 'react-toastify/dist/ReactToastify.css';
 // Wrapping `ToastContainer` in a local component (rather than using
 // `react-toastify`'s `ToastContainer` directly as a `client:*` island in an
 // `.astro` file) keeps it on the same react-toastify module instance as the
-// `toast()` callers.
+// `toast()` callers when running in dev mode.
 //
-// In `astro dev`, a bare framework import referenced directly by a client
-// directive in an `.astro` file is served as raw source
-// (`/node_modules/react-toastify/dist/index.mjs`), while imports inside
-// `.tsx`/`.jsx` islands go through Vite's dependency optimizer
-// (`/node_modules/.vite/deps/react-toastify.js`). Those are two different
-// module instances, each with its own module-level toast store, so
-// `toast()` calls never reach the `ToastContainer` and no toast is shown.
-// The production build bundles everything into one chunk, so it only breaks
-// in dev. See https://github.com/loculus-project/loculus/issues/6850
 export default function AppToastContainer() {
     return <ToastContainer />;
 }
--- a/website/src/components/common/AppToastContainer.tsx
+++ b/website/src/components/common/AppToastContainer.tsx
@@ -1,0 +1,21 @@
+import { ToastContainer } from 'react-toastify';
+
+import 'react-toastify/dist/ReactToastify.css';
+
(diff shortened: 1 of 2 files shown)


tombch:
Nice, just tested and it seems to work!

theosanderson:
@theosanderson-agent get CI passing

theosanderson-agent:
Fixed the failing **Check format and types** job. The `eslint` `import/order` rule flagged that `AppToastContainer` (a `common/` import) was ordered before the `Navigation/` imports in `BaseLayout.astro`. Moved it after `Navigation/cleanOrganism` to satisfy the rule. Verified locally with `eslint` and `prettier --check` — both pass. Pushed as 92258f3c0.

tombch (approved):
Tested locally and looks good to me, nice

PR #6853, opened 2026-07-07
feat(website): normalise case before accession lookup on the /seq page

### LINKED ISSUE #6852: Case insensitive accession search on front page
I just tried to search for an accession as `pp_00123` instead of `PP_00123` and got "no accession found" - it would be nice to be case insensitive when the accession prefix is all-uppercase.

theosanderson-agent:
## Summary

Closes #6852.

Searching for an accession with the wrong letter case (e.g. `pp_00123` instead of `PP_00123`) previously returned "no data found".

## How it works

When the configured accession prefix is all-uppercase (which is the case for the standard Loculus prefixes like `PP_`, `LOC_`, etc.), accessions are effectively case-insensitive — the only letters in a Loculus accession are the fixed prefix followed by digits. In that situation the `/seq` page:

1. Uppercases the incoming `accessionVersion`.
2. If that changed the value, issues a redirect to the canonical URL (`/seq/PP_00123`), so the address bar and all downstream links/downloads use consistent casing.

If a deployment configures a prefix that itself contains lowercase letters, the input is left untouched to avoid breaking a mixed-case scheme.

🚀 Preview: https://feat-case-insensitive-acc.loculus.org

### DIFF
--- a/website/src/pages/seq/[accessionVersion]/index.astro
+++ b/website/src/pages/seq/[accessionVersion]/index.astro
@@ -18,10 +18,6 @@ import { getSegmentNames } from '../../../utils/sequenceTypeHelpers';
 
 const rawAccessionVersion = Astro.params.accessionVersion!;
 
-// Accessions are case-insensitive when the configured prefix is all-uppercase.
-// Normalize the case here (rather than at each entry point, e.g. the search
-// box) so that any way of reaching this page resolves the same entry, then
-// redirect to the canonical URL so the address bar and downstream links match.
 const accessionVersion = normalizeAccessionCase(rawAccessionVersion, getWebsiteConfig().accessionPrefix);
 if (accessionVersion !== rawAccessionVersion) {
     return Astro.redirect(`/seq/${accessionVersion}`);
--- a/website/src/pages/seq/[accessionVersion]/index.astro
+++ b/website/src/pages/seq/[accessionVersion]/index.astro
@@ -13,9 +13,20 @@ import { type Group } from '../../../types/backend';
 import { parseAccessionVersionFromString } from '../../../utils/extractAccessionVersion.ts';
 import { getAccessToken } from '../../../utils/getAccessToken';
 import { getMyGroups } from '../../../utils/getMyGroups';
+import { normalizeAccessionCase } from '../../../utils/normalizeAccessionCase.ts';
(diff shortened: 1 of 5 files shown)


website/src/pages/seq/[accessionVersion]/index.astro
```diff
@@ -13,9 +13,16 @@ import { type Group } from '../../../types/backend';
...
 import { getMyGroups } from '../../../utils/getMyGroups';
+import { normalizeAccessionCase } from '../../../utils/normalizeAccessionCase.ts';
 import { getSegmentNames } from '../../../utils/sequenceTypeHelpers';
 
-const accessionVersion = Astro.params.accessionVersion!;
+const rawAccessionVersion = Astro.params.accessionVersion!;
+
+const accessionVersion = normalizeAccessionCase(rawAccessionVersion, getWebsiteConfig().accessionPrefix);
+if (accessionVersion !== rawAccessionVersion) {
+    return Astro.redirect(`/seq/${accessionVersion}`);
```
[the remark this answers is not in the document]
Comment (theosanderson) [reply website/src/pages/seq/[accessionVersion]/index.astro:23]: (think this is ok for now)

website/src/utils/normalizeAccessionCase.ts
tombch, line 1: I am wondering, would it not be preferable for this to substitute in the correct casing for any prefixes, uppercased or otherwise? I feel like if the prefix was `Loc_` and someone searches `loc_`, that should also be correctly redirected

theosanderson: TBH I think we should probably mandate uppercase prefixes - our accessions use upper case characters in the "id" portion anyway, and IMO accessions should be all one case. But that's a bigger change so I didn't make it yet. I'll make an issue now. The change here is lower impact and doesn't have downsides as I see it.

tombch: Yeah, uppercase prefix that can be searched case insensitively sounds like the ideal approach in the long run. Ok makes sense!

tombch (approved):

website/src/utils/normalizeAccessionCase.ts
theosanderson: forgot to link the issue :) https://github.com/loculus-project/loculus/issues/6857

PR #6855, opened 2026-07-07
fix(website): Disable submit button for submission/revision/edit during file uploads

### LINKED ISSUE #6766: MVP: Can submit before file uploads finish, silently dropping files
Reported by @theosanderson. Reproducible on `main` and **independent of the bulk-revision file-reuse work**.

## Symptom

On the submission page for an organism with file uploads, you can click **"Upload and proceed to Approval"** while files are still uploading. The submission goes through, but the entry ends up with only the files that had finished uploading at that moment (and possibly none) — the rest are **silently dropped**, with no error.

## Root cause

The submit button is only disabled while the submit request itself is pending — it is not gated on the file-upload state:

```tsx
// website/src/components/Submission/DataUploadForm.tsx:218
<Button name='submit' type='submit'
    onClick={(e) => void handleSubmit(e)}
    alsoDisabledIf={isPending}        // <-- only the submit request, not the uploads
>
    ... Upload and proceed to Approval
</Button>
```

`FolderUploadComponent` reports the completed `fileMapping` upward (via `setFileMapping`) only once its uploads reach the `UploadCompleted` state. So if the user submits mid-upload, `handleSubmit` sends whatever `fileMapping` is currently set (partial or `undefined`), and the backend attaches only those files.

## Suggested fix

Surface an "uploading in progress" state from `FolderUploadComponent` / `ExtraFilesUpload` up to `DataUploadForm`, and include it in the submit button's disabled condition (so submit is blocked until every file upload has reached `UploadCompleted`). Ideally also show a hint explaining why submit is disabled. Consider guarding `handleSubmit` itself as a backstop.

tombch:
resolves #6766

### PR Checklist


tombch:
Superseded by #7050 and #7053

PR #6874, opened 2026-07-09
feat(deposition): no longer special case broker for manifest creation

anna-parker:
While working on https://github.com/loculus-project/loculus/pull/6816 I realized that although we test on a non-broker account we are still able to supply the authors field to the manifest and have a successful submission to ENA dev. There is a lot of special casing in the code because this used to fail. It seems like this works now so it makes sense to remove the special casing. 

I left the `is_broker` config field but we can now even remove it, what are your thoughts?

### PR Checklist


ena-submission/scripts/deposition_dry_run.py
```diff
@@ -167,15 +166,14 @@ def local_ena_submission_generator(
...
+        create_manifest(manifest_object, dir=directory)
         logger.info(
             "You can submit the assembly to ENA using the command: \n"
             "ena-webin-cli -username $ena_submission_username "
             "-password $ena_submission_password -context genome "
             "-manifest assembly/manifest.tsv -submit "
             f"-centername {center_name}"
             "\n Remember to submit with -test if you do not want to submit to production"
-            "\n Remember to add `is_broker=true` to config"
             "if you are submitting as a broker"
```
anna-parker:
```suggestion
```

corneliusroemer (dismissed):
I trust you tested we no longer need this :) I'd remove the config option if it's no longer used.

Note there are merge conflicts so big can't solve it in web editor.

corneliusroemer (commented):
Actually I'm no longer so sure it's good to remove the broker flag. IIUC you only checked submission to dev - and there the assembly doesn't process fully.

Per docs, there are genuine broker-only fields. The question is whether prod ENA fails if you submit with those fields as non-broker.

Maybe we should write email to helpdesk to ask for clarification if authors would just be dropped or if there was real error.

Claude:

<details>

The PR description (anna-parker) says:

▎ "although we test on a non-broker account we are still able to supply the authors field to the manifest and have a successful submission to ENA dev."

So the entire evidence base for "the broker special-casing isn't necessary" is ENA dev (their test/staging endpoint) — exactly the case you were worried about. There is no evidence in the PR that a non-broker account can set AUTHORS/ADDRESS against production ENA.

What the ENA docs / source actually say

1. AUTHORS and ADDRESS are genuinely broker-only fields. The Webin-CLI source labels them verbatim: "For submission brokers only. Submitter's names..." / "...Submitter's address". They aren't even listed on the standard genome-assembly manifest doc page — only in the data-brokering FAQ, which says a broker account is required to supply the author list.
2. The broker check is enforced server-side, not by the client. Webin-CLI does not check broker status — it just forwards AUTHORS/ADDRESS to whichever endpoint the -test flag selects (wwwdev.ebi.ac.uk vs www.ebi.ac.uk). So "it worked on dev" tells you nothing definitive about production; the acceptance/rejection decision lives in ENA's closed-source submission service, which can differ between the two hosts.
3. Whether dev and prod enforce this identically is undocumented. I (via a research agent, cross-checking official docs, GitHub, biostars, helpdesk posts) could not find any statement confirming or denying that dev skips the check while prod enforces it. It's genuinely unverifiable from public sources — it would need an ENA helpdesk question (support@ebi.ac.uk) or an actual non-broker submission against production.

What this means for the code

After the PR, create_manifest_object always calls get_address(...) and always populates authors (via the manifest field mapping), and create_manifest writes those lines unconditionally. So every deployment now sends broker-only fields.

The nuance that matters for your risk assessment:
- Pathoplexus prod is a broker account (you broker submissions on behalf of users), so for your instance this PR is safe and even correct.
- The risk is for other Loculus deployments using a non-broker account. The is_broker flag existed precisely so non-broker instances could omit these fields. Notably, no values file in the repo ever set enaIsBroker: true, and the default was False — so the default distributed behavior was "don't send authors/address," and this PR flips that to "always send them."

Recommendation

The safe answer is: the broker guard may still be necessary for non-broker production submissions, and "it works on dev" doesn't prove otherwise. Before relying on removal, I'd do one of:

1. Ask ENA helpdesk whether production rejects AUTHORS/ADDRESS from a non-broker Webin account (definitive, ~1 email).
2. Keep the is_broker flag as a safety valve for non-broker deployments (revert my cleanup commit and possibly the parent PR too), even if Pathoplexus itself always runs as a broker.
3. Merge as-is only if the project is comfortable that every real deployment submitting to prod is a broker.

</details>

Issue #6760, opened 2026-06-24
MVP Raw reads: documentation


anna-parker:
Ensure raw reads submission flow is clearly documented on the website and in ppx docs

1. Add a how to upload raw reads doc for PPX like the how to submit sequences doc: https://pathoplexus.org/docs/how-to/upload-sequences
2. Ensure that there is proper documentation on the format of raw reads on the submission page itself with a link to the raw reads submission page (like for sequence and metadata submission above)
<img width="2080" height="1460" alt="Image" src="https://github.com/user-attachments/assets/358e8c09-31d2-4e45-a366-ad01f4fd262c" />
3. On the metadata page add a new section for metadata that is recommended for raw reads using the issue https://github.com/loculus-project/loculus/pull/6634

PR #6876, opened 2026-07-09
feat(prepro): Add deacon

maverbiest:
Partially resolves https://github.com/loculus-project/loculus/issues/6758

### Summary

This PR incorporates [Deacon](https://github.com/bede/deacon#quickstart) into the preprocessing pipeline. Deacon is used to check submitted raw reads files for host contamination and rejects files where the proportion of raw reads above a threshold configured via `preprocessing.configFile.deacon_max_host_proportion` in the values.yaml.

### Details
- Memory cap for preprocessing pods is raised to accomodate loading the 3.1G panhuman index into RAM (set it at 8G for now)
- `FileIdAndName` -> `FileIdAndNameAndReadUrl` (same as the backend type). This is because prepro now needs to download raw reads from S3 so it needs the URLs
- At startup, preprocessing downloads the `panhuman-1.k31w15.idx` that's needed for host depletion
- When preprocessing gets a batch of sequences to process, a deacon server is started. The server is shut down again after a batch is processed. This prevents the 3.1G index from sitting in RAM when it's not used. Downside is that it will need to be read from disk for each batch of sequences (100 seqs for most organisms), which takes ~6 seconds
- Deacon filter is run on each (pair of) raw reads file(s), a summary of the run is output as json and loaded in to a `DeaconSummary`. This information is used for deciding to reject a file or not
- A new test file is added to cover the new functionality (Deacon calls are mocked, however)

### Manual testing

I monitored the prepro logs in Grafana to confirm that:
- Index gets fetched once at prepro startup
- Deacon server starts/stops as expected when a batch of unprocessed sequences is received from the backend
- Deacon filter runs

I also started a preview to check that:
- Raw reads files without host sequences are not rejected
- Raw reads files with too many host sequences get rejected
- Raw reads file is attached to submission and downloadable from the sequence details page after release

Here, the first submission has file `ERR17072040.fastq.gz`, which is just west-nile sequences. The second submission (with the error) has the same file where I added 10000 reads sourced from the human reference genome:

grafik

When the sequence is released, the raw reads file is accessible on the sequence details page:

grafik

### PR Checklist
~- [ ] All necessary documentation has been adapted.~ -> will be added in https://github.com/loculus-project/loculus/issues/6760


maverbiest:
@claude review

preprocessing/nextclade/src/loculus_preprocessing/file_processing_functions.py
```diff
@@ -74,4 +92,113 @@ def validate_raw_reads_submission(
...
+                logger.error(message)
+                errors.append(
+                    ProcessingAnnotation.from_single(
+                        name=file.name, type=AnnotationSourceType.FILE, message=message
+                    )
+                )
+                continue
+
+            summary_json_path = os.path.join(tmp_dir, f"{file.fileId}_summary.json")
+            deacon_summary = run_deacon_filter(deacon_index, file_name_internal, summary_json_path)
```
anna-parker: are you running deacon for each file individually? Paired reads should be run together: https://github.com/bede/deacon#quickstart

maverbiest: Sure I'll update! This was just for simplicity as it shouldn't make a difference

kubernetes/loculus/values_preview_server.yaml
```diff
@@ -156,3 +156,23 @@ defaultResources:
...
+  dummy-organism:
+    enabled: false
+  dummy-organism-with-files:
+    enabled: false
+  ebola-sudan:
+    enabled: false
+  enteroviruses:
+    enabled: false
+  not-aligned-organism:
+    enabled: false
```
corneliusroemer, line 178: This would disable all preview organisms but west nile, which we don't want.

So needs to be reverted before going on main

kubernetes/loculus/values.schema.json
```diff
@@ -814,6 +814,14 @@
                     "type": "string",
                     "description": "Minimizer used for nextclade sort (if require_nextclade_sort_match or if segments are to be assigned using nextclade sort)"
                   },
+                  "deacon_max_host_proportion": {
```
anna-parker, line 817: we probably need to configure two thresholds.
1. Which percentage of reads are flagged
2. How many bp are flagged

preprocessing/nextclade/src/loculus_preprocessing/datatypes.py
```diff
@@ -265,3 +267,19 @@ class MoleculeType(StrEnum):
 class Topology(StrEnum):
     LINEAR = "linear"
     CIRCULAR = "circular"
+
+
+@dataclass
+class DeaconSummary:
```
anna-parker, line 273: its also important to know how many bp were flagged

preprocessing/nextclade/tests/test_file_processing.py
```diff
@@ -0,0 +1,192 @@
...
+@pytest.fixture
+def config() -> Config:
+    return Config(
+        submission_file_categories=[FileCategory.RAW_READS],
+        deacon_max_host_proportion=MAX_HOST_PROPORTION,
+    )
+
+
+@pytest.fixture
+def mock_deacon(monkeypatch):
```
anna-parker, line 31: It might actually be good to create a little deacon index and example raw reads for testing. We also do that with nextclade as I noticed if we dont actually run nextclade in tests we miss things

maverbiest:
Closing as we merged another branch with Deacon functionality

PR #6920, opened 2026-07-16
feat(website): improve SeqSet citation metadata

theosanderson-agent:
## Summary

SeqSet citation author lists now retain up to ten names and use an ellipsis when additional contributors exist. This keeps citations readable without hiding normal-sized author lists.

The citation model now retains journal metadata from Crossref, persists it in the database, exposes it through the API, and displays it in public and administrative citation views. Manually curated citations can also supply a journal, including DOI metadata lookups.

## Why

Citation views previously rendered every contributor, causing unusually long paper author lists to dominate the page. The backend also discarded Crossref journal metadata, leaving users without a key publication detail.
org.loculus.backend.service.crossref.CrossRefServiceTest --console=plain`

## Preview evidence

Filled manual citation form

Saved citation showing journal and truncated authors

Sequence details “Cited in” section with truncated authors and journal

🚀 Preview: https://agent-seqset-citation-aut.loculus.org

### DIFF
--- a/website/src/components/SeqSetCitations/formatCitationContributors.ts
+++ b/website/src/components/SeqSetCitations/formatCitationContributors.ts
@@ -1,0 +1,16 @@
+import type { CitationContributor } from '../../types/seqSetCitation';
+
+const MAX_DISPLAYED_CONTRIBUTORS = 10;
+
+const formatContributor = ({ givenName, surname }: CitationContributor) =>
+    [givenName, surname].filter((name) => name).join(' ');
+
+export const formatCitationContributors = (contributors: CitationContributor[]) => {
+    const formattedContributors = contributors.map(formatContributor).filter((name) => name);
+
+    if (formattedContributors.length <= MAX_DISPLAYED_CONTRIBUTORS) {
+        return formattedContributors.join(', ');
+    }
+
+    return `${formattedContributors.slice(0, MAX_DISPLAYED_CONTRIBUTORS).join(', ')}, ...`;
+};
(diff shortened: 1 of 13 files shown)


website/src/components/SeqSetCitations/formatCitationContributors.ts
```diff
@@ -0,0 +1,16 @@
...
+    [givenName, surname].filter((name) => name).join(' ');
+
+export const formatCitationContributors = (contributors: CitationContributor[]) => {
+    const formattedContributors = contributors.map(formatContributor).filter((name) => name);
+
+    if (formattedContributors.length <= MAX_DISPLAYED_CONTRIBUTORS) {
+        return formattedContributors.join(', ');
+    }
+
+    return `${formattedContributors.slice(0, MAX_DISPLAYED_CONTRIBUTORS).join(', ')}, ...`;
```
tombch, line 15: Could this possibly display a `Show more` / `Show less` button when there's over ten authors, similar to how the nucleotide mutations are displayed?

theosanderson: I don't think this is the best UX in this case. It's not a pattern I've seen in any other list of citations. I think in future we might even limit this further to be just a single line of authors.

Google Scholar:
image

Nature reference list:
image

If people want an expanded list they will naturally click through to the publication. Having too many different interactive elements I think adds to the mental load of the page. For mutations, and for authors of the sequence itself, there is nothing to click through to.

tombch: Imo it wouldn't add much to the mental load of the page, but yeah I can see this would not line up with other existing citation displays. In which case, truncating sounds good!

tombch (approved):
Looks good to me!

Issue #5820, opened 2026-01-09
MVP Revise in bulk reusing existing files


chaoran-chen:
When revising entries in bulk, it should be possible to reuse previously uploaded files without the need to upload them again.

It's not obvious to me how the UI should look like and would appreciate suggestions.

theosanderson:
I think we plan to move to a model in which when revising metadata entries in bulk you can supply any number of columns, and if you miss out a column it stays as is from previous version. Equivalently, maybe this file feature for revisions is the same as the current bulk submission UI and if you don't upload files, or you upload a directory which doesn't contain a folder for a particular submissionId, the files stay as in the previous version. The idea would be that if you wanted to delete the files associated with a sequence you would upload with a submission ID with an empty folder (I'm not entirely sure if we can detect that or not though), and if you wanted to replace them you would supply a filled folder.

anna-parker:
I have been discussing the next steps here with @tombch, we think we should allow users to do the following:
1. delete all existing files (basically the current set up)
2. modify existing files
3. keep all existing submitted files as is 

We discussed the first step would be to support deleting and modifying. 
1. Add `delete all linked files` button when `files` are enabled to make deletion clear in the UI
2. Add the upload files folder component (same as for submission), users would have to re-upload all files submitted previously to keep them in the revision and could modify and delete specific files

anna-parker:
The second, more complicated step would be to allow submitters to keep all existing files. This is more complicated as we need to know which files were linked to each previous submission.

1. Add the fileMapping to the `get-original-metadata` endpoint for API users. This means an additional fileMapping.json will be downloaded with original metadata (when `files` is enabled)
2. User must now either upload fileMapping or "discard fileMapping" (could then upload a brand new folder - page would show folder as before)

anna-parker:
As an even more fancy third step, when submitters click the "reuse files [fileMapping]" button this could open the folder component (as is currently implemented) and allow users to select a submissionId, see the files uploaded for that submissionId and delete and upload additional files for that submissionId

theosanderson:
Also heard from @tombch about the discussion - which generally sounded good. I agree that the hashes thing is over-complicated at least for an initial version.

Here is one version I thought I heard from Tom that would sound good to me:

[...]

PR #6922, opened 2026-07-17
feat!(website, backend): Update bulk file submission/revision format to use metadata file

### LINKED ISSUE #6883: MVP: feat(website, backend): Update bulk file submission format
## Backstory

As noted by @tombch in https://loculus.slack.com/archives/C0B95CH8BEG/p1783590528630039 our current model does not allow `/` in submissionIds when users submit files. This can be challenging for backward compatibility (e.g. ingest submits with '/' and existing direct submissions have an '/').

This prompted a discussion on whether we want to continue with our current bulk file submission format, as it also was problematic for bulk revision. Currently we require a folder with a subfolder with the same name as the `submissionId`. 

We had a long Zoom discussion about the bulk upload and revision of raw reads and subsequent slack discussion in https://loculus.slack.com/archives/C05G172HL6L/p1783952073472099. We also asked microbio info how users would like to submit data. The leading option was to have file mapping info in a separate manifest or column. The current approach with subfolders was disliked. The manifest option did win over columns but we believe having the fileMapping in a metadata.tsv column is easier for us to implement and more inline with our current `fastaIds` column structure to link multi-segmented submissions.

## Solution

The current consensus is for a metadata columns called say `files.raw_reads` that will at submission contain a space-separated list of filenames of read files, e.g. `SAG_123.fq SAG_456.fq`. (Reads can then be supplied just as a bulk set of files). Each fileCategory will need its own column.

At bulk revision when you download the existing data in order to revise it, the metadata will contain `SAG_123.fq:8723478-21454-134324 SAG_456.fq:123223-5666-324342`  where those IDs are references to the existing files. If you leave them as is [90% use-case] the files will remain the same. If you wanted to delete you just delete the relevant string. If you wanted to add a further file you would change to `SAG_123.fq:8723478-21454-134324 SAG_456.fq:123223-5666-324342 my_new_file.fq`  and upload `my_new_file.fq`

At any stage (e.g. initial submission) you can also manually upload files with the endpoints and directly use the `file_name:file_id` style.

We discussed how we could move to manifests (winner of https://microbial-bioinfo.slack.com/archives/CB0HYT53M/p1783599228294719 - and since we discussed an extra vote and now winning by 2) later in a pretty backwards compatible way.

## Submission Flow (Website)
1. Submitters upload metadata, sequences and one folder with all their raw reads. The metadata file should contain a column called `files.raw_reads` with a space-separated list of fileNames of files in the folder that should be linked to that entry.
2. Website requests predesigned URLs for each file in the folder and adds the UUID to the column. Website sends the sequences.fasta and metadata.tsv to the backend with some simple validation that all the files in the folders are in the column and that all the column-values have been uploaded. No additional `fileMapping` JSON creation is now required.
3. Backend receives the sequences.fasta and metadata.tsv, confirms these are linked properly and the format is correct (backend will assert that every file in the `files.raw_reads` column needs to have a UUID linked to it.
4. Preprocessing receives a JSON with metadata, sequences and fileIds, it downloads files and does basic validation.

## Submission Flow (API)
1. API users request presigned URLs for each file, storing the fileIds.
2. API users prepare files for submission: metadata.tsv and sequences.fasta, manifest.tsv, adding  the UUID and fileNames to the `files.raw_reads` column.
3. Backend receives the sequences.fasta and metadata.tsv, confirms these are linked properly and the format is correct and uploads to tables.
4. Preprocessing receives a JSON with metadata, sequences and fileIds, it downloads files and does basic validation.

## Revision Flow (Website)

1. Users download the metadata.tsv and sequences.fasta files (as part of original data). The metadata.tsv will contain a `files.raw_reads` column with a space separated list of fileName and fileIds of existing files, i.e. `file_name:file_id` style.
2. To delete a file they remove it from the column, to change a file they remove the fileId and upload a file by the same name, to add a new file they update the `files.raw_reads` column with the fileName and add the file to an upload folder. When not modifying raw reads users should not touch the `files.raw_reads` column.
3. Website parses the input. If new files are provided the website will request presigned URLs, upload the new files and add the UUID to the metadata.tsv, it will then upload the sequences.fasta and metadata.tsv to the backend.
Steps 4 and 5 are the same as in submission.

## Revision Flow (API)

1. Users download the metadata.tsv and sequences.fasta files (as part of original data). The metadata.tsv will contain a `files.raw_reads` column with a space separated list of fileName and fileIds of existing files, i.e. `file_name:file_id` style.
2. To delete a file they remove it from the column, to change a file they remove the fileId and upload a file by the same name, to add a new file they request a presigned URL from the backend, upload the file to the presigned URL and update the `files.raw_reads` column with the fileName and fileId. When not modifying raw reads users should not touch the `files.raw_reads` column.
Steps 4 and 5 are the same as in submission.

## Migration for users of file sharing
Existing users of the file sharing feature should update their pipelines to support having one flat folder with all files to be submitted or batch their file submissions to reuse fileNames across sequences.

tombch:
resolves #6883 and #5820

### Breaking changes

**Loculus deployments that carry out bulk submissions with extra files enabled will require users to update their bulk submission metadata file to include files.{category} column(s) that list the files for each submission. Details for how to do this for the current subfolder structure are described below in the 'Examples' section under 'Bulk uploads with subfolders'.**

Additionally **white space is no longer allowed in fileNames**, for loculus users already using the files feature we suggest confirming no fileNames include white space and running the migration listed in the details below.

<details>

Way to find all entries where the fileName has whitespace in processed and submitted data:

```
SELECT
    se.accession,
    f.key AS file_type,
    obj ->> 'name' AS name
FROM public.sequence_entries_preprocessed_data se,
     jsonb_each(se.processed_data -> 'files') AS f(key, value),
     jsonb_array_elements(f.value) AS obj
WHERE obj ->> 'name' ~ '\s'
  AND processed_data -> 'files' IS NOT NULL
  AND processed_data -> 'files' <> 'null'::jsonb
ORDER BY se.accession;
```

and

```
SELECT
    se.accession,
    f.key AS file_type,
    obj ->> 'name' AS name
FROM public.sequence_entries se,
     jsonb_each(se.submitted_data -> 'files') AS f(key, value),
     jsonb_array_elements(f.value) AS obj
WHERE obj ->> 'name' ~ '\s'
  AND submitted_data -> 'files' IS NOT NULL
  AND submitted_data -> 'files' <> 'null'::jsonb
ORDER BY se.accession;
```

This can then be modified to an underscore in the submitted data as follows (the processed data can be changed by a similar command or by reprocessing the data)

```
UPDATE public.sequence_entries se
SET submitted_data = jsonb_set(
    se.submitted_data,
    '{files}',
    (
        SELECT jsonb_object_agg(
            f.key,
            (
                SELECT jsonb_agg(
                    CASE
                        WHEN obj ->> 'name' ~ '\s'
                        THEN jsonb_set(
                            obj,
                            '{name}',
                            to_jsonb(
                                regexp_replace(
                                    obj ->> 'name',
                                    '\s+',
                                    '_',
                                    'g'
                                )
                            )
                        )
                        ELSE obj
                    END
                )
                FROM jsonb_array_elements(f.value) AS obj
            )
        )
        FROM jsonb_each(se.submitted_data -> 'files') AS f(key, value)
    )
)
WHERE se.submitted_data -> 'files' IS NOT NULL
  AND se.submitted_data -> 'files' <> 'null'::jsonb
  AND EXISTS (
      SELECT 1
      FROM jsonb_each(se.submitted_data -> 'files') AS f(key, value),
           jsonb_array_elements(f.value) AS obj
      WHERE obj ->> 'name' ~ '\s'
  );
```
</details>

### Summary

This PR restructures how files are linked to submissions by the website, as well as how file mappings are sent from the website to the backend, in both bulk and form submissions/revisions. Previously, bulk and form website submission/revisions attached a separate file mapping object. Now, users add a list of files to the submitted metadata tsv and the website adds file mappings for each submission before sending the files to the backend for submission/revision.

New `files.{category}` columns are introduced within the metadata, and uploaded files must be linked to within here - this can be done simply by adding a space separated list of: the file names (`name1 name2 name3`, if uploading a flat folder of files) or the file names with the file paths ( `name1::path1 name2::path2 name3::path3`, if uploading a folder containing subfolders - allowing for different submissions to use the same file names within a bulk upload).
 
#### Bulk uploads with a flat folder of files

Bulk uploads of files can be carried out with a single flat folder of uniquely named files - the user just needs to link these files to the correct submission in the `metadata.tsv`. For example, with this folder of `rawReads`:

```
rawReads /
  id_1_reads_1.fastq
  id_1_reads_2.fastq
  id_2_reads_1.fastq
  id_2_reads_2.fastq
  ...
``` 

The user must add the following `files.rawReads` column to their `metadata.tsv`:

 ```
id    ...  files.rawReads
id_1  ...  id_1_reads_1.fastq id_1_reads_2.fastq
id_2  ...  id_2_reads_1.fastq id_2_reads_2.fastq
...
```

A user could also choose to give multiple output files the same name across submissions using the `name::path` syntax:

```
id    ...  files.rawReads
id_1  ...  reads_1.fastq::id_1_reads_1.fastq reads_2.fastq::id_1_reads_2.fastq
id_2  ...  reads_1.fastq::id_2_reads_1.fastq reads_2.fastq::id_2_reads_2.fastq
...
```

#### Bulk uploads with subfolders

The existing bulk uploads mechanism for extra files relied on (for each category) the user uploading a folder containing subfolders, where each subfolder was the submission ID for the files. An example folder for bulk `rawReads` upload:

```
rawReads /
  id_1 /
    reads_1.fastq
    reads_2.fastq
  id_2 /
    reads_1.fastq
    reads_2.fastq
  ...
``` 

This structure can still remain the same - or a user can now upload any subfolder structure. But what is now required is a `files.rawReads` column in the `metadata.tsv`, with each file in a submission referenced by `name::path` where `name` is the desired output file name in Loculus, and `path` is the path to the file in the uploaded folder:

```
id    ...  files.rawReads
id_1  ...  reads_1.fastq::id_1/reads_1.fastq reads_2.fastq::id_1/reads_2.fastq
id_2  ...  reads_1.fastq::id_2/reads_1.fastq reads_2.fastq::id_2/reads_2.fastq
...
```

#### Submission

1. At initial file submission, users can now choose between uploading a folder, or select individual files:

image

2. Subfolders can be submitted using `fileName::filePath` in the metadata file:

image

3. Files referenced in metadata but not uploaded are indicated with a warning and files that are not referenced in the metadata but are uploaded are also warned on:

image

This 'warning' is blocking - submission can't continue until they are resolved (an error toast would be displayed at submit).

#### Revision

1. The `files.{category}` field is included in the downloaded original metadata for revisions, each file has a linked fileId:
image

2. Single sequences can be revised via the edit page: can add more files and delete individual files:
image

3. Bulk revision of sequences and metadata without modifying the files just by not modifying the `files.{category}` field. Allows deletion of files by removing files from the `files.{category}` field, allows renaming of a file by reusing the fileId, allows adding new files (similar to initial submission).

original:
image

revised: 
image

image

### PR Checklist

🚀 Preview: https://bulk-file-submissions.loculus.org

### CODE REVIEW

File: integration-tests/tests/specs/features/file-sharing.spec.ts [hunk 1]
```diff
@@ -8,7 +8,8 @@ import { BulkSubmissionPage, SingleSequenceSubmissionPage } from '../../pages/su
 
 const ORGANISM_NAME = 'Test organism (with files)';
 const ORGANISM_URL_NAME = 'dummy-organism-with-files';
-const RAW_READS = 'rawReads';
+const RAW_READS = 'raw_reads';
```
tombch: Commenting as needs addressing - `rawReads` vs `raw_reads`

website/src/components/Submission/FileUpload/ColumnMapping.ts
```diff
@@ -97,11 +98,21 @@ export class ColumnMapping {
...
             indices.push(headersInFile.findIndex((sourceHeader) => sourceHeader === sourceCol));
         });
+
+        // Include file category columns, as these will not be present in the column mapping
+        headersInFile.forEach((sourceHeader, sourceIndex) => {
+            if (sourceHeader.startsWith(FILES_HEADER_PREFIX) && !mappedSourceColumns.has(sourceHeader)) {
+                headers.push(sourceHeader);
+                indices.push(sourceIndex);
+            }
+        });
```
tombch: Another thing to note - the files category columns currently appear as unmapped in the column mapper - they should probably be recognised as input fields similar to `fastaIds` field

website/src/components/Submission/DataUploadForm.tsx
```diff
@@ -103,20 +125,42 @@ const InnerDataUploadForm = ({
...
+                    onError('Cannot submit: metadata file is still being processed.');
+                    return;
+                }
+
+                if (submissionFileMapping.isErr()) {
+                    onError(submissionFileMapping.error.message);
+                    return;
+                }
+
+                const fileLinkage = getFileLinkage(submissionFileMapping.value, fileMapping);
```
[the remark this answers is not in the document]
tombch: I think this is ok to recompute the file linkage synchronously on submit, the alternative of using the memoised file linkage but with a `fileLinkage!` to assert its not actually undefined seems more risky.

backend/src/main/kotlin/org/loculus/backend/model/SubmitModel.kt
```diff
@@ -174,11 +178,21 @@ class SubmitModel(
                 }
             }
 
-            submissionParams.files?.let { submittedFiles ->
-                submissionMetrics.timeWritePhase(endpoint, organism, VALIDATE_FILE_MAPPING_PHASE) {
-                    val fileSubmissionIds = submittedFiles.keys
-                    validateSubmissionIdSetsForFiles(metadataSubmissionIds, fileSubmissionIds)
-                    validateFileGroupOwnership(submittedFiles, submissionParams, uploadId)
+            submissionMetrics.timeWritePhase(endpoint, organism, VALIDATE_FILE_MAPPING_PHASE) {
+                // File mappings in submissionParams can contain submission Ids not present in the metadata
```
anna-parker, line 185: Ah do I understand correctly that the backend actually still supports the old submission format and will accept having the files in a fileMapping JSON? 

Sounds like a good rollout strategy! It might be good to still have a follow up PR where the JSON is completely removed to ensure the removal will be smooth - I started doing that locally now and does seem like it will still be a bit of a change!

tombch: Yeah, currently I haven't removed any of the separate file mapping stuff from the backend, but for single and bulk submission/revision the website now sends the file mapping within the metadata tsv. The `submit-edited-data` endpoint didn't have to change how it handled the file mapping as that one is completely JSON for everything

backend/src/main/kotlin/org/loculus/backend/utils/MetadataEntry.kt
```diff
@@ -78,6 +85,78 @@ fun extractAndValidateFastaIds(record: CSVRecord, submissionId: String, recordNu
...
+
+    if (fileName.isEmpty()) {
+        throw UnprocessableEntityException(
+            "In metadata file: record #$recordNumber with id '$submissionId': " +
+                "file entry '$token' in column '$header' is missing a file name. Expected format 'fileName${FILE_NAME_ID_SEPARATOR}fileId'.",
+        )
+    }
+
+    // TODO: Update when moving away from UUIDs to more user-friendly file IDs
+    val fileId = try {
```
anna-parker, line 149: we discussed adding a specific length check for now to improve the error message in case part of the UUID is accidentally deleted

backend/src/main/kotlin/org/loculus/backend/model/SubmitModel.kt
```diff
@@ -174,11 +178,21 @@ class SubmitModel(
...
+                // This is implicitly validated for the file mappings within the metadata column
+                // TODO: This can be removed once file mappings JSON support is removed
```
anna-parker, line 187: The backend tests actually all still rely on the fact that the fileMapping can be passed in submission and do not check that the files will also be parsed from the metadata. I guess this PR is for an intermittent state where both submission options work but might still make sense to update at least some of the tests - I ended up doing that in https://github.com/loculus-project/loculus/pull/7012  (when the fileMapping JSON is completely removed)

backend/src/main/kotlin/org/loculus/backend/utils/MetadataEntry.kt
theosanderson: (I guess it's only worth investing in this if we don't plan to do the user-friendly IDs any time soon?)

theosanderson: but maybe it's quick enough it doesn't matter :)

anna-parker:
@tombch and I discussed some blockers for merging this PR:
- [x] Docs that need updating: https://loculus.org/for-users/submit-extra-files/, also backend endpoints in the FilesController need updating (ah claude actually did this for me in https://github.com/loculus-project/loculus/pull/7012/changes/5c693fb7f263759200898563cdea5859df7bf513)
- [x] Adding a migration guide for users using the current feature in the PR description
- [x] Ensure files.rawReads shows as mapped to a keyword in the column mapping:
image

website/src/components/Submission/FormOrUploadWrapper.tsx
```diff
@@ -71,6 +77,28 @@ export const FormOrUploadWrapper: FC<FormOrUploadWrapperProps> = ({
...
+        const controller = new AbortController();
+        void (async () => {
+            if (!metadataFile) {
+                setSubmissionFileMapping(undefined);
+                return;
+            }
+            const text = columnMapping
+                ? await (await columnMapping.applyTo(metadataFile)).text()
+                : await metadataFile.text();
+            if (!controller.signal.aborted) {
```
anna-parker: its probably nicer to use `if (controller.signal.aborted) return;` to have less indented code

website/src/components/Submission/FormOrUploadWrapper.tsx
```diff
@@ -71,6 +77,28 @@ export const FormOrUploadWrapper: FC<FormOrUploadWrapperProps> = ({
     // The columnMapping can be null; if null -> don't apply mapping.
     const [columnMapping, setColumnMapping] = useState<ColumnMapping | null>(null);
 
+    useEffect(() => {
+        const controller = new AbortController();
```
anna-parker: Im not sure we actually need a controller I think just a boolean called cancelled would be enough (it needs to be set to true in the cleanup function)

website/src/components/Submission/FormOrUploadWrapper.tsx
tombch: Updated!

website/src/components/Submission/FormOrUploadWrapper.tsx
tombch: Updated!

anna-parker:
Side note: we now restrict metadata fields from starting with `files.` (maybe we already did this implicitly), this should probably be documented in loculus administrator docs

website/src/components/Submission/DataUploadForm.tsx
```diff
@@ -66,14 +77,25 @@ const InnerDataUploadForm = ({
 
     const { submit, revise, isPending } = useSubmitFiles(accessToken, organism, clientConfig, onSuccess, onError);
     const [fileFactory, setFileFactory] = useState<FileFactory | undefined>(undefined);
-    const [fileMapping, setFileMapping] = useState<FilesBySubmissionId | undefined>(undefined);
+    const [fileMapping, setFileMapping] = useState<FileMapping<ResolvedSubmissionFile> | undefined>(undefined);
```
anna-parker: A meta-level comment but I wonder if there is a clearly way to name the fileMapping generated from the metadata.tsv (called `submissionFileMapping` atm) and the fileMapping coming from the uploaded files called (`fileMapping`)

website/src/components/Submission/DataUploadForm.tsx
```diff
@@ -103,20 +125,55 @@ const InnerDataUploadForm = ({
...
-        let fileMappingWithSubmissionId = fileMapping;
-        // for single submission, use the submissionID that the user gave in the form
-        if (extraFilesEnabled && inputMode === 'form' && fileMapping !== undefined) {
-            fileMappingWithSubmissionId = { [submissionId!]: Object.values(fileMapping)[0] };
+        let finalMetadataFile = metadataFile;
+
+        if (extraFilesEnabled) {
+            if (inputMode === 'form') {
+                if (fileMapping !== undefined) {
+                    const finalSubmissionFileMapping = new Map([[submissionId!, fileMapping]]);
```
anna-parker: can you maybe keep the comment that there is only one submissionID in the form case - I got confused reading this!

website/src/components/Submission/FileUpload/fileMapping.ts
```diff
@@ -0,0 +1,467 @@
...
+export async function applyFileMappings(
+    metadataFile: File,
+    resolvedSubmissionFileMapping: SubmissionFileMapping<ResolvedSubmissionFile>,
+): Promise<Result<File, Error>> {
+    const text = await metadataFile.text();
+    const parsed = Papa.parse<string[]>(text, { delimiter: '\t', skipEmptyLines: true });
+    if (parsed.data.length === 0) return err(new Error('Please provide a non-empty metadata file.'));
+
+    // If there are no resolved file entries, return the original file
+    if (resolvedSubmissionFileMapping.size === 0) return ok(metadataFile);
```
anna-parker: I think we can do this check before calling papa.parse

website/src/components/Submission/FileUpload/fileMapping.ts
```diff
@@ -0,0 +1,467 @@
...
+
+    const columnNames = parsed.data[0];
+    const columns = columnNames.map((column, index) => ({ name: column, index }));
+    const rows = parsed.data.slice(1);
+
+    // Ensure a column exists for every category present in the resolved mapping
+    const categories = new Set<string>();
+    for (const fileMapping of resolvedSubmissionFileMapping.values())
+        for (const category of fileMapping.keys()) categories.add(category);
```
anna-parker:
```suggestion
    const categories = new Set(
        [...resolvedSubmissionFileMapping.values()].flatMap((fileMapping) => [...fileMapping.keys()])
    );
```

anna-parker: Is this actually required by the backend? I assume if the file category is in the `resolvedSubmissionFileMapping` and not in the metadata this would mean either there are no files for that catgory (in which case do we need to upload an empty column) or we uploaded files without a corresponding metadata column which I believe will also error

tombch: The reason for having this was to support form submissions with the new file mappings - they derive a single row metadata TSV from the form values, and that doesn't include a files column. So for that case files columns needed to be created before applying the mapping. Although yeah there should be a comment explaining this, or now I think about it, an alternative could be to create the files columns in `EditableMetadata.getMetadataTsv`

website/src/components/Submission/FileUpload/fileMapping.ts
```diff
@@ -0,0 +1,467 @@
...
+    const resolvedEntries: ResolvedEntry[] = [];
+
+    for (const [category, pathMapping] of linkageMapping) {
+        const categoryLinkage: CategoryLinkage = {
+            linked: [],
+            reused: [],
+            missing: [],
+            orphaned: [],
+            shadowed: [],
+        };
```
anna-parker: I think this could be a subfunction: 
```
export function createCategoryLinkage(): CategoryLinkage {
  return {
    [LinkageType.LINKED]: [],
    [LinkageType.REUSED]: [],
    [LinkageType.MISSING]: [],
    [LinkageType.ORPHANED]: [],
    [LinkageType.SHADOWED]: [],
  };
}
```

website/src/components/Submission/FileUpload/ColumnMapping.ts
tombch: File category columns are now added as input fields in `getGroupedInputFields` and `getOrderedTemplateInputFields`, so they appear in metadata documentation and column mapping

website/src/components/Submission/FileUpload/FolderUploadComponent.tsx
anna-parker, line 1: Probably a silly question but where I noticed the fileMapping created for these files from the folder includes a file.name - where is the name set? I would think we only know file.path at this point? but maybe I am missing sth

anna-parker: ah ok the name is basically the name of the file without the subpaths. We actually dont use it though as we join the files based on paths now and not the names - so actually we could drop this here I think?

website/src/components/Submission/FileUpload/fileMapping.ts
anna-parker, line 1: Some structural higher level thoughts I have been having:
- the fileMapping produced by the folderComponent should be from filePath to fileID
- the fileMapping produced by the metadata is a map from the submissionID to another map from the fileName to the filePath OR the fileName to the fileID (if no filePath or fileID is provided filePath is set to fileName) - filePath doesnt actually have to be unique
- files with a fileID do not need to be mapped, files with a path must be joined to the filePaths in the folderComponent

website/src/components/Submission/FileUpload/fileMapping.ts
tombch: This was added in #7022

website/src/components/Submission/FileUpload/fileMapping.ts
tombch: This has now been added!

website/src/components/Submission/DataUploadForm.tsx
tombch: This has now been updated to have `getSingleSubmissionFileMapping` introduced first in which IMO makes this clear

website/src/components/Edit/EditPage.tsx
```diff
@@ -118,12 +137,19 @@ const InnerEditPage: FC<EditPageProps> = ({
...
             });
         } else {
-            const fileMappingForEdit = fileMappingForSubmission?.[dataToEdit.submissionId] ?? null;
+            const fileMappingForEdit =
+                extraFilesEnabled && fileMapping !== undefined
+                    ? Object.fromEntries(
+                          [...fileMapping].map(([category, files]) => [
+                              category,
+                              [...files.entries()].map(([path, fileId]) => ({ fileId, name: path })),
+                          ]),
```
tombch: this code is nearly a duplicate of getSingleSubmissionFileMapping, I will update

website/src/components/Submission/FileUpload/fileMapping.ts
tombch: Also just applied the suggestion above

anna-parker: ah ok!

website/src/components/Submission/FileUpload/FolderUploadComponent.tsx
anna-parker: I think this has now been refactored - closing for now

website/src/components/Submission/FileUpload/fileMapping.ts
anna-parker: also resolving as my changes have been merged in :-)

website/src/pages/[organism]/submission/template/index.ts
```diff
@@ -57,7 +57,10 @@ export const GET: APIRoute = async ({ params, request }) => {
 };
 
 function createTsvTemplate(organism: string, action: UploadAction): ArrayBuffer {
-    const columnNames = Array.from(getMetadataTemplateFields(organism, action).keys());
+    // Only default template fields are included in the TSV template
```
Comment (anna-parker) [inline website/src/pages/[organism]/submission/template/index.ts]: Can this refactor be done in a different PR? this is basically a way to drop `getMetadataTemplateFields` and use `getOrderedTemplateInputFields` instead which is unrelated to this PR IMO

website/src/components/Edit/EditPage.tsx
anna-parker: also kinda lines 70-80

website/src/components/Submission/DataUploadForm.tsx
```diff
@@ -273,6 +331,61 @@ export const InputModeTabs = ({
...
+        count: number;
+        message: string;
+    }[] = useMemo(() => {
+        if (categoryLinkage === undefined) return [];
+        return [
+            {
+                key: 'linked',
+                icon: '✓',
+                color: 'text-green-500',
+                count: Array.from(new Set(categoryLinkage.linked.map((file) => file.path))).length,
```
anna-parker, line 350: why doesnt `categoryLinkage.linked.length` not work here?

website/src/components/Submission/FileUpload/fileMapping.ts
```diff
@@ -0,0 +1,483 @@
...
+const FILE_NAME_ID_SEPARATOR = ':';
+const FILE_ENTRY_REGEX = /^([^:]+)(?:::([^:]+)|:([^:]+))?$/;
+
+type SubmissionId = string;
+type FileCategory = string;
+type FilePath = string;
+type FileName = string;
+type FileId = string;
+
+type DeclaredFile = {
```
anna-parker, line 20: so declared and reused are files defined in the metadata correct? maybe we can comment this

anna-parker: oh sorry its below! will resolve my comment

website/src/components/Submission/FileUpload/fileMapping.ts
```diff
@@ -0,0 +1,483 @@
...
+ * An entry containing the submission ID, file category, and resolved file.
+ * Used to build a resolved submission file mapping, which contains the file IDs for all files which are linked or reused.
+ */
+type ResolvedEntry = {
+    submissionId: SubmissionId;
+    category: FileCategory;
+    file: LinkedFile | ReusedFile;
+};
+
+type ResolvedFile = {
```
anna-parker, line 71: what is the resolvedFile? how is it different than a linkedFile?

anna-parker: its kinda just a linkedFile without the path?

website/src/components/Submission/FileUpload/fileMapping.ts
```diff
@@ -0,0 +1,483 @@
...
+
+        const linkedPaths = new Set<FilePath>();
+        const shadowedPaths = new Set<FilePath>();
+
+        for (const [submissionId, categoryMapping] of submissionFileMapping) {
+            const files = categoryMapping.get(category);
+            if (files === undefined) continue;
+
+            for (const file of files.values()) {
+                switch (file.type) {
```
anna-parker, line 186: yes this is very nice to read - I like this refactor!

website/src/components/Submission/FileUpload/fileMapping.ts
anna-parker: ah ok its a linked or reusedfile without the path - yeah this makes sense - might be cool to add a comment here!

website/src/components/Submission/FileUpload/fileMapping.ts
```diff
@@ -0,0 +1,483 @@
...
+    const text = await metadataFile.text();
+    const parsed = Papa.parse<string[]>(text, { delimiter: '\t', skipEmptyLines: true });
+    if (parsed.data.length === 0) return err(new Error('Please provide a non-empty metadata file.'));
+
+    const columnNames = parsed.data[0];
+    const columns = columnNames.map((column, index) => ({ name: column, index }));
+    const rows = parsed.data.slice(1);
+
+    // Ensure a column exists for every category present in the resolved mapping
+    const categories = new Set(
```
anna-parker: hmm... I dont think we need to look at the values of each submissionId at this point as we already ensured that each submissionId will have each category in its fileMapping in resolveFileMappings, we could just look at the first entry to save computation. Alternatively, this information should be in the config `submissionDataTypes.files?.categories?.map((c) => c.name)`- I wonder if generally the list of categories should always come from the config to reduce needing to get the categories in each mapping.

tombch: ah yeah good point, I will update to use the submissionDataTypes categories

website/src/components/Submission/FileUpload/fileMapping.spec.ts
```diff
@@ -0,0 +1,449 @@
...
+                ...emptyDetails,
+                missing: [metadataEntry],
+                orphaned: [otherUpload],
+            });
+            expect(submissionFileMapping.size).toBe(0);
+        });
+    });
+
+    describe('orphaned', () => {
+        it('when an upload folder entry has no metadata entry on its path', () => {
```
anna-parker: the test name is a bit odd, maybe "uploaded file with no corresponding metadata entry"?

website/src/components/Submission/FileUpload/fileMapping.spec.ts
```diff
@@ -0,0 +1,449 @@
...
+        );
+    });
+
+    it('reports orphaned files', () => {
+        expect(getLinkageErrors(detailsOf([], ['b.txt']))).toBe(
+            getLinkageErrorMessage(LinkageType.ORPHANED, 'raw', 'b.txt'),
+        );
+    });
+
+    it('reports shadowed uploads separately from genuinely unreferenced ones', () => {
```
anna-parker: Im not sure this test is really needed

website/src/components/Submission/FileUpload/fileMapping.spec.ts
```diff
@@ -0,0 +1,449 @@
...
+            metadataFile([
+                ['id', 'country'],
+                ['e1', 'CH'],
+            ]),
+            merged,
+        );
+        expect(await linesOf(result._unsafeUnwrap())).toEqual(['id\tcountry\tfiles.raw', 'e1\tCH\ta.txt:id-a']);
+    });
+
+    it('leaves rows with no files for the category untouched', async () => {
```
anna-parker: hmm... do we actually want this behavior? this should never happen right? either the metadata referenced a reused file or `parseSubmissionFileMapping` should have failed?

website/src/components/Submission/FileUpload/FolderUploadComponent.tsx
```diff
@@ -345,202 +344,184 @@ export const FolderUploadComponent: FC<FolderUploadComponentProps> = ({
...
+            }
 
-        // Check for collisions with existing files
-        const submissionId = formSubmissionId ?? DUMMY_SUBMISSION_ID;
-        const existingFiles = fileUploadState.files[submissionId];
-        const uniqueSelectedNames = new Set(selectedFileNames);
-        const existingFileCollisions = existingFiles.filter((file) => uniqueSelectedNames.has(file.name));
+            // Check for collisions with existing files
+            const filePaths = new Set(awaiting.map((file) => file.path));
+            const collisions = fileUploadState.files.filter((file) => filePaths.has(file.path));
```
anna-parker, line 375: hmm shouldnt we be checking for collisions before we set the state to `awaitingUrls`?

anna-parker: side thought: all the steps after the state `awaitingUrls` are handled in the `useEffect` for the folder component, it seems a bit odd to do this separately for the files - this can be done in a follow up PR

integration-tests/tests/specs/features/file-sharing.spec.ts
```diff
@@ -294,3 +402,212 @@ test('single revise seq with files via edit page', async ({ page, groupId, tmpDi
...
+    await searchPage2.openModalByRoleAndName('link', `${accession}.${version + 2}`);
+    await searchPage2.checkAllFileContents({
+        [file2Name]: EBOLA_SUDAN_SMALL_FASTQ(1),
+        [addedFileName]: EBOLA_SUDAN_SMALL_FASTQ(2),
+    });
+    await expect(page.getByRole('link', { name: file1Name })).toHaveCount(0);
+    await searchPage2.closeDetailsModal();
+});
+
+test('bulk revise can reuse, replace, discard and add files', async ({ page, groupId, tmpDir }) => {
```
anna-parker, line 489: it would probably be good to keep the subfolders for this test

integration-tests/tests/specs/features/file-sharing.spec.ts
```diff
@@ -27,11 +32,18 @@ const SEQUENCING_INSTRUMENT = 'Illumina MiSeq';
...
 const ID_2 = 'sub2';
 const FILES_SINGLE = { 'testfile.fastq': EBOLA_SUDAN_SMALL_FASTQ(1) };
-const FILES_DOUBLE = {
+const FILES_DOUBLE: Record<string, string> = {
     'file1.fastq': EBOLA_SUDAN_SMALL_FASTQ(1),
     'file2.fastq': EBOLA_SUDAN_SMALL_FASTQ(2),
 };
 
+// Tests upload files in subfolders grouped by submissionId
+// TODO: Update integration tests to move away from submission Id subfolders
```
anna-parker: TODO sounds good!

website/src/components/Submission/FileUpload/FolderUploadComponent.spec.tsx
```diff
@@ -47,20 +48,29 @@ const defaultProps = {
     onError: mockOnError,
 };
 
-// In the case of previous uploads, they are keyed by a real submission id, not the dummySubmissionId
-const submissionId = 'SUBMISSION_ID_123';
+// The fileMapping is keyed by category, then by file path, to its file ID. Previous uploads have no
+// upload path, so (within a single entry, where names are unique) they are keyed by their name.
+const fileMappingOf = (files: { fileId: string; name: string }[]) =>
+    new Map([['extraFiles', new Map(files.map((f) => [f.name, f.fileId]))]]);
```
anna-parker: didnt we change that this a map from path to fielId not name to fileId?

website/src/components/Submission/FileUpload/FolderUploadComponent.spec.tsx
```diff
@@ -71,273 +81,373 @@ describe('FolderUploadComponent', () => {
...
+        it('renders upload folder button when the mapping has no files for the category', () => {
+            render(
+                <FolderUploadComponent
+                    {...defaultProps}
+                    inputMode='form'
+                    fileMapping={new Map([['extraFiles', new Map()]])}
+                />,
+            );
+
+            expect(screen.getByTestId('extraFiles')).toBeInTheDocument();
```
anna-parker, line 101:
```suggestion
expect(screen.getByText('Upload folder')).toBeInTheDocument();     expect(screen.getByTestId('extraFiles')).toBeInTheDocument();
```

website/src/components/Submission/FileUpload/FolderUploadComponent.spec.tsx
```diff
@@ -71,273 +81,373 @@ describe('FolderUploadComponent', () => {
...
-        vi.mocked(multipartUpload.uploadPart).mockImplementation(async () => {
-            uploadCount++;
-            if (uploadCount === 1) {
-                await new Promise((resolve) => setTimeout(resolve, 10));
-            } else {
-                await new Promise(() => {});
-            }
-            return '"etag"';
+            await userEvent.upload(input, file);
+            await waitFor(() => expect(screen.getByText('test.txt')).toBeInTheDocument());
```
anna-parker: in this case it should actually display `submission1/test.txt` I believe?

website/src/components/Submission/FileUpload/FolderUploadComponent.spec.tsx
```diff
@@ -71,273 +81,373 @@ describe('FolderUploadComponent', () => {
...
-        vi.mocked(multipartUpload.uploadPart).mockReturnValue(new Promise(() => {}));
+            await userEvent.click(screen.getByRole('button', { name: 'Cancel' }));
+            await waitFor(() =>
+                expect(screen.queryByText(/are you sure you want to discard/i)).not.toBeInTheDocument(),
+            );
+            expect(screen.getByText('file-a.txt')).toBeInTheDocument();
+            expect(screen.getByText('file-b.txt')).toBeInTheDocument();
+        });
+
+        it('discard all files is not disabled during upload or file upload failure', async () => {
```
anna-parker, line 428: I actually feel like the better behavior would be to allow the user to discard the exact file that had an error not all files - but perhaps this isnt feasible?

tombch: Yeah this is definitely doable, in the previous PR I had for blocking submit during uploads I changed it to still be able to complete if files error and then you can discard the error. so can definitely add this. I feel like this could come in the follow up PR to handle the blocking submit stuff?

website/src/components/Submission/FileUpload/fileMapping.ts
tombch: Added a comment!

website/src/components/Submission/DataUploadForm.tsx
tombch: Added a comment!

website/src/components/Submission/FileUpload/FolderUploadComponent.tsx
tombch: We do in this case? For the `handleFileSelect`, the file upload state only gets set to `awaitingUrls` if it was `undefined`, so no collisions can occur. If the file upload state wasn't undefined we check for collisions, and if its > 0, display confirmation dialog, only then does the inline `addFiles` function run that requests the file urls and then sets the state to `uploadInProgress`. I guess what this does mean is that when hitting `add additional files` there is no visual state currently for those files awaiting urls - i think this could be updated in the future

website/src/components/Submission/FileUpload/fileMapping.ts
tombch: I updated the `EditableMetadata.getMetadataTsv` to add empty file categories to its tsv so we don't need this function to add them anymore. I have kept the check for file categories being present, as i think its useful as a stopgap - its not expensive to check and if we ever somehow had file mapping categories not present in the metadata it would mean those mappings weren't included, which would be really bad

website/src/components/Edit/EditPage.tsx
tombch: I had a look and they're all slightly different, the pain point is because I decided to use Maps when the other types are Records so its converting to/from those. Could update this but I dont think its worth doing now

website/src/pages/[organism]/submission/template/index.ts
Comment (tombch) [reply website/src/pages/[organism]/submission/template/index.ts]: Yep, this has been reverted and getMetadataTemplateFields updated to have file categories

website/src/components/Submission/FileUpload/FolderUploadComponent.spec.tsx
tombch: The folder and text are in different elements so it cant exactly match on that, but changed to match on title instead

website/src/components/Submission/FileUpload/FolderUploadComponent.spec.tsx
tombch: Updated

website/src/components/Submission/FileUpload/fileMapping.spec.ts
tombch: I removed most of the linkage error message tests as IMO they were unnecessary

website/src/components/Submission/FileUpload/fileMapping.spec.ts
tombch: updated

integration-tests/tests/specs/features/file-sharing.spec.ts
tombch: This has been updated now! The `prepareTmpDirForBulkUpload` can accept a folder with a mix of either just files or subfolders containing files

website/src/components/Submission/FileUpload/FolderUploadComponent.spec.tsx
tombch: Resolving this as we determined this can be done in a follow-up PR

anna-parker:
@claude review

integration-tests/tests/specs/features/file-sharing.spec.ts
```diff
@@ -151,16 +187,82 @@ test('bulk submit 1 seq with a 35 MB FASTQ file', async ({ page, groupId, tmpDir
...
+            metadataFileEntries: filesColumnCell(Object.keys(FILES_DOUBLE), ID_1),
+            uploadedFiles: file1,
+            error: `referenced in metadata but not uploaded: ${ID_1}/${file2Name}`,
+        },
+        {
+            metadataFileEntries: filesColumnCell(Object.keys(file1), ID_1),
+            uploadedFiles: FILES_DOUBLE,
+            error: `uploaded but not referenced in metadata: ${ID_1}/${file2Name}`,
+        },
+        // TODO: Test shadowed files, which requires uploading files without submission ID
```
anna-parker: link to (follow up) issue or remove

website/src/config.ts
```diff
@@ -317,6 +339,9 @@ export function getGroupedInputFields(
         groups.get(header)!.push({ ...field });
     });
 
+    const fileCategoryFields = inputMode === 'bulk' ? getFileCategoryInputFields(schema) : [];
```
anna-parker, line 338: sorry why is this only added when bulk submission is selected again?

anna-parker: ah this makes sense sorry!

website/src/components/Edit/EditPage.tsx
```diff
@@ -63,11 +68,16 @@ const InnerEditPage: FC<EditPageProps> = ({
...
-        extraFilesEnabled && dataToEdit.submittedData.files
-            ? { [dataToEdit.submissionId]: dataToEdit.submittedData.files }
-            : undefined,
-    );
+    const [fileMapping, setFileMapping] = useState<FileMapping | undefined>(() => {
+        const previousFiles = dataToEdit.submittedData.files;
+        if (!previousFiles) return undefined;
+        return new Map(
+            Object.entries(previousFiles).map(([category, files]) => [
+                category,
```
anna-parker, line 76: can you maybe add a comment that for the edit page no subfolders are allowed and thus the name is the same as the path?

website/src/components/Edit/EditPage.tsx
```diff
@@ -118,12 +138,18 @@ const InnerEditPage: FC<EditPageProps> = ({
...
             }
             submitRevision({
-                metadataFile,
+                metadataFile: finalMetadataFile,
                 sequenceFile,
-                fileMapping: fileMappingForSubmission,
             });
         } else {
-            const fileMappingForEdit = fileMappingForSubmission?.[dataToEdit.submissionId] ?? null;
+            let fileMappingForEdit = null;
```
anna-parker: maybe we can type this as `FilesByCategory | null` to be clearer

anna-parker: then backend.ts just needs to have `export type FilesByCategory = z.infer<typeof filesByCategory>;` as well

anna-parker (approved):
Looks great! (probably good to still fix the 3 things claude flagged as well, but these are small changes)

Updating with discussed points:
1. we should either switch from ' ' to ',' as a separator or add that loculus users should confirm no files have spaces in their name before updating to this commit (we should add an postSQL command to run to check this)
2. we can switch the order of the checks
3. small docs update

other minor issues can be disregarded

anna-parker:
Way to find all entries where the fileName has whitespace in processed and submitted data:
```
SELECT
    se.accession,
    f.key AS file_type,
    obj ->> 'name' AS name
FROM public.sequence_entries_preprocessed_data se,
     jsonb_each(se.processed_data -> 'files') AS f(key, value),
     jsonb_array_elements(f.value) AS obj
WHERE obj ->> 'name' ~ '\s'
  AND processed_data -> 'files' IS NOT NULL
  AND processed_data -> 'files' <> 'null'::jsonb
ORDER BY se.accession;
```
and
```
SELECT
    se.accession,
    f.key AS file_type,
    obj ->> 'name' AS name
FROM public.sequence_entries se,
     jsonb_each(se.submitted_data -> 'files') AS f(key, value),
     jsonb_array_elements(f.value) AS obj
WHERE obj ->> 'name' ~ '\s'
  AND submitted_data -> 'files' IS NOT NULL
  AND submitted_data -> 'files' <> 'null'::jsonb
ORDER BY se.accession;
```
This can then be modified to an underscore in the submitted data as follows (the processed data can be changed by a similar command or by reprocessing the data)
```
UPDATE public.sequence_entries se
SET submitted_data = jsonb_set(
    se.submitted_data,
    '{files}',
    (
        SELECT jsonb_object_agg(
            f.key,
            (
                SELECT jsonb_agg(
                    CASE
                        WHEN obj ->> 'name' ~ '\s'
                        THEN jsonb_set(
                            obj,
                            '{name}',
                            to_jsonb(
                                regexp_replace(
                                    obj ->> 'name',
                                    '\s+',
                                    '_',
                                    'g'
                                )
                            )
                        )
                        ELSE obj
                    END
                )
                FROM jsonb_array_elements(f.value) AS obj
            )
        )
        FROM jsonb_each(se.submitted_data -> 'files') AS f(key, value)
    )
)
WHERE se.submitted_data -> 'files' IS NOT NULL
  AND se.submitted_data -> 'files' <> 'null'::jsonb
  AND EXISTS (
      SELECT 1
      FROM jsonb_each(se.submitted_data -> 'files') AS f(key, value),
           jsonb_array_elements(f.value) AS obj
      WHERE obj ->> 'name' ~ '\s'
  );
```

website/src/components/Edit/EditPage.tsx
tombch: Done! Thanks

website/src/components/Edit/EditPage.tsx
tombch: Done!

integration-tests/tests/specs/features/file-sharing.spec.ts
tombch: Just realised this test was easy to add now that the `uploadExternalFiles` test function was refactored to support flat folders so this has been added!

website/src/components/Submission/FileUpload/fileMapping.spec.ts
tombch: Yeah this was changed, it no longer adds a column if the category is missing from the metadata but throws an error and this has a test now

tombch:
Ok all comments have been resolved... as there was a couple changes, I will ask for one (hopefully final) @claude review

corneliusroemer:
There's a new integration test flake (race) that seems to have been triggered by changes here: https://github.com/loculus-project/loculus/blob/1050a4ad301d3fe9feefc7c587e240b4eb159278/integration-tests/tests/specs/features/file-sharing.spec.ts#L484-L488

> The race: in `file-sharing.spec.ts:486` the test replaces `file2.fastq`, then calls `expectExtraFileUploaded(RAW_READS, 'file2.fastq')`. That assertion is a no-op for a replaced file: the `test-id` is keyed on filename, and `getStatusIcon` renders the identical `✓` for both previousUpload and uploaded. So it matches the stale pre-existing row and passes instantly (14 ms in firefox), long before the replacement upload starts. The test then clicks Submit mid-upload, `handleSubmit`'s new `validateFileUploadStates` sees `uploadInProgress`, and instead of the confirmation dialog you get the toast — visible verbatim in the error-context snapshot: "Please wait for all files to finish uploading before submitting."

I noticed when looking into integration test failure on main, when both chromium and firefox failed on same test, same step was suspicious:
- https://github.com/loculus-project/loculus/actions/runs/32494192105/job/96808654000
- https://github.com/loculus-project/loculus/actions/runs/32494192105/job/96808653808

I'm working on a fix.

PR #6924, opened 2026-07-20
feat(actions): add zika taxon id to dataset mirror workflow

arthurshem86:
In order to ingest Zika data from NCBI, we need to add zika to the set of organisms for which we periodically pull data from NCBI so that it can accessed in Pathoplexus. For this reason I added the zika taxon id to the set of configured organisms.


maverbiest (commented):
Looks good, @arthurshem86 thank you! Could you please add a brief description to the PR to say what you're adding and why it's needed? 

Also, we use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) for PRs on Loculus. This is a convention on how we call PRs that we merge into main. Could you please have a look at that and update the PR name? It would probably be something like `feat(actions): add zika taxon id to dataset mirror workflow`

Then I'll approve and we can merge this in to Loculus :)

arthurshem86:
Thanks @maverbiest . In order to ingest Zika data from NCBI, we need to add `zika` to the set of organisms for which we periodically pull data from NCBI so that it can accessed in Pathoplexus. For this reason I added the zika taxon id to the set of configured organisms.

maverbiest (approved):

Issue #3511, opened 2025-01-11
get-released-data endpoint times out when there are too many sequences


anna-parker:
We are trying to use Loculus as a temporary backend in genspectrum, I was now performing some tests on our deployment and I see that calls to the `get-released-data` endpoint fail with the error: 
```
curl -X 'GET' \
  'https://api.loculus.staging.genspectrum.org/backend/influenza-a/get-released-data' \
  -H 'accept: application/x-ndjson' | zstdcat > influenza_a_heuristic_groupings.ndjson
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 1040M    0 1040M    0     0  6442k      0 --:--:--  0:02:45 --:--:-- 8883k
curl: (92) HTTP/2 stream 1 was not closed cleanly: INTERNAL_ERROR (err 2)
```
For reference calls to organisms with less data points succeed, here I query an organism with over 300k sequences, with up to 8 segments each. In total there are approx 1.1M sequences.

anna-parker:
Calling LAPIS also leads to an internal server error: 
```
curl -X 'GET' \
  'https://api.loculus.staging.genspectrum.org/influenza-a/sample/details?dataFormat=JSON&downloadAsFile=true' \
  -H 'accept: application/json' > influenza_a_submissionIds.json
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   570    0   570    0     0     37      0 --:--:--  0:00:15 --:--:--   143
```
```
{"error":{"type":"about:blank","title":"Internal Server Error","status":500,"detail":"Could not connect to silo: class java.io.IOException HTTP/1.1 header parser received no bytes"},"info":{"dataVersion":null,"requestId":"247b69ac-ea12-4a8b-8ab5-ba4e30b7c7b9","requestInfo":"Influenza A on api.loculus.staging.genspectrum.org at 2025-01-11T10:12:45.813011669","reportTo":"Please report to https://github.com/GenSpectrum/LAPIS/issues in case you encounter any unexpected issues. Please include the request ID and the requestInfo in your report.","lapisVersion":"0.3.10"}}
```

fengelniederhammer:
[...]

Issue #2778, opened 2024-09-13
Put get-released-data record-count and result-stream in one transaction 


corneliusroemer:
See:
https://github.com/loculus-project/loculus/blob/bc59479d50aef18e0f1eae762b6364b0133ea9ba/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt#L272-L275

This was discussed in PR, @anna-parker tested (IIUC) that this can happen (by introducing an exaggeratedly long delay between count and streaming through sleep - it is unlikely to cause any harm in practice, and the PR makes things in no way worse than without it.

https://github.com/loculus-project/loculus/pull/2773

chaoran-chen:
As this means that the value in the entry can potentially be wrong, I'd rather classify it as a bug. If there are few sequences and unfrequent new submissions, I agree that it is not a big problem. If we go to an outbreak / pandemic situtation with milliions of sequences and frequent uploads, this changes.

PR #6926, opened 2026-07-20
fix(backend): spool released exports safely

### LINKED ISSUE #5198: When calculating number of rows in get-released-data, cache to disk instead of requesting from db twice
Right now in get-released-data we get the db to go through all sequences twice - once just to count, and a second time to actually get the sequences.

The count itself seems expensive - it takes 30s for mpox, for example. This could maybe be improved - but it'd be easier to just get the full response once, write to disk (while counting), then sending the result from disk to client.

https://github.com/loculus-project/loculus/blob/5d0fc3e147f44302ccd46156b56acb770432978e/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt#L326-L336

https://github.com/loculus-project/loculus/blob/5d0fc3e147f44302ccd46156b56acb770432978e/backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt#L684-L691

This would likely reduce db load and speed up released-data requests, both desirable.

clonder:
materialize each released-data export from a single successful run of the released-record stream. count records while writing the temporary file so `X-Total-Records` matches the served bytes and the database transaction closes before network transfer begins.

because the backend now prepares the complete export before sending the first byte, increased the read timeouts for the bundled SILO and CLI clients and configure bounded spool storage in the deployment.

Resolves #5198  
Related to #3511

🚀 Preview: https://fix-get-released-data-sin.loculus.org


backend/src/main/kotlin/org/loculus/backend/config/BackendSpringConfig.kt
anna-parker, line 1: @clonder and @corneliusroemer I heard you guys did some perf of this new feature - could you maybe summarize your results in the PR description? would be nice to have this for future reference :-)

backend/src/main/kotlin/org/loculus/backend/service/submission/StreamSpoolService.kt
```diff
@@ -0,0 +1,300 @@
...
+            val output = file.outputStream()
+            trackedBytes -= trackedFileSizes.put(file, 0) ?: 0
+            output
+        }
+        val checkedFileStream = FreeSpaceCheckingOutputStream(
+            fileStream,
+        ) { bytes, write -> write(file, bytes, write) }.buffered()
+        return when (compressionFormat) {
+            CompressionFormat.ZSTD -> ZstdCompressorOutputStream(checkedFileStream)
+            null -> checkedFileStream
```
chaoran-chen, line 145: As the data can be very large, I think that we should always compress them when writing to disk. If the user has requested compressed data, the compressed file can be directly returned. If the user does not want compressed data, they would be decompressed again when returning to the user.

backend/src/main/kotlin/org/loculus/backend/config/BackendSpringConfig.kt
chaoran-chen, line 1: Thanks @clonder for the PR and @corneliusroemer for the proposal. Like Anya, I'd also be quite interested in seeing the performance impacts; maybe let's merge #6807 first and use it to compare the performances?

clonder:
hey @anna-parker @chaoran-chen - ran the comparison against main, here's what came out

setup: 10k real mpox sequences (~196kb each, so a 4.4GB export uncompressed / 72MB zstd) same db:

edit: updated the numbers - my first run was through `bootRun`, which caps the JIT and was inflating everything ~2x, so these are re-measured from the built jar.

| | main | this PR |
|---|---|---|
| total time (zstd) | 17.6s | 18.1s |
| time to first byte | 0.55s | 18.0s |
| db transaction held open (slow client) | 84s | 16s |
| 6 concurrent requests | all 6 @ ~57s | 4 @ ~37s + 2x 503 |
| X-Total-Records / Content-Length | can drift | exact |

so honestly - **it's ~3% slower**. two things I didn't expect:

- the count from #5198 that supposedly "takes 30s for mpox" is now 12-15ms. maybe the V1.23/V1.24 view+index work already fixed that, so removing the second pass doesn't really buy us anything
- ~12s of the 17.5s is just zstd decompression + NDJSON serialization in the jvm. postgres only does ~5.8s. this PR doesn't touch that path at all

what it fixes i think: the db transaction isn't held open while the client downloads anymore (84s → 16s, i.e. just the query duration). over a real network that's the difference between a transaction scoped to the query and one scoped to the slowest consumer - probably also what's behind the aborts in #3511. plus the count/length headers are exact now (#2778).

cost is time-to-first-byte going 0.55s → 18s, which is why the SILO/CLI timeouts went up, plus the ~3%.

@chaoran-chen always compressing the spool sounds like a good idea to me

edit: updated the numbers - the first run was through bootRun, which caps the JIT and inflated everything ~2x.

corneliusroemer:
As we discussed ~3 weeks ago, the extra complexity is not worth it given it doesn't improve performance in the end.

clonder:
closing as discussed, ~3% slower end to end and the complexity isn't worth it. numbers and takeaways are in the knowledge base now: https://github.com/loculus-project/loculus/blob/kb/spooling-investigations/kb/2026-08-23_Spooling_investigations.md

PR #6930, opened 2026-07-20
feat(prepro): new conditional metadata field requirements flag

### LINKED ISSUE #6886: MVP prepro - add support for fields being required only in certain circumstances, e.g. when raw_reads are provided
> the sequencingInstrument field needs to be required when raw reads are submitted, otherwise it is not required. 

 _Originally posted by @anna-parker in [#6634](https://github.com/loculus-project/loculus/pull/6634/changes#r3559676872)_ see details in slack: https://loculus.slack.com/archives/C0B95CH8BEG/p1783691455787169 - this relates to the fields we need to add here: https://github.com/loculus-project/loculus/pull/6634

maverbiest:
resolves #6886 

We currently have `required: true`, which makes a metadata field required. However, some fields are only required in context. With us adding support for raw-reads, an example came up where we only need `sequencingInstrument` to be provided when raw reads files are submitted (so we can submit to ENA): https://github.com/loculus-project/loculus/issues/6886

## Summary

- New optional `requiredWhen` config key on metadata fields: a list of conditions under
  which the field is required. Three condition types are supported:
  - `files.<category>` — required when files of that category are submitted
    (e.g. `files.raw_reads`)
  - `processed.<field>` — required when another metadata field has a non-null value after processing
  - `<field>` — required when another metadata field has a non-null input value
- `required` and `requiredWhen` are mutually exclusive, enforced both in the Helm
  `values.schema.json` (fails `helm lint`/install) and in the preprocessing config
  validator.
- Enabled by default in `defaultOrganismConfig`: `sequencingInstrument` is now
  `requiredWhen: [files.raw_reads]`.
- INSDC-ingested sequences are exempt from `requiredWhen` checks (just like they already are for `required: true`)

## Alternatives considered

I thought about extending the existing `required` config field to cover this behaviour. In theory, we could rework this so that `required: ["always"]` captures what is currently achieved by `required: true`, and then also allow for stuff like `required: ["files.raw_reads"]`. However, this would be a breaking change and would make checking whether a field is required more annoying in prepro than is the case currently.

## Manual testing

Created a preview with a requiredWhen condition that `sequencingInstrument` should be provided with raw reads, and another requirement saying that `geoLocCity` is required when `geoLocCountry` is provided (this last one has been reverted).

I then submitted test data to a preview which showed that (also see image below):
1) sequence with raw_reads and no `sequencingInstrument` is rejected
2) sequence with `geoLocCountry` but no `geoLocCity` is rejected
3) sequence with raw_reads and `sequencingInstrument: Illumina` gets accepted

grafik

### PR Checklist
- [x] All necessary documentation has been adapted: values schema is updated

🚀 Preview: https://metadata-requirements.loculus.org

### DIFF
--- a/kubernetes/loculus/values.schema.json
+++ b/kubernetes/loculus/values.schema.json
@@ -346,6 +346,16 @@
             },
             "errorMessage": "When rangeOverlapSearch is set, type must be numeric or date ('int', 'float', 'number', or 'date')"
           }
+        },
+        {
+          "if": {
+            "properties": { "required": { "const": true } },
+            "required": ["required"]
+          },
+          "then": {
+            "not": { "required": ["requiredWhen"] },
+            "errorMessage": "A field with required: true cannot also set requiredWhen (it is already always required)."
+          }
         }
       ]
     },
--- a/kubernetes/loculus/values.schema.json
+++ b/kubernetes/loculus/values.schema.json
@@ -33,7 +33,7 @@
           "type": "boolean",
           "description": "Whether the field is required by backend."
         },
-        "required_when": {
+        "requiredWhen": {
           "groups": ["metadata"],
           "type": "array",
           "description": "Array of conditions under which this metadata field is required. Conditions can be the presence of a file category, specified as files.<file_category>, or another metadata fields, specified as processed.<field_name>"
--- a/kubernetes/loculus/values.schema.json
+++ b/kubernetes/loculus/values.schema.json
@@ -36,7 +36,7 @@
         "requiredWhen": {
           "groups": ["metadata"],
           "type": "array",
-          "description": "Array of conditions under which this metadata field is required. Conditions can be the presence of a file category - specified as files.<file_category>, or another metadata field - specified as <field_name>"
+          "description": "Array of conditions under which this metadata field is required. Conditions can be: the presence of a file category - specified as files.<file_category>; the input value of another metadata field - specified as <field_name>; or the processed value of another metadata field - specified as processed.<field_name>"
         },
         "desired": {
           "groups": ["metadata"],
--- a/kubernetes/loculus/values.schema.json
+++ b/kubernetes/loculus/values.schema.json
@@ -36,7 +36,7 @@
         "requiredWhen": {
           "groups": ["metadata"],
           "type": "array",
-          "description": "Array of conditions under which this metadata field is required. Conditions can be: the presence of a file category - specified as files.<file_category>; the input value of another metadata field - specified as <field_name>; or the processed value of another metadata field - specified as processed.<field_name>"
+          "description": "Array of conditions under which this metadata field is required. When multiple conditions are provided they are ORed; a submission is rejected if at least one condition is not met. Conditions can be: the presence of a file category - specified as files.<file_category>; the input value of another metadata field - specified as <field_name>; or the processed value of another metadata field - specified as processed.<field_name>"
         },
         "desired": {
           "groups": ["metadata"],
--- a/kubernetes/loculus/values.schema.json
+++ b/kubernetes/loculus/values.schema.json
@@ -36,7 +36,7 @@
         "requiredWhen": {
           "groups": ["metadata"],
           "type": "array",
-          "description": "Array of conditions under which this metadata field is required. Conditions can be the presence of a file category - specified as files.<file_category>, or another metadata field - specified as processed.<field_name>"
+          "description": "Array of conditions under which this metadata field is required. Conditions can be the presence of a file category - specified as files.<file_category>, or another metadata field - specified as <field_name>"
         },
         "desired": {
           "groups": ["metadata"],
@@ -354,7 +354,7 @@
           },
           "then": {
             "not": { "required": ["requiredWhen"] },
-            "errorMessage": "A field with required: true cannot also set requiredWhen (it is already always required)."
+            "errorMessage": "A field with 'required: true' cannot also set 'requiredWhen' (it is already always required)."
           }
         }
       ]
--- a/kubernetes/loculus/values.schema.json
+++ b/kubernetes/loculus/values.schema.json
@@ -36,7 +36,7 @@
         "requiredWhen": {
           "groups": ["metadata"],
           "type": "array",
-          "description": "Array of conditions under which this metadata field is required. Conditions can be the presence of a file category, specified as files.<file_category>, or another metadata fields, specified as processed.<field_name>"
+          "description": "Array of conditions under which this metadata field is required. Conditions can be the presence of a file category - specified as files.<file_category>, or another metadata field - specified as processed.<field_name>"
         },
         "desired": {
           "groups": ["metadata"],
--- a/kubernetes/loculus/values.schema.json
+++ b/kubernetes/loculus/values.schema.json
@@ -33,6 +33,11 @@
           "type": "boolean",
           "description": "Whether the field is required by backend."
         },
+        "required_when": {
+          "groups": ["metadata"],
+          "type": "array",
+          "description": "Array of conditions under which this metadata field is required. Conditions can be the presence of a file category, specified as files.<file_category>, or another metadata fields, specified as processed.<field_name>"
+        },
         "desired": {
           "groups": ["metadata"],
           "type": "boolean",
--- a/preprocessing/nextclade/src/loculus_preprocessing/config.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/config.py
@@ -271,17 +271,23 @@ def validate_required_when(config: Config) -> None:
                         f"condition referencing non-existing metadata field '{field}'"
                     )
                     raise ValueError(msg)
+                if field == output_field:
+                    msg = (
+                        f"invalid configuration: field '{output_field}' lists itself "
+                        "in `requiredWhen`"
+                    )
+                    raise ValueError(msg)
             elif condition.startswith(FILES_PREFIX):
                 category = condition.removeprefix(FILES_PREFIX)
                 if category not in FileCategory:
                     msg = (
-                        f"invalid configuration: field '{output_field}' has a required_when "
-                        f"condition referencing unknown file category '{category}'. "
+                        f"invalid configuration: field '{output_field}' has a requiredWhen "
+                        f"condition referencing unknown file category '{category}'."
                     )
                     raise ValueError(msg)
             else:
                 msg = (
-                    f"invalid configuration: field '{output_field}' has an invalid required_when "
+                    f"invalid configuration: field '{output_field}' has an invalid requiredWhen "
                     f"condition '{condition}'. Conditions must start with "
                     f"'{PROCESSED_PREFIX}' or '{FILES_PREFIX}'."
                 )
--- a/preprocessing/nextclade/src/loculus_preprocessing/config.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/config.py
@@ -1,5 +1,6 @@
 import argparse
 import graphlib
+import itertools
 import logging
 import os
 from enum import StrEnum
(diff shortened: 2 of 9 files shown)


maverbiest:
@claude review

maverbiest:
> Potential issue: requiredWhen doesn't honor the INSDC ingest group exemption

Here I wasn't quite sure what to do - should we exempt INSDC ingested sequences from the conditional requirements? I guess it's true that they're already exempt from `required: true` annotation at the moment so might make sense to also do that here

EDIT: INSDC ingested sequences now also don't get rejected based on required_when conditions

kubernetes/loculus/values.schema.json
```diff
@@ -33,6 +33,11 @@
           "type": "boolean",
           "description": "Whether the field is required by backend."
         },
+        "requiredWhen": {
+          "groups": ["metadata"],
+          "type": "array",
+          "description": "Array of conditions under which this metadata field is required. Conditions can be the presence of a file category - specified as files.<file_category>, or another metadata field - specified as processed.<field_name>"
```
anna-parker: If we are adding support for `processed.` we should probably also add support for normal (preprocessing) input metadata fields

anna-parker: Yeah this actually adds in some bugs, now `requiredWhen` only gets triggered if the processed value exists, however if the input field was given but errored/warned it will be set to None and `requiredWhen` will not be triggered and no error will be flagged

maverbiest: Yeah fair, so adding support for checking on the input field will give both options: a condition on whether the output field exists (`processed.`) or just whether the input is provided at all (metadata field without the prefix)

Or do you think it would be better to only check the input fields so that the requirements apply to what a submitter provides, regardless of if preprocessing succeeds for fields or not?

anna-parker: Let me go through the code - but yeah I think adding support for both would be best!

maverbiest: Ah yes I'm working on this now, will push some commits in a little while!

anna-parker: suggestion in https://github.com/loculus-project/loculus/pull/6969 :-)

maverbiest: Thanks! That's quite close to what I have locally, I'll reconcile them a bit and let you know once I've pushed here :)

preprocessing/nextclade/tests/test_metadata_processing_functions.py
```diff
@@ -756,20 +850,61 @@ def test_preprocessing(test_case_def: Case, config: Config, factory_custom: Proc
...
+    config = generate_config_with_deps()
+    factory_custom = ProcessedEntryFactory(all_metadata_fields=list(config.processing_spec.keys()))
     test_case = test_case_def.create_test_case(factory_custom)
-    processed_entry = process_single_entry(test_case, config_dependency)
+    processed_entry = process_single_entry(test_case, config)
     verify_processed_entry(processed_entry, test_case.expected_output, test_case.name)
 
-    wrong_order = tuple(
-        ["depends_on_A"] + [i for i in config_dependency.processing_order if i != "depends_on_A"]
+
```
anna-parker, line 964: nice tests!

kubernetes/loculus/values.schema.json
anna-parker: ah we should probably make it explicit in the docs that these conditions are not AND but OR

maverbiest: Alright I pushed my commits and updated the docs

preprocessing/nextclade/tests/metadata_dependency.yaml
```diff
@@ -14,3 +14,25 @@ processing_spec:
...
+      - B
+  B:
+    function: identity
+    inputs:
+      input: B
+  required_when_processed_B:
+    function: identity
+    inputs:
+      input: required_when_processed_B
+    required_when:
```
anna-parker, line 33: It would be cool to have an example where there are multiple fields in the `required_when`.

maverbiest: I added this now!

preprocessing/nextclade/src/loculus_preprocessing/config.py
```suggestion
    Dependencies are derived from input fields and requiredWhen fields in `config.processing_spec`.
```

preprocessing/nextclade/src/loculus_preprocessing/config.py
```diff
@@ -244,6 +249,49 @@ def generate_argparse_from_model(config_cls: type[BaseModel]) -> argparse.Argume
...
+                        f"condition referencing unknown file category '{category}'."
+                    )
+                    raise ValueError(msg)
+                continue
+            field = (
+                condition.removeprefix(PROCESSED_PREFIX)
+                if condition.startswith(PROCESSED_PREFIX)
+                else condition
+            )
+            if (referenced_spec := config.processing_spec.get(field)) is None:
```
anna-parker: Its a bit weird to only validate that field exists for `processed.field` and not for `field` or `files.field`.

However, I understand that if we do this then the config will fail with these values.yaml changes as `files.rawReads` doesnt exist (only on the preview).

Either I would remove the validation here (it will fail in `get_processing_order` (message might need improvement) or I would add validation for all three cases and not make the values.yaml changes (at least in this PR)

preprocessing/nextclade/src/loculus_preprocessing/config.py
```diff
@@ -257,25 +305,30 @@ def get_processing_order(config: Config) -> tuple[str, ...]:
...
                 continue
-            dependency = input.replace(METADATA_DEPENDENCY_PREFIX, "", 1)
+            dependency = input.removeprefix(PROCESSED_PREFIX)
+
             if dependency not in config.processing_spec:
-                raise ValueError(
-                    f"invalid configuration: metadata field '{output_field}' requested non-existing field '{dependency}' as input"
+                msg = (
+                    f"invalid configuration: metadata field '{output_field}' requested "
+                    f"non-existing field '{dependency}' as input."
```
```suggestion
                    f"non-existing field '{dependency}' as input or as a requiredWhen condition."
```

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -416,30 +418,86 @@ def get_output_metadata(
...
+        if unprocessed.files and unprocessed.files.get(FileCategory(file_category)):
+            error_message = (
+                f"Metadata field `{output_field}` is required when "
+                f"`{file_category}` files are provided."
+            )
+    elif condition.startswith(PROCESSED_PREFIX):
+        field_name = condition.removeprefix(PROCESSED_PREFIX)
+        if not null_per_backend(output_metadata.get(field_name)):
+            error_message = (
+                f"Metadata field `{output_field}` is required when `{field_name}` is provided."
```
anna-parker: this actually isnt quite accurate - if the required field is the processed field its only required when the processed field exists.

preprocessing/nextclade/tests/test_metadata_processing_functions.py
```diff
@@ -717,6 +729,218 @@
...
+                    ["multi_dep"],
+                    ["multi_dep"],
+                    "Metadata field `multi_dep` is required when `A` is provided.",
+                ),
+            ]
+        ),
+        expected_warnings=[],
+    ),
+    Case(
+        name="multi_dep_fails_when_all_present",
```
anna-parker:
```suggestion
        name="missing_multi_dep_fails_when_all_requiredWhen_condition_present",
```

preprocessing/nextclade/tests/test_metadata_processing_functions.py
```diff
@@ -717,6 +729,218 @@
...
+                    ["required_when_input_A"],
+                    ["required_when_input_A"],
+                    "Metadata field `required_when_input_A` is required when `A` is provided.",
+                ),
+            ]
+        ),
+        expected_warnings=[],
+    ),
+    Case(
+        name="multi_dep_fails_when_one_present",
```
anna-parker:
```suggestion
        name="missing_multi_dep_fails_when_one_requiredWhen_condition_present",
```

preprocessing/nextclade/tests/test_metadata_processing_functions.py
```diff
@@ -687,6 +689,10 @@
     "Count4th, EwanMcGregor, Count4th",
 ]
 
+RAW_READS_FILES = {
+    FileCategory.RAW_READS: [FileIdAndName(fileId="file-raw-reads", name="reads.fastq.gz")]
+}
+
 test_metadata_dependency_test_definitions = [
     Case(
         name="metadata_dependency",
```
anna-parker:
```suggestion
        name="metadata_dependencies_all_present",
```

anna-parker: I might just add the `raw_reads_prerequisite_present` case here to have less tests (coverage is still there :-) )

anna-parker (commented):
this looks great - I have some minor nits about test names and some warning messages, I think it would be best to extend the validation and then remove the requiredWhen files from the values.yaml :-)

preprocessing/nextclade/src/loculus_preprocessing/config.py
maverbiest: Yes good point! I will rework this a bit and remove the `requiredWhen` from the main values.yaml

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
maverbiest: True! But I wasn't sure if we want to make that distinction in the user-facing message

preprocessing/nextclade/src/loculus_preprocessing/config.py
maverbiest: I updated the validation now to have a clearer distinction between the `processed.<field>` and `<field>` cases. 

I am now actually debating if we want validation on whether a file category is enabled on a specific organism, as it seems to me that it would require much more duplication in the values.yaml in situations were a filetype is enabled only for certain organisms. It's also not super clear to me what failure mode we'd be guarding against by adding it

To me it doesn't seem so bad to just validate that the FileCategory in a requiredWhen exists and not check if the FileCategory is enabled on the organism. But lmk if you disagree and I will add it!

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
maverbiest: Now reworded to:

```python
            error_message = (
                f"Metadata field `{output_field}` is required when `{field_name}` exists."
            )
```

anna-parker:
could you rebase on main?

preprocessing/nextclade/src/loculus_preprocessing/config.py
anna-parker: I would then opt to just remove the validation entirely of existing metadata fields and categories entirely, as otherwise these are two different checks. You are now checking if the metadata field exists for that specific organism and if there is code for that file type - not if that file category exists for that organism. File categories like metadata are configured on an organism level.

The requiredWhen check only throws an error if the field mentioned in the requiredWhen condition is supplied and the field with the condition does not exist. If you add a field that doesnt exist into the requiredWhen condition the condition will never be triggered.

maverbiest: Yeah you're right, I removed the existance validations. The only thing we lose is that somone could mistype the name of an input metadata field so the requiredWhen condition will silently do nothing, but probably not too big of a deal. Like you said, the processed.X conditions will still be caught by `get_processing_order`

anna-parker: I think this is fine for now!

anna-parker (approved):
awesome thanks!

PR #6940, opened 2026-07-21
fix(website): render login prompt instead of blank page on released sequences page

### LINKED ISSUE #6939: Released sequences page renders a blank page when logged out
## Summary

The "Released sequences" submission page returns an **empty HTTP 200 with a zero-byte body** when the user is not logged in, so the browser shows a completely blank white page instead of the "You need to be logged in" prompt that every other submission page shows.

## Reproduction

Log out (or open a preview instance you have never authenticated against) and request:

```
/<organism>/submission/<groupId>/released
```

Observed on `main.loculus.org` as well as on preview deployments:

| Path | Response |
| --- | --- |
| `/ebola-sudan/submission/1/released` | `200`, **0 bytes**, blank page |
| `/ebola-sudan/submission/1/review` | `200`, ~30 KB, renders "You need to be logged in" |
| `/ebola-sudan/submission/1/revise` | `200`, ~30 KB, renders "You need to be logged in" |
| `/ebola-sudan/submission/1/submit` | `200`, ~30 KB, renders "You need to be logged in" |

The response has no `Content-Type`, no `Location` and no body — just a bare chunked `200`.

## Cause

`getGroups` in `website/src/utils/submissionPages.ts` reports the logged-out case with an HTTP status of `200`:

```ts
return err({
    type: 'not_logged_in',
    status: 200,
});
```

`released.astro` is the only submission page that early-returns on that error and forwards the status into a body-less response:

```ts
const groupsResult = await getGroupsAndCurrentGroup(Astro.params, Astro.locals.session);
if (groupsResult.isErr()) {
    return new Response(undefined, { status: groupsResult.error.status });
}
```

So `not_logged_in` becomes a literal empty `200`.

The sibling pages (`review`, `revise`, `submit`, `index`) never early-return. They pass `groupsResult` down to `SubmissionPageWrapper`, whose `.match()` already has a `case 'not_logged_in'` arm rendering `NeedToLogin`. Because `released.astro` returns before the wrapper is reached, that arm can never fire.

`released.astro` is also the only place in the codebase that reads `error.status`, so the `status` field on these error types is effectively dead weight that exists only to enable this bug.

## Impact

- Anyone hitting a released-page URL without a session sees a blank page with no explanation and no way to log in. This is easy to run into on fresh preview deployments, and on shared/bookmarked links after a session expires.
- Because it is a `200` with an empty body rather than a `4xx`, it is invisible to uptime and error monitoring.

## Expected behaviour

The released sequences page should render the same "You need to be logged in" prompt as the other submission pages, and `group_not_found` should render "Group not found" rather than a blank page.

## Notes

This is long-standing rather than a recent regression — `git log -L` traces the early-return to a27d203f6 (April 2024), when the submission page paths were restructured.

theosanderson-agent:
Fixes #6939.

https://fix-released-page-blank-w.loculus.org/ebola-sudan/submission/1/released?column_dataUseTerms=true

vs

https://main.loculus.org/ebola-sudan/submission/1/released?column_dataUseTerms=true

## The problem

Requesting `/<organism>/submission/<groupId>/released` while logged out returns an **empty HTTP 200 with a zero-byte body**, so the browser renders a blank white page. Every other submission page under the same route renders a "You need to be logged in" prompt:

| Path | Before |
| --- | --- |
| `/ebola-sudan/submission/1/released` | `200`, **0 bytes**, blank page |
| `/ebola-sudan/submission/1/review` | `200`, ~30 KB, login prompt |
| `/ebola-sudan/submission/1/revise` | `200`, ~30 KB, login prompt |
| `/ebola-sudan/submission/1/submit` | `200`, ~30 KB, login prompt |

This is reproducible on `main.loculus.org`, not just on preview deployments, and it is long-standing — `git log -L` traces it to a27d203f6 (April 2024), when the submission paths were restructured.

## Why it happened

Two things combined.

`getGroups` reports the logged-out case with an HTTP status of `200`:

```ts
return err({ type: 'not_logged_in', status: 200 });
```

And `released.astro` was the only submission page that early-returned on a groups error, forwarding that status into a body-less response:

```ts
if (groupsResult.isErr()) {
    return new Response(undefined, { status: groupsResult.error.status });
}
```

So `not_logged_in` became a literal empty `200`. `SubmissionPageWrapper` already handles this case — its `.match()` has a `case 'not_logged_in'` arm rendering `NeedToLogin`, plus a `group_not_found` arm — but the early return meant the wrapper was never reached, so those arms could not fire.

Because it surfaced as a `200` rather than a `4xx`, the failure was also invisible to error monitoring.

## The change

**`released.astro`** now follows the same shape as `review.astro`: `groupsResult` is passed to `SubmissionPageWrapper`, which renders the login prompt or the group error, and `SearchFullUI` is rendered only when a group is actually available.

The one wrinkle versus the sibling pages is that this page does server-side work that depends on the group — `performLapisSearchQueries` needs `group.groupId` for `hiddenFieldValues`. That work moved into a `loadReleasedSequences(group)` helper called only in the `isOk()` branch, so the logged-out path no longer touches LAPIS at all. Everything group-independent (schema, reference genomes, config, parsed query params) is still computed unconditionally in the frontmatter.

**`submissionPages.ts`**: the `status` field on `GetGroupsError` / `GetGroupsAndCurrentGroupError` is removed. The early return in `released.astro` was its only reader anywhere in the codebase, so it was dead weight whose sole effect was to make this bug expressible.

Also converted the adjacent unknown-organism branch from `return { statusCode: 404, body: 'Organism not found' }` to a real `new Response('Organism not found', { status: 404 })` — Astro frontmatter requires a `Response`, so the plain object was never going to work as intended.

## Behaviour after this change

- Logged out → "You need to be logged in to submit sequences." (matches review/revise/submit)
- Logged in, group you are not a member of → "Group not found"
- Logged in, valid group → unchanged; the released sequences table renders as before

Note that error cases now return `200` with a rendered explanation rather than propagating `404`/`500` status codes. That is a deliberate trade for consistency with the sibling submission pages, which all behave this way.

### DIFF
--- a/website/src/pages/[organism]/submission/[groupId]/released.astro
+++ b/website/src/pages/[organism]/submission/[groupId]/released.astro
@@ -11,6 +11,7 @@ import {
     getWebsiteConfig,
 } from '../../../../config';
 import { GROUP_ID_FIELD, VERSION_STATUS_FIELD } from '../../../../settings';
+import type { Group } from '../../../../types/backend';
 import { versionStatuses } from '../../../../types/lapis';
 import { getAccessToken } from '../../../../utils/getAccessToken';
 import { parseUrlSearchParams } from '../../../../utils/parseUrlSearchParams';
@@ -18,18 +19,11 @@ import { performLapisSearchQueries } from '../../../../utils/serversideSearch';
 import { getGroupsAndCurrentGroup } from '../../../../utils/submissionPages';
 
 const groupsResult = await getGroupsAndCurrentGroup(Astro.params, Astro.locals.session);
-if (groupsResult.isErr()) {
-    return new Response(undefined, { status: groupsResult.error.status });
-}
-const { currentGroup: group } = groupsResult.value;
 
 const { organism: cleanedOrganism } = cleanOrganism(Astro.params.organism);
 
 if (!cleanedOrganism) {
-    return {
-        statusCode: 404,
-        body: 'Organism not found',
-    };
+    return new Response('Organism not found', { status: 404 });
 }
 
 const clientConfig = getRuntimeConfig().public;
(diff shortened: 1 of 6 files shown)


theosanderson:
@codex review

website/src/pages/[organism]/submission/[groupId]/released.astro
```diff
@@ -11,25 +11,19 @@ import {
...
-const { currentGroup: group } = groupsResult.value;
 
 const { organism: cleanedOrganism } = cleanOrganism(Astro.params.organism);
 
 if (!cleanedOrganism) {
-    return {
-        statusCode: 404,
-        body: 'Organism not found',
-    };
+    return new Response('Organism not found', { status: 404 });
```
Comment (tombch) [inline website/src/pages/[organism]/submission/[groupId]/released.astro]: Not an issue introduced in this PR but just realised - this code is never hit. The `organismValidatorMiddleware` validates and cleans the organism for all of the pages, and if it did not do this (and we hit this old code which didn't use `Response`) it would have returned a 500.

There's a few other pages that have the previous pattern as well (which would also throw an error if not for the middleware). I guess we could either remove the duplicate `cleanOrganism` checks, or ensure they return `new Response(undefined, { status: 404 })` - as when it returns a string 'Organism not found' it doesn't show the login page (only the literal text 'organism not found'). I could open a PR for this if useful!

Comment (theosanderson) [reply website/src/pages/[organism]/submission/[groupId]/released.astro]: I haven't looked into the details of this but happy to trust your interpretation. One thing that may or may not be relevant here is that sometimes the typechecker requires branches that we know in practice will never be called. Feel free to push to this PR, or make a PR or whatever you like.

Comment (tombch) [reply website/src/pages/[organism]/submission/[groupId]/released.astro]: Ok cool, I just opened a PR to add to this one!

tombch (approved):
Looks good to me!

PR #6942, opened 2026-07-21
fix(website): compute restricted-until date bounds in UTC

### LINKED ISSUE #6941: Selectable/default restricted use until date exceeds 1 year
The backend checks that the specified restrictedUntil date is not more than 1 year from today but the applied timezone may be different than on the website. I wanted to submit sequences at 1:19am CEST on 20 July 2026 and the default and max. selectable restriction date was displayed as 20 July 2027 but the server rejected it with the error message "Failed to submit sequence enries: The date 'restrictedUntil' most not exceed 1 year from today."

Image

theosanderson-agent:
Fixes #6941

## The problem

The backend validates `restrictedUntil` against **today in UTC** — `DateProvider` hard-codes `TimeZone.UTC`, and `DataUseTermsPreconditionValidator.checkThatRestrictedUntilDateValid` rejects anything later than `now.plus(1, YEAR)`.

The website, however, derived the default and maximum restriction date from the **browser's local date**: `dateTimeInMonths` was `DateTime.now().plus({ months })`. For any user in a timezone ahead of UTC, those two disagree for part of the day, and the website then offers a maximum date that the backend refuses.

That is exactly the reported scenario. At 01:19 CEST on 20 July 2026 (= 23:19 UTC on 19 July):

| | today | max `restrictedUntil` |
|---|---|---|
| Website (local, Europe/Berlin) | 2026-07-20 | **2027-07-20** |
| Backend (UTC) | 2026-07-19 | 2027-07-19 |

So the date picker pre-selected and permitted 2027-07-20, and submission failed with `The date 'restrictedUntil' must not exceed 1 year from today.` — with no date the user could obviously blame, since the picker itself said the date was allowed.

There is a mirror-image bug on the lower bound. Both date pickers used `minDate={new Date()}`, i.e. the local date. For a user *behind* UTC (say 18:00 in Los Angeles, which is already the next day in UTC), the picker allows picking the local "today", which the backend has already moved past and rejects with `The date 'restrictedUntil' must be in the future, …`.

## The fix

Add a `utcToday()` helper in `website/src/utils/DateTimeInMonths.tsx` that returns the current **UTC** calendar date, and build `dateTimeInMonths()` on top of it. Both `DataUseTermsSelector` and `DateChangeModal` now use it for `minDate` instead of `new Date()`.

One subtlety worth flagging for review: the helper deliberately returns the UTC calendar date *expressed in the local zone* (`DateTime.utc().startOf('day').setZone('local', { keepLocalTime: true })`). A plain `DateTime.utc()` would not have fixed anything, because the flowbite `Datepicker` takes JS `Date` objects and renders them in the local zone — `DateTime.utc().toJSDate()` would land back on the local calendar date and reproduce the original off-by-one. Keeping the local time and swapping the zone means `.toJSDate()` and `.toFormat('yyyy-MM-dd')` both yield the UTC date, which is what the backend compares against.

Because the UTC date only ever moves forward relative to a page that was loaded earlier, there is no remaining race: a bound computed when the page loaded is, at worst, one day stricter than what the backend will accept by the time the user submits.

## Tests

New `website/src/utils/DateTimeInMonths.spec.ts` freezes the clock and the luxon default zone to cover both directions — a timezone a day ahead of UTC (Europe/Berlin at the exact instant from the issue) and one a day behind (America/Los_Angeles) — plus a check that the JS `Date` round-trip used by the date pickers preserves the calendar date. All four tests fail against the previous implementation.

Full website suite passes (709 tests), along with `check-types` and `format`.

## Note on scope

This fixes the website, which is where the issue was reported and labelled. The backend contract is unchanged, so any other API client that computes its bound in local time can still hit the same rejection. If you'd rather make the backend tolerant of client/server timezone skew — e.g. allowing `now + 1 year + 1 day` — that would be a separate, complementary change; happy to add it if you want it.


theosanderson:
@theosanderson-agent would the test be clearer if you used e.g. Thailand rather than Berlin?

theosanderson-agent:
Yes — switched to `Asia/Bangkok` (UTC+7) for the ahead-of-UTC case and `Pacific/Honolulu` (UTC-10) for the behind-UTC case, in 458bf197f.

Two things got clearer:

- Neither zone observes daylight saving, so the offset is a fixed, obvious constant. Berlin is UTC+2 in July but UTC+1 in January, and Los Angeles shifts the same way, so a reader had to know the season to work out what the frozen instant meant locally.
- The frozen instant moved from `23:19Z` to `20:00Z`. With Berlin that only worked because 23:19Z is within 41 minutes of midnight, which reads like an arbitrary edge case; with Bangkok it is unambiguously 03:00 the next day, four hours clear of the boundary.

Also hoisted the shared `freezeTime` helper and `afterEach` cleanup out of the two `describe` blocks, since both suites now use the same fixtures. All 4 tests pass.

theosanderson:
@theosanderson-agent address reviewer comments

theosanderson-agent:
Both review nits addressed in 1afc9d69d.

**File name.** Renamed `website/src/utils/DateTimeInMonths.tsx` → `website/src/utils/utcDates.ts`. Two things were off: the name only described one of the two exports now that `utcToday` lives there, and the `.tsx` extension claimed JSX the file has never contained. `utcDates.ts` matches the camelCase `.ts` convention the other plain helpers in `src/utils` use (`parseFasta.ts`, `sequenceTypeHelpers.ts`, …), with `.tsx`/PascalCase reserved for components like `CustomTooltip.tsx`. The spec file moved alongside it.

**Import inconsistency.** All four import sites now use the `.ts` extension. `DateChangeModal.tsx` was the extensionless one; the other three already carried an extension. Note this inconsistency is repo-wide rather than specific to this PR (e.g. `getAccessToken` and `sequenceTypeHelpers` are each imported both ways elsewhere), so I only normalised the sites this PR touches rather than doing a sweep.

No behaviour change — pure rename plus import updates. `CI=1 npm run test` (709 tests, 67 files), `npm run check-types` (0 errors) and `npm run format` all pass locally.

website/src/utils/utcDates.ts
```diff
@@ -0,0 +1,13 @@
+import { DateTime } from 'luxon';
+
+/**
+ * The backend validates `restrictedUntil` against the current date in UTC, so the website has to use
+ * the UTC date as well. Otherwise users in timezones ahead of UTC are offered a maximum date that the
+ * backend rejects, and users in timezones behind UTC cannot pick the backend's "today".
+ *
+ * The returned DateTime holds the UTC calendar date but is placed in the local zone, so that
+ * converting it to a JS `Date` for the date pickers preserves that calendar date.
+ */
```
chaoran-chen, line 6: Not sure whether we need this long explanation here, if we want to explain this concept, it would be better to have it somewhere where we actually have code related to restrictedUtil. I'd probably just remove the comment but no big deal.

chaoran-chen (approved):
Thanks very much for the quick fix!

website/src/utils/utcDates.ts
theosanderson: @theosanderson-agent address

theosanderson-agent: Trimmed it in 30a54fa44 — dropped the `restrictedUntil`/backend rationale (that belongs with the code that uses it, or in the PR description) and kept only the bit about the local-zone placement, since that trick is surprising when reading this function on its own.

PR #6947, opened 2026-07-22
Fix (currently unused) not found responses

tombch:
Adds to #6940 by fixing the other 404 response handling, and updates the released page 404 to use `undefined` so that the base astro 404 page is displayed (this is an existing pattern in the website).

This code isn't actually hit due to the `organismValidatorMiddleware`, but good to have for type checking and best for it to work in case that middleware was ever removed/changed.

### Screenshot

With the organism validator middleware disabled, and an incorrect organism in the URL, this is what would happen before:

image

And now:

image

### PR Checklist

### DIFF
--- a/website/src/pages/[organism]/search/index.astro
+++ b/website/src/pages/[organism]/search/index.astro
@@ -27,10 +27,7 @@ const hiddenFieldValues = {
 const { organism: cleanedOrganism } = cleanOrganism(Astro.params.organism);
 
 if (!cleanedOrganism) {
-    return {
-        statusCode: 404,
-        body: 'Organism not found',
-    };
+    return new Response(undefined, { status: 404 });
 }
 
 const clientConfig = getRuntimeConfig().public;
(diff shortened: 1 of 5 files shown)


website/src/pages/[organism]/search/index.astro
```diff
@@ -27,10 +27,7 @@ const hiddenFieldValues = {
 const { organism: cleanedOrganism } = cleanOrganism(Astro.params.organism);
 
 if (!cleanedOrganism) {
-    return {
-        statusCode: 404,
-        body: 'Organism not found',
-    };
+    return new Response(undefined, { status: 404 });
```
Comment (theosanderson) [inline website/src/pages/[organism]/search/index.astro:32]: Even if the new code is more correct it looks less clear so maybe a comment is needed to explain why we don't need a nice error message and whatever
```suggestion
    // This branch should never be reached due to middleware checking in advance, but is needed for Typecheck
    return new Response(undefined, { status: 404 });
```

theosanderson (approved):

website/src/pages/[organism]/search/index.astro
Comment (tombch) [reply website/src/pages/[organism]/search/index.astro:32]: Yep I added comments for each!

Issue #6963, opened 2026-07-22
Default to single sequence revision and guide through bulk revision


theosanderson:
Default to single sequence revision to keep it simple for users with lower tech skills

Also on the bulk page provide a guide at the top for how to get original metadata

theosanderson:
@theosanderson-agent create a PR that makes the single sequence form version of revision the default and also the left-most option in the menu, with bulk revision to the right of it. Name on menu can be "Revise single sequence" - we don't need the reference to the form now. Other menu name can be "Upload bulk revisions".

theosanderson-agent:
Opened #6964 (draft).

Revision now defaults to `inputMode=form`, so `/<organism>/submission/<groupId>/revise` lands on the single-sequence form. The tabs on the revise page are `Revise single sequence` (left, default) and `Upload bulk revisions` (right) — the "using a form" wording is gone. Submission is untouched and still defaults to bulk.

Also updated: `routes.revisePage` default, the `Revise` card description on the submission portal, the revise-sequences docs page (now distinguishes the two flows and points at the bulk tab), and three integration tests that relied on the revise page opening in bulk mode.

Not done: the second part of this issue — a guide at the top of the bulk page for how to get the original metadata. Happy to do that as a follow-up.

PR #6964, opened 2026-07-22
feat(website): default revision to the single sequence form

theosanderson-agent:
Closes part of #6963.

## What this changes

Revising a single sequence via the form is now the default experience on the revise page, so users who want to fix one sequence no longer have to discover a tab and never have to build a metadata TSV. Bulk revision is still one click away.

Concretely:

- **The revise page defaults to `inputMode=form`.** Visiting `/<organism>/submission/<groupId>/revise` with no `inputMode` parameter (or with an unrecognised value) now lands on the single-sequence form; you get the bulk upload flow with `?inputMode=bulk`. This is the mirror image of the previous behaviour, and it is scoped to revision only — submission still defaults to bulk upload, since new submissions are much more often bulk.
- **The tab order is flipped for revision, and the labels are new.** On the revise page the tabs now read `Revise single sequence` (left, default) and `Upload bulk revisions` (right). The old label mentioned the form as an implementation detail (`Revise individual sequence entry using a form`); that reference is dropped. Submission tabs keep their existing order and labels.
- `InputModeTabs` was refactored from two hard-coded anchors into a small per-action list of `{inputMode, label}` rendered in order, which is what lets revision and submission differ in both order and wording without duplicating the anchor markup.
- The submission portal card for `Revise` said "Upload revisions for existing sequences", which no longer describes where the link takes you; it now says "Update sequences that you have already submitted."

## Screenshots

All screenshots are of the West Nile Virus revise flow for a group with sequences, at a 1280px viewport.

### The revise page you land on, before and after

Previously, `/<organism>/submission/<groupId>/revise` opened on the bulk upload tab, and the way to revise one sequence was the second tab, labelled `Revise individual sequence entry using a form`:

Revise page before this change: the bulk upload tab is selected, and the tabs read "Upload bulk sequences" and "Revise individual sequence entry using a form"

Now the same URL opens on the single-sequence form. The tab order is reversed for revision and both labels are new, so the bulk route is still visible and one click away:

Revise page after this change: the "Revise single sequence" tab is selected and shows the accession search, with "Upload bulk revisions" as the second tab

### The bulk tab is unchanged apart from its label and position

`?inputMode=bulk` (or clicking the second tab) gives the same bulk upload page as before:

Revise page with the "Upload bulk revisions" tab selected, showing the unchanged sequence file and metadata file upload panels

### Submission portal card wording

The `Revise` card used to promise an upload, which is no longer where the link goes.

Before:

Submission portal before: the Revise card reads "Upload revisions for existing sequences."

After:

Submission portal after: the Revise card reads "Update sequences that you have already submitted."

## Other updates

- `routes.revisePage(...)` defaults its `inputMode` argument to `'form'` to match the page. All existing callers that care (the "Revise this sequence" button on the sequence details page, and the edit page) already pass `'form'` explicitly.
- `SubmissionRouteUtils.parseToRoute` for `revise` now treats anything other than `bulk` as `form`, with new unit tests covering both the explicit and default cases.
- Three integration tests exercised the bulk revision flow by navigating to the revise page without an `inputMode`, and would now land on the form. They pass `'bulk'` explicitly: the two template-download / bulk-upload tests in `revise-sequence.spec.ts` and the bulk file-sharing revision test in `file-sharing.spec.ts`.
- The user docs for revising sequences describe the bulk flow throughout. They now open with a short paragraph distinguishing the two routes, and the "Submitting the revision" section tells you to select the `Upload bulk revisions` tab.

## Not included

The second half of #6963 — a guide at the top of the bulk revision page explaining how to obtain the original metadata — is not addressed here. That content partly exists in the docs already ("Downloading originally submitted data for revision"), so surfacing it in the UI is worth doing as its own change.

## Test plan

- `CI=1 npm run test` (website) passes, including the new `SubmissionRoute` cases.
- `npm run check-types` passes with 0 errors.
- `npm run format` clean in both `website` and `integration-tests`.
- Integration tests were not run locally (they need a deployed cluster); the affected specs were updated to pin the bulk mode they rely on.

### DIFF
--- a/website/src/pages/[organism]/submission/[groupId]/index.astro
+++ b/website/src/pages/[organism]/submission/[groupId]/index.astro
@@ -56,7 +56,7 @@ async function getSequenceCounts(organism: string, groupId: number) {
                     },
                     {
                         title: 'Revise',
-                        description: 'Upload revisions for existing sequences.',
+                        description: 'Update sequences that you have already submitted.',
                         route: routes.revisePage(organism, group.groupId),
                         icon: F7Arrow2Circlepath,
                     },
--- a/website/src/pages/[organism]/submission/[groupId]/index.astro
+++ b/website/src/pages/[organism]/submission/[groupId]/index.astro
@@ -56,7 +56,7 @@ async function getSequenceCounts(organism: string, groupId: number) {
                     },
                     {
                         title: 'Revise',
-                        description: 'Update sequences that you have already submitted.',
+                        description: 'Update sequences that you have already released.',
                         route: routes.revisePage(organism, group.groupId),
                         icon: F7Arrow2Circlepath,
                     },
--- a/docs/src/content/docs/for-users/revise-sequences.md
+++ b/docs/src/content/docs/for-users/revise-sequences.md
@@ -4,7 +4,9 @@ title: Revise sequences
 
 Sequences can be corrected or updated after they have been submitted to Loculus. Submitting these changes ("revisions") will cause the [version](../../reference/glossary/#version) of the sequence to be incremented, and previous versions of the metadata and sequence data can always be accessed via previous version numbers.
 
-The process of submitting revisions is very similar to original submission. The main difference is that you must provide an `accession` column in the metadata file, which contains the Loculus [accession number(s)](../../reference/glossary/#accession) assigned when the sequences were submitted originally.
+There are two ways to revise sequences. To change a single sequence, use the **Revise single sequence** page, look up the accession you want to revise, and edit its values directly in the form. To revise many sequences at once, switch to the **Upload bulk revisions** tab and upload a metadata file and a FASTA file, as described below.
+
+The bulk process is very similar to original submission. The main difference is that you must provide an `accession` column in the metadata file, which contains the Loculus [accession number(s)](../../reference/glossary/#accession) assigned when the sequences were submitted originally.
 
 ## Downloading originally submitted data for revision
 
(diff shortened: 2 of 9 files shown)


theosanderson:
@theosanderson-agent update PR description with screenshots (can use the preview perhaps)

tombch (commented):
I can see the rationale for prioritising bulk for submission and form for revision, but I also can't help feel like the tabs should be in the same order on both the submit and revise pages 🙃

theosanderson:
I think defaulting to a right tab is more confusing than the inconsistency

website/src/pages/[organism]/submission/[groupId]/index.astro
```diff
@@ -56,7 +56,7 @@ async function getSequenceCounts(organism: string, groupId: number) {
                     },
                     {
                         title: 'Revise',
-                        description: 'Upload revisions for existing sequences.',
+                        description: 'Update sequences that you have already submitted.',
```
Comment (tombch) [inline website/src/pages/[organism]/submission/[groupId]/index.astro]: Could this be `Update sequences that you have already released` instead? As I'm thinking that could clearly distinguish from editing an unreleased submission

Comment (tombch) [inline website/src/pages/[organism]/submission/[groupId]/index.astro]:
```suggestion
                        description: 'Update sequences that you have already released.',
```

tombch (approved):
Aside from one comment, looks good to me!

website/src/pages/[organism]/submission/[groupId]/index.astro
Comment (theosanderson) [reply website/src/pages/[organism]/submission/[groupId]/index.astro]: nice catch

Comment (theosanderson) [reply website/src/pages/[organism]/submission/[groupId]/index.astro]: @theosanderson-agent implement (and fix any tests affected)

Comment (theosanderson-agent) [reply website/src/pages/[organism]/submission/[groupId]/index.astro]: Done in 80ab5ec — changed the Revise description to "Update sequences that you have already released." No tests referenced the old string, so nothing else needed updating.

anna-parker:
Sorry to jump in - Im very fine with switching the default tab but I also agree that it would be good to keep the tab order consistent, is that still possible?

theosanderson:
@theosanderson-agent make a second PR with tab order reversed and add a preview tag

theosanderson:
I really think there is a significant chance people fail to find the bulk revision option if we start on the right tab: I think that not loading on the left most tab is really unexpected.

theosanderson-agent:
Done — opened #6973 as an alternative to this PR. It keeps the single-sequence form as the default revise mode but restores the original tab order (**Upload bulk revisions** left-most, **Revise single sequence** second) and adds a small **Preview** badge to the single-sequence tab. It's a draft based on this branch; type-check and route tests pass. The two PRs are mutually exclusive presentations of the same feature.

PR #6966, opened 2026-07-22
feat(config): use Traefik 3 for loculus preview server (included in kubernetes version bump from v1.31 to v1.32)

anna-parker:
### PR Checklist

🚀 Preview: https://update-traefik.loculus.org

### DIFF
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -2695,7 +2695,7 @@ runDevelopmentMainDatabase: true
 runDevelopmentS3: true
 developmentDatabasePersistence: false
 enforceHTTPS: true
-traefikVersion: 2
+traefikVersion: 3
 registrationTermsMessage: >
   You must agree to the <a href="http://main.loculus.org/terms">terms of use</a>.
 
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -2695,7 +2695,7 @@ runDevelopmentMainDatabase: true
 runDevelopmentS3: true
 developmentDatabasePersistence: false
 enforceHTTPS: true
-traefikVersion: 3
+traefikVersion: 2
 registrationTermsMessage: >
   You must agree to the <a href="http://main.loculus.org/terms">terms of use</a>.
 
--- a/kubernetes/loculus/values_preview_server.yaml
+++ b/kubernetes/loculus/values_preview_server.yaml
@@ -1,4 +1,5 @@
 createTestAccounts: true
+traefikVersion: 3
 secrets:
   smtp-password:
     type: sealedsecret


kubernetes/loculus/values.yaml
theosanderson, line 1: This should definitely be marked as a breaking change

anna-parker: I updated the description :-)

theosanderson: I guess I would at least tell people in the PR to set to 2 to maintain compatibility?

theosanderson: > he traefik version default stays as v2. Only for users using this values.yaml might this lead to changes

No this isn't right? The values.yaml sets the default for the chart

theosanderson: If we were stressed about documenting it we could just update `values_e2e` and `values_preview_server` for now - but I'm also fine with us merging if we make PR have some guidance

theosanderson (approved):

kubernetes/loculus/values.yaml
anna-parker: Hmm... then maybe I dont understand what values.schema.json default 2 does

anna-parker: As I want to fix main I will just update the description and check later

theosanderson: > what values.schema.json default 2 does

essentially nothing, I think - just documentation. Agree that it may be unhelpful to have that second "default". @theosanderson-agent , pls correct me if wrong.

theosanderson-agent: Mostly right, with one wrinkle worth knowing.

For Helm itself, `default` in `values.schema.json` is inert: Helm only uses the schema for *validation*, it never injects defaults into the values. So the effective default is entirely `values.yaml` — which this PR changes to `3`. Nothing breaks by leaving the schema saying `2`.

But it is not purely decorative either: our docs site reads this file. `docs/package.json` copies `values.schema.json` into `docs/src/`, and `docs/src/components/SchemaDocs.astro` renders a table with a **Default** column straight from `definition.default` for every property carrying a `groups` key. `traefikVersion` has `"groups": ["general"]`, so it shows up in the published config reference — and after this PR it will publicly document the default as `2` while the chart actually ships `3`.

So I would bump the schema default to `3` in the same PR rather than delete it (deleting it just leaves a blank cell in the docs table). One line:

```json
"traefikVersion": {
  "groups": ["general"],
  "type": "integer",
  "enum": [2, 3],
  "default": 3,
  ...
}
```

The general "two defaults is confusing" point stands though — the schema default is a hand-maintained copy of `values.yaml` with nothing enforcing that they agree, and this is exactly the drift it invites. Might be worth a CI check that every schema `default` matches the corresponding `values.yaml` value, if we hit this again.

theosanderson: @theosanderson-agent make a follow up PR for that documentation bump

theosanderson-agent: Done: #6967 (draft) — one line, `"default": 2` → `"default": 3` in `kubernetes/loculus/values.schema.json`.

It targets `main` and should be merged after this PR, since merging it first would just flip the drift the other way (docs saying 3 while the chart still ships 2). Docs-only either way. `helm lint` passes and prettier reports no reformatting needed.

I left the "two defaults are confusing" point as a note in that PR rather than acting on it — a CI check that every schema `default` matches the value at the same key in `values.yaml` would prevent a repeat, but it felt like scope creep for a one-line fix.

theosanderson: (ah sorry I misunderstood your plans and thought you planned to merge as is - maybe we don't need follow up PR :) )

PR #6978, opened 2026-07-24
feat!(config): Loculus now expects traefik v3 by default (+ upgrades the k3s version used in integration tests)

anna-parker:
This PR makes Loculus default to expect Traefik v3 rather than Traefik v2.

It also upgrades the k3s version used by default in the deploy script (used by integration tests and for local development), using the same version used on loculus previews and pathoplexus. Upgrade from `v1.31.14` to `v1.35.6, which is what necessitates the change to v3 since later versions are bundled with traefik v3 rather than v2.

### Breaking Changes
Pathoplexus does not need to make any changes. Other users of Loculus can continue to stay on earlier versions of traefik (which are bundled in early versions of k3s) but to do so should overwrite the declared traefik version in their values.yaml to `traefikVersion: 2` in order to continue to use the previous k3s version. However, we would recommend Loculus users to upgrade their traefik version, as traefik v2 may be deprecated or have support removed in future.

### DIFF
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -2696,7 +2696,7 @@ runDevelopmentMainDatabase: true
 runDevelopmentS3: true
 developmentDatabasePersistence: false
 enforceHTTPS: true
-traefikVersion: 2
+traefikVersion: 3
 registrationTermsMessage: >
   You must agree to the <a href="http://main.loculus.org/terms">terms of use</a>.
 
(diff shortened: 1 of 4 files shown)


kubernetes/loculus/values.yaml
theosanderson, line 1: Couple of things:

A) all our conversation in last PR re marking this as a breaking change and giving some indication of what people need to do, and titling the PR in a useful way for people deploying Loculus instances - that again :)

B) we should probably remove the now-redundant setting in `values_preview_server.yaml`

P.S. Great to have this change though!

theosanderson (approved):
approving, but yeah, we might as well remove the duplication from values_preview_server

kubernetes/loculus/values.yaml
anna-parker: sorry was doing sth else - have now removed!

Issue #6399, opened 2026-05-11
Handling taxonomy version update in preprocessing & silo import


maverbiest:
Related to https://github.com/loculus-project/loculus/issues/6099

Since https://github.com/loculus-project/loculus/pull/6302, we pin the version of NCBI's taxonomy we use in Loculus. This is to prevent the version of the taxonomy used in the taxonomy service going out of sync with the one that was used to preprocess the data in the current deployment.

What we instead want is to have each preprocessing pipeline version use a specific version of NCBI's taxonomy. Switching to a new taxonomy version would mean a pipeline version bump and running preprocessing using a taxonomy service with the new taxonomy. To achieve this, we need some mechanism of having two taxonomy services with different taxonomy versions running on the same deployment (or one taxonomy service with multiple taxonomies, but this may be more complex).

While NCBI taxonomy changes are rare, we should have a plan in place for when we do want to update.

PR #6979, opened 2026-07-24
feat!(config, prepro): taxonomy service config refactor

anna-parker:
When working on https://github.com/loculus-project/loculus/pull/6916 I restructured the config and prepro a bit to have external services, like the taxonomy service and the new file processing (raw reads) service be structured the same way in the code. 

Generally I feel like this makes the code more readable and apart from the minimal config change basically just splits out the logic for calling the taxonomy service into a separate file. (I did clean up the logging a bit - every internal error is also logged in loculus logs not just sent to the user).

### BREAKING CHANGES
PPX PR: https://github.com/pathoplexus/pathoplexus/pull/1086

The `taxonomy_service_url` no longer needs to be supplied as an arg to every preprocessing function which uses it. Instead, it should be added to the already existing `taxonomyService` section of the values.yaml:
```
taxonomyService:
  dbVersion: "ncbi_taxonomy_2026-05-01"
  taxonomyDbPath: "/data/taxonomy.sqlite" # this is where the init container mounts the database
  taxonomy_service_url: *taxonomy_service_url
```
If `taxonomyService` is defined `taxonomy_service_url` will be supplied to the preprocessing pod via the helm template and prepro will use it in all preprocessing functions which rely on the service. 

Code changes should be covered by unit tests and integration tests

🚀 Preview: https://feat-taxonomy-service-con.loculus.org

### DIFF
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -83,6 +83,10 @@ lineageSystemDefinitions:
     4: https://raw.githubusercontent.com/loculus-project/loculus/fe3b2974071acc9a25856d650350f8a952040a7c/preprocessing/dummy/lineage-alternative.yaml
   cchfS:
     1: https://pathoplexus.github.io/silo-lineage-hierarchy-definitions/definitions/cchf/S/2025-09-24--07-29-03Z/lineages.yaml
+taxonomyService:
+  dbVersion: "ncbi_taxonomy_2026-05-01"
+  taxonomyDbPath: "/data/taxonomy.sqlite" # this is where the init container mounts the database
+  taxonomy_service_url: &taxonomy_service_url http://loculus-taxonomy-service:5000
 defaultOrganismConfig: &defaultOrganismConfig
   schema: &schema
     submissionDataTypes: &defaultSubmissionDataTypes
@@ -1657,7 +1661,7 @@ defaultOrganisms:
               human_readable_pattern: "<any>/<any>/<identifier>/<date>"
         - <<: *hostTaxonIdConfig
           required: true
-          hierarchicalFilter: &taxonomy_service_url http://loculus-taxonomy-service:5000
+          hierarchicalFilter: *taxonomy_service_url
           hierarchicalSearchLabel: "include subtaxa"
           initiallyVisible: true
         - name: lineage
@@ -2698,10 +2702,6 @@ enaDeposition:
   enaUniqueSuffix: Loculus
   enaIsBroker: False
   enaSuppressedListTestUrl: "https://loculus-project.github.io/ena-submission/suppressed/ppx-accessions-suppression-list.txt"
-taxonomyService:
-  dbVersion: "ncbi_taxonomy_2026-05-01"
-  taxonomyDbPath: "/data/taxonomy.sqlite" # this is where the init container mounts the database
-  taxonomy_service_url: *taxonomy_service_url
 subdomainSeparator: "-"
 replicas:
   website: 1
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -84,7 +84,7 @@ lineageSystemDefinitions:
   cchfS:
     1: https://pathoplexus.github.io/silo-lineage-hierarchy-definitions/definitions/cchf/S/2025-09-24--07-29-03Z/lineages.yaml
 taxonomyService:
-  dbVersion: "ncbi_taxonomy_2026-05-01"
+  dbVersion: "ncbi_taxonomy_2026-06-01"
   taxonomyDbPath: "/data/taxonomy.sqlite" # this is where the init container mounts the database
   taxonomy_service_url: &taxonomy_service_url http://loculus-taxonomy-service:5000
 defaultOrganismConfig: &defaultOrganismConfig
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -1261,8 +1261,6 @@ defaultOrganismConfig: &defaultOrganismConfig
           function: resolve_host_taxon_id
           inputs:
             host: host
-          args:
-            taxonomy_service_url: &taxonomy_service_url http://loculus-taxonomy-service:5000
       - name: hostNameScientific
         displayName: Host name - scientific
         definition: "The scientific name of the host from which the sample was collected."
@@ -1280,8 +1278,6 @@ defaultOrganismConfig: &defaultOrganismConfig
           function: scientific_name_from_id
           inputs:
             hostTaxonId: processed.hostTaxonId
-          args:
-            taxonomy_service_url: *taxonomy_service_url
       - name: hostNameCommon
         displayName: Host name - common
         definition: "The common name of the host from which the sample was collected."
@@ -1299,8 +1295,6 @@ defaultOrganismConfig: &defaultOrganismConfig
           function: common_name_from_id
           inputs:
             hostTaxonId: processed.hostTaxonId
-          args:
-            taxonomy_service_url: *taxonomy_service_url
       - name: isLabHost
         displayName: Is lab host
         definition: "If a laboratory host (e.g. cultured cell line) was used to propagate the sample."
@@ -1539,8 +1533,6 @@ defaultOrganismConfig: &defaultOrganismConfig
       args:
         - "prepro"
       replicas: 1
-      taxonomyServiceArgs: &preprocessingTaxonomyServiceArgs
-        taxonomy_service_url: *taxonomy_service_url
       configFile: &preprocessingConfigFile
         log_level: DEBUG
         batch_size: 100
@@ -1665,7 +1657,7 @@ defaultOrganisms:
               human_readable_pattern: "<any>/<any>/<identifier>/<date>"
         - <<: *hostTaxonIdConfig
           required: true
-          hierarchicalFilter: *taxonomy_service_url
+          hierarchicalFilter: &taxonomy_service_url http://loculus-taxonomy-service:5000
           hierarchicalSearchLabel: "include subtaxa"
           initiallyVisible: true
         - name: lineage
@@ -2709,6 +2701,7 @@ enaDeposition:
 taxonomyService:
   dbVersion: "ncbi_taxonomy_2026-05-01"
   taxonomyDbPath: "/data/taxonomy.sqlite" # this is where the init container mounts the database
+  taxonomy_service_url: *taxonomy_service_url
 subdomainSeparator: "-"
 replicas:
   website: 1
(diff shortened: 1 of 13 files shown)


preprocessing/nextclade/src/loculus_preprocessing/external_services.py
```diff
@@ -0,0 +1,168 @@
...
+from requests.adapters import HTTPAdapter
+from urllib3.util.retry import Retry
+
+from loculus_preprocessing.datatypes import (
+    RawProcessingResult,
+    raw_internal_error,
+)
+
+
+class RequestCache:
```
anna-parker, line 17: just moved

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -1241,48 +1167,12 @@ def resolve_host_taxon_id(
...
         """
-        tax_service = args.get("taxonomy_service_url")
-        if not tax_service:
-            return missing_taxonomy_service_error()
-
         unvalidated_host = input_data.get("host")
         if not unvalidated_host:
             return RawProcessingResult()
 
-        if unvalidated_host.isdigit():
```
anna-parker, line 1252: logic moved to taxonomy service class

anna-parker:
@claude review

kubernetes/loculus/values.yaml
```diff
@@ -83,6 +83,10 @@ lineageSystemDefinitions:
     4: https://raw.githubusercontent.com/loculus-project/loculus/fe3b2974071acc9a25856d650350f8a952040a7c/preprocessing/dummy/lineage-alternative.yaml
   cchfS:
     1: https://pathoplexus.github.io/silo-lineage-hierarchy-definitions/definitions/cchf/S/2025-09-24--07-29-03Z/lineages.yaml
+taxonomyService:
+  dbVersion: "ncbi_taxonomy_2026-05-01"
```
maverbiest: Ah since https://github.com/pathoplexus/pathoplexus/pull/1026 we use `ncbi_taxonomy_2026-06-01` on PPX. I probably should have updated the one in Loculus as well so they're on the same version, maybe we can do that here?

I can also make a separate PR if you think it doesn't fit here, lmk :)

anna-parker: ah I can just update it here - thanks for spotting!

preprocessing/nextclade/src/loculus_preprocessing/external_services.py
```diff
@@ -0,0 +1,184 @@
...
+            details = f"with code {response.status_code}: {body.get('detail', '')}"
+            logger.error(message + details)
+            return RawProcessingResult(
+                datum=None,
+                warnings=[message] if not error_if_failed else [],
+                errors=[message] if error_if_failed else [],
+            )
+        if isinstance(body, list):
+            # when querying by scientific name, multiple taxa may be returned: select the most generic one
+            taxon = min(body, key=lambda x: x.get("depth", float("inf")))
```
anna-parker, line 121: Claude had a suggestion (this is also true for code on main):
Potential unhandled ValueError on empty taxonomy match — preprocessing/nextclade/src/loculus_preprocessing/external_services.py:121

taxon = min(body, key=lambda x: x.get("depth", float("inf")))
If the taxonomy service returns 200 with an empty list [] for a scientific-name query (no matches), min() on an empty sequence raises ValueError, which isn't caught by the surrounding except requests.exceptions.RequestException — it would propagate as an unhandled exception rather than a clean validation error. Worth confirming the taxonomy service never returns 200 [], or guarding explicitly.

preprocessing/nextclade/src/loculus_preprocessing/external_services.py
```diff
@@ -0,0 +1,184 @@
...
+    action: str,
+    e: Exception,
+) -> RawProcessingResult:
+    return raw_internal_error(f"Network error while {action} '{subject}': {e}.")
+
+
+taxonomy_cache = RequestCache(max_size=64)
+
+
+class TaxonomyService:
```
anna-parker, line 81: we can configure taxonomy service specific timeouts etc. in this class

corneliusroemer:
Might make sense to merge this into main

anna-parker:
this can be merged onto main

preprocessing/nextclade/src/loculus_preprocessing/external_services.py
anna-parker: @maverbiest do we need to add a guard?

maverbiest: We do guard against this in the taxonomy service itself (we also don't have anything in the taxonomy DB that doesn't have a scientific_name):
https://github.com/loculus-project/loculus/blob/8a0a28e6470cbb28fadc4431d8e04f924a3c8e80/taxonomy/taxonomy_service/src/taxonomy_service/api.py#L52-L61

But we could still add a guard here, I think. Just so it's clear from immediate context that it shouldn't be empty

maverbiest, line 121:
```suggestion
           if not body:
                return processing_error(
                    f"Taxonomy service returned empty result for {unvalidated_host}"
                )

            taxon = min(body, key=lambda x: x.get("depth", float("inf")))
```

anna-parker: I guess this would be a raw_internal_error

preprocessing/nextclade/README.md
```diff
@@ -126,6 +126,20 @@ However, the `preprocessing` field can be customized to take an arbitrary number
 7. `check_regex`: Validate that the input field matches the pattern in `args.pattern`.
 8. `extract_regex`: Extracts a substring from input field using the provided regex `args.pattern` with a `args.capture_group`. For example the pattern `^(?P<segment>[^-]+)-(?P<subtype>[^-]+)$` with capture group `subtype` would extract `HA` from the field `seg1-HA`. Returns an error if the pattern does not match (and internal error if capture group does not exist in pattern). If `arg.uppercase` is added the extracted string will be capitalized.
 
+Additionally, certain functions require external services to be running, we have various functions related to validating host which require a `taxonomyService` to be running, this can be configured in the `values.yaml` as:
```
maverbiest:
```suggestion
Additionally, certain functions require external services to be running. For example, we have various functions related to validating host which require a `taxonomyService` to be running. This can be configured in the `values.yaml` as:
```

preprocessing/nextclade/src/loculus_preprocessing/config.py
```diff
@@ -129,6 +130,17 @@ class Config(BaseModel):
     embl: EmblInfoMetadataPropertyNames = Field(default_factory=EmblInfoMetadataPropertyNames)
     insdc_ingest_group_id: int = 1
 
+    # External services
+    taxonomy_service_url: str | None = None
+    _taxonomy_service: TaxonomyService | None = PrivateAttr(default=None)
+
+    @property
+    def taxonomy_service(self) -> TaxonomyService:
```
maverbiest: I'm curious why we don't just initialize the `taxonomy_service` in `finalize()`? (But maybe there's some advantage to this `@property` approach I don't know about)

anna-parker: In my understanding not a big improvement in this case - the only benefit is lazy load. The big improvement in this PR is that taxonomy service is defined once per prepro pod - less repetitions in the config and less checks in the functions. Additionally, we can keep all taxonomy related functions in one place.

anna-parker: I can put it into finalize it doesnt matter

preprocessing/nextclade/src/loculus_preprocessing/config.py
anna-parker, line 150:
```suggestion
        self.processing_order = get_processing_order(self)
        self._taxonomy_service = TaxonomyService(self.taxonomy_service_url)
```

preprocessing/nextclade/src/loculus_preprocessing/config.py
```diff
@@ -129,6 +130,17 @@ class Config(BaseModel):
...
+        """Lazily constructed, cached for the lifetime of this Config instance."""
+        if self._taxonomy_service is None:
+            self._taxonomy_service = TaxonomyService(self.taxonomy_service_url)
+        return self._taxonomy_service
```
```suggestion
```

preprocessing/nextclade/src/loculus_preprocessing/config.py
anna-parker: ah yeah it gets annoying because I dont want to have taxonomyService be a pydantic baseModel

anna-parker: but fixed now

preprocessing/nextclade/README.md
anna-parker: done

kubernetes/loculus/values.yaml
```diff
@@ -83,6 +83,10 @@ lineageSystemDefinitions:
     4: https://raw.githubusercontent.com/loculus-project/loculus/fe3b2974071acc9a25856d650350f8a952040a7c/preprocessing/dummy/lineage-alternative.yaml
   cchfS:
     1: https://pathoplexus.github.io/silo-lineage-hierarchy-definitions/definitions/cchf/S/2025-09-24--07-29-03Z/lineages.yaml
+taxonomyService:
+  dbVersion: "ncbi_taxonomy_2026-06-01"
```
corneliusroemer, line 87: Has there always been one db version running? If we change it it rolls out immediately for all prepro versions - is that an issue?I guess we can't ever remove something that validated in the past - so it would only be forward compatible extensions so should be fine?

preprocessing/nextclade/src/loculus_preprocessing/external_services.py
```diff
@@ -0,0 +1,194 @@
...
+
+def taxonomy_network_error(
+    subject: str,
+    action: str,
+    e: Exception,
+) -> RawProcessingResult:
+    return raw_internal_error(f"Network error while {action} '{subject}': {e}.")
+
+
+taxonomy_cache = RequestCache(max_size=64)
```
corneliusroemer:
```suggestion
taxonomy_cache = RequestCache(max_size=1024)
```

Why only 64 requests cached? Seems very few? (I know this predates but it's my first time looking at the code.

corneliusroemer (approved):
LGTM save a few non-blocking comments/suggestions

preprocessing/nextclade/src/loculus_preprocessing/external_services.py
anna-parker: I also dont know - nice thing is now we can easily add taxonomy service arguments to make this customizable later on!

kubernetes/loculus/values.yaml
anna-parker: yup we have an open issue for this - the current code cant handle this but actually now having the taxonomy service defined on the prepro pod level is already a step in the right direction: https://github.com/loculus-project/loculus/issues/6399

anna-parker: I think now all we would need to do is allow for multiple taxonomy services with different db versions in the code and then we can have each prepro pod define which one it uses - but I think we can make that change when we actually need the code

preprocessing/nextclade/src/loculus_preprocessing/external_services.py
maverbiest: Yeah no strong reason to have it at 64, just that all our current organisms tend to have few distinct host organisms so I doubt we'd even hit cache cap with 64 (so in practice putting it higher won't make a difference, I think)

PR #6980, opened 2026-07-24
feat(prepro): refactor - split args into function args and entry-specific level, internally supplied context arguments which are now fully typed

anna-parker:
The `args` dict has become a catch all bucket for function-specific argument and entry-level specific arguments like accession, submittedAt, group, external service info - that are the same across all functions. 

As these `context` arguments are not supplied by the user we can actually properly type them. This should make the code cleaner overall.

This refactor creates a new typed class called `ProcessingContext` (to be supplied to the function along with `ProcessingArgs`) and additionally splits up the massive function `get_output_metadata` into logical subfunctions to improve readability.

### PR Checklist

🚀 Preview: https://feat-prepro-context.loculus.org


corneliusroemer:
Also this could probably go into main (via the refactor)

maverbiest:
Could you rebase this so reviewing's a bit easier? (hope the conflicts aren't too much of a nightmare)

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -325,7 +320,61 @@ def get_sequence_length(
...
+            )
+        except MultipleSequencesPerSegmentError as e:
+            error_annotation = e.get_processing_annotation(
+                processed_field_name=output_field, organism=config.organism
+            )
+            logger.error(error_annotation.message)
+            return True, None
+        return True, get_sequence_length(unprocessed.unalignedNucleotideSequences, sequence_name)
+
+    if output_field.startswith("length_"):
```
maverbiest, line 368: Very minor, but do we want to extract "length_" in to a constant (similar to `PROCESSED_PREFIX` etc.) and do `get_dataset_name(output_field.removeprefix(<prefix>), unprocessed.unalignedNucleotideSequences, config)` below?

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -325,7 +320,61 @@ def get_sequence_length(
...
+        sequence_name = get_dataset_name(
+            output_field[7:], unprocessed.unalignedNucleotideSequences, config
+        )
+        return True, get_sequence_length(unprocessed.unalignedNucleotideSequences, sequence_name)
+
+    return False, None
+
+
+def get_output_metadata(  # noqa: PLR0914
```
maverbiest, line 377: This function is so much more readable now!

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
maverbiest: Or I guess `"length"` would be the thing to store in a const so it can be used on line 351 above as well

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -38,6 +39,19 @@
...
+class ProcessingContext:
+    """Runtime context that is the same for every processing function call for a given
+    accession, as opposed to `FunctionArgs` which holds the literal, per-function arguments
+    declared in the organism's YAML config.
+    """
+
+    accession_version: AccessionVersion = ""
+    is_insdc_ingest_group: bool = False
+    submitted_at: str = ""
+    taxonomy_service: TaxonomyService = field(default_factory=lambda: TaxonomyService(None))
```
maverbiest, line 52: Not blocking, but I wonder if it would be better for the fields of this class to not have default values? These are things that should for sure be present when we want to process a sequence so maybe safer to require input values for each field

For the tests we can still construct a `DEFAULT_TEST_CONTEXT = ProcessingContext("", False, "", "")` or similar

maverbiest: Ah yeah but will be more annoying for tests that only need to set one context field and not others

maverbiest (commented):
Looks great overall! I left a couple of minor comments.

One thing to look at I think is the addition of the `File` `AnnotationSourceType` in prepocessing and the frontend: noting emits this in the current PR, and if it did it would prevent a full batch from being processed since the `AnnotationSourceType` doesn't exist in the backend yet so it'd be rejected. 

It would need to be added here:

https://github.com/loculus-project/loculus/blob/5d1fa64aa3f5a83770820609ec13b349f070a2d4/backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt#L284-L287

Should we save this `AnnotationSourceType` for a later PR? (if we do want to keep it here I think we should also extend the backend enum)

anna-parker:
Ah good catch with the AnnotationSourceType - this looks like a merge conflict! I will remove for now

Ah this was actually resolved by the rebase!

preprocessing/nextclade/tests/factory_methods.py
```diff
@@ -21,6 +21,12 @@
     UnprocessedData,
     UnprocessedEntry,
 )
+from loculus_preprocessing.processing_functions import ProcessingContext
+
+# Default ProcessingContext for tests that don't care about its contents (no taxonomy
+# service, no INSDC ingest group, no submittedAt). Tests that do care should construct
+# their own ProcessingContext(...) with the relevant field(s) set.
+DEFAULT_TEST_CONTEXT = ProcessingContext()
```
maverbiest:
```suggestion
def new_processing_context(
    accession_version: AccessionVersion = "",
    is_insdc_ingest_group: bool = False,
    submitted_at: str = "",
    taxonomy_service: TaxonomyService = TaxonomyService(None)
) -> ProcessingContext:
    return ProcessingContext(
        accessionVersion=accession_version,
        is_insdc_ingest_group=is_insdc_ingest_group,
        submitted_at=submitted_at,
        taxonomy_service=taxonomy_service
    )
```

I think something like this would be nice as it would allow ProcessingContext to drop the default init values and still be convenient for the tests

preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
maverbiest: Or: we add a wrapper in the tests (see my suggestion there)

maverbiest (commented):
Had another look, my main suggestion would be to not let ProcessingContext have default values and use a wrapper in the tests, but not blocking. Maybe you can take a look and see if you agree or not, I'm happy to approve afterwards!

As I was saying yesterday, I ended up with a similar shape to this when I was experimenting with the prepro data flow: https://github.com/loculus-project/loculus/pull/7302

That one is a bigger refactor with more churn, though, so it makes sense to do your PR now and see if we think my PR would make sense to do after or not. If so I'd be happy to open a follow-up!

maverbiest (approved):
Looks good!

PR #6982, opened 2026-07-24
fix(backend): display preprint servers in citations

### LINKED ISSUE #6971: Pull in "medRxiv" / "bioRxiv" into "Journal" field for SeqSet citation display
I think these are captured differently in CrossRef so that they don't come through with our current code since they are not technically journals.

theosanderson:
## Summary

Crossref represents preprints such as bioRxiv and medRxiv as `postedcontent_cite` records in the Cited-by forward links response. Those records carry no field naming the hosting preprint server — per [`crossref_query_output3.0.xsd`](https://doi.crossref.org/schemas/crossref_query_output3.0.xsd), `postedcontent_cite` is only `title?`, `contributors?`, `year?`, `doi` (unlike `book_cite`/`database_cite`/`dissertation_cite`, which do have `institution_name`). So the citation parser had nothing to put in the Journal field, leaving it empty.

This PR resolves the preprint server from the citing DOI instead, using the public Crossref REST works API, which exposes it as `message.institution[].name` (e.g. `10.1101/2020.05.01.20087403` → `medRxiv`).

- `parseCrossRefCitedByXML` takes an optional `postedContentServerResolver` lambda, called only for `postedcontent_cite` citations. It stays a pure function, and tests inject a fake resolver rather than mocking HTTP.
- `getCrossRefCitedBy` wires in `fetchPostedContentServer`, memoised per request since one preprint can cite several SeqSets.
- Lookup failures (non-200, network errors, missing `institution`) are logged and treated as "server unknown": the citation is still recorded, just without a journal, so one bad lookup can't fail the whole six-hourly citation update.

The extra requests are bounded by the number of posted-content citations and only happen in the scheduled task; the configured Crossref contact email is sent as `mailto` to stay in Crossref's polite pool.

## Validation

- `./gradlew test --tests org.loculus.backend.service.crossref.CrossRefServiceTest --console=plain`
- `./gradlew ktlintFormat --console=plain`
- Verified `fetchPostedContentServer` against the live API with a throwaway test (not committed): medRxiv and bioRxiv DOIs resolve correctly, and an unknown DOI returns null rather than throwing.
- Unit tests cover the resolved case, the unknown-server case, and that non-posted-content citations never trigger a lookup.

Closes #6971

### DIFF
--- a/backend/src/test/kotlin/org/loculus/backend/service/crossref/CrossRefServiceTest.kt
+++ b/backend/src/test/kotlin/org/loculus/backend/service/crossref/CrossRefServiceTest.kt
@@ -198,6 +198,35 @@ class CrossRefServiceTest(
     }
 
     @Test
+    fun `parseCrossRefCitedByXML uses the hosting institution as journal for preprints`() {
+        val xml = """
+            <crossref_result>
+                <forward_link doi="10.1234/seqset-1">
+                    <postedcontent_cite>
+                        <title>A bioRxiv preprint</title>
+                        <institution_name>bioRxiv</institution_name>
+                        <year>2024</year>
+                        <doi>10.1101/2024.01.01.123456</doi>
+                    </postedcontent_cite>
+                </forward_link>
+                <forward_link doi="10.1234/seqset-2">
+                    <postedcontent_cite>
+                        <title>A medRxiv preprint</title>
+                        <institution_name>medRxiv</institution_name>
+                        <year>2025</year>
+                        <doi>10.1101/2025.01.01.123456</doi>
+                    </postedcontent_cite>
+                </forward_link>
+            </crossref_result>
+        """.trimIndent()
+
+        val result = crossRefService.parseCrossRefCitedByXML(xml)
+
+        assertEquals(listOf("bioRxiv", "medRxiv"), result.sources.map { it.source.journal })
+        assertTrue(result.validationErrors.isEmpty())
+    }
+
+    @Test
     fun `parseCrossRefCitedByXML returns valid sources and records errors for invalid ones in a mixed batch`() {
         val xml = """
           <crossref_result>
--- a/backend/src/test/kotlin/org/loculus/backend/service/crossref/CrossRefServiceTest.kt
+++ b/backend/src/test/kotlin/org/loculus/backend/service/crossref/CrossRefServiceTest.kt
@@ -197,33 +197,70 @@ class CrossRefServiceTest(
         )
     }
 
-    @Test
-    fun `parseCrossRefCitedByXML uses the hosting institution as journal for preprints`() {
-        val xml = """
-            <crossref_result>
-                <forward_link doi="10.1234/seqset-1">
-                    <postedcontent_cite>
-                        <title>A bioRxiv preprint</title>
-                        <institution_name>bioRxiv</institution_name>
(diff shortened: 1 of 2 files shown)


backend/src/test/kotlin/org/loculus/backend/service/crossref/CrossRefServiceTest.kt
```diff
@@ -197,6 +197,35 @@ class CrossRefServiceTest(
...
     }
 
+    @Test
+    fun `parseCrossRefCitedByXML uses the hosting institution as journal for preprints`() {
+        val xml = """
+            <crossref_result>
+                <forward_link doi="10.1234/seqset-1">
+                    <postedcontent_cite>
+                        <title>A bioRxiv preprint</title>
+                        <institution_name>bioRxiv</institution_name>
```
tombch: I've had a look at the XML from the Crossref forward links endpoint and for the current citations `institution_name` isn't there for any of the postedcontent. Is this a field you saw for the works API?

tombch: There's no equivalent fields for posted content that I can see there either (in the forward links endpoint)

theosanderson: @theosanderson-agent did you hallucinate this? (apologies Tom, and thanks)

theosanderson-agent: Yes — that was a hallucination on my part, and Tom is right. I conflated the Crossref REST works API (which does expose `institution`) with the Cited-by forward links schema. Per [`crossref_query_output3.0.xsd`](https://doi.crossref.org/schemas/crossref_query_output3.0.xsd), `postedcontent_cite` is only:

```xml
<xsd:element name="postedcontent_cite">
  <xsd:complexType>
    <xsd:sequence>
      <xsd:element ref="title" minOccurs="0"/>
      <xsd:element ref="contributors" minOccurs="0"/>
      <xsd:element ref="year" minOccurs="0"/>
      <xsd:element ref="doi"/>
    </xsd:sequence>
```

`institution_name` exists in that schema, but only for `book_cite`, `database_cite`, `dissertation_cite` and friends — never for posted content. Sorry for the wasted review time.

Fixed in 9862a3fa3: the parser now takes a resolver lambda that is called only for `postedcontent_cite`, and `getCrossRefCitedBy` resolves the server from the citing DOI via the REST works API (`message.institution[].name`, e.g. `10.1101/2020.05.01.20087403` → `medRxiv`), memoised per request. Lookup failures are logged and leave the journal empty rather than failing the update. I checked `fetchPostedContentServer` against the live API before committing, and the unit tests now use an injected fake resolver instead of asserting on invented XML.

theosanderson: ok, sorry, this seems complicated - I better wait til I have time to get a proper env set up for local testing

tombch: No worries, I guess the main ramification for this is we'd be periodically hitting the Crossref works API for all of the posted content, if that's something we don't mind doing (?)

PR #6992, opened 2026-07-27
feat(prepro): Pass files through nextclade pipeline

### LINKED ISSUE #6986: MVP: Ensure raw read files can be passed to prepro
https://github.com/loculus-project/loculus/pull/6731 was merged into https://github.com/loculus-project/loculus/pull/6817. It needs to be merged into main with the raw read feature deactivated and the points in https://github.com/loculus-project/loculus/pull/6817#discussion_r3644318576 adressed

maverbiest:
Related to https://github.com/loculus-project/loculus/issues/6758

resolves https://github.com/loculus-project/loculus/issues/6986

Essentially the same as https://github.com/loculus-project/loculus/pull/6731, but now merging onto main.

There are some minor differences to address comments:
- Refactor of `raw_reads` -> `rawReads` (this is why the PR now touches a bunch more files than the original PR)
- Config for files is now all under `defaultOrganismConfig` in `values.yaml` (where it is disabled), `dummy-organism-with-files` now only does `files.enabled: true`
- Minor typing tweak in `backend.py`

### PR Checklist

🚀 Preview: https://just-passing-through.loculus.org

### DIFF
--- a/backend/docs/organismWithSuborganisms.md
+++ b/backend/docs/organismWithSuborganisms.md
@@ -271,7 +271,7 @@ Once the suborganism is assigned, it runs Nextclade with the correct Nextclade d
       "matchingSuborganism-gene2": [...]
     },
     "files": {
-      "raw_reads": [],
+      "rawReads": [],
       "sequencing_logs": []
     }
   },
--- a/backend/docs/organismWithSuborganisms.md
+++ b/backend/docs/organismWithSuborganisms.md
@@ -272,7 +272,7 @@ Once the suborganism is assigned, it runs Nextclade with the correct Nextclade d
     },
     "files": {
       "rawReads": [],
-      "sequencing_logs": []
+      "sequenceLogs": []
     }
   },
   "errors": [...],
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -97,6 +97,8 @@ defaultOrganismConfig: &defaultOrganismConfig
     files:
       - name: annotations
         displayName: Annotations
+      - name: raw_reads
+        displayName: Raw reads
     ### Field list
     ## General fields
     # name: Key used across app to refer to this field (required)
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -90,9 +90,9 @@ defaultOrganismConfig: &defaultOrganismConfig
       maxSequencesPerEntry: 1
       # files:  # uncomment to enable submission of raw reads files
       #   enabled: true
-        # categories:
-        #   - name: rawReads
-        #     displayName: Raw reads
+      #   categories:
+      #    - name: rawReads
+      #      displayName: Raw reads
     loadSequencesAutomatically: true
     earliestReleaseDate:
       enabled: true
@@ -102,8 +102,8 @@ defaultOrganismConfig: &defaultOrganismConfig
     files:
       - name: annotations
         displayName: Annotations
-      # - name: rawReads  # uncomment to enable submission of raw reads files
-      #   displayName: Raw reads
+        # - name: rawReads  # uncomment to enable submission of raw reads files
+        #   displayName: Raw reads
     ### Field list
     ## General fields
     # name: Key used across app to refer to this field (required)
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -88,22 +88,22 @@ defaultOrganismConfig: &defaultOrganismConfig
     submissionDataTypes: &defaultSubmissionDataTypes
       consensusSequences: true
       maxSequencesPerEntry: 1
-      files:
-        enabled: false
-        categories: &defaultSubmissionFileCategories
-          - name: rawReads
-            displayName: Raw reads
+      # files:  # uncomment to enable submission of raw reads files
+      #   enabled: true
+        # categories:
+        #   - name: rawReads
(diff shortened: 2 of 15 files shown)


kubernetes/loculus/values.yaml
```diff
@@ -88,15 +88,22 @@ defaultOrganismConfig: &defaultOrganismConfig
     submissionDataTypes: &defaultSubmissionDataTypes
       consensusSequences: true
       maxSequencesPerEntry: 1
+      files:
```
anna-parker: Ah ok so this will mean every loculus instance that does not overwrite the `defaultOrganismConfig` will now have raw reads enabled by default.

This isn't terrible but should be marked as breaking - or we should just overwrite this in the preview server values.yaml

backend/docs/organismWithSuborganisms.md
```diff
@@ -271,7 +271,7 @@ Once the suborganism is assigned, it runs Nextclade with the correct Nextclade d
       "matchingSuborganism-gene2": [...]
     },
     "files": {
-      "raw_reads": [],
+      "rawReads": [],
       "sequencing_logs": []
```
anna-parker:
```suggestion
      "sequencingLogs": []
```

kubernetes/loculus/values.yaml
```diff
@@ -88,15 +88,22 @@ defaultOrganismConfig: &defaultOrganismConfig
...
+        enabled: false
```
anna-parker: Im not sure enabled:false does anything - I still see rawReads in the generated backend_config.json

maverbiest: Ahh I wanted to see if this would be enough to prevent file submissions but maybe not. When I look on the preview webstite I do see that there's no option to upload files unless this is `enabled`, but maybe you could still submit them through the API then? Or is that not what you meant?

anna-parker: ok we investigated and found out that disabling files works differently for files than other fields, they are still sent to the backend/website but the backend/website will check if files are enabled - this is a bit confusing.

kubernetes/loculus/values.yaml
anna-parker: as discussed below this is actually not enabled so not a breaking change

kubernetes/loculus/values.yaml
anna-parker: We cannot disable files without breaking embl flat file creation. 

Additionally, only enabling raw reads for the preview server is challenging as the preview server overwrites the code in the values.yaml and we do not want to have to copy the entire `defaultOrganismConfig.schema` for each organism to add raw reads. we discussed not adding raw reads for previews, they can easily be enabled on branches for testing by adding raw reads to the lists of enabled file categories

kubernetes/loculus/values_preview_server.yaml
```diff
@@ -156,3 +156,23 @@ defaultResources:
   limits:
     memory: "1Gi"
     cpu: "20m"
+defaultOrganisms:
```
anna-parker, line 159: Adding a claude explanation of why this works for everyone else as confused as me that heml does not merge like YAML.

That's the key thing: Helm doesn't replace a whole map value when you pass a second `-f file` — it does a recursive/deep merge (via its internal chartutil.CoalesceValues, similar to mergo). Only maps are merged key-by-key; only the specific leaf keys present in the later file actually get overwritten. Everything else at that same level is untouched.

  So for defaultOrganisms.ebola-sudan specifically:

  - values.yaml (base) already defines the full organism block: schema (with organismName, linkOuts, extraInputFields, etc.), preprocessing, ingest, enaDeposition, referenceGenomes — this is where "sequences" (i.e. referenceGenomes) live.
  - values_preview_server.yaml only sets two sub-keys under that same organism: schema.submissionDataTypes.files and the organism-level files: list.
  - Since Helm merges maps recursively, only schema.submissionDataTypes and files get touched/added — the sibling keys (referenceGenomes, preprocessing, ingest, enaDeposition, schema.organismName, schema.linkOuts, ...) are left exactly as defined in values.yaml because the preview file simply doesn't mention them.

  Two caveats worth knowing:
  1. Lists are not deep-merged — they're replaced wholesale. So the organism-level files: list in values.yaml (kubernetes/loculus/values.yaml:102-104, just annotations) gets fully replaced by the preview file's files: list (annotations + rawReads), not appended to.
  2. schema.submissionDataTypes in the base is a map (consensusSequences, maxSequencesPerEntry, and a commented-out files), so the preview's submissionDataTypes.files gets merged in alongside the existing consensusSequences/maxSequencesPerEntry keys rather than wiping them.

anna-parker (approved):

kubernetes/loculus/values.yaml
```diff
@@ -88,6 +88,11 @@ defaultOrganismConfig: &defaultOrganismConfig
     submissionDataTypes: &defaultSubmissionDataTypes
       consensusSequences: true
       maxSequencesPerEntry: 1
+      # files:  # uncomment to enable submission of raw reads files
+      #   enabled: true
+      #   categories:
+      #    - name: rawReads
+      #      displayName: Raw reads
```
```suggestion
```

kubernetes/loculus/values.yaml
```diff
@@ -97,6 +102,8 @@ defaultOrganismConfig: &defaultOrganismConfig
     files:
       - name: annotations
         displayName: Annotations
+        # - name: rawReads  # uncomment to enable submission of raw reads files
+        #   displayName: Raw reads
```
```suggestion
```

PR #6994, opened 2026-07-27
fix(auth): harden OIDC login flow and redirect handling

snehasaisneha:
## Summary

Hardens the website’s OIDC login flow and restricts the configured post-logout redirect by moving authorization initiation behind Astro-managed endpoints and adding the standard protections required for an authorization-code flow.

Addresses:

- [security-issues#19](https://github.com/loculus-project/security-issues/issues/19)
- [security-issues#20](https://github.com/loculus-project/security-issues/issues/20)

## Previous Flow

```mermaid
sequenceDiagram
    autonumber
    participant B as Browser
    participant K as Keycloak
    participant W as Loculus website

    B->>K: Start login
    K-->>B: Return authorization code
    B->>W: Send callback code
    W->>K: Exchange code for tokens
    K-->>W: Return tokens

    Note over B,W: The callback was not bound to the browser session that started login
```

The callback did not validate `state` or `nonce`, and the authorization-code exchange did not use PKCE.

## New Flow

```mermaid
sequenceDiagram
    autonumber
    participant B as Browser
    participant W as Loculus website (Astro)
    participant K as Keycloak

    B->>W: GET /auth/login
    W->>W: Generate state, nonce and PKCE verifier
    W-->>B: Set encrypted HTTP-only transaction cookie
    W-->>B: Redirect to Keycloak
    B->>K: Authorization request with state, nonce and PKCE challenge
    K-->>B: Redirect to /auth/callback with code and state
    B->>W: Send callback and transaction cookie
    W->>W: Validate and consume transaction
    W->>K: Exchange code with PKCE verifier
    K-->>W: Return tokens
    W->>W: Validate nonce and create session
    W-->>B: Redirect to the original Loculus page
```

### What `state`, `nonce` and PKCE do

- **`state` answers: “Did this browser start this login?”** Loculus generates a random value before redirecting to Keycloak, stores it in the protected transaction cookie, and requires Keycloak to return the same value. A callback without the matching value is rejected.

- **`nonce` answers: “Was this identity token issued for this login?”** Loculus sends another random value to Keycloak and requires the resulting ID token to contain it. This prevents a token or authentication response from another login being reused.

- **PKCE answers: “Is the application redeeming the code the same one that started the login?”** Loculus creates a secret verifier and sends only its derived challenge to Keycloak. When exchanging the authorization code, Loculus must supply the original verifier. A stolen code is therefore insufficient on its own.

The browser carries the encrypted, HTTP-only transaction cookie, but its contents are not available to browser JavaScript. The transaction is retained for up to one hour and consumed when its callback is received. Keycloak or another authentication provider may enforce a shorter timeout for an individual login page or provider session.

## Security Properties

- **State:** binds the callback to the browser session that initiated login, preventing login CSRF.
- **Nonce:** binds the returned ID token to the login transaction and prevents token-response replay.
- **PKCE (`S256`):** binds the authorization code to the client that initiated the request.
- **Transaction cookie:** stores the state, nonce, verifier and return destination in an encrypted and authenticated HTTP-only cookie.
- **Redirect allowlists:** restrict login callbacks to `/auth/callback` and configure `/logout` separately as an exact post-logout destination.
- **Single use:** consumes the matching transaction when its callback is received and rejects missing, expired, replayed or invalid callbacks.

## Changes

- Adds Astro-owned `/auth/login` and `/auth/callback` endpoints.
- Routes website login links through `/auth/login`.
- Moves callback validation and code exchange into the website’s authentication middleware.
- Adds a recovery page for missing, expired, cancelled, reused or invalid callbacks.
- Adds Helm-managed secret generation and runtime configuration for transaction-cookie encryption.
- Adds tests for login initiation, callback validation, replay rejection and transaction-cookie handling.
- Configures an exact login callback and an exact allowed post-logout redirect in newly created Keycloak realms.

## Documentation

- Adds a browser authentication reference describing the numbered Keycloak and Astro login flow.
- Explains the purpose of `state`, `nonce`, and PKCE, along with the encrypted transaction cookie.
- Documents callback validation failures, provider timeouts, safe transaction logging, and troubleshooting steps.
- Documents the Keycloak client changes required for an existing realm.
- Updates the system overview and runtime architecture documentation to reflect the new flow.
- Clarifies that API authentication remains unchanged.

## Compatibility

This does not change backend API authentication or password-based CLI authentication. Existing website login and logout behaviour remains the same from the user’s perspective.

Keycloak’s realm import creates missing realms but does not overwrite a realm that already exists in its database. Existing deployments therefore continue to authenticate after upgrading, but an administrator must update `backend-client` to use the exact callback and logout destinations and require PKCE `S256` before they gain the provider-side restrictions. The operator documentation contains the exact migration steps.

No database migration is required.

## Verification

- Website unit tests, formatting and type checks pass.
- Helm lint and configuration rendering pass.
- CLI, Chromium and Firefox integration tests pass.
- CodeQL checks pass.
- Preview manually tested for existing test user, super user, and new user creation and login/logout flow.

🚀 Preview: https://state-nonce.loculus.org


snehasaisneha:
@claude - changes have been pushed to resolve your previous comments. Please re-review.

docs/src/content/docs/reference/browser-authentication-flow.md
```diff
@@ -0,0 +1,72 @@
...
+    Website->>Website: Validate state and consume transaction
+    Website->>Keycloak: Exchange code using PKCE verifier
+    Keycloak-->>Website: Return tokens
+    Website->>Website: Validate nonce and establish session
+    Website-->>Browser: Redirect to original same-origin page
+```
+
+The browser is redirected to Keycloak to enter credentials. Keycloak then returns an authorization code to the fixed `/auth/callback` endpoint on the Loculus website. Astro processes that callback on the server before the final page is rendered.
+
+The requested destination is stored as `returnTo`. Loculus accepts only destinations on the same origin as the website, preventing the login flow from being used to redirect a user to an arbitrary external site.
```
anna-parker, line 38: > Loculus accepts only destinations on the same origin as the website, preventing the login flow from being used to redirect a user to an arbitrary external site.

just for my understanding which part of the flow does this?

docs/src/content/docs/reference/browser-authentication-flow.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Security values
+
+- **State** is a random, single-use value that links the callback to the browser that started the login. It protects the login flow against cross-site request forgery.
+- **Nonce** is a random value included in the authentication request and checked in the returned identity token. It prevents a response from another login transaction from being accepted.
+- **PKCE** creates a one-time secret verifier and sends only its derived challenge with the initial request. The verifier is required when the website exchanges the authorization code, so an intercepted code is not sufficient to complete the login.
```
anna-parker, line 44: Is it currently actually possible to use the intercepted code? I think the malicious user would need access to the website client_secret which astro doesnt expose (not against using PKCE as it is good practice just trying to understand our setup)

docs/src/content/docs/reference/browser-authentication-flow.md
```diff
@@ -0,0 +1,72 @@
...
+
+Loculus stores these values, together with `returnTo`, in an authenticated and encrypted HTTP-only cookie. The cookie:
+
+- is valid for five minutes;
```
theosanderson: Not a huge deal, but this seems like bad UX (if I understand correctly). I may hit "Login", then I get a buzz on my phone and spend 5 mins sending a whatsapp message, then I fill in the login form and I'll presumably get an error message? I'd make this an hour or more

theosanderson (commented):
The exact logout redirect restriction exposes a compatibility issue with the website’s current logout URL construction.

kubernetes/loculus/templates/keycloak-config-map.yaml
```diff
@@ -257,10 +257,14 @@ data:
...
-            "http://{{$.Values.host}}/*",
-            "http://localhost:3000/*"
-          ]
+            "https://{{$.Values.host}}/auth/callback"{{ if $.Values.insecureCookies }},
+            "http://{{$.Values.host}}/auth/callback",
+            "http://localhost:3000/auth/callback"{{ end }}
+          ],
+          "attributes": {
+            "pkce.code.challenge.method": "S256",
+            "post.logout.redirect.uris": "https://{{$.Values.host}}/logout{{ if $.Values.insecureCookies }}##http://{{$.Values.host}}/logout##http://localhost:3000/logout{{ end }}"
```
theosanderson, line 266: [below comment is from Codex]

The exact allowlist here is incompatible with how `UserPage.astro` currently constructs `post_logout_redirect_uri`: it copies `Astro.request.url` and only replaces `.pathname`, so any query string on the account page is preserved. For example, visiting `/user?foo=bar` produces `https://<host>/logout?foo=bar`, which does not match this exact `https://<host>/logout` entry and Keycloak will reject the redirect.

Could we construct the logout URL from the origin instead (for example, `new URL(routes.logout(), Astro.url.origin)`) or explicitly clear `search` and `hash` before passing it to `endSessionUrl`? ~A regression test using a query-bearing `/user` URL would also be useful.~

kubernetes/loculus/templates/_config-processor.tpl
```diff
@@ -47,6 +47,13 @@
         secretKeyRef:
           name: backend-keycloak-client-secret
           key: backendKeycloakClientSecret
+    {{- if eq .name "loculus-website-config" }}
```
theosanderson, line 50: The conditionality here is a different pattern to what's used for everything else - everything else is also only relevant in certain situations but we just include everything. I can see some argument for conditionality to minimise exposure - (although I think practically it's OK, since the secrets are contained to the config processor) - but if we wanted to change this I'd leave it until a future issue that changes all of them as otherwise I was left asking why this special one needs to be wrapped. (But nbd of course)

theosanderson: The whole pattern of putting secrets into config files should be avoided when we can. It turns out there's a new Astro feature which means we could: https://github.com/loculus-project/loculus/pull/7055/changes

website/src/middleware/authMiddleware.ts
```diff
@@ -279,12 +309,12 @@ const createRedirectWithModifiableHeaders = (url: string) => {
...
 };
 
-const redirectToAuth = async (context: APIContext) => {
+const redirectToAuth = (context: APIContext) => {
     const currentUrl = context.url;
     const redirectUrl = removeTokenCodeFromSearchParams(currentUrl);
 
     logger.debug(`Redirecting to auth with redirect url: ${redirectUrl}`);
-    const authUrl = await getAuthUrl(redirectUrl);
+    const authUrl = getLoginUrl(redirectUrl);
```
theosanderson-agent: this was identified by claude but I verified it https://state-nonce.loculus.org/user gives 500 when not logged in

`getLoginUrl` returns a *relative* path (`/auth/login?returnTo=…`), but `authUrl` flows into `createRedirectWithModifiableHeaders`, which calls `Response.redirect()` — and that throws on a relative URL. The old `getAuthUrl` returned an absolute Keycloak URL, which is why the comment removed from it read `// Beware: relative url does not work with Redirect.response()`. That warning went away along with the behaviour that satisfied it.

Driving the middleware directly with a logged-out context for `/user`:

```
TypeError: Invalid URL
 ❯ Proxy.redirect .../Response.ts:392:14
 ❯ createRedirectWithModifiableHeaders src/middleware/authMiddleware.ts:308:31
 ❯ redirectToAuth src/middleware/authMiddleware.ts:320:12
 ❯ Module.<anonymous> src/middleware/authMiddleware.ts:116:16
```

This hits every enforced-login route — `/user`, `/<organism>/user`, `/<organism>/my_sequences` — so a logged-out visit, or a session expiring while a user sits on "My sequences", returns a 500 instead of redirecting to login.

One-line fix:

```suggestion
    const authUrl = new URL(getLoginUrl(redirectUrl), context.url.origin).toString();
```

The e2e suite misses this because `AuthPage.login()` always starts from `/` and clicks the nav Login link, which goes to `/auth/login` directly; the only test that visits `/user` (`integration-tests/tests/pages/group.page.ts:89`) is already authenticated. Worth adding coverage for the logged-out case.

docs/src/content/docs/reference/browser-authentication-flow.md
theosanderson: In conversation with claude it pointed out that for the registration path - involving an email verification - this is especially short

kubernetes/loculus/templates/keycloak-config-map.yaml
```diff
@@ -257,10 +257,14 @@ data:
...
+          }
```
theosanderson-agent, line 267: _Drafted with the help of Claude._

Worth calling out in the PR description: the realm config map is only ever applied when the realm doesn't yet exist, so none of these Keycloak-side changes reach an instance that has already been deployed.

`keycloak-deployment.yaml:113` starts Keycloak with `--import-realm`, which reads `/opt/keycloak/data/import/` at boot but skips any realm already present in the database. That applies to the whole `loculus` realm definition, not just this client block — every change made to `keycloak-config-map.yaml` since a given instance's first boot is inert on that instance. This PR is just the first time that has security consequences. (The `valuesHash` pod annotation doesn't change it: it's gated on `runDevelopmentKeycloakDatabase && !developmentDatabasePersistence`, so it only covers throwaway dev databases, and a restart wouldn't re-import anyway.)

Concretely, on an existing instance `backend-client` keeps `https://<host>/*` and `http://<host>/*`, and gains neither `pkce.code.challenge.method` nor the post-logout allowlist.

Nothing breaks, which is what makes this easy to miss. The old wildcard still covers `/auth/callback`, and Keycloak honours a `code_challenge` even when the client doesn't mandate one, so `state`, `nonce` and PKCE all keep working end to end. What's silently absent is the server-side half — the narrowed redirect-URI allowlist and the exact post-logout destination, which is the part that addresses security-issues#20. Someone reading the PR description would reasonably assume that upgrading applies it.

Rather than adding an import-strategy override or a partial-import job to this PR, I'd suggest noting the caveat in the description and adding a short section admins can follow if they want the restriction on an existing realm. `for-administrators/user-administration.md` already covers reaching the admin console and selecting the `loculus` realm, so it could pick up from there.

E.g.

**Per-client, via the admin console** — Clients → `backend-client`:

- Settings → Valid redirect URIs: replace the wildcards with `https://<host>/auth/callback`
- Settings → Valid post logout redirect URIs: `https://<host>/logout`
- Advanced → Proof Key for Code Exchange Code Challenge Method: `S256`

docs/src/content/docs/introduction/system-overview.md
```diff
@@ -13,7 +13,7 @@ Loculus has a modular architecture and consists of several sub-services:
...
 - **LAPIS(es):** [LAPIS](https://github.com/GenSpectrum/LAPIS) provides a convenient interface to SILO, offering a lightweight web API and additional data and compression formats. For each SILO instance, there is a corresponding LAPIS instance.
-- **Website:** The frontend application of Loculus accesses the APIs of the backend server and LAPIS. It uses the backend server for everything related to data submission and LAPIS for searching and downloading released data. For logins and registrations, users are redirected to Keycloak.
+- **Website:** The frontend application of Loculus accesses the APIs of the backend server and LAPIS. It uses the backend server for everything related to data submission and LAPIS for searching and downloading released data. For browser login and registration, the website starts an OpenID Connect transaction and redirects the user to Keycloak. The website then validates Keycloak's response before establishing the session. See the [browser authentication flow](../../reference/browser-authentication-flow/) for details.
```
chaoran-chen: I'm not sure about the change here as this is document is to give a quite high-level overview and this sounds quite technical already. Could this be phrased a bit simpler?

docs/src/content/docs/reference/browser-authentication-flow.md
```diff
@@ -0,0 +1,72 @@
+---
+title: Browser authentication flow
+description: How the Loculus website signs users in through OpenID Connect
+---
+
+Loculus uses [Keycloak](https://www.keycloak.org/) as its OpenID Connect (OIDC) provider. Keycloak authenticates the user, while the Loculus website starts the login transaction, validates the response and establishes the website session.
+
+This page describes browser login. It does not change [authentication via the API](../../for-users/authenticate-via-api/), which uses a separate flow.
```
chaoran-chen:
```suggestion
This page describes browser login. [Authentication via the API](../../for-users/authenticate-via-api/) follows a separate flow.
```

docs/src/content/docs/reference/browser-authentication-flow.md
chaoran-chen: Indeed, I think the current UX is not great. If staying at the login screen for more than 5 minutes, one gets afterwards:

Image

I think that the time should be extended. If one exceeds the time, one should get redirected back to the website with a useful message (like "Your session has expired, please click on login again.") - although what I find confusing here is that the authentication, at that point, will have been successfully performed by Keycloak already, so if one clicks on login, one gets directly logged in without needing to enter any credentials, that seems confusing and it's not clear to me how much security the time limt adds.

kubernetes/loculus/templates/keycloak-config-map.yaml
chaoran-chen: To clarify what "Could we construct the logout URL from the origin instead" means: do you mean adapting  `UserPage.astro` to not attach any `?foo=bar`? Or do you mean adapting the keycloak config map here? (Do we have any use case for `/logout?foo=bar`?)

theosanderson: Yes I think Codex means "adapting UserPage.astro to not attach any ?foo=bar"

Afaik, we don't have a use case for `/logout?foo=bar`

docs/src/content/docs/reference/browser-authentication-flow.md
snehasaisneha: **(Comment by Codex)**

Handled in `/auth/login`: it resolves `returnTo` against `context.url.origin` and rejects it unless the origins match. Only that validated value is stored in the transaction cookie and used after the callback. I expanded the reference text and added regression coverage for both an ordinary cross-origin URL and `//attacker.test/...`.

docs/src/content/docs/reference/browser-authentication-flow.md
snehasaisneha: **(Comment by Codex)**

Good question. `backend-client` is configured in Keycloak as a public client, so the value Astro currently passes as `client_secret` is not a reliable security boundary for this authorization-code exchange. PKCE binds the code to the verifier held in the encrypted transaction cookie; `state` separately prevents a callback from being injected into a browser that did not start it. I rewrote the docs to make that relationship clearer.

docs/src/content/docs/reference/browser-authentication-flow.md
snehasaisneha: **(Comment by Codex)**

The Loculus transaction lifetime is now one hour, and invalid or expired callbacks go to a dedicated failure page with a clean retry action. Manual preview testing also surfaced the independent provider timer discussed below: Keycloak may expire one login page earlier, show a Loculus-themed timeout message, and restart its part of the flow while the outer Loculus transaction remains valid. The docs now distinguish those timers. I have left the provider-specific page timeout unchanged here because institutional IdPs will also enforce their own timeouts.

kubernetes/loculus/templates/keycloak-config-map.yaml
snehasaisneha: **(Comment by Codex)**

Fixed in `UserPage.astro`: the post-logout destination is now constructed from the origin and the canonical `/logout` path, so query parameters and fragments from `/user` cannot leak into it. The integration regression test deliberately begins at `/user?source=account-page` and verifies the exact logout flow.

kubernetes/loculus/templates/_config-processor.tpl
snehasaisneha: **(Comment by Codex)**

Agreed that this should not become a one-off secret-handling pattern. PR #7055 is now stacked directly on `state_nonce` and moves the website secrets to Astro server environment variables. I am keeping the minimal conditional in this PR so the security fix remains deployable by itself, with the cross-cutting cleanup isolated in the stacked PR.

website/src/middleware/authMiddleware.ts
snehasaisneha: **(Comment by Codex)**

Fixed by resolving the relative result of `getLoginUrl()` against `context.url.origin` before passing it to `Response.redirect()`. Added middleware coverage for a logged-out request to `/user`, including the expected absolute `/auth/login?...` destination. The preview and full browser matrix now pass.

kubernetes/loculus/templates/keycloak-config-map.yaml
snehasaisneha: **(Comment by Codex)**

Agreed. The PR body now calls out this compatibility boundary, and the operator guide contains the exact `backend-client` steps for existing realms: narrow the login callback, set the exact post-logout destination, and require PKCE `S256`. Existing realms keep authenticating without the action, but do not gain those provider-side restrictions. I kept this as an explicit upgrade step instead of adding automatic mutation of persisted Keycloak realm state.

docs/src/content/docs/introduction/system-overview.md
snehasaisneha: **(Comment by Codex)**

Simplified this back to the system-overview level: the website redirects users to Keycloak and processes the result, with the detailed flow left to the linked reference page.

docs/src/content/docs/reference/browser-authentication-flow.md
snehasaisneha: **(Comment by Codex)**

Applied, thank you.

docs/src/content/docs/reference/browser-authentication-flow.md
snehasaisneha: As Codex said above, this check happens before the browser is redirected to Keycloak. [`website/src/pages/auth/login.ts`](https://github.com/loculus-project/loculus/blob/ce84e58abd82f9afbd5ae9b880ee8a995e372247/website/src/pages/auth/login.ts#L8-L29) validates that `returnTo` points to the same Loculus website and stores it with the login transaction.

After Keycloak returns to the registered `/auth/callback` endpoint, [`website/src/middleware/authMiddleware.ts`](https://github.com/loculus-project/loculus/blob/ce84e58abd82f9afbd5ae9b880ee8a995e372247/website/src/middleware/authMiddleware.ts#L90-L95) retrieves that validated destination and redirects the user there. This prevents a crafted Loculus login link from sending someone to an external website after they sign in.

docs/src/content/docs/for-administrators/user-administration.md
```diff
@@ -28,6 +28,18 @@ Use the Keycloak admin password that you have configured previously
 
 The Loculus-specific users are stored in the `loculus` realm which can be selected in the selection box in the left navigation bar.
 
+## Update browser authentication settings on an existing realm
```
anna-parker, line 31: if this is sth maintainers should do when rolling this PR out it would be great to mark it as breaking :-)

docs/src/content/docs/reference/browser-authentication-flow.md
snehasaisneha: To simplify further:

In the standard Loculus Keycloak configuration, `backend-client` is registered as a "public client" in [`kubernetes/loculus/templates/keycloak-config-map.yaml`](https://github.com/loculus-project/loculus/blob/ce84e58abd82f9afbd5ae9b880ee8a995e372247/kubernetes/loculus/templates/keycloak-config-map.yaml#L253-L266). This means Keycloak does not require the application to authenticate with a client secret when redeeming an authorization code.

Astro does pass `backendKeycloakClientSecret` when constructing its OIDC client in [`website/src/utils/clientMetadata.ts`](https://github.com/loculus-project/loculus/blob/ce84e58abd82f9afbd5ae9b880ee8a995e372247/website/src/utils/clientMetadata.ts#L4-L20). However, because `backend-client` is registered as public in Keycloak, that secret is not enforced as the security boundary for the token exchange.

Before this PR, someone who obtained a valid, unused authorization code could therefore attempt to redeem it using the public client ID and registered callback URL. The code is short-lived and HTTPS makes interception difficult, so this is not necessarily an easy attack, but an intercepted code could be used without an additional transaction-specific secret.

With this PR, [`website/src/pages/auth/login.ts`](https://github.com/loculus-project/loculus/blob/ce84e58abd82f9afbd5ae9b880ee8a995e372247/website/src/pages/auth/login.ts#L24-L40) sends Keycloak a challenge derived from a private PKCE verifier. [`website/src/middleware/authMiddleware.ts`](https://github.com/loculus-project/loculus/blob/ce84e58abd82f9afbd5ae9b880ee8a995e372247/website/src/middleware/authMiddleware.ts#L253-L266) must provide the original verifier when redeeming the code. An intercepted code is therefore no longer sufficient by itself.

The current [OAuth 2.0 Security Best Current Practice](https://datatracker.ietf.org/doc/html/rfc9700#section-2.1.1) requires public clients to use PKCE.

anna-parker: thanks!

website/src/routes/routes.ts
anna-parker, line 1: codex flagged -  not sure it is true:

The state/nonce cookie is per-response and shared across the page. If a user opens a page (cookie set
  for redirect A), then navigates/opens another tab (cookie overwritten for redirect B), the first page's
  stale login link will now fail state validation and dump them at an error. Just a UX edge, not a
  security hole — but worth a fallback that re-initiates the flow cleanly on state mismatch.

docs/src/content/docs/reference/browser-authentication-flow.md
```diff
@@ -0,0 +1,76 @@
...
+## Callback validation failures
+
+When the middleware cannot complete the transaction, `/auth/callback` redirects to a login-failure page without creating a session. Removing the callback parameters from the visible URL also prevents an invalid authorization code and `state` from remaining in the address bar. Common causes include:
+
+- the login was started more than one hour earlier;
+- the callback was refreshed or reused after its transaction had already been consumed;
+- the transaction cookie is missing or cannot be decrypted; or
+- OIDC validation or retrieval of the user's information failed.
+
+Use **Try logging in again** on the failure page instead of refreshing or reopening a failed callback URL. This starts a new OIDC transaction and returns to the account page after a successful login. Server logs record a non-secret transaction identifier and one of the following reasons:
```
anna-parker, line 69: do we do this?

integration-tests/tests/specs/auth/logout.spec.ts
```diff
@@ -16,7 +16,9 @@ test.describe('Logout Flow', () => {
         await authPage.createAccount(testAccount);
         await page.waitForLoadState('networkidle');
 
-        await authPage.logout();
```
anna-parker, line 19: I noticed authPage.logout() is also called in 'should login with valid credentials' - is there a reason it wasnt changed there?

kubernetes/loculus/templates/keycloak-config-map.yaml
```diff
@@ -257,10 +257,14 @@ data:
           "publicClient": true,
           "directAccessGrantsEnabled": true,
           "redirectUris": [
-            "https://{{$.Values.host}}/*",
-            "http://{{$.Values.host}}/*",
-            "http://localhost:3000/*"
-          ]
+            "https://{{$.Values.host}}/auth/callback"{{ if $.Values.insecureCookies }},
+            "http://{{$.Values.host}}/auth/callback",
+            "http://localhost:3000/auth/callback"{{ end }}
```
theosanderson-agent, line 262: Gating the `http://localhost:3000` redirect URI on `insecureCookies` breaks the documented local install path.

`values.yaml` defaults `insecureCookies: false`, and `values_e2e_and_dev.yaml` (which sets it to `true`) is only picked up by `deploy.py` under `--dev` / `--for-e2e`. The `helm install` command in `my-first-loculus.md` sets neither `insecureCookies` nor `host`:

```
helm template kubernetes/loculus --set environment=local --set branch=latest
```

renders

```json
"redirectUris": [
  "https:///auth/callback"
],
```

and `post.logout.redirect.uris` becomes `https:///logout`. On `main` the list always contained `http://localhost:3000/*`. `login.ts` builds `redirect_uri` from `context.url.origin`, so a user following the docs and opening `http://localhost:3000` gets `Invalid parameter: redirect_uri` from Keycloak on login, and logout is rejected the same way. The same happens for `generate_local_test_config.sh --from-live` against a freshly imported realm.

Suggest keying the localhost entries on `environment == "local"` (or always including them, as before) rather than on `insecureCookies`, and dropping the `http://{{ host }}` entries independently if the intent is to stop allowing plain-HTTP callbacks on real hosts.

kubernetes/loculus/templates/loculus-website-config.yaml
```diff
@@ -29,7 +29,8 @@ data:
         "public": {
           {{- template "loculus.publicRuntimeConfig" . -}}
         },
-        "backendKeycloakClientSecret" : "[[backendKeycloakClientSecret]]"
+        "backendKeycloakClientSecret" : "[[backendKeycloakClientSecret]]",
+        "oidcTransactionCookieSecret" : {{ if .Values.testconfig }}"test-oidc-transaction-cookie-secret"{{ else }}"[[oidcTransactionCookieSecret]]"{{ end }}
```
theosanderson-agent, line 33: This breaks `generate_local_test_config.sh --from-live`, which is a documented workflow in `website/README.md`.

`deploy.py` only sets `testconfig=true` on the non-from-live branch. With `--from-live` it sets `environment=server` and `usePublicRuntimeConfigAsServerSide=true` instead, so this line renders the literal `"[[oidcTransactionCookieSecret]]"`. The config processor does run locally, but it only substitutes placeholders that have a matching `LOCULUSSUB_*` env var, and those are only injected in-cluster by `_config-processor.tpl`. Nothing sets them on a dev machine, so the placeholder lands unchanged in `website/tests/config/runtime_config.json`.

`[[oidcTransactionCookieSecret]]` is 31 characters and the new schema in `runtimeConfig.ts` requires `min(32)`, so `getRuntimeConfig()` throws on the first request and the local website serves nothing. `[[backendKeycloakClientSecret]]` has always been left unsubstituted in the same way, but its check is `min(5)` so the placeholder passes validation and the wrong value is never used locally.

Options: have `deploy.py` also pass `testconfig=true` on the from-live branch (there is no reason a locally generated config should ever carry a real secret), or key the fallback here on `disableWebsite` which local config generation always sets.

website/src/utils/getAuthUrl.ts
anna-parker, line 1: getAuthUrl has been renamed getLoginUrl - can we rename it back or change the name of this file?

website/src/middleware/authMiddleware.ts
```diff
@@ -296,6 +327,7 @@ function removeTokenCodeFromSearchParams(url: URL): string {
     newUrl.searchParams.delete('code');
     newUrl.searchParams.delete('session_state');
     newUrl.searchParams.delete('iss');
+    newUrl.searchParams.delete('state');
```
theosanderson-agent, line 330: Is this helper still needed? Its only caller is `redirectToAuth`, which runs for logged-out requests to the protected routes (`/user`, `/<organism>/user`, `/<organism>/my_sequences`). Now that `login.ts` always sets `redirect_uri` to `/auth/callback`, Keycloak never sends `code`/`state`/`session_state`/`iss` to those pages, so there is nothing for this to strip.

The other three places that build a login link (`Navigation.astro`, `NeedToLogin.astro`, `group/[groupId]/index.astro`) pass `Astro.url.toString()` through unmodified, so this path is the only one that sanitises `returnTo`.

Adding `state` to the list is also new here. Nothing on a protected page reads a `state` query param today, but it will silently drop one if that ever changes.

Suggest deleting the helper and having `redirectToAuth` call `getLoginUrl(context.url.toString())` like the other callers.

website/src/utils/getAuthUrl.ts
```diff
@@ -2,33 +2,25 @@ import { KeycloakClientManager } from './KeycloakClientManager';
 import { realmPath } from './realmPath.ts';
 import { routes } from '../routes/routes';
 
-export const getAuthUrl = async (redirectUrl: string) => {
-    const logout = routes.logout();
-    if (redirectUrl.endsWith(logout)) {
-        redirectUrl = redirectUrl.replace(logout, routes.userOverviewPage());
+export const getLoginUrl = (returnTo: string) => {
```
anna-parker, line 5: I found it a bit confusing to have to use `new URL(returnTo, 'https://loculus.invalid')` - could we instead pass the origin as well and then we can check for same origin here as well? (please let me know if this isnt possible for some reason)

```
export const getLoginUrl = (returnTo: string, origin: string) => {
      const fallback = routes.userOverviewPage();

      let resolved: URL;
      try {
          resolved = new URL(returnTo, origin);
      } catch {
          return routes.authLogin(fallback);
      }

      // Open-redirect guard: only ever return to our own origin.
      if (resolved.origin !== origin) {
          return routes.authLogin(fallback);
      }

      // Don't bounce the user straight back to an auth exit page.
      const authExitPaths: string[] = [routes.logout(), routes.authLoginFailed()];
      const returnToPath = authExitPaths.includes(resolved.pathname)
          ? fallback
          : resolved.pathname + resolved.search + resolved.hash;

      return routes.authLogin(returnToPath);
  };
```

kubernetes/loculus/templates/keycloak-config-map.yaml
anna-parker, line 1: We should update the website docs, it was previously possible to run the website locally and talk to a preview using `../generate_local_test_config.sh --from-live --live-host preview.loculus.org`, this is nolonger possible on previews as `insecureCookies` is set to false by default, in order for this to work on previews we need to set `insecureCookies: true`. Alternatively we could keep `insecureCookies` the default and mark that in production this should be marked false (I dont mind either way).

Another (preexisting) issue is that `.Values.testconfig` is not configured in the values.yaml and is set to false by default everywhere except for when using `--from-live`, as cookie secrets are created when `testconfig` is false (default) there is a mismatch when the website is run locally, leading to errors. This can be resolved by adding a cookie when running the website locally:

```
LOCULUSSUB_oidcTransactionCookieSecret=loculus-local-dev-oidc-cookie-secret-0123456789 \
    ../generate_local_test_config.sh --from-live --live-host state-nonce.loculus.org
  npm run start
```

kubernetes/loculus/templates/keycloak-config-map.yaml
anna-parker: Oh cool, I also found this while testing locally! https://github.com/loculus-project/loculus/pull/6994#discussion_r3924578680

kubernetes/loculus/templates/loculus-website-config.yaml
anna-parker: same here! https://github.com/loculus-project/loculus/pull/6994#discussion_r3924578680

kubernetes/loculus/templates/keycloak-config-map.yaml
theosanderson: I think somehow or other we should avoid the need to update the docs but instead make this work

website/src/pages/auth/callback.ts
anna-parker, line 1: As a PKCE and auth newbie I was a bit confused why this file exists, claude explained keycloak requires a redirect endpoint and so this needs to exist and be registered in keycloak.

I asked if this is standard and claude said it was but we have an odd code split (which is preexisting) and it would be more standard to do the following:


> Putting the code→token exchange in authMiddleware and leaving callback.ts as a sad-path stub is not the common pattern. The conventional split is:
> 
>   - Callback route does the whole OIDC dance: read code/state → consume + validate the transaction cookie
>     → exchange code for tokens → set cookies → return context.redirect(returnTo); any failure →
>     redirect(authLoginFailed()).
>   - Middleware only hydrates context.locals.session from cookies (with refresh) and enforces login on guarded routes. It has no idea /auth/callback is special.
> 
> I (claude) would suggest: Moving the exchange into callback.ts and slimming the middleware to session-hydration + route-guarding

website/src/middleware/authMiddleware.ts
```diff
@@ -225,21 +229,48 @@ async function getUserInfo(token: TokenCookie, client: BaseClient) {
...
+    token: TokenCookie;
+    transactionId: string;
+    returnTo: string;
+};
+
+export async function getTokenFromParams(context: APIContext, client: BaseClient): Promise<CallbackResult | undefined> {
     const params = client.callbackParams(context.url.toString());
-    logger.debug(`Keycloak callback params: ${JSON.stringify(params)}`);
+    if (params.code === undefined && params.error !== undefined) {
+        consumeAuthRequest(context.cookies, params.state);
```
anna-parker, line 241: maybe we should comment that as a side effect consumeAuthRequest removes the state from our store

website/src/utils/authRequestCookies.ts
```diff
@@ -0,0 +1,137 @@
...
+            .filter((entry): entry is [string, StoredAuthRequest] => entry[1] !== undefined)
+            .sort(([, left], [, right]) => right.expiresAt - left.expiresAt)
+            .slice(0, maxConcurrentTransactions),
+    );
+    writeStore(cookies, boundedStore);
+}
+
+export function consumeAuthRequest(cookies: AstroCookies, state: string | undefined): AuthRequest | undefined {
+    const store = activeTransactions(unseal(cookies.get(AUTH_TRANSACTIONS_COOKIE)?.value));
+    if (state === undefined) {
```
anna-parker, line 117: store can be unsealed after confirming state is not undefined to reduce the amount of unsealing

docs/src/content/docs/reference/browser-authentication-flow.md
anna-parker: we do this

StefHarmens:
Do we have an ETA for this PR? we require features that are in this PR for our environment.

corneliusroemer:
I think there's a gap in the server-side `state` check that undoes it on existing/upgraded realms.

`consumeAuthRequest` looks up the transaction with `store[state]`, but `store` comes from `Object.fromEntries(...)` so it inherits `Object.prototype`, and `state` is the raw callback value. A callback with `state=constructor` (or `__proto__`, `toString`, `hasOwnProperty`, `valueOf`) resolves to an inherited property, so `transaction === undefined` is never true and the check is bypassed. The "transaction" it returns is `{}`-shaped, so `nonce` and `codeVerifier` are both `undefined` — and openid-client skips the nonce check when the stored nonce and the ID token claim are both absent, and omits `code_verifier` from the grant when it's undefined. So none of state/nonce/PKCE actually gates the callback.

On a freshly imported realm the new `pkce.code.challenge.method: S256` attribute stops the crafted authorization request, so this is only latent there. However, on a realm upgraded in place Keycloak skips re-importing an existing realm, so `backend-client` keeps `redirectUris: https://<host>/*` and no server-side PKCE requirement, and the chain can be walked end to end — which is login CSRF / session fixation, the thing `state` is there to prevent. Worth noting the callback branch now runs *before* the cookie check, so unlike the pre-PR code it can also swap an already-logged-in user onto someone else's session.

One-line fix plus regression tests in a PR I'm making, based on this branch. Guard the lookup with `Object.prototype.hasOwnProperty.call(store, state)` — not `Object.hasOwn`.

PR #7015, opened 2026-07-31
fix(backend): Add current date to etag that's used in get-released-data

### LINKED ISSUE #7013: Include current date in get-released-data ETag so newly OPEN sequences don't wrogly show as RESTRICTED in SILO
`get-released-data` decides open vs. restricted data use terms by comparing today's date to restricted_until (`computeDataUseTerm` in `ReleasedDataModel.kt`), but the ETag is the timestamp of the last write to a released-data table (`getLastDatabaseWriteETag` in `ReleasedDataModel.kt`). 

A RESTRICTED DUT expiry writes nothing, so after midnight `get-released-data` starts returning OPEN for newly OPEN sequences while the ETag is unchanged so it doesn't trigger a reingest of data by SILO. This means the website will keep showing a sequence as RESTRICTED until the hard refresh time is reached (~27 hours at most) or something changes in the resleased data table.

## Suggested fix

Append the current UTC date to the ETag on the `get-released-data` endpoint: "\"$lastUpdateTime|${dateProvider.getCurrentDate()}\"".

maverbiest:
resolves #7013

The ETag `get-released-data` returns is only updated when something is written to released data tables. However, a sequence's DUT can change from RESTRICTED to OPEN on a new day if the restricted until date is reached. This does not result in a database write so the ETag does not change, but `get-released-data` will start returning DUT OPEN for the sequence as it is computed from the current date and the restricted until date each time the endpoint is called.

This can cause a mismatch between the DUT reported by the endpoint and what is show in SILO/LAPIS, since SILO only imports new data when the ETag changes. This was noticed as a mismatch between data returned by the endpoint and what was visible on the website (website reflects LAPIS state): https://loculus.slack.com/archives/C0757PTR607/p1785492200559359.

## Fix

This PR adds a `getLastDatabaseWriteETagWithDate` function to be used in the `get-released-data` endpoint controller. The only difference is that it adds the current date to the ETag. I'm still keeping the old ETag implementation around as it's also used by `extract-unprocessed-data`, where a change in DUT doesn't matter as far as I can tell? (let me know if this is not the case, happy to update)

This does mean that all organisms will have a SILO refresh when the UTC date changes. This is some extra load but hopefully shouldn't be too bad.

### PR Checklist

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
@@ -335,7 +335,7 @@ open class SubmissionController(
             ),
             Header(
                 name = "eTag",
-                description = "Last database write Etag",
+                description = "Last database write Etag, combined with the current date",
                 schema = Schema(type = "integer"),
             ),
         ],
@@ -343,8 +343,8 @@ open class SubmissionController(
     @ApiResponse(
         responseCode = "304",
         description =
-        "No database changes since last request " +
-            "(Etag in HttpHeaders.IF_NONE_MATCH matches lastDatabaseWriteETag)",
+        "No database changes since last request, and the date has not changed " +
+            "(Etag in HttpHeaders.IF_NONE_MATCH matches lastDatabaseWriteETagWithDate)",
     )
     @GetMapping("/get-released-data", produces = [MediaType.APPLICATION_NDJSON_VALUE])
     fun getReleasedData(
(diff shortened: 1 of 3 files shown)


anna-parker:
side note: I realized the `/extract-unprocessed-data` etag is not scoped by organism, we should ideally add organism scope and a test there - can be a follow up PR :-)

anna-parker (approved):
looks great! sorry it took me so long to review, due to the extra load this will produce every day at midnight I will quickly check in with others if we are ok to merge this - but I think it is fine!

(Otherwise we wait 27h until a hard refresh, meaning we will end up waiting approx 13h until open sequences also appear as open on SILO after being released)

PR #7017, opened 2026-07-31
feat!(deposition): split and move out functions to be reused in raw reads, improve manifest creation code, fix bug in assembly revision code

anna-parker:
refactor in preparation of https://github.com/loculus-project/loculus/pull/6816

- splits and moves some functions out of the assembly file into generic helper files so that functions can be reused in raw reads
- improves manifest creation code (raw reads also require a manifest with slightly different fields)
- fix bug in assembly revision code: if `allow_revision_with_manifest_changes` is true we also want the pipeline to revise assemblies automatically (instead of sending us a slack notification) if the manifest metadata changed. I added this to the ena workflow tests.

## Breaking Change
Technically `manifest_fields_mapping` has been renamed `assembly_manifest_fields_mapping` - however as Pathoplexus is the only user of the ENA code and we havent configured it there no changes are required.

### PR Checklist
I added additional unit tests for the manifest generation and manifest field diff code and additionally I added a new ena workflow test that covers the bug case (workflow tests talk to ENA dev directly, sadly ENA dev does not process assemblies so part of the full process can only be covered in production code but the workflow tests document this)

### DIFF
--- a/ena-submission/ENA_submission.md
+++ b/ena-submission/ENA_submission.md
@@ -519,7 +519,7 @@ Revisions to a study or sample should be submitted the same way as original sequ
 
 ## 2. [Revising Assemblies](https://ena-docs.readthedocs.io/en/latest/update/assembly.html)
 
-It appears that all fields that were explicitly set via the manifest must be updated via an email. This includes changes to any field that is in the `manifest_fields_mapping` field of the default.yaml. Additionally, study and sample reference must stay the same and chromosome names cannot be changed (but new ones can be added).
+It appears that all fields that were explicitly set via the manifest must be updated via an email. This includes changes to any field that is in the `assembly_manifest_fields_mapping` field of the default.yaml. Additionally, study and sample reference must stay the same and chromosome names cannot be changed (but new ones can be added).
 However, unlike the alias a NEW `ASSEMBLYNAME` is required (cannot be the same as the assemblyname of the previous version).
 
 Currently we automate revision of studies and assemblies, if a manifest update is required the pipeline will not update the assembly but set the state of assembly submission to `HAS_ERRORS` and document the reason for the errors in the database. We will then receive a slack notification and will have to manually send an email to ENA to update the manifest.
--- a/ena-submission/ENA_submission.md
+++ b/ena-submission/ENA_submission.md
@@ -519,7 +519,7 @@ Revisions to a study or sample should be submitted the same way as original sequ
 
 ## 2. [Revising Assemblies](https://ena-docs.readthedocs.io/en/latest/update/assembly.html)
 
-It appears that all fields that were explicitly set via the manifest must be updated via an email. This includes changes to any field that is in the `assembly_manifest_fields_mapping` field of the default.yaml. Additionally, study and sample reference must stay the same and chromosome names cannot be changed (but new ones can be added).
+Study and sample accessions must stay the same in a revision and chromosome names cannot be changed (but new ones can be added). In the past, all metadata fields that were set via the manifest had to be revised via an email. ENA has now informed us we can revise these fields in a normal automated revision but we have had issues with it, so we have a `allow_revision_with_manifest_changes` flag which can enable/disable automated manifest revisions.
 However, unlike the alias a NEW `ASSEMBLYNAME` is required (cannot be the same as the assemblyname of the previous version).
 
 Currently we automate revision of studies and assemblies, if a manifest update is required the pipeline will not update the assembly but set the state of assembly submission to `HAS_ERRORS` and document the reason for the errors in the database. We will then receive a slack notification and will have to manually send an email to ENA to update the manifest.
--- a/ena-submission/config/defaults.yaml
+++ b/ena-submission/config/defaults.yaml
@@ -29,7 +29,7 @@ min_between_publicness_checks: 720
 log_level: DEBUG
 metadata_fallback_fields:
   specimenCollectorSampleId: submissionId
-manifest_fields_mapping:
+assembly_manifest_fields_mapping:
   authors:
     loculus_fields: [authors]
     function: reformat_authors
@@ -41,8 +41,7 @@ manifest_fields_mapping:
     default: "Unknown"
   coverage:
     loculus_fields: [depthOfCoverage]
-    type: int
-    default: 1
+    default: "1"
   run_ref:
     loculus_fields: [insdcRawReadsAccession]
 #ena_checklist: ERC000033 - do not use until all fields are mapped to ENA accepted options
--- a/ena-submission/src/ena_deposition/create_assembly.py
+++ b/ena-submission/src/ena_deposition/create_assembly.py
@@ -373,11 +373,12 @@ def can_be_revised(config: Config, db_engine: Engine, submission_row: Submission
     return not manifest_fields_changed(config, db_engine, submission_row, last_version_entry)
 
 
-def is_flatfile_data_changed(
+def has_assembly_data_changed(
     config: Config, db_engine: Engine, submission_row: SubmissionTableEntry
 ) -> bool:
     """
-    Check if change in sequence or flatfile metadata has occurred since last version.
+    Check if there have been changes in the: sequence, flatfile or manifest
+    (if config.allow_revision_with_manifest_changes is true) metadata since last version.
     """
     last_entry = get_last_entry(db_engine, submission_row.pkey)
 
@@ -486,7 +487,8 @@ def assembly_table_create(db_engine: Engine, config: Config):
             logger.debug(f"Entry {row.accession} is a revision, checking if it can be revised")
             if not can_be_revised(config, db_engine, submission_row):
                 continue
-            if not is_flatfile_data_changed(config, db_engine, submission_row):
+            if not has_assembly_data_changed(config, db_engine, submission_row):
+                # Do not revise assembly if there are no changes
                 update_assembly_results_with_latest_version(db_engine, seq_key)
                 continue
 
--- a/ena-submission/src/ena_deposition/create_assembly.py
+++ b/ena-submission/src/ena_deposition/create_assembly.py
@@ -377,8 +377,10 @@ def has_assembly_data_changed(
     config: Config, db_engine: Engine, submission_row: SubmissionTableEntry
 ) -> bool:
     """
-    Check if there have been changes in the: sequence, flatfile or manifest
-    (if config.allow_revision_with_manifest_changes is true) metadata since last version.
+    Check if there have been changes since last version in: 
+    - sequence
+    - flatfile
+    - manifest metadata (iff config.allow_revision_with_manifest_changes==True)
     """
     last_entry = get_last_entry(db_engine, submission_row.pkey)
 
--- a/ena-submission/src/ena_deposition/create_assembly.py
+++ b/ena-submission/src/ena_deposition/create_assembly.py
@@ -41,8 +41,6 @@ from .notifications import SlackConfig, send_slack_notification, slack_conn_init
 from .submission_db_helper import (
     AccessionVersion,
     AssemblyTableEntry,
-    ProjectTableEntry,
-    SampleTableEntry,
     Status,
     StatusAll,
     SubmissionTableEntry,
--- a/ena-submission/src/ena_deposition/create_assembly.py
+++ b/ena-submission/src/ena_deposition/create_assembly.py
@@ -377,7 +377,7 @@ def has_assembly_data_changed(
     config: Config, db_engine: Engine, submission_row: SubmissionTableEntry
 ) -> bool:
     """
-    Check if there have been changes since last version in: 
+    Check if there have been changes since last version in:
     - sequence
     - flatfile
     - manifest metadata (iff config.allow_revision_with_manifest_changes==True)
--- a/ena-submission/src/ena_deposition/create_assembly.py
+++ b/ena-submission/src/ena_deposition/create_assembly.py
@@ -21,9 +21,12 @@ from .ena_submission_helper import (
     create_ena_assembly,
     create_flatfile,
     create_manifest,
-    get_authors,
     get_description,
     get_ena_analysis_process,
+    linked_accession_diff,
+    manifest_fields_diff,
+    resolve_manifest_field,
+    resolve_required_manifest_field,
     retry_failed_submissions_for_matching_errors,
     set_accession_does_not_exist_error,
 )
@@ -48,6 +51,7 @@ from .submission_db_helper import (
     find_conditions_in_db,
     find_errors_or_stuck_in_db,
     find_waiting_in_db,
+    get_last_entry,
     is_latest_revision,
     is_revision,
     previous_version,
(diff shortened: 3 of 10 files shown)


ena-submission/src/ena_deposition/create_assembly.py
```diff
@@ -103,66 +107,6 @@ def get_segment_order(unaligned_sequences: dict[str, str | None]) -> list[str]:
     return sorted(segment_order)
 
 
-def get_address(config: Config, submission_row: SubmissionTableEntry) -> str | None:
```
anna-parker, line 106: moved

ena-submission/src/ena_deposition/create_assembly.py
```diff
@@ -103,66 +107,6 @@ def get_segment_order(unaligned_sequences: dict[str, str | None]) -> list[str]:
...
-        address.city,
-        address.state,
-        address.country,
-    ]
-    address_string = ", ".join([x for x in address_list if x])
-    logger.debug("Created address from group_info")
-    return address_string
-
-
-def get_assembly_values_in_metadata(config: Config, metadata: dict[str, str]) -> dict[str, str]:
```
anna-parker, line 133: replaced with resolve_required_manifest_field

ena-submission/src/ena_deposition/create_assembly.py
```diff
@@ -497,8 +405,16 @@ def is_flatfile_data_changed(db_engine: Engine, submission_row: SubmissionTableE
                 f"for {submission_row.accession}. (Maybe other fields changed as well)"
             )
             return True
+    if config.allow_revision_with_manifest_changes and manifest_fields_diff(
```
anna-parker, line 408: this was actually a bug - if `allow_revision_with_manifest_changes` is true we also want to revise assemblies if the manifest metadata changed

ena-submission/src/ena_deposition/create_assembly.py
```diff
@@ -534,39 +451,6 @@ def update_assembly_results_with_latest_version(db_engine: Engine, seq_key: Acce
     )
 
 
-def get_project_and_sample_results(
```
anna-parker, line 537: moved

ena-submission/config/defaults.yaml
```diff
@@ -41,8 +41,7 @@ manifest_fields_mapping:
     default: "Unknown"
   coverage:
     loculus_fields: [depthOfCoverage]
```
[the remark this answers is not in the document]
anna-parker: this is fine

ena-submission/src/ena_deposition/ena_submission_helper.py
```diff
@@ -519,52 +521,108 @@ def create_manifest(
...
+) -> str:
+    """Like resolve_manifest_field, but for required fields."""
+    value = resolve_manifest_field(field_details, metadata)
+    if value is None:
+        msg = (
+            "Manifest field resolved to None despite being required: "
+            f"loculus_fields={field_details.loculus_fields}, no default configured."
+        )
+        raise ValueError(msg)
+    return value
```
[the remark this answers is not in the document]
anna-parker: I got claude to write some unit tests for `manifest_fields_diff`, `resolve_required_manifest_field` gets tested when testing `create_manifest`.

ena-submission/src/ena_deposition/create_assembly.py
anna-parker: I added a test for this!

ena-submission/ENA_submission.md
```diff
@@ -519,7 +519,7 @@ Revisions to a study or sample should be submitted the same way as original sequ
 
 ## 2. [Revising Assemblies](https://ena-docs.readthedocs.io/en/latest/update/assembly.html)
 
-It appears that all fields that were explicitly set via the manifest must be updated via an email. This includes changes to any field that is in the `manifest_fields_mapping` field of the default.yaml. Additionally, study and sample reference must stay the same and chromosome names cannot be changed (but new ones can be added).
+It appears that all fields that were explicitly set via the manifest must be updated via an email. This includes changes to any field that is in the `assembly_manifest_fields_mapping` field of the default.yaml. Additionally, study and sample reference must stay the same and chromosome names cannot be changed (but new ones can be added).
```
maverbiest: Is still accurate? Didn't we hear from ENA that we can submit revisions for manifest fields (author list) without having to reach out via email?

ena-submission/src/ena_deposition/ena_submission_helper.py
```diff
@@ -519,52 +521,108 @@ def create_manifest(
...
+) -> dict[str, str]:
+    """If the submitter provided a biosampleAccession/bioprojectAccession (e.g. because reads
+    are linked to an already-existing sample/project), check it still matches the accession
+    used for the previous version."""
+    previous_accessions = {
+        "biosampleAccession": previous_sample_accession,
+        "bioprojectAccession": previous_study_accession,
+    }
+    differing_fields = {}
+    for field, previous_accession in previous_accessions.items():
```
maverbiest, line 622: Not blocking, but since it's only two fields I wonder if it would be more readable to fetch them explicity rather than making a dict and iterating? Somrthing like (and then same for bioproject):

```python
new_sample_accession = submission_row.seq_metadata.get("biosampleAccession")
if new_sample_accession and new_sample_accession != previous_sample_accession:
            differing_fields["biosampleAccession"] = f"Last: {previous_sample_accession}, New: {new_sample_accession}"
```

ena-submission/src/ena_deposition/ena_submission_helper.py
```diff
@@ -519,52 +521,108 @@ def create_manifest(
...
+        value = get_authors(values[0] or "")
+    else:
+        value = ", ".join(str(v) for v in values if v) or None
+
+    if value is not None:
+        return value
+    return field_details.default
+
+
+def resolve_required_manifest_field(
```
maverbiest, line 573: Probably beyond the scope of this PR and I also haven't looked at this code a lot, so might not be a good suggestion. But could it make sense to add a `required: bool` field to `ManifestFieldDetails` and have one `resolve_manifest_field` function that checks this and handles required vs not required? 

I guess now you need to check whether a `AssemblyManifest` field allows for `None` or not and call the right function based on that

ena-submission/src/ena_deposition/ena_submission_helper.py
```diff
@@ -519,52 +521,108 @@ def create_manifest(
...
+
+def manifest_fields_diff(
+    assembly_manifest_fields_mapping: dict[str, ManifestFieldDetails],
+    submission_row: SubmissionTableEntry,
+    last_version_entry: SubmissionTableEntry,
+) -> dict[str, str]:
+    differing_fields = {}
+    for field, mapping in assembly_manifest_fields_mapping.items():
+        try:
+            last_value = resolve_manifest_field(mapping, last_version_entry.seq_metadata)
```
maverbiest, line 595: Are the mappings we check here always non-required? Or does it not matter at this point since this just gets a diff?

ena-submission/src/ena_deposition/ena_submission_helper.py
```diff
@@ -519,52 +521,108 @@ def create_manifest(
...
-            if not is_broker:
-                logger.error("Cannot set authors field for non broker")
-                msg = "Cannot set authors field for non broker"
+        for field in fields(manifest):
+            value = getattr(manifest, field.name)
+            if value is None:
+                continue
+            if isinstance(value, list):
+                f.writelines(f"{field.name.upper()}\t{val}\n" for val in value)
+                continue
```
maverbiest, line 545:
```suggestion

```

I don't think `AssemblyManifest` has any `list` fields, but maybe I missed something?

ena-submission/src/ena_deposition/submission_db_helper.py
```diff
@@ -562,3 +562,50 @@ def previous_version(engine: Engine, seq_key: AccessionVersion) -> int | None:
     )
     all_versions = sorted(row.version for row in rows)
     return all_versions[-2]
+
+
+def get_last_entry(db_engine: Engine, accession_version: AccessionVersion) -> SubmissionTableEntry:
+    version_to_revise = previous_version(db_engine, accession_version)
```
maverbiest, line 568: Should we check if `version_to_revise` is `None` and raise here? Or are we not too worried about that happening here since it'll probably be caught by `last_version_rows` being length zero, which is checked below?

ena-submission/ENA_submission.md
anna-parker: They told us this but we weren't actually able to revise the manifest (then old and new authors showed) so we have a flag `allow_revision_with_manifest_changes` to enable or disable automated manifest revisions - I will update the readme

anna-parker:
```suggestion
Study and sample accessions must stay the same in a revision and chromosome names cannot be changed (but new ones can be added). In the past, all metadata fields that were set via the manifest had to be revised via an email. ENA has now informed us we can revise these fields in a normal automated revision but we have had issues with it, so we have a `allow_revision_with_manifest_changes` flag which can enable/disable automated manifest revisions.
```

ena-submission/src/ena_deposition/ena_submission_helper.py
anna-parker: I think this is actually quite clean like this so I think I will keep it!

ena-submission/src/ena_deposition/ena_submission_helper.py
anna-parker: the reason there are 2 functions is for the typing - `resolve_required_manifest_field` returns a string, `resolve_manifest_field` can return None

maverbiest (commented):
This is looking good, I left some minor comments and questions but tbf I don't know this part of the code super well yet so please disregard whatever doesn't make sense :)

ena-submission/src/ena_deposition/ena_submission_helper.py
anna-parker: I think this is the cleanest option without having `# type: ignore or cast(str, ...) at every "required" call site` so I will resolve this for now - but if you can find a better option please let me or make a new PR!

ena-submission/src/ena_deposition/ena_submission_helper.py
anna-parker: as this is just for the diff it doesnt matter :-)

ena-submission/src/ena_deposition/submission_db_helper.py
anna-parker: yeah this should be covered by the check below :-)

ena-submission/src/ena_deposition/ena_submission_helper.py
anna-parker: ah very true! However, `RawReadsManifest` (which will be added in the follow up PR) does have list entries, so Id like to already add it now, but could be added later

ena-submission/config/defaults.yaml
```diff
@@ -29,7 +29,7 @@ min_between_publicness_checks: 720
 log_level: DEBUG
 metadata_fallback_fields:
   specimenCollectorSampleId: submissionId
-manifest_fields_mapping:
+assembly_manifest_fields_mapping:
```
corneliusroemer, line 32: Is this a breaking config change?

ena-submission/src/ena_deposition/create_assembly.py
```diff
@@ -210,7 +153,7 @@ def create_manifest_object(
 
     flat_file = create_flatfile(config, metadata, ena_organism, unaligned_nucleotide_sequences, dir)
 
-    assembly_values = get_assembly_values_in_metadata(config, metadata)
+    assembly_manifest_fields_mapping = config.assembly_manifest_fields_mapping
```
corneliusroemer, line 156: We could type the assembly_manifest_fields_mapping itself then below we could access with dot notation (and it'd be typo-proof.

ena-submission/src/ena_deposition/ena_submission_helper.py
maverbiest: Makes sense!

maverbiest (approved):

ena-submission/src/ena_deposition/create_assembly.py
```diff
@@ -457,27 +373,19 @@ def can_be_revised(config: Config, db_engine: Engine, submission_row: Submission
...
 
 
-def is_flatfile_data_changed(db_engine: Engine, submission_row: SubmissionTableEntry) -> bool:
+def has_assembly_data_changed(
+    config: Config, db_engine: Engine, submission_row: SubmissionTableEntry
+) -> bool:
     """
-    Check if change in sequence or flatfile metadata has occurred since last version.
+    Check if there have been changes in the: sequence, flatfile or manifest
+    (if config.allow_revision_with_manifest_changes is true) metadata since last version.
```
corneliusroemer: I'm not sure I understand this comment completely - why is the paren between `manifest` and `metadata`? Is this what you mean:

```suggestion
    Check if there have been changes since last version in: 
    - sequence
    - flatfile
    - manifest metadata (iff config.allow_revision_with_manifest_changes==True)
```

corneliusroemer (approved):
Nice to refactor some of the long functions! Check comments for (non-blocking) suggestions

ena-submission/config/defaults.yaml
anna-parker: I guess technically (its only used inside the defaults.yaml) but yeah I can make a comment

corneliusroemer:
Few more things Claude flagged (run locally):
-  The new integration test doesn't exercise the bug fix. get_revisions(modify_manifest=True) changes authors, but authors is also a flatfile field, so has_assembly_data_changed returns True at the flatfile check and never reaches the new manifest_fields_diff branch. I ran the same probe on both branches: an authors-only change revises on main and the PR (no difference), while sequencingInstrument/depthOfCoverage-only changes flip False → True. So the test is green on main too. Fix is one line — mutate sequencingInstrument instead.
-  Missing or empty authors now silently writes AUTHORS\t; — get_authors("") returns the bare ";", which isn't None so it defeats the default fallback. Both cases raised loudly on main (measured). Regression from fail-loudly to depositing garbage at ENA. One rewrite of resolve_manifest_field (resolve → default → then apply function) also kills an IndexError on empty loculus_fields and the values[0]-drops-extra-fields issue @claude[bot] raised.

ena-submission/src/ena_deposition/create_assembly.py
anna-parker: exactly

ena-submission/src/ena_deposition/ena_submission_helper.py
corneliusroemer, line 545:
```suggestion
```

ena-submission/src/ena_deposition/create_assembly.py
anna-parker: we discussed this is a bigger change so I wont add it here

Issue #6842, opened 2026-07-06
MVP Migrate/curate sequencing-related metadata of submitted sequences to ENA-allowed enums


anna-parker:
ENA only accepts a list of valid enums for specific metadata fields and will error submissions where these are not provided. For these fields we should offer the same options that ENA provides but this means curating data where these fields have already been entered to fit the list of valid fields.

anna-parker:
done in https://github.com/pathoplexus/pathoplexus/pull/1091

PR #6634, opened 2026-06-09
feat(config, deposition): give submitters options for ENA raw reads fields

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/4007

For future raw read submission ENA asks for additional fields (we have already added the required fields), in a follow up PR it would be good to ensure these are properly marked as desired when submitting raw reads in our docs.

These are the fields ENA asks for during raw reads, submission, fields marked in bold we do not currently have a way for users to submit to us.

- STUDY: Study accession or unique name (alias - we get from bioprojectAccession)
- SAMPLE: Sample accession or unique name (alias - we get from biosampleAccession)
- NAME: Unique experiment name (we create when submitting
- PLATFORM: See [permitted values](https://ena-docs.readthedocs.io/en/latest/submit/reads/webin-cli.html#permitted-values-for-platform) -> options already listed in `sequencingInstrument` 
- INSTRUMENT: See [permitted values](https://ena-docs.readthedocs.io/en/latest/submit/reads/webin-cli.html#permitted-values-for-instrument) -> options already listed in `sequencingInstrument` 
- **INSERT_SIZE**: Insert size for paired reads (optional)
- LIBRARY_NAME: Library name (optional) -> not added
- **LIBRARY_SOURCE**: See [permitted values](https://ena-docs.readthedocs.io/en/latest/submit/reads/webin-cli.html#permitted-values-for-library-source) -> can be set to `OTHER`
- **LIBRARY_SELECTION**: See [permitted values](https://ena-docs.readthedocs.io/en/latest/submit/reads/webin-cli.html#permitted-values-for-library-selection) e.g. PCR - can be set to `unspecified`
- LIBRARY_STRATEGY: See [permitted values](https://ena-docs.readthedocs.io/en/latest/submit/reads/webin-cli.html#permitted-values-for-library-strategy) e.g. WGS or WGA - can be set to `OTHER` -> options already listed under `sequencingAssayType`
- DESCRIPTION: free text library description (optional - can be mapped to `sequencingProtocol`, `rawSequenceDataProcessingMethod` and `ampliconPcrPrimerScheme`)

🚀 Preview: Add `preview` label to enable

PR #7029, opened 2026-08-04
feat!(config): Add options for sequencingInstrument and sequencingAssayType

maverbiest:
Based on https://github.com/loculus-project/loculus/pull/6634, but only adds options for `sequencingInstrument` and `sequencingAssayType` metadata fields.

Related to https://github.com/loculus-project/loculus/issues/6842.

The options lists are taken from the allowed values listed here: https://ena-docs.readthedocs.io/en/latest/submit/reads/webin-cli.html#metadata-validation. Note that we exclude the `unspecified` value; ENA makes the PLATFORM field required if INSTRUMENT is `unspecified`, and since we're merging the INSTRUMENT and PLATFORM fields into one field on our side we can't support that.

#### Options appear in the submission form:

grafik

#### Incorrect values are rejected. 
Although note that there is an error because of the incorrect value and then also an error because `sequencingInstrument` is required when raw reads are provided (this is checked on the output field). Not strictly wrong just a bit noisy

grafik

## BREAKING CHANGES
The values.yaml has already been updated on PPX and the db surgery performed, for other users switching from a string to a list of defined options a db surgery should be performed in advance, see https://github.com/pathoplexus/pathoplexus/pull/1091 for an example.

### PR Checklist

🚀 Preview: https://seq-metadata-options.loculus.org

### DIFF
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -1073,7 +1073,7 @@ defaultOrganismConfig: &defaultOrganismConfig
           # ena-webin-cli -context reads -fields
           # ena-webin-cli=9.0.3
           # Last checked: 09.06.2026
-          # Fields listed in INSTRUMENT
+          # Fields listed in INSTRUMENT (excluding `unspecified`)
           - name: "454 GS"
           - name: "454 GS 20"
           - name: "454 GS FLX"
--- a/kubernetes/loculus/values.yaml
+++ b/kubernetes/loculus/values.yaml
@@ -1056,7 +1056,7 @@ defaultOrganismConfig: &defaultOrganismConfig
         orderOnDetailsPage: 2060
       - name: sequencingInstrument
         ontology_id: GENEPIO:0001452
-        definition: The model of the sequencing instrument used.
+        definition: The model of the sequencing instrument used. (required when raw reads file are provided)
         guidance: Select the sequencing instrument from the pick list.
         example: MinION
         displayName: Sequencing instrument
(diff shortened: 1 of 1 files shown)


kubernetes/loculus/values.yaml
```diff
@@ -1056,7 +1056,7 @@ defaultOrganismConfig: &defaultOrganismConfig
         orderOnDetailsPage: 2060
       - name: sequencingInstrument
         ontology_id: GENEPIO:0001452
-        definition: The model of the sequencing instrument used. (required when raw reads file are provided)
+        definition: The model of the sequencing instrument used. (required when raw reads files are provided)
```
anna-parker: raw read files?

anna-parker (commented):
Not quite sure the best order to merge this in. In theory we can just add these options for loculus right away and merge this onto Loculus.

However, I guess it would probably be best to keep Loculus and PPX somewhat in sync. So maybe we can also create a PPX PR with these changes and then test the db surgery on staging - if that succeeds we can merge this onto Loculus :-)

kubernetes/loculus/values.yaml
```diff
@@ -1076,12 +1175,65 @@ defaultOrganismConfig: &defaultOrganismConfig
...
-        example: whole genome sequencing assay [OBI:0002117]
+        definition: The overarching sequencing technique that was used to determine the sequence of a biomaterial, also called library strategy in the INSDC. (required when raw read files are provided)
+        guidance: Select the sequencing technology from the pick list.
+        desired: true
+        example: WGS
         displayName: Sequencing assay type
         header: Sequencing
-        orderOnDetailsPage: 2120
+        orderOnDetailsPage: 2130
+        requiredWhen:
```
anna-parker: ah this field isnt required in my understanding? If users dont supply it we can fallback to OTHER when submitting to ENA. Only `sequencingInstrument` is a requirement. This is also why the integration tests are failing

maverbiest: Ah alright got it! Will update

anna-parker (approved):

PR #7033, opened 2026-08-05
feat(helm): add support for podSecurityContext and containerSecurityContext configurations

### LINKED ISSUE #7014: [Helm] Add support for custom podSecurityContext and containerSecurityContext configurations
When deploying Loculus in more compliance sensitive environments it is often a (hard) requirement that workloads provide pod-level and container-level security policies (such as `runAsNonRoot`, `readOnlyRootFilesystem`, `allowPriviligeEscalation`, or dropping Linux capabilities).

Currently, the Helm chart does not provide a way to pass custom configurations to deployment workloads.

I'd propose adding top-level `podSecurityContext` and `containerSecurityContext` configuration blocks in values.yaml with two phases:
1. **Component-specific overrides** such as `backend`, `keycloak`, `lapis`, `silo`.
2. **default fallback** which applies to all workloads if no component specific override is provided.

This would look somewhat like the following:
```yaml
podSecurityContext:
  default:
    runAsNonRoot: true
    fsGroup: 1000
  backend:
    runAsUser: 1001


containerSecurityContext:
  default:
    allowPrivilegeEscalation: false
    readOnlyRootFilesystem: true
    capabilities:
      drop:
        - ALL
  silo-importer:
    readOnlyRootFilesystem: false  # Component-specific override
```

Reference: [Security context kubernetes documentation](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/)

Would this proposed structure work for you? If yes, I can open a PR for you to review in the coming days

florianzwagemaker:
resolves #7014 

This PR introduces security context configuration support to the helm chart, allowing pod-level and container-level security contexts to be customized.
This will allow Loculus to be deployed in k8s environments where specific security policies are enforced (such as running as a specific user, or user/group constraints).

Here we add two root-level properties in the helm chart (values.yaml and schema):  
* `podSecurityContext`: which applies specific security contexts on a pod-level (applies to all containers in a pod)
* `containerSecurityContext`: applies specific security contexts on a container level (per container)

Both properties support a `default` entry that applies to all workloads, as well as per-component overrides (such as backend, silo, or lapis) A component specific entry will take precedence over the default entry.

> [!NOTE]
> Regarding overrides: there is no deep merge between `default` and a component-specific entry. If a component has or needs its own entry, that entry is used in full and the `default` block will be ignored for that component. In other words, if it's required to change a field for one component, the complete security context for that component must be specified.

### Example 1: Defaults applied to all components

```yaml
podSecurityContext:
  default:
    runAsNonRoot: true
    runAsUser: 1000
    fsGroup: 2000
    seccompProfile:
      type: RuntimeDefault

containerSecurityContext:
  default:
    allowPrivilegeEscalation: false
    readOnlyRootFilesystem: true
    capabilities:
      drop: ["ALL"]
```

This will result in all workloads to render in the helm chart with these security contexts (for example, the silo pod):
```yaml
spec:
  template:
    spec:
      securityContext:
        runAsNonRoot: true
        runAsUser: 1000
        fsGroup: 2000
        seccompProfile:
          type: RuntimeDefault
      containers:
      - name: silo
        securityContext:
          allowPrivilegeEscalation: false
          readOnlyRootFilesystem: true
          capabilities:
            drop: ["ALL"]
```

### Example 2: Defaults applied, with one component needing different settings

For this example we state that the `silo-importer` container needs to write to the filesystem for logging purposes, as a result it cannot use the default `readOnlyRootFilesystem: true`.
Because there's no deep merge, we will have to specify the full security context for that component like below:
```yaml
podSecurityContext:
  default:
    runAsNonRoot: true
    runAsUser: 1000
    fsGroup: 2000

containerSecurityContext:
  default:
    allowPrivilegeEscalation: false
    readOnlyRootFilesystem: true
    capabilities:
      drop: ["ALL"]
  silo-importer:
    allowPrivilegeEscalation: false
    readOnlyRootFilesystem: false # full block is identical except for readOnlyRootFilesystem here
    capabilities:
      drop: ["ALL"]
```

This will result in all containers getting the `default` container security context, except for `silo-importer` which will get its own block with a writable root filesystem. 
```yaml
spec:
  template:
    spec:
      securityContext:
        runAsNonRoot: true
        runAsUser: 1000
        fsGroup: 2000
        seccompProfile:
          type: RuntimeDefault
      containers:
        - name: silo
          securityContext:
            allowPrivilegeEscalation: false
            readOnlyRootFilesystem: true
            capabilities:
              drop: ["ALL"]
        - name: silo-importer
          securityContext:
            allowPrivilegeEscalation: false
            readOnlyRootFilesystem: false
            capabilities:
              drop: ["ALL"]
```

A new `_security-context.tpl` template with `loculus.podSecurityContext` and `loculus.containerSecurityContext` helpers has been added. And all relevant deployment templates are updated to include the helpers & blocks.

The configuration blocks added to `values.yaml` are empty by default to retain backwards compatibility.
Currently, no security contexts will be applied unless explicitly configured.


chaoran-chen:
Thanks @florianzwagemaker! (I've created a mirror of the PR to run the CI / preview: #7117)

florianzwagemaker:
Below is a brief list of changes based on the initial claude review comments in #7117 :

1. Added groups/type metadata to the new `podSecurityContext`/`containerSecurityContext` schema properties so they render in the generated docs.
2. Removed the stale `silo-importer` and `config-processor` keys from `podSecurityContextSpec` as these only exist on a container level. They're kept in `containerSecurityContextSpec`
3. Split init-container security context includes from their app containers. So the security context for `ingest-init` is (and can now be) defined separately from the `ingest` container.
4. Added the missing security-context includes to `raw-reads-processing-deployment.yaml` as these were added after this initial pr.

I intentionally left out changes to the `loculus.configProcessor` helper as the same changes necessary are already included in #7072 

I did not add any more CI coverage here now to keep the PR small, and we can propose reasonable defaults for the security contexts afterwards. I did however test the security contexts locally and as an easy example setting the values below will result in several containers and init-containers not starting because `readOnlyRootFilesystem: true` prevents them from writing to their filesystems.. So we can confirm that this works as intended. Setting proper defaults will require a lot more testing though.
```yaml
podSecurityContext:
  default:
    runAsNonRoot: true
    runAsUser: 1000
    fsGroup: 2000
    seccompProfile:
      type: RuntimeDefault

containerSecurityContext:
  default:
    allowPrivilegeEscalation: false
    readOnlyRootFilesystem: true
    capabilities:
      drop: ["ALL"]
```

PR #7036, opened 2026-08-05
fix(website): keep Tools drop-down inside the viewport on mobile

### LINKED ISSUE #6988: Tools drop-down outside of viewport on mobile
Tools drop-down is mostly outside of viewport on mobile and one can't scroll it into view either.

Image

corneliusroemer:
Claude-created PR to fix #6988:

The Tools drop-down panel was positioned with plain CSS (`absolute right-0 w-64`), so its right edge was pinned to the right edge of the Tools button. On narrow viewports the button wraps onto its own line at the left of the toolbar, which put most of the 256px-wide panel off the left edge of the screen - and because the overflow is to the left of the document origin, it cannot be scrolled into view.

Switch the panel to Headless UI's `anchor` prop, which positions it with floating-ui. It still prefers right-alignment under the button (`bottom end`, 8px gap), but now shifts and flips to stay inside the viewport and caps its width/height to the available space, scrolling internally instead of running past the fold.

Measured in Chromium with the surrounding toolbar reproduced at a 360px mobile viewport:

  before  x=-131.9  right=124.1  bottom=908  (viewport 360x560)
  after   x=8.0     right=264.0  bottom=552, panel scrolls internally

Desktop positioning is unchanged (identical box at 768px). Also adds `z-50`, matching the other menus in the codebase.

Fixes #6988

### Screenshot

Correct positioining now on mobile:
Google Chrome 2026-08-05 15 57 32

And still fine on desktop:
Google Chrome 2026-08-05 15 57 42

🚀 Preview: https://fix-tools-dropdown-mobile.loculus.org

### DIFF
--- a/website/src/components/SearchPage/DownloadDialog/LinkOutMenu.tsx
+++ b/website/src/components/SearchPage/DownloadDialog/LinkOutMenu.tsx
@@ -215,7 +215,15 @@ export const LinkOutMenu: FC<LinkOutMenuProps> = ({
                     <IwwaArrowDown className='ml-2 h-5 w-5' aria-hidden='true' />
                 </MenuButton>
 
-                <MenuItems className='absolute right-0 mt-2 w-64  origin-top-right rounded-md bg-white shadow-lg ring-1 ring-black/5 focus:outline-hidden'>
+                {/*
+                 * `anchor` positions the panel with floating-ui, which keeps it inside the viewport:
+                 * it prefers right-alignment under the button, but shifts/flips and caps its size when
+                 * there isn't enough room - without it the panel hangs off-screen on narrow viewports.
+                 */}
+                <MenuItems
+                    anchor={{ to: 'bottom end', gap: 8, padding: 8 }}
+                    className='z-50 w-64 rounded-md bg-white shadow-lg ring-1 ring-black/5 focus:outline-hidden'
+                >
                     <div className='py-1'>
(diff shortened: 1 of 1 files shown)


tombch (approved):
Tested and this looks good to me! Thanks

website/src/components/SearchPage/DownloadDialog/LinkOutMenu.tsx
```diff
@@ -215,7 +215,15 @@ export const LinkOutMenu: FC<LinkOutMenuProps> = ({
                     <IwwaArrowDown className='ml-2 h-5 w-5' aria-hidden='true' />
                 </MenuButton>
 
-                <MenuItems className='absolute right-0 mt-2 w-64  origin-top-right rounded-md bg-white shadow-lg ring-1 ring-black/5 focus:outline-hidden'>
+                {/*
+                 * `anchor` positions the panel with floating-ui, which keeps it inside the viewport:
+                 * it prefers right-alignment under the button, but shifts/flips and caps its size when
+                 * there isn't enough room - without it the panel hangs off-screen on narrow viewports.
```
corneliusroemer, line 221:
```suggestion
                 * there isn't enough room - without it the panel hangs off-screen on narrow viewports.
                 * Prevents: https://github.com/loculus-project/loculus/issues/6988
```

PR #7049, opened 2026-08-07
chore(deps-dev): bump to playwright with new firefox version

maverbiest:
Clone of https://github.com/loculus-project/loculus/pull/7007 to see if Firefox integration test flakiness can be addressed.

Dependabot is trying to bump playwright from 1.61.1 to 1.62.1, which has Firefox version 153.0  (was 151.0 before: https://github.com/microsoft/playwright/releases).

The integration test for Firefox failed repeatedly and unpredictably on that branch on testing steps that create a new user through the Auth page, specifically when trying to check the 'I agree' check box (screenshots below). There is a playwright issue where people report clicking checkboxes in playwright as a source of flakiness (though not all exactly the same problem, see https://github.com/microsoft/playwright/issues/13470)

This PR wraps the `this.page.getByLabel('I agree').check()` in an `expect(...).toPass()` so it retries. While this does seem to resolve flakiness, the underlying cause is still not clear to me. Especially because we have different checkboxes that are `check()`-ed without retrying in the integration tests where this issue doesn't arise (although those are not in the keycloakify page).

### Screenshots

Error message in CI logs of https://github.com/loculus-project/loculus/pull/7007

grafik

#### Playwright screenshots

Action:

grafik

The error logs:

```
Error: locator.check: Clicking the checkbox did not change its state
Call log:
  - waiting for getByLabel('I agree')
    - locator resolved to <input type="checkbox" id="termsAccepted" name="termsAccepted" aria-invalid="false" class="kcCheckboxInputClass"/>
  - attempting click action
    - waiting for element to be visible, enabled and stable
    - element is visible, enabled and stable
    - scrolling into view if needed
    - done scrolling
    - performing click action
    - click action done
    - waiting for scheduled navigations to finish
    - navigations have finished
 ```

Before:

grafik

After:

grafik

(the blue square that's added seems to be locator added by playwright, not the actual page, `__playwright_target__ marker`)

### PR Checklist


corneliusroemer (approved):
Thanks for investigating and making a fix PR. Sounds reasonable to me, weird that we suddenly need it but so be it.

corneliusroemer:
I looked at the trace again and I noticed that when selecting "before" the click location is slightly to the right of the checkbox. Maybe that's why the check goes nowhere - some people on the linked issue suggested the click was in the wrong place.

image

Doesn't matter though if your workaround works.

maverbiest:
@corneliusroemer huh that's weird, is that from the failed run https://github.com/loculus-project/loculus/actions/runs/31168223702?pr=7007 ? Because when I looked at that trace the dot was right over the checkbox, also in "before":

grafik


I might just go ahead and merge this anyway since it seems to resolve the issue

PR #7050, opened 2026-08-10
refactor/fix(website): Block submissions proceeding while files are still uploading

### LINKED ISSUE #6766: MVP: Can submit before file uploads finish, silently dropping files

tombch:
resolves #6766, replaces #6855

### Summary

- Lifts `FileUploadState` out of each `FolderUploadComponent` and into a single state of `Map<FileCategoryName, FileUploadState>` for the submission/revision/edit components.
- From here, adds validation for all file categories to be in state `uploadCompleted` before proceeding with submission.

### Screenshot

Attempting (multiple times) to submit while a file is still uploading:

image
 
### PR Checklist

🚀 Preview: https://upload-state-refactor.loculus.org

### DIFF
--- a/website/src/components/Submission/FileUpload/FolderUploadComponent.spec.tsx
+++ b/website/src/components/Submission/FileUpload/FolderUploadComponent.spec.tsx
@@ -1,14 +1,17 @@
 import { render, screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { err, ok } from 'neverthrow';
+import { useState, type ComponentProps } from 'react';
 import { describe, expect, it, vi, beforeEach } from 'vitest';
 
 import { FolderUploadComponent } from './FolderUploadComponent';
 import { type FileMapping } from './fileMapping';
+import { type FileUploadState } from './fileUpload';
 import * as multipartUpload from '../../../utils/multipartUpload';
 
 const mockRequestMultipartUpload = vi.fn();
 const mockCompleteMultipartUpload = vi.fn();
+const mockOnError = vi.fn();
 
 vi.mock('../../../services/backendClient', () => {
     return {
@@ -31,8 +34,28 @@ vi.mock('../../../utils/multipartUpload', async () => {
     };
 });
 
-const mockSetFileMapping = vi.fn();
-const mockOnError = vi.fn();
+let fileMapping: FileMapping | undefined;
+
+const FolderUploadComponentWithState = ({
+    initialState,
+    initialMapping,
+    ...props
+}: Omit<ComponentProps<typeof FolderUploadComponent>, 'fileUploadState' | 'setFileUploadState' | 'setFileMapping'> & {
+    initialState?: FileUploadState;
+    initialMapping?: FileMapping;
+}) => {
+    const [fileUploadState, setFileUploadState] = useState(initialState);
+    const [mapping, setMapping] = useState<FileMapping | undefined>(initialMapping);
+    fileMapping = mapping;
+    return (
+        <FolderUploadComponent
+            {...props}
+            fileUploadState={fileUploadState}
+            setFileUploadState={setFileUploadState}
+            setFileMapping={setMapping}
+        />
+    );
+};
 
 const defaultProps = {
     fileCategory: {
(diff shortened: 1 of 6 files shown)


website/src/components/Submission/FileUpload/fileUpload.ts
```diff
@@ -0,0 +1,87 @@
...
+};
+
+export type Uploaded = {
+    type: 'uploaded';
+    fileId: string;
+    path: string;
+    size: number;
+};
+
+export type PreviousUpload = {
```
anna-parker, line 31: I just realized... does this state only exist for files that were uploaded beforehand and are stored in the db and thus dont actually have a path but a name?

website/src/components/Submission/FileUpload/fileUpload.ts
```diff
@@ -0,0 +1,87 @@
...
+};
+
+type UploadCompleted = {
+    type: 'uploadCompleted';
+    files: (Uploaded | PreviousUpload)[];
+};
+
+export type FileUploadState = AwaitingUrlState | UploadInProgressState | UploadCompleted;
+
+export const getInitialFileUploadStates = (files: FilesByCategory): Map<string, FileUploadState> =>
```
anna-parker: maybe we could make it clearer that this sets all submitted files to the state 'previousUpload', i.e. with `getPreviouslyUploadedStates`?

website/src/components/Submission/FileUpload/FolderUploadComponent.spec.tsx
```diff
@@ -43,59 +63,46 @@ const defaultProps = {
     accessToken: 'test-token',
     clientConfig: { backendUrl: 'http://test-backend', lapisUrls: {} },
     groupId: 1,
-    fileMapping: undefined,
-    setFileMapping: mockSetFileMapping,
-    onError: mockOnError,
+    onError: vi.fn(),
 };
 
-// The fileMapping is keyed by category, then by file path, to its file ID. Previous uploads have no
```
anna-parker, line 51: ah yes... I think renaming path to name for previous uploads would make this clear and then no comment would be needed!

website/src/components/Submission/FileUpload/fileUpload.ts
anna-parker: ah ok this is annoying to change to .name as the code always calls with .path and type aliases cant have getter and setter like in python dataclasses... so feel free to ignore this!

website/src/components/Submission/FileUpload/FolderUploadComponent.spec.tsx
anna-parker: see comment above

website/src/components/Submission/FileUpload/fileUpload.ts
anna-parker: but maybe we could comment here directly that for this type the path is a bare fileName

anna-parker (approved):
awesome thanks!

website/src/components/Submission/FileUpload/fileUpload.ts
tombch: Done!

website/src/components/Submission/FileUpload/fileUpload.ts
tombch: Yeah I've added some comments for this. I'm also now thinking this shouldn't need to be commented - we could probably have a discriminating union for file states that are form upload vs for bulk upload. Because if we are in form upload, we know all the files can be referred to by name only rather than needing a path. Something to follow up on!

PR #7070, opened 2026-08-13
fix(backend): clearer error for JWT without preferred_username, and resolve spring-boot-4 conflict with main

corneliusroemer-agent:
Follow-up to #6974, targeting its `spring-boot-4` branch. Two things: the review comment about the unsafe `!!` on the JWT username claim, and the merge conflict that branch currently has with `main`.

## 1. Clearer failure mode for a JWT without `preferred_username`

The Spring Security 7 `JwtAuthenticationToken` constructor takes a non-null `name`, which forced a `!!` in `KeycloakAuthenticationConverter`:

```kotlin
jwt.getClaimAsString(StandardClaimNames.PREFERRED_USERNAME)!!,
```

If a token ever arrives without the claim, that throws a bare `NullPointerException` inside the authentication converter — most likely surfacing as an unhandled 500 with nothing in the message pointing at the cause. Replaced with:

```kotlin
requireNotNull(jwt.getClaimAsString(StandardClaimNames.PREFERRED_USERNAME)) {
    "JWT missing preferred_username claim"
},
```

Same control flow, same non-null guarantee for the constructor — the only change is that the failure now names itself. Keycloak always sets this claim, so this is defensive rather than a fix for an observed bug.

Note there is a parallel unsafe cast at `AuthenticatedUser.kt:27` (`claims[PREFERRED_USERNAME] as String`) with the same property. Deliberately left alone here to keep this PR to what the review comment asked for.

## 2. Merge conflict with `main`

`spring-boot-4` currently conflicts with `main` on `backend/gradle.lockfile`. This PR merges `origin/main` in and resolves it.

The lockfile was **regenerated**, not hand-merged: `./gradlew dependencies --write-locks` against the merged `build.gradle`. With `lockAllConfigurations()` and `LockMode.STRICT`, a hand-stitched union of the two sides can easily be internally inconsistent, and STRICT mode only surfaces that at CI time.

Every version difference between the regenerated lockfile and the pre-merge `spring-boot-4` one traces to a dependabot bump that landed on `main`, with no unexplained drift:

| Dependency | before | after |
| --- | --- | --- |
| `org.jsoup:jsoup` | 1.22.2 | 1.23.1 |
| `org.keycloak:keycloak-admin-client` (+ `keycloak-client-common-synced`) | 26.0.11 | 26.0.12 |
| `software.amazon.awssdk:*` | 2.47.1 | 2.51.3 |

Each matches the corresponding change to `backend/build.gradle` coming from `main`. The merge also picks up main's Gradle wrapper bump (9.6.1 → 9.7.0).

**Because of the merge commit, the diff of this PR against `spring-boot-4` includes everything on `main` since that branch last synced** — not just the two backend changes above. The reviewable content is the `SecurityConfig.kt` hunk and the lockfile resolution.

## Verification

Run against the merged tree with a JDK 21 toolchain:

- `./gradlew compileKotlin compileTestKotlin ktlintMainSourceSetCheck` — BUILD SUCCESSFUL. Compiling test sources matters here: with STRICT locking, test configurations are where an incomplete lockfile blows up.
- The build left `gradle.lockfile` untouched afterwards, confirming the lock state is complete and stable rather than partially written.

The full `./gradlew test` suite was not run locally (Docker-dependent); CI covers it.


corneliusroemer-agent:
Superseded by #7071, which uses a branch in this repo instead of a fork so CI runs with full access.

PR #7071, opened 2026-08-13
fix(backend): return 401 for a JWT without preferred_username, and resolve spring-boot-4 conflict with main

corneliusroemer-agent:
Follow-up to #6974, targeting its `spring-boot-4` branch. Two things: the review comment about the unsafe `!!` on the JWT username claim, and the merge conflict that branch had with `main`.

## 1. A JWT without `preferred_username` now gives 401, not 500

Spring Security 7's `JwtAuthenticationToken` constructor takes a non-null `name`, which forced a `!!` in `KeycloakAuthenticationConverter`. This PR first used `requireNotNull`, which names the failure but keeps throwing `IllegalArgumentException` — still a 500. It now throws `InvalidBearerTokenException`, so the request is rejected as a bad credential:

```kotlin
jwt.getClaimAsString(StandardClaimNames.PREFERRED_USERNAME)
    ?: throw InvalidBearerTokenException("Token is missing the 'preferred_username' claim"),
```

Why the status code changes, from the Spring Security 7.1.0 bytecode:

- `JwtAuthenticationProvider.authenticate`'s exception table covers only offsets 0–13 — the `getJwt(...)` decode step. The converter runs at offset 16, **outside** that range, so whatever it throws propagates unwrapped.
- `BearerTokenAuthenticationFilter` catches only `AuthenticationException`. An `NPE`/`IllegalArgumentException` isn't one, so it escapes the filter chain → container 500.
- `InvalidBearerTokenException` extends `OAuth2AuthenticationException` extends `AuthenticationException` (not deprecated in 7.1.0), so it is caught → 401 with the `WWW-Authenticate: Bearer` header, via the existing `LoggingAuthenticationEntryPoint`.

On reachability: `preferred_username` is supplied by Keycloak's `profile` client scope, so a validly signed token issued without that scope carries no claim. Keycloak sets it in practice, so this is defensive — the point is the failure mode, not an observed bug.

No test added: the failure is a one-line guard and the trade-off wasn't judged worth an `@EndpointTest` context. Note `AuthenticatedUser.kt:27` has the same unchecked assumption (`claims[PREFERRED_USERNAME] as String`), deliberately left alone here.

## 2. Merge conflict with `main`

`spring-boot-4` conflicted with `main` on `backend/gradle.lockfile`. Resolved by merging `origin/main` and **regenerating** the lockfile (`./gradlew dependencies --write-locks`) rather than hand-stitching it — with `lockAllConfigurations()` and `LockMode.STRICT`, a hand-merged union can be internally inconsistent in ways that only surface at CI time. Every version difference traces to a dependency bump already on `main` (keycloak-admin-client 26.0.11→26.0.12, awssdk 2.47.1→2.51.3, jsoup 1.22.2→1.23.1).

Verified the regeneration preserved the Jackson pinning from #6974: `spring-boot-jackson` still absent, `spring-boot-jackson2:4.1.0` still present.

## Note on the Jackson 2/3 mix (context for the `build.gradle` TODO)

While reviewing #6974: Jackson 3 (`tools.jackson:3.1.4`) is on the runtime classpath despite the `spring-boot-jackson` exclusion. The exclusion is not at fault — it works as intended. `gradlew dependencyInsight` shows the source is **Flyway 12**, which migrated to Jackson 3 (12.0.0 is the last Jackson 2 release):

```
tools.jackson.core:jackson-databind:3.1.1 -> 3.1.4
\--- org.flywaydb:flyway-core:12.4.0
```

So excluding the `tools.jackson` group isn't an option — it would break DB migrations at startup. Nor is downgrading Flyway attractive: pinning schema-migration tooling four minors behind what Boot 4.1 manages, to avoid a duplicate JSON library, then unpinning it again at the Jackson 3 migration.

This matters because Spring MVC 7 selects default converters by probing the classpath (`ClassUtils.isPresent("tools.jackson.databind.ObjectMapper")` in `WebMvcConfigurationSupport`), and Flyway makes that probe true. What keeps Jackson 2 serialising the API is auto-configuration precedence — `spring-boot-jackson2` contributing the converter while `spring-boot-jackson` is excluded — not an unambiguous classpath, which isn't achievable here. Verified empirically that Jackson 2 is winning today: released-data JSON from the #6974 preview is identical to `main` apart from environment-specific values.

No change proposed here; recorded so #5753 has the full picture.

## Testing

`./gradlew compileKotlin compileTestKotlin ktlintCheck` passes, and the full backend suite passes locally against a real Postgres + Minio via `USE_NONDOCKER_INFRA=true`: **567 tests across 74 classes, 0 failures, 1 skipped**. That includes the existing `GIVEN invalid authorization token WHEN performing action THEN returns 401 Unauthorized` case, i.e. the ordinary bad-token path still behaves as before.

## Prompts behind this PR

For reference, the prompts that produced this work (verbatim):

1. `review this pr in depth for me: https://github.com/loculus-project/loculus/pull/6974`
2. `can you check i made a pr in 7071 does that fix 1. correctly? [...] what do you think about jackson-3 autoconfig? how to best avoid this mix of versions in your view?`
3. `or should we for now go back with flyway to a version that uses jackson2?`
4. `can you do 7071 the way you think is best?` / `just push to that pr`
5. `i don't think we need to test the exception it's ok`

## PR Checklist
- [x] Any manual testing that has been done is documented (see Testing above).


tombch (approved):
This looks good to me, thanks!

If I understand correctly, the PR comment about Jackson 2 and 3 is basically saying that Jackson 3 _is_ pulled in and used by flyway 12, but Jackson 2 is still unambiguously what is being used in our app (i.e. for converting from/to kotlin objects to http requests/responses, serialising to JSONB columns, etc)?

corneliusroemer:
> If I understand correctly, the PR comment about Jackson 2 and 3 is basically saying that Jackson 3 is pulled in and used by flyway 12, but Jackson 2 is still unambiguously what is being used in our app (i.e. for converting from/to kotlin objects to http requests/responses, serialising to JSONB columns, etc)?

Yeah that's my understanding, Flyway pulls in Jackson 3 as well but Jackson 2 takes preference for us.

PR #7072, opened 2026-08-13
feat(helm): allow setting custom container repositories for all containers

### LINKED ISSUE #7016: [Helm] Allow configuring custom container registries for all containers
In many enterprise and secure Kubernetes environments (especially where the security department wields the scepter), clusters are often blocked from pulling container images directly from public registries like Docker Hub, Quay, or GHCR.

Instead, all images must be pulled from an internal container registry (e.g., Azure Container Registry, AWS ECR, Harbor). These internal registries typically act as pull-through caches or internal mirrors that run required vulnerability scans before allowing workloads to execute.

Currently, image registries are partially hardcoded or lack a flexible global registry override mechanism, making enterprise deployments cumbersome without manually modifying chart templates.

I propose extending the existing `images:` configuration block in `values.yaml` (and consolidating `imagePullPolicy` under `images.pullPolicy`) to support a custom global registry prefix along with per-image overrides.

This pattern supports four deployment scenarios:

1. **Public registries (Default - zero breaking changes):**
`images.registry: ""` renders default upstream repository URIs as-is.
2. **Pull-through caches (ACR / ECR / Harbor):**
Setting a global `images.registry: "myacr.azurecr.io"` prepends the registry host to the full upstream image path (e.g., `myacr.azurecr.io/ghcr.io/loculus-project/backend:latest`).
3. **Flat internal registries (Re-pushed images):**
Overriding component `repository` fields strips upstream domain hosts when enterprise pipelines mirror images to flat internal namespaces.
4. **Mixed / Global with per-image exemptions:**
Component-level `registry: ""` overrides allow specific images to bypass the global registry prefix when necessary.

```yaml
# Proposed values.yaml structure
images:
  pullPolicy: Always
  registry: "myacr.azurecr.io"  # Global pull-through cache prefix
  pullSecrets: []
  
  backend:
    repository: ghcr.io/loculus-project/backend
  
  keycloak:
    registry: ""  # Exemption: pulls directly without the global prefix
    repository: quay.io/keycloak/keycloak
    tag: "23.0"

  website:
    repository: "ghcr.io/loculus-project/website"
    tag: "commit-f339e6c"
    pullPolicy: IfNotPresent
```

Would this structure work for you? If yes, I can open a PR for you to review in the coming days

florianzwagemaker:
resolves #7016

This PR introduces the ability to set custom container repository overrides for all used containers throughout the Loculus helm chart.
This will allow Loculus to be deployed in Kubernetes environments where clusters are blocked from pulling container images directly from public registries (such as DockerHub, Quay, or GHCR), and instead require images to be pulled from an internal registry (such as Azure ACR, or AWS ECR) that can act as a pull-through cache or internal mirror.

Previously, most image repositories were hardcoded in the chart templates, making such deployments cumbersome without manually modifying the chart.  

Here we introduce a new `loculus.image` helper in `_loculus-docker-tag.tpl` that resolves the full `repository:tag` structure for a named image in the `images:` block, falling back to defaults when no override is provided. Additionally, a minimal `loculus.imagePullPolicy` helper is added in the same file that will ensure safe lookup of `imagePullPolicy` overrides and fallback behavior.
All relevant deployment templates are updated to use the new helpers, keeping default container repositories/tags as is but allowing to be overridden where necessary.

> [!NOTE]  
> Images (and templating files) for the **preprocessing** and **ingest** are untouched as these could already be changed on a per-organism basis elsewhere in the `values.yaml` file.
> However, please note that autoapprove (which runs the ingest container image) is now exposed as its own `images.autoapprove` key and is not tied to the per-organism ingest image.

An example `images:` block containing *every* container can be seen below (ex preprocessing/ingest), here also some keys are changed to a custom value for illustrative purposes. In the actual `values.yaml` in this PR, the `images:` block has not been changed.

```yaml
images:
  loculusSilo:
    repository: "ghcr.io/loculus-project/loculus-silo"
    pullPolicy: Always
  lapis:
    repository: "ghcr.io/genspectrum/lapis"
    tag: "v0.8.0@sha256:4da71d94aceb36b7ea5963aa2e7b0566567c6423fe7fdd39448e69aebda2a8bb"
    pullPolicy: Always
  website:
    repository: "ghcr.io/loculus-project/website"
    pullPolicy: IfNotPresent
  backend:
    repository: "ghcr.io/loculus-project/backend"
    pullPolicy: IfNotPresent
  configProcessor:
    repository: "ghcr.io/loculus-project/config-processor"
  autoapprove:
    repository: "ghcr.io/loculus-project/ingest"
  docs:
    repository: "ghcr.io/loculus-project/docs"
  enaSubmission:
    repository: "ghcr.io/loculus-project/ena-submission"
  enaSubmissionFlyway:
    repository: "ghcr.io/loculus-project/ena-submission-flyway"
  keycloakify:
    repository: "ghcr.io/loculus-project/keycloakify"
  taxonomyService:
    repository: "ghcr.io/loculus-project/taxonomy-service"
  rawReadsProcessing:
    repository: "ghcr.io/loculus-project/raw-reads-processing-service"
  keycloak:
    repository: "custom_registry.io/keycloak/keycloak"
    tag: "23.0"
  postgres:
    repository: "custom_registry.io/postgres/postgres"
    tag: "15.12"
  minio:
    repository: "minio/minio"
    tag: "latest"
  alpine:
    repository: "alpine"
    tag: "3"
  busybox:
    repository: "busybox"
    tag: "latest"
  kubectl:
    repository: "alpine/kubectl"
    tag: "1.36.0"
```

within the templates, the new helper is used as follows with `lapis` as an example:

```yaml
- name: lapis
  image: {{ include "loculus.image" (dict "name" "lapis" "defaultRepository"
"ghcr.io/genspectrum/lapis" "values" $.Values) }}
  imagePullPolicy: {{ include "loculus.imagePullPolicy" (dict "name" "lapis" "values" $.Values) }}
```

Here, the default container repository is already specified (ghcr.io/genspectrum/lapis) so it is no longer a requirement to provide this in `values.yaml`.  
All values set within `values.yaml` now function purely as an override of the default values set within the relevant template(s).

The tag is taken from the `values.yaml`, but if none is present it will fall back to the `loculus.dockerTag` helper which will default to 'latest' or the sha-commit.
(for lapis specifically, this is set in `values.yaml` to `v0.8.0@sha256:4da71d94aceb36b7ea5963aa2e7b0566567c6423fe7fdd39448e69aebda2a8bb`)

The imagePullPolicy, through the helper, will always first look if there's an override in the `images:` block, and if not, uses the default imagePullPolicy property.

This is the default behavior for all loculus-owned containers.
However, for external containers (e.g. keycloak) the helper is used as follows:

```yaml
- name: keycloak
  image: {{ include "loculus.image" (dict "name" "keycloak" "defaultRepository" "quay.io/keycloak/keycloak" "defaultTag" "23.0" "values" $.Values) }}
```

Here a `defaultTag` is also provided within the template itself. This will be used as the fallback when no override is provided within the `values.yaml`.

The `images:` block in `values.schema.json` has been extended to cover all containers, and repository/pullPolicy are no longer *required* in an image spec since defaults are provided by the helper.

The `images:` block in `values.yaml` is unchanged, so this is not a breaking change and existing deployments will render identical manifests.


florianzwagemaker:
Below a brief list of changes based on the initial claude review comments in https://github.com/loculus-project/loculus/pull/7116

1. removed the stale `dockerTag` parameter from the `loculus.configProcessor` template. As well as cleanup of the readme and the various caller templates.
2. Added descriptions to `values.schema.json` so defaults are more easily discoverable.
3. containers for raw reads processing (i.e. deaconIndex) were added after this initial pr so they're included in the pattern as well now.
4. Added the `imagePullPolicy` keys to all and includes to all external/helper/init containers. With this all used containers (except per-pathogen configurations (prepro/ingest)) follow the same pattern.
5. Added a `defaultTag` for Lapis in the template caller (`lapis-deployment.yaml`) which defaults to version `0.8.0`. The version with the full hash is still present in `values.yaml` so this is the versiontag that will actually be used in real deployments that use this values file. However, if this is actively omitted, the Lapis container will at least fall back to version `0.8.0` instead of `latest`, a branch-name, or a commit-hash.

PR #7075, opened 2026-08-14
feat!(backend, preprocessing): lease-based preprocessing work queue

clonder:
### What

Before, every worker tried to claim sequences by inserting rows. With 64 workers, about half of ~1M claim attempts lost the race, and the contention made requests so slow that workers hit their 30s timeout and crash-looped (101 restarts in the baseline run). 

Now every sequence is already in the queue as an `UNPROCESSED` row. A worker
claims a batch atomically (`FOR UPDATE SKIP LOCKED`) and gets a lease: a claim
ID + expiry time. If the worker dies, the lease expires and the batch goes
back to the queue. Submitting results requires the matching claim ID, so a
worker that lost its lease can't overwrite anyone else's work.

The second half of this PR batches result writes and defers the
`table_update_tracker` bump to once per batch. Without it, every queue state
change fired the tracker trigger, which updates one shared row, so all
workers serialized on that single row lock.

### Result (500k sequences, 64 workers)

- Claim attempts: **1,008,177 → 25,468**, zero lost races
- Worker restarts: **101 → 0** — first fully VALID run

| Stage | Before | Queue only | Queue + batched results |
|---|---|---|---|
| Upload | 2269s | 2309s | 2334s |
| Preprocessing (drain) | 2217s | 2145s | **832s** (2.7×) |
| Approval | 237s | 236s | 245s |
| LAPIS/SILO | 742s | 722s | 741s |
| **Total workload** | **5508s** | **5475s** | **4186s** |

### notes

- **Organism column on the queue table.** The alternative 
  is a join-based claim: filter on `sequence_entries.organism` inside the
  locking query. It works, but the claim then stops being a single index-only
  scan: every candidate row needs a heap fetch from `sequence_entries` before
  it can be locked or skipped, which couples claim path back to the
  main table, so this PR removes it. With the column, the
  claim (`... WHERE organism = ? AND pipeline_version = ? AND
  processing_status = 'UNPROCESSED' ... FOR UPDATE SKIP LOCKED LIMIT n`) is
  served entirely by one partial index. Other variants (dedicated queue table,
  partitioning by organism) still store organism next to the queue rows. Drift is not possible in practice: the enqueue trigger is the only writer of the column, organism never changes for an accession, and it is NOT NULL after migration. Same denormalization as proposed in #4584.
- **The `status_check` constraint** The queue's
  partial indexes filter on specific status values, so an unknown status
  written by app code would silently disappear from the queue. The CHECK turns
  that silent bug into an immediate error. A PostgreSQL enum would achieve the
  same: CHECK is easier to evolve in migrations (`NOT VALID` + `VALIDATE`,
  no enum ALTER restrictions) and performs identically.
- **A lease (timeout + claim ID).** Once a worker owns a batch exclusively,
  it can die, and those sequences would stay `IN_PROCESSING` forever.
  The lease reuses the existing
  `reset-stale-in-processing-after-seconds` setting (60s): an expired batch
  goes back to the queue, and the claim-ID check on submit rejects the old
  worker's late result, so a batch that outlives its lease is simply redone,
  never written twice. (Main's stale-reset already requeues after 60s, but
  nothing checks ownership on submit, so a slow batch can be silently
  processed and written twice today.) In the measured
  runs the mean batch took ~10s on a fully saturated host, so expiry is
  rare.

### Breaking change

External preprocessing pipelines must send back the `processingAttemptId`
they received when claiming work. Migrations V1.35–V1.38.

Benchmark data included (runs 03, 04).

Note: the `results/` files are here only so the numbers can be checked during review - they will be removed before merging.

### PR Checklist
  - TODO update preprocessing/specification.md
  - benchmark runs in the PR


kubernetes/loculus/templates/loculus-backend.yaml
```diff
@@ -9,6 +9,8 @@ metadata:
     argocd.argoproj.io/sync-options: Replace=true
 spec:
   replicas: {{$.Values.replicas.backend}}
+  strategy:
+    type: Recreate
```
fhennig, line 13: Why was this done? Maybe for rolling out this particular change we don't want to do a rolling update, but generally we still can, can't we?

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -1203,8 +1498,19 @@ class SubmissionDatabaseService(
...
+            },
+        ) {
+            it[processingStatusColumn] = UNPROCESSED.name
+            it[processingAttemptIdColumn] = null
+            it[leaseUntilColumn] = null
+            it[startedProcessingAtColumn] = null
+            it[finishedProcessingAtColumn] = null
+            it[processedDataColumn] = null
+            it[errorsColumn] = null
+            it[warningsColumn] = null
```
fhennig, line 1513: We have a pre-existing issue here. 

We're updating this for all preprocessing versions. If someone does an edit mid-version-change, then any already running preprocessing will not be able to submit for this sequence.

AFAICT the preprocessing version is just not part of the editedSequenceEntryData so I see no obvious quick fix for this.

Maybe we could create an issue instead? It's also not urgent but a wart we could try fixing.

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -181,96 +221,170 @@ class SubmissionDatabaseService(
...
+        }
 
-        table.batchInsert(sequenceEntries, ignore = true, shouldReturnGeneratedValues = false) {
-            this[table.accessionColumn] = it.accession
-            this[table.versionColumn] = it.version
-            this[table.pipelineVersionColumn] = pipelineVersion
-            this[table.processingStatusColumn] = IN_PROCESSING.name
-            this[table.startedProcessingAtColumn] = now
+        val currentTransaction = TransactionManager.current()
+        currentTransaction.exec("LOCK TABLE sequence_entries IN SHARE ROW EXCLUSIVE MODE")
```
fhennig, line 291: the `LOCK TABLE` is potentially an issue. It locks this very central table, for the whole transaction, which is the duration of the whole HTTP call. This could potentially take quite long, and during it no edits can be made.

I'm not entirely sure how "bad" this issue potentially is, what do you think? It could be a couple of seconds right, where no one can do any edits on the whole site, if I got it right.

preprocessing/nextclade/src/loculus_preprocessing/backend.py
```diff
@@ -148,20 +146,18 @@ def fetch_unprocessed_sequences(
         f"request id: {response.headers.get('x-request-id')}"
     )
     match response.status_code:
-        case HTTPStatus.NOT_MODIFIED:
```
fhennig, line 151: This case doesn't exist anymore. Now we always call `fetchUnprocessedEntriesAndUpdateToInProcessing` in the backend - What is the idle load like?

The function gets hit constantly, and actually does aquire DB locks etc; it's not significantly more expensive. And it gets hit every second by multiple pods at the same time.

backend/src/main/resources/db/migration/V1.35__expand_preprocessing_work_queue.sql
```diff
@@ -0,0 +1,40 @@
...
+    ADD COLUMN organism TEXT,
+    ADD COLUMN processing_attempt_id UUID,
+    ADD COLUMN lease_until TIMESTAMP WITHOUT TIME ZONE,
+    ALTER COLUMN started_processing_at DROP NOT NULL;
+
+UPDATE sequence_entries_preprocessed_data sepd
+SET organism = se.organism
+FROM sequence_entries se
+WHERE se.accession = sepd.accession
+  AND se.version = sepd.version;
```
fhennig, line 11: Some of the other migrations like 1.28 include extensive comments on what was changed and why, I think it'd be good to have at least some description here on each particular change as to why it was done.

For example I'm not sure right away why we now have the `organism` column (it's probably sensible!), I think it'd be nice to have the changes documented here from the perspective of the database.

backend/docs/db/schema.sql
theosanderson, line 1: Thanks a lot for the PR - looks an important thing to address - we should probably get a few sets of eyes on it since it touches a bunch of places

backend/src/main/resources/db/migration/V1.38__defer_preprocessed_data_tracker.sql
```diff
@@ -0,0 +1,29 @@
...
+        timezone('UTC', CURRENT_TIMESTAMP)
+    FROM changed_rows
+    WHERE changed_rows.processing_status = 'PROCESSED'
+    GROUP BY changed_rows.organism, changed_rows.pipeline_version
+    ON CONFLICT (table_name, organism, pipeline_version)
+    DO UPDATE SET last_time_updated = timezone('UTC', CURRENT_TIMESTAMP);
+
+    RETURN NULL;
+END;
+$$ LANGUAGE plpgsql;
```
fhennig, line 29: What is this doing?

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -457,40 +586,174 @@ class SubmissionDatabaseService(
...
+                "Processed result batch contains duplicate accession version " +
+                    duplicate.key.displayAccessionVersion(),
+            )
+        }
+
+        val maximumRowsPerStatement =
+            (org.loculus.backend.utils.DatabaseConstants.POSTGRESQL_PARAMETER_LIMIT - 10) /
+                STORED_RESULT_PARAMETERS_PER_ROW
+        storedResults.chunked(maximumRowsPerStatement).forEach { chunk ->
+            updateProcessedDataChunk(
```
fhennig, line 641: Could we use `chunkedForDatabase` here instead?

fhennig (changes requested):
First round, I think I'll have to look more

emmahodcroft:
Thank you for working on this Stacy!
Agree with Theo on getting more eyes on it. 

Just one thing for next time - for stuff at this level (breaking changes, schema/architecture rework), it's good to open an issue first, before digging into deep work, so there's a chance for the team to weigh in on the approach itself in case there's any question about how to introduce such changes/weighups of different ways to do things. In addition, it can easy and good to post the issue in #team-developers to ensure people see it and can weigh in if interested. Many times there may not be any reply or comment - and that's fine! But it's a chance to ensure people can weigh in if they do have any thoughts/concerns - before you get too deep into implementing, which can be frustrating!

chaoran-chen:
Hi @theosanderson and @emmahodcroft, thanks for the comments. We have not intended to merge the PR right now but created it to facilitate review (first by Felix and/or me) and marking it as ready-for-review allows Claude to review as well. Sorry that this was not commuicated clearly enough and that is on me (I told Stacy to mark it as ready).

The code is a direct output of the investigations of removing bottlenecks to submitting one million sequences and has been part of the branch https://github.com/loculus-project/loculus/tree/perf/million-upload-baseline (which has been mentioned a few weeks ago in our meeting) and for this specific task, I do think that having the code and the performance measurements is essential for the discussion. (By the way, I think the performance improvement is quite impressive: lowering the preprocessing time from 2217s to 832s; this does not mean that we should merge this immediately and there are probably also other approaches to address the bottleneck, but I think it is worth acknowledging!)

kubernetes/loculus/templates/loculus-backend.yaml
clonder: was only for this rollout. old pods claim by inserting IN_PROCESSING without organism and the lease fields, so they break after the migration. will fix it, make it a values flag defaulting to rolling update.

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
clonder: agreed, i don't see a quick fix either and this pr doesn't make it worse. will open an issue for it.

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
clonder: yeah you're right. measured it: 12461 ms of blocked writes at 500k, and the tx also covers the claim and the whole streaming. it only happens once per new pipeline version, not per request, but still bad. will fix it - lock only around the one insert, backfill outside of it.

preprocessing/nextclade/src/loculus_preprocessing/backend.py
clonder: no idle number yet, will measure it properly. one thing to watch if we ever put the 304 back: with the update trigger gone a requeued stale lease doesn't bump the tracker, so a worker would sit on 304 forever and never pick those up. will add backoff on empty polls instead.

backend/src/main/resources/db/migration/V1.35__expand_preprocessing_work_queue.sql
clonder: yeah fair, will fix it - organism, check vs enum, the indexes and the lease columns all get documented in the migration.

backend/src/main/resources/db/migration/V1.38__defer_preprocessed_data_tracker.sql
clonder: sorry, two things in one file. it drops the per statement insert/update tracker triggers - each one upserted the same single row, so all workers queued on that row lock - and rewrites the function to read the new organism column instead of joining sequence_entries. the delete trigger from v1.28 stays and still uses it. this part is where the speedup actually was, so i'm splitting it out into its own pr. will fix the comments there.

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
clonder: yes, will fix it. needs a small adapter since it returns List<R> and the loop returns Unit. 6 params per row is right.

clonder:
split this one. the queue alone moved the drain from 2217 s to 2145 s, batching the
result writes moved it to 832 s, so the speed is in the cheap half and it needs
nothing breaking. that half is now #7133.

what stays here is the lease and the claim id, which is really a correctness fix.
the batching commits and results/ files here go away when i rebase.

PR #7076, opened 2026-08-14
perf(preprocessing): set Nextclade jobs

clonder:
Makes the number of Nextclade jobs per preprocessing pod configurable via `nextclade_jobs` (was hardcoded to 1, default stays 1). Per profiling the gain is marginal and only shows without a CPU limit, but it does not hurt.

### PR Checklist
  - values.schema.json
  - 500k sequence runs with 32 and 64 pods, jobs 1 and 4


benchmarks/preprocessing-large-queue/full-run/results/05-nextclade-pods64-jobs4/500k-20260801T091408Z/pg-stat-statements.tsv
anna-parker, line 1: I don't understand what these files are and how they can be used in future - I am not sure they should be committed into the main repo?

anna-parker: It would be great to add a README describing how to run these tests. The files here appear to include machine specific results. I dont think the exact results should be committed, only files necessary to run the tests if we think they could be useful in future.

anna-parker: I'm also a bit confused by your description what are the 64 pods? Are these 64 preprocessing pods processing data from the same organism?

corneliusroemer:
Per profiling results this has only marginal benefits - but also doesn't harm.

I'd cut the tests and profiling files, this is a very simple change with low consequences if broken that the extra test lines are not worth it IMHO.

More jobs don't really help as long as there's a CPU limit - it only helps if there's no limit and if Nextclade is actually rate limiting which it doesn't really seem to be.

Issue #2683, opened 2024-09-01
Make max memory and max file size configured through one param


corneliusroemer:
Right now we set max memory for the backend and max upload size independently. 

This way they can become out of sync, as seen in https://github.com/loculus-project/loculus/issues/1899

We should tie the two configs into one - or at least add clear comment referring to each other where they are set.

chaoran-chen:
I added the discussion label as it is not clear to me what exactly should be done, whether it is possible to do in an accurate/reliable way and whether it is a backend, kubernetes-specific or documentation task.

corneliusroemer:
There are two things:

1. For now we should document clearly in the config files where the values are set that max file upload size and backend max memory need to be adjusted in sync
2. We can possibly avoid the need to mention that if we set them both through one single value, but 1 should happen until 2 is done (if it's done)

Issue #2682, opened 2024-09-01
Confirm that backend respond with 413 if the upload file is too large (and backend doesn't die OOM)


corneliusroemer:
Follow-up from: https://github.com/loculus-project/loculus/issues/1899

Ideally this would be a test in the test suite - but manual is ok as a starter.

Issue #1899, opened 2024-05-14
Backend should send 4XX HTTP error instead of 500 when client uploads too large file


corneliusroemer:
When the backend approaches the hard memory limit it is set by kubernetes because it receives a large request, it should return an appropriate HTTP error saying "payload too large" as opposed to responding with HTTP 500.

Ideally, we should also communicate the rough max size that the backend can accept.

Repro:
- Upload all of mpox (7k sequences) to a backend limited to 1GB RAM
- Observe result

Example logs:

<details>

```
2024-05-14 17:32:24,265 ERROR [http-nio-8079-exec-3] [mpox] - org.loculus.backend.controller.ExceptionHandler: Caught unexpected exception: Handler dispatch failed: java.lang.OutOfMemoryError: Java heap space
jakarta.servlet.ServletException: Handler dispatch failed: java.lang.OutOfMemoryError: Java heap space
	at org.springframework.web.servlet.DispatcherServlet.doDispatch(DispatcherServlet.java:1104)
	at org.springframework.web.servlet.DispatcherServlet.doService(DispatcherServlet.java:979)
	at org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:1014)
	at org.springframework.web.servlet.FrameworkServlet.doPost(FrameworkServlet.java:914)
	at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:590)
	at org.springframework.web.servlet.FrameworkServlet.service(FrameworkServlet.java:885)
	at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:658)
	at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:206)
	at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:150)
	at org.apache.tomcat.websocket.server.WsFilter.doFilter(WsFilter.java:51)
	at org.apache.catalina.core.ApplicationFilterChain.internalDoFilter(ApplicationFilterChain.java:175)
	at org.apache.catalina.core.ApplicationFilterChain.doFilter(ApplicationFilterChain.java:150)
	at org.springframework.web.filter.AbstractRequestLoggingFilter.doFilterInternal(AbstractRequestLoggingFilter.java:289)
	at org.springframework.web.filter.OncePerRequestFilter.doFilter(OncePerRequestFilter.java:116)
[...]

PR #7077, opened 2026-08-14
fix(backend/deployment): increase backend max ram to -XX:MaxRAMPercentage=75 up from 25

clonder:
### What

The backend container has 3Gi memory limit but the JVM heap defaults to
25% of RAM (~768 MiB), so most of memory reserved for the pod could never
actually be used. This sets `-XX:MaxRAMPercentage=75`, giving the backend
~2.3Gi of heap.

### PR Checklist
  - one flag
  - deployment config
  - benchmark runs 05-10 ran with it

🚀 Preview: https://chore-backend-heap.loculus.org


kubernetes/loculus/templates/loculus-backend.yaml
```diff
@@ -98,7 +98,7 @@ spec:
             {{- end }}
           env:
             - name: JVM_OPTS
-              value: -XX:+UseContainerSupport -XX:+UseG1GC -XX:MaxHeapFreeRatio=5 -XX:MinHeapFreeRatio=2
+              value: -XX:+UseContainerSupport -XX:MaxRAMPercentage=75.0 -XX:+UseG1GC -XX:MaxHeapFreeRatio=5 -XX:MinHeapFreeRatio=2
```
chaoran-chen: With the heap size increasing, the available RAM for non-heap would be smaller, right? Could this cause issues under some circumstances?

fhennig (approved):
If you've been testing with this configuration and it's fine, I'd say we can merge it!

kubernetes/loculus/templates/loculus-backend.yaml
clonder: Yeah if something needs a lot of direct buffers, 750Mi could get tight and then the pod just gets OOM killed, so we could lower the percentage or bump the 3Gi limit if that happens. Didnt see that in the 500k benchmark run but its not real traffic though

corneliusroemer:
I remember I carefully tested the heap settings some time ago to reduce RAM usage - it'd be good to check git blame or similar to find out what my reasoning was back then. The changes here could cause increase in RAM use.

This is the PR that set the current JVM args: 
- https://github.com/loculus-project/loculus/pull/1901

Also related:
- https://github.com/loculus-project/loculus/issues/1899
- https://github.com/loculus-project/loculus/issues/2683
- https://github.com/loculus-project/loculus/issues/2682

corneliusroemer:
We could make the JVM args configurable so that people can give as much RAM as they want.

With this PR here I see an increase in backend container memory usage on the preview by around ~300 MB up from ~400MB to 700MB. Across 16 previews that's +4.5 GB which we should be ok with but worth watching.

image

[Grafana URL](https://grafana.loculus.org/explore?schemaVersion=1&panes=%7B%22ly9%22:%7B%22datasource%22:%22decyzhq5ockqoe%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22container_memory_usage_bytes%7Bcontainer%3D%5C%22backend%5C%22,%20namespace%3D%5C%22prev-main%5C%22%7D%22,%22range%22:true,%22instant%22:true,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22decyzhq5ockqoe%22%7D,%22editorMode%22:%22builder%22,%22legendFormat%22:%22__auto%22,%22useBackend%22:false,%22disableTextWrap%22:false,%22fullMetaSearch%22:false,%22includeNullMetadata%22:true%7D,%7B%22refId%22:%22B%22,%22expr%22:%22container_memory_usage_bytes%7Bcontainer%3D%5C%22backend%5C%22,%20namespace%3D%5C%22prev-chore-backend-heap%5C%22%7D%22,%22range%22:true,%22instant%22:true,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22decyzhq5ockqoe%22%7D,%22editorMode%22:%22builder%22,%22legendFormat%22:%22__auto%22,%22useBackend%22:false,%22disableTextWrap%22:false,%22fullMetaSearch%22:false,%22includeNullMetadata%22:true%7D%5D,%22range%22:%7B%22from%22:%22now-30m%22,%22to%22:%22now%22%7D%7D%7D&orgId=1)

fhennig:
configurable sounds good!

PR #7078, opened 2026-08-14
perf(backend): insert initial data use terms in one statement

clonder:
### what

setting the initial data use terms for an upload was one insert per accession, so a
500k submission ran 500,000 single row inserts. the upload is one request served by
one backend thread, so database time there is close to wall time.

the rows are now inserted with one `INSERT ... SELECT` straight from
`metadata_upload_aux_table`, which is where the accessions already are. exposed
reports how many rows the statement wrote, so the count is checked against the
number of mapped rows and a mismatch aborts the transaction.

### numbers

two 500k runs in the same configuration, 32 preprocessing replicas and 4 nextclade
jobs:

| | before | after |
|---|---:|---:|
| inserting the terms | 500,000 statements, 1,927,529 ms | 1 statement, 3,011 ms |
| upload stage | 2352 s | 350 s |

both are single runs, and the measured code issued the same statement through a raw
prepared statement rather than through exposed, so read these as evidence for the
set based approach rather than for this exact commit.

### notes

`setNewDataUseTerms` authorized the groups of the affected entries in chunks, which
the set based insert cannot do. rows of an original upload all carry the submitted
group, so the group is authorized once for the whole upload instead.

the audit entry now records the upload id and the row count instead of every
accession, and the aux rows are deleted afterwards, so the accessions are no longer
recoverable from the audit log alone. that is a deliberate trade, happy to change it
if you would rather keep the list.

this pr was rebased. it used to sit on the branch of #7074, which was closed in
favour of #7096, so it now holds only the data use terms change against main. the
benchmark result files that were in it are gone, and the raw prepared statement was
replaced by exposed's own insert from select.

### PR Checklist
  - no user facing or api change
  - `UploadDatabaseServiceTest`, one test for the happy path and one for the count mismatch rollback
  - the two benchmark runs above, plus the backend suite locally

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/service/datauseterms/DataUseTermsDatabaseService.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/service/datauseterms/DataUseTermsDatabaseService.kt
@@ -1,7 +1,12 @@
 package org.loculus.backend.service.datauseterms
 
+import org.jetbrains.exposed.v1.core.QueryParameter
 import org.jetbrains.exposed.v1.core.eq
+import org.jetbrains.exposed.v1.core.stringParam
+import org.jetbrains.exposed.v1.datetime.dateTimeParam
 import org.jetbrains.exposed.v1.jdbc.batchInsert
+import org.jetbrains.exposed.v1.jdbc.insert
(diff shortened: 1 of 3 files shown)


fhennig (approved):
LGTM!

PR #7080, opened 2026-08-17
fix(website): Add display names for output file categories in review page

tombch:
### Summary

- Adds display names for output file categories on the review page.
- Removes the `filesEnabled` prop as this is now redundant with file categories passed down (it was computed as `length of file categories > 0`).

#### Before

image

#### After

image

### PR Checklist


website/src/components/ReviewPage/ReviewCard.tsx
```diff
@@ -126,7 +128,12 @@ export const ReviewCard: FC<ReviewCardProps> = ({
                 dataToView={data}
                 referenceGenomesInfo={referenceGenomesInfo}
             />
-            <FilesDialog isOpen={isFilesDialogOpen} onClose={() => setFilesDialogOpen(false)} dataToView={data} />
+            <FilesDialog
```
anna-parker: I guess this should only exist when `filesEnabled` is true? then `fileCategories` will never be null in the FilesDialog? (if `filesEnabled` is false then the files button doesn't show up even)

anna-parker (approved):
generally looks good - only one comment/suggestion!

website/src/components/ReviewPage/ReviewCard.tsx
tombch: Thanks, updated

PR #7082, opened 2026-08-18
feat(backend): scope the `extract-unprocessed-data` etag by organism

anna-parker:
resolves #

The `extract-unprocessed-data` endpoint is always called on an organism-level and this should be reflected in the etag. 

Additionally, when adding tests I noticed we do return the etag if the response is 304. I asked claude and this is apparently against convention, see https://datatracker.ietf.org/doc/html/rfc7232 (we worked around this by reusing the etag we used in the call):
```
    The server generating a 304 response MUST generate any of the
   following header fields that would have been sent in a 200 (OK)
   response to the same request: Cache-Control, Content-Location, Date,
   ETag, Expires, and Vary.
```

### PR Checklist


maverbiest (approved):
Looks good to me!

PR #7089, opened 2026-08-18
fix(backend): disallow revoking an already revoked sequence entry

### LINKED ISSUE #1946: You can revoke a sequence that's already been revoked

corneliusroemer-agent:
Resolves #1946

Now revocation gets blocked if latest version already revoked (even if not yet propagated to LAPIS):
image

<details>

## Background

This started from an observation that released data contains accessions with **several revocation versions stacked in a row**, with no non-revocation version in between. The question was whether the backend really permits that, since one would naively expect it not to.

It does. `SubmissionDatabaseService.revoke()` only required that the latest version of an accession be in status `APPROVED_FOR_RELEASE`. An approved revocation satisfies exactly that: in `sequence_entries_view` the status expression checks `released_at IS NOT NULL` **before** the `is_revocation -> 'PROCESSED'` branch, so once a revocation is approved it reports `APPROVED_FOR_RELEASE` like any other version. There was no `isRevocation` check anywhere in `AccessionPreconditionValidator`, and `git log -L` over the precondition block shows there never was one. So revocation-on-revocation was allowed by omission rather than by design, and no test covered it.

The website side of this (it offered the revoke button on earlier, non-revocation versions of an already revoked accession, which is the likely route by which the stacked revocations were created) is **split out separately**. This PR is backend-only.

## Changes

- New `andThatLatestVersionsAreNotRevocations()` in `AccessionPreconditionValidator`, modelled on the neighbouring `andThatSequenceEntriesHaveNoErrors()`, applied in `revoke()`. It is placed after the state check so an *unapproved* revocation still yields the clearer "not in state `APPROVED_FOR_RELEASE`" error.
- Selecting `is_revocation` in the two precondition queries is required, not defensive: Exposed throws on reading a column absent from the select list, and the new method lives on the shared `CommonPreconditions`.
- `REVOKE_DESCRIPTION` updated, since the precondition is part of the endpoint's documented contract.
- New test `WHEN revoking an already revoked entry THEN throws an unprocessableEntity error`.

Because `AccessionPreconditions` filters on `isMaxVersion`, this checks precisely "the latest version is a revocation", not the broader "this accession was ever revoked" — so revising after a revocation is unaffected. That is deliberate: `ReleasedDataModel` defines `REVISED` as "not the highest version, and no higher version is a revocation", i.e. the version-status logic already anticipates non-revocation versions above a revocation. Only stacking revocations was the hole.

## Verification

- Full backend suite: 74 classes, **581 tests, 0 failures**; `ktlintCheck` clean.
- The new test was confirmed to **fail against the pre-fix code** and pass after, so it is not vacuous.
- Checked that no existing E2E test breaks: all four revoke-touching specs in `integration-tests/` revoke exactly once on a fresh `.1`, so none now receives a 422. `SubmissionController` is the sole caller of the service method — no ingest or automated path revokes.

## Note on existing data

This prevents new stacked revocations; it does not clean up accessions that already have them. Those remain internally consistent — `computeVersionStatus` marks all but the last revocation `REVOKED` and the last `LATEST_VERSION` — so it is cosmetic noise in the release rather than corruption, but worth deciding separately whether to tidy.

</details>


website/src/components/SequenceDetailsPage/SequenceDataUI.tsx
```diff
@@ -59,6 +62,11 @@ export const SequenceDataUI: FC<Props> = ({
 
     const isMyGroup = myGroups.some((group) => group.groupId === groupId);
 
+    // Revoking always targets the latest version of an accession, not the version being displayed,
+    // so the revoke button must be hidden whenever the latest version is already a revocation.
+    // Revisions are unaffected: they are based on the latest non-revocation version.
```
corneliusroemer:
```suggestion
```

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -1017,6 +1017,7 @@ class SubmissionDatabaseService(
             thatAccessionsExist(accessions)
                 .andThatUserIsAllowedToEditSequenceEntries(authenticatedUser)
                 .andThatSequenceEntriesAreInStates(listOf(Status.APPROVED_FOR_RELEASE))
+                .andThatSequenceEntriesAreNotRevocations()
```
theosanderson: It's arguably partly an existing issue for `AreInStates`, but this naming is really confusing. A sequence entry is the thing referred to by a bare accession without a version. So it can't be a "revocation", it can only have already been revoked. So this should be `andThatLatestVersionsAreNotRevocations`

backend/src/main/kotlin/org/loculus/backend/service/submission/AccessionPreconditionValidator.kt
```diff
@@ -171,6 +173,25 @@ class AccessionPreconditionValidator(
             return this
         }
 
+        /**
+         * Rejects entries whose selected version is a revocation. Reached via [thatAccessionsExist],
+         * the selected version is the latest version of each accession.
+         */
```
corneliusroemer:
```suggestion
```

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
corneliusroemer: I've renamed as suggested, but worth  noting that the function only checks what it's given - it doesn't check it's the latest version. The reason it's given the latest version here is that `thatAccessionsExist` queries for the latest version. This is a bit of a latent footgun but not sure this PR is the right place to refactor.

theosanderson: yes, agree

tombch (approved):
Looks good to me!

corneliusroemer:
Thanks for approving and merging latest main @tombch - that was what I was about to do just to discover you'd done so already, great!

PR #7090, opened 2026-08-18
test(integration): ignore stochastic 502 console errors

corneliusroemer-agent:
We sometimes get `502 (Bad Gateway)` — probably via traefik, potentially when SILO/LAPIS gets new data (see https://github.com/GenSpectrum/LAPIS/issues/1699). Any Playwright test that happens to load a page in that window fails the console-error check in `console-warnings.fixture.ts`, so this shows up as stochastic CI flakiness unrelated to the change under test.

This adds the 502 message to the existing `harmlessMessages` filter list.


corneliusroemer:
(Probably) no longer necessary as #7092 should fix the root cause

Issue #5478, opened 2025-11-19
Export all cluster logs - e.g. traefik, not just loculus ones in integration tests


corneliusroemer:
Trying to figure out why we got 502 Bad Gateway in an integration test.

I can't see any traefik logs, for example, might be nice to export those as well.

PR #4469, opened 2025-06-20
test(website): fail on console warnings

theosanderson:
## Summary
- watch for warning level console messages in Playwright tests
- fail a test if any warnings were emitted

## Testing
- `npm run format`

------
https://chatgpt.com/codex/tasks/task_e_68555a044428832592e1df0c241a17da

PR #7092, opened 2026-08-18
fix(deployment): retire pooled traefik connections before backends close them

### LINKED ISSUE #5550: LAPIS (or traefik) occasionally responds with 502 Bad Gateway in integration tests
In integration tests we sometimes get a response of 502 Bad Gateway when calling LAPIS: https://github.com/loculus-project/loculus/actions/runs/19710238594/job/56468578070#step:19:91

We might want to not fail tests on this console error for now.

See related issues to improve logging to be able to figure out what causes those 502s:
- https://github.com/loculus-project/loculus/issues/5478
- https://github.com/loculus-project/loculus/issues/5479

As the 502 is transient, it might happen when SILO loads a new data version?

Image

corneliusroemer-agent:
Fixes #5550.

E2E tests have been plagued by occasional 502 Bad Gateway errors on LAPIS requests for a while. With help of Claude I've finally figured out why. Traefik kept idle HTTP connections alive for longer than LAPIS - so it could happen that traefik sends a POST request just before LAPIS tells it that it's terminating the connection. By the time the request arrives at LAPIS, LAPIS has shut down the connection and this causes traefik to then return 502 to the client.

Traefik would retry and this would self-heal if these requests were GET, but all of our data query requests are POST and those are by definition not retry-safe so traefik doesn't retry.

We can fix this in 2 ways: a) shorten the keep alive time on traefik to be below LAPIS or b) lengthen the keep alive of LAPIS. Either is doable but the advantage of a is that we can apply the same fix to all the other servers we have (backend, astro). For Astro in particular it would not be as simple as setting an argument/env variable in our Helm template, hence I ended up implementing a - the shortening of traefik's keep alive time.

The same 502 issue can happen to backend/astro - the reason we don't see it in E2E test CI is that there traefik only handles LAPIS requests. In real deployments, however, traefik sits in front of more requests - so it's good to fix it for all servers.

# Claude PR description below

Integration tests occasionally fail on a `502 (Bad Gateway)` console error from a LAPIS query. The guess in #5550 was that SILO was swapping in a new data version — [GenSpectrum/LAPIS#1699](https://github.com/GenSpectrum/LAPIS/issues/1699) was filed for that and closed as unreproducible, and I think that was right. In the run I looked at, LAPIS was healthy throughout: `Restart Count: 0`, readiness returning 200 in 3–35 ms either side of the failure, other queries served in the same second. The 502'd request never reached LAPIS at all.

It's traefik, and it's a timeout inversion. Traefik keeps idle upstream connections in a pool for [`idleConnTimeout`, default 90s](https://doc.traefik.io/traefik/routing/overview/#transport-configuration), inherited from Go's [`DefaultTransport`](https://github.com/golang/go/blob/go1.24.0/src/net/http/transport.go#L53). Our backends all drop idle connections far sooner, and nobody picked these numbers relative to each other:

- LAPIS + backend (Tomcat): **20s** — [`keepAliveTimeout` defaults to `connectionTimeout`](https://tomcat.apache.org/tomcat-10.1-doc/config/http.html)
- website (Astro → Node): **5s** — [`server.keepAliveTimeout`](https://nodejs.org/api/http.html#serverkeepalivetimeout)
- taxonomy / raw-reads / ena-deposition (uvicorn): **5s** — [`timeout_keep_alive`](https://github.com/encode/uvicorn/blob/0.52.4/uvicorn/config.py#L229)

So traefik reuses a connection that Tomcat is closing at that instant, Go reports `io.EOF`, and traefik [turns that into a 502](https://github.com/traefik/traefik/blob/v3.3.4/pkg/proxy/httputil/proxy.go#L130-L149) (a timeout would be 504, no healthy endpoints 503). Go normally retries straight through this, but only for [replayable requests: no body, and GET/HEAD/OPTIONS/TRACE](https://github.com/golang/go/blob/go1.24.0/src/net/http/request.go#L1534-L1546). Every LAPIS query from the website is a POST with a body, so none qualify — which is why GETs and CORS preflights never show this and LAPIS queries do.

Nothing exotic about our stack: AWS documents the identical failure for ALB with the same advice — [*"Check whether the keep-alive duration of the target is shorter than the idle timeout value of the load balancer"*](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-troubleshooting.html#http-502-issues). What's unusual is that our tests assert on zero console errors (#4469, #4555), so a ~0.1% background error rate becomes a red build.

## The change

A `ServersTransport` with `idleConnTimeout: 3s`, under the shortest backend keep-alive (5s), so traefik always retires a connection before a backend can close it. Attached per ingress via `traefik.ingress.kubernetes.io/service.serverstransport` rather than cluster-wide, since we don't control traefik's static config everywhere.

`maxIdleConnsPerHost: 200` looks redundant but isn't. `forwardingTimeouts` gets [`SetDefaults()` applied before the CRD's fields](https://github.com/traefik/traefik/blob/v3.3.4/pkg/provider/kubernetes/crd/kubernetes.go#L369-L370) so `dialTimeout` keeps its 30s default, but `maxIdleConnsPerHost` is copied straight from the spec and Go reads 0 as [`DefaultMaxIdleConnsPerHost` = 2](https://github.com/golang/go/blob/go1.24.0/src/net/http/transport.go#L60). Without that line we'd quietly drop from a 200-connection pool to a 2-connection one.

`lapis-redirect-ingress` is left alone on purpose: it redirects before forwarding and its backend is a placeholder.

## Testing

Reproduced outside k8s, traefik in front of a server that closes idle keep-alive connections, 8 parallel POSTs per burst with the gap tuned to land on the keep-alive boundary:

| | 502 rate |
|---|---|
| traefik default (90s) | 9 / 480 — 1.88% |
| with this `ServersTransport` | 0 / 480 |

Traefik's API confirms it applies (`serversTransport: shortidle@file` vs `<none>`) and the mock backend drops to `idle_closes: 0` — traefik is now always the one closing first. The reproduced 502 is byte-identical to CI's: `Bad Gateway`, `Content-Length: 11`, `Access-Control-Allow-Origin: *`, no `Server` header. `helm lint` passes; chart renders for `environment=local` and `=server`.

## Notes

- If this holds, #7090 shouldn't be needed. #6917/#6368 suppress a *preflight* failure blamed on LAPIS, but traefik answers preflights itself and OPTIONS is retried anyway — different cause, those stay.
- #4945 suppresses `ERR_INCOMPLETE_CHUNKED_ENCODING`, which is the same event landing after the response has started, too late for traefik to convert to a 502. This may reduce it.
- The same race can hit backend POSTs wherever `loculus-backend-ingress` puts Tomcat behind traefik, so it's not only a test-flakiness fix.
- This was painful to find from CI artifacts: traefik logs the cause at DEBUG only, the access log shows a bare `502`, and we only collect `-l app=loculus` pods. #5478 would have made it quick.


kubernetes/loculus/templates/servers-transport.yaml
```diff
@@ -0,0 +1,23 @@
...
+Traefik keeps idle upstream connections in a pool for `idleConnTimeout` (default 90s, inherited
+from Go's http.DefaultTransport). Our backends close idle keep-alive connections much sooner
+(Node/Astro and uvicorn: 5s, Tomcat: 20s). When traefik dispatches a request onto a pooled
+connection at the instant the backend closes it, Go reports io.EOF and traefik answers 502 --
+and Go only replays bodyless GET/HEAD/OPTIONS/TRACE, so POSTs (every LAPIS query) are not
+retried. Retiring pooled connections before any backend can close them removes the race.
+*/}}
+{{- define "loculus.serversTransportRef" -}}
+{{ .Release.Namespace }}-short-idle-conn-timeout@kubernetescrd
+{{- end }}
```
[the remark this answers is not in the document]
corneliusroemer-agent: Agreed, moved it to `_servers-transport.tpl` in 322fd3c8a — `servers-transport.yaml` is now just the manifest.

kubernetes/loculus/templates/servers-transport.yaml
```diff
@@ -0,0 +1,23 @@
...
+apiVersion: {{ $traefikApiVersion }}
+kind: ServersTransport
+metadata:
+  name: short-idle-conn-timeout
+spec:
+  # Unlike forwardingTimeouts, this is not defaulted when a ServersTransport is attached: Go reads
+  # 0 as DefaultMaxIdleConnsPerHost (2), which would shrink the pool from traefik's own default.
+  maxIdleConnsPerHost: 200
```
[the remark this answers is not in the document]
corneliusroemer-agent: Good catch, and it checks out: the CRD path builds a fresh `dynamic.ServersTransport` rather than merging, and `ForwardingTimeouts.SetDefaults()` only fills in `dialTimeout` and `idleConnTimeout`. So `responseHeaderTimeout` happens to match the static default (0/no timeout) by luck, and the TLS fields are moot while every upstream is plain HTTP — but that stops being true the moment someone adds an HTTPS backend or a static `serversTransport`. Written into the manifest comment in 322fd3c8a.

theosanderson:
@codex review

@theosanderson-agent consider the pros and cons of this change

kubernetes/loculus/templates/servers-transport.yaml
```diff
@@ -0,0 +1,17 @@
+{{- $traefikApiVersion := ternary "traefik.io/v1alpha1" "traefik.containo.us/v1alpha1" (eq (int $.Values.traefikVersion) 3) -}}
```
theosanderson:
```suggestion
{{- $traefikApiVersion := ternary "traefik.io/v1alpha1" "traefik.containo.us/v1alpha1" (ne (int $.Values.traefikVersion) 2) -}}
```

Something like this seems a bit less likely to break on traefik v4?

theosanderson-agent:
Pros and cons, having read the diff and the surrounding chart.

## In favour

- **Right layer for the fix.** One knob covers all six upstream runtimes at once, including the ones we don't build (LAPIS, keycloak, minio, `http-server` for docs). Option (b) has to be re-derived per runtime and isn't reachable at all for Astro/Node, where `keepAliveTimeout` isn't exposed as a CLI flag or env var.
- **It targets the failure class that can't be papered over.** Go replays only bodyless GET/HEAD/OPTIONS/TRACE, so no retry setting — traefik's `retry` middleware included — could safely cover LAPIS/backend POSTs. Removing the race is the only option that helps the requests that actually break.
- **Blast radius is proxy→upstream only.** Client-facing keep-alive is untouched, so browsers/CDNs see no change.
- **`maxIdleConnsPerHost: 200` is the non-obvious part and it's right.** Omitting it would have silently traded a 90s-idle 200-conn pool for a 3s-idle 2-conn pool, which would have been a much worse regression than the bug being fixed.
- **Not only a CI-flake fix.** `loculus-backend-ingress` puts Tomcat behind traefik for real user POSTs (submission), so this removes user-visible 502s too.
- Cheap, declarative, trivially revertible, no new CRD dependency (the chart already requires traefik's `Middleware`).

## Against / risks

1. **The invariant spans repos and nothing enforces it.** `3s` is safe only while every upstream's keep-alive stays above it. Four of those upstreams are images we don't control; if an upgrade drops one to 2s, the symptom is intermittent 502s again, not a failing build. The comment naming the current values is the mitigation, and it needs to stay accurate.
2. **It replaces rather than merges with traefik's static `serversTransport`** (correctly flagged in the template). For us that's theoretical — all upstreams are plaintext in-cluster, and `forwardingTimeouts` still gets `SetDefaults()` — but Loculus is deployed by third parties who run their own traefik. Anyone with a customised default transport (custom `rootCAs`, `insecureSkipVerify`, tuned `responseHeaderTimeout`) loses it for every Loculus router with no signal. If that ever comes up, the escape hatch is exposing the transport through `values.yaml` rather than hardcoding.
3. **Per-ingress opt-in has to be remembered forever.** The partial removes the duplicated string, not the omission risk, and `helm-schema-lint` won't catch a missing annotation. There's already one gap: `loculus-docs-ingress` (`docs-preview.yaml`) doesn't carry it, and it runs `http-server` on Node — same 5s default. Preview-only and static GETs, which Go *does* replay, so effectively harmless, but it's exactly the drift this pattern invites. Worth either adding the line or a render-time check that every Ingress with a real backend has the annotation. (`lapis-redirect-ingress` genuinely doesn't need it — agreed.)
4. **More connection churn.** Idle conns retire after 3s instead of 90s, so low-traffic paths pay a fresh TCP handshake per request. In-cluster plaintext, so sub-millisecond, and busy paths still reuse from the 200-conn pool. Real, but not worth optimising against.
5. **It's a margin, not a proof.** Both timers start at the same event (end of previous response), so 3s vs 5s leaves 2s of slack — comfortable, but its safety is entirely point 1.
6. **It doesn't fix deploy-time 502s.** Pod rollouts and graceful shutdown still sever in-flight upstream connections. A green E2E run after this shouldn't be read as "all 502s are gone" — the residue would have a different cause.

## Verdict

Worth merging. The diagnosis is well-evidenced and the reproduction (9/480 → 0/480) is the kind of before/after that's hard to argue with. The two follow-ups I'd consider are the CI guard for point 3 and the one-line docs-ingress annotation; happy to push either to this branch if you want them here rather than separately.

kubernetes/loculus/templates/servers-transport.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+Beware: this replaces traefik's cluster-wide static serversTransport for these routers instead of
+merging with it, so changes made there won't apply and need mirroring here.
+*/}}
+apiVersion: {{ $traefikApiVersion }}
+kind: ServersTransport
+metadata:
+  name: short-idle-conn-timeout
+spec:
+  maxIdleConnsPerHost: 200 # not defaulted here; 0 would mean Go's DefaultMaxIdleConnsPerHost (2)
```
theosanderson: I found this comment confusing, as I didn't know what "not defaulted here" would mean in this context. I asked Claude, and now get it. But IMO this would be clearer without the comment. If something is specified one assumes there's a reason.

theosanderson (approved):
nice

kubernetes/loculus/templates/servers-transport.yaml
corneliusroemer: Great catch! I'll see if we use eq anywhere else

kubernetes/loculus/templates/servers-transport.yaml
```diff
@@ -0,0 +1,17 @@
...
```
corneliusroemer:
```suggestion
  maxIdleConnsPerHost: 200 # matches traefik defaults
```

kubernetes/loculus/templates/servers-transport.yaml
corneliusroemer: I've edited the comment to say this is the traefik default

PR #7093, opened 2026-08-18
Revert "feat(backend): scope the `extract-unprocessed-data` etag by organism"

theosanderson:
Reverts loculus-project/loculus#7082 - merge conflict breaking CI

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/controller/SubmissionController.kt
@@ -224,7 +224,7 @@ open class SubmissionController(
             )
         }
 
-        val lastDatabaseWriteETag = releasedDataModel.getLastDatabaseWriteETag(organism = organism)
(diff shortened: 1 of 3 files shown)


corneliusroemer:
I've already got Claude on fixing the problem which would save a revert back&forth - but then Claude is flaky right now 😢

corneliusroemer-agent:
#7094 is an alternative to this revert: the breakage is a one-line merge race, not a problem with #7082 itself.

#7082 added `.eTag(lastDatabaseWriteETag)` to the 304 path of `getReleasedData`, while #7087 had meanwhile renamed that local to `lastDatabaseWriteETagWithDate`. Using the renamed variable fixes the compile and keeps the feature. That value is also what #7082's own test asserts, since it is the etag that matched `If-None-Match`.

Verified locally on top of current main: `compileKotlin`, `compileTestKotlin` and `ktlintCheck` clean, and the full backend suite passes (74 classes, 568 tests, 0 failures, 1 skipped).

PR #4813, opened 2025-08-05
fix(backend): handle PostgreSQL parameter limit for large batch submi…

theosanderson:
…ssions

- Add chunking to prevent exceeding PostgreSQL's 65,535 parameter limit when processing large batches
- Create shared DatabaseConstants with POSTGRESQL_PARAMETER_LIMIT and SAFE_CHUNK_SIZE constants
- Add extension functions for safe database chunking operations
- Update UploadDatabaseService to chunk data use terms and accession validation operations
- Update FilesDatabaseService to chunk all inList operations
- Add test case for submitting 100,000 sequences to verify the fix

This prevents internal server errors when submitting very large batches of sequences.

🤖 Generated with [Claude Code](https://claude.ai/code)

resolves #1613

### Screenshot
image

### PR Checklist
- [ ] All necessary documentation has been adapted.
- [x] The implemented feature is covered by appropriate, automated tests.

I manually tested that I could submit 100k sequences to the dummy organism (it took a while)

🚀 Preview: https://largebatch.loculus.org

PR #7096, opened 2026-08-19
perf(backend): assign accessions in one statement

corneliusroemer-agent:
This is an alternative to https://github.com/loculus-project/loculus/pull/7074 with a smaller diff and without unnecessary test.

### What

During submission the backend assigned accessions with one `UPDATE` statement per sequence, so
500k single-row statements per upload. This assigns them all in a single statement instead.

`generateNewAccessionsForOriginalUpload` already builds two equal-length lists — the submission ids
read from `metadata_upload_aux_table`, and the accessions freshly drawn from `accession_sequence`.
Both are now bound as `text[]` parameters and zipped server-side by Postgres' two-argument `unnest`:

```sql
UPDATE metadata_upload_aux_table AS m
SET
    accession = generated.accession,
    version = 1
FROM unnest(?::text[], ?::text[]) AS generated(submission_id, accession)
WHERE
    m.upload_id = ?
    AND m.submission_id = generated.submission_id
```

`unnest(a, b)` emits one row per position, pairing `a[i]` with `b[i]`, which is exactly the pairing
`submissionIds.zip(nextAccessions)` produced in Kotlin. The join then matches those rows against the
aux table on `(upload_id, submission_id)`, which is its primary key.

### Why arrays rather than a generated `VALUES` list

The natural way to batch this is `FROM (VALUES (?,?), (?,?), …)`, but that spends **two bind
parameters per row**. PostgreSQL's extended query protocol caps a statement at 65,535 parameters
(the Bind message counts them in an `int16`), so that form has to be chunked, and each chunk size
produces a different SQL string to parse and plan.

An array is **one bind parameter regardless of how many elements it holds**. So this statement takes
three parameters — two arrays and `upload_id` — for any upload size:

| | placeholders | statements for 500k rows | SQL text |
|---|---|---|---|
| one `UPDATE` per row (before) | 3 | 500,000 | fixed |
| `VALUES (?,?), …` | 2N + 1 | 50 (at 10k/chunk) | grows with N |
| `unnest(?::text[], ?::text[])` | 3 | 1 | fixed |

No chunking, no `DatabaseConstants.POSTGRESQL_PARAMETER_LIMIT` arithmetic, and one cached plan.

### Why raw SQL

Exposed has no DSL for "update N rows to N different values". Checked against `exposed-core-1.0.0`:
`BatchUpdateStatement`'s only constructor takes `IdTable<?>`, and `MetadataUploadAuxTable` is a plain
`Table` with a composite primary key; the JDBC DSL offers `batchInsert`, `batchReplace` and
`batchUpsert` but no `batchUpdate`. `batchUpsert` would work mechanically but would force restating
every NOT NULL column and rewrite whole rows instead of two columns.

The sibling method `associateRevisedDataWithExistingSequenceEntries` already does raw SQL inside
`transaction { }` for the same reason, and this follows that shape.

### Result

Measured against PostgreSQL 16 on a table built from the `V1__init.sql` DDL, 500k rows under the
target `upload_id` plus 50k under another so `upload_id` isn't trivially the whole table:

| Variant | Time |
|---|---|
| `unnest`, 1 statement | **2.6 s** |
| `VALUES`, 50 chunks of 10k | 4.9 s |

The `VALUES` form hash-joins each chunk against a full scan of the table, so 50 chunks means 50
passes; the array form does one.

For context this statement is not where the upload phase spends its time — in a 500k benchmark it is
single-digit seconds out of a ~2200 s upload phase. Against the per-row version it removes 500k
round trips, which is the change that matters; against a chunked `VALUES` version it is mostly about
the code being simpler.

### On the query plan

`unnest` over a *parameter* has no row estimate from statistics, so the concern is that the planner
falls back to its ~100-row default and picks a nested loop over 500k iterations. Verified both plan
modes on PG 16:

- **Custom plan** (what pgjdbc uses below its `prepareThreshold` of 5 executions, and this statement
  runs once per submission): PG's `unnest` support function reads the bound array's length and
  estimates 500,000 rows. Correct hash join.
- **Forced generic plan**: it does pick the nested loop — and still completes in 2.4 s, because each
  iteration is a primary-key index scan.

So neither mode regresses.

### A note on the length invariant

Multi-argument `unnest` pads with NULLs rather than erroring when the arrays differ in length, which
would silently write NULL accessions. That cannot happen here — the existing
`submissionIds.size != nextAccessions.size` check runs before the update, and
`getNextSequenceNumbers` itself throws if the sequence returns the wrong count.

### PR Checklist

  - internal change, no behaviour or API change
  - No new tests. Accession assignment on submit is already covered: every `SubmitEndpointTest`
    submit asserts the returned submissionId→accession mapping, `SubmissionJourneyTest` carries those
    accessions through the lifecycle, and `sequence_entries.accession` is NOT NULL so any row that
    failed to receive an accession would break the subsequent insert rather than pass silently.
    `SubmitLargeBatchTest` (100k sequences, `RUN_EXTRA_TESTS=true`) covers this at scale — it was
    added by #4813, the previous parameter-limit change.
- [x] Any manual testing that has been done is documented
  - `./gradlew test` for `*Submit*`, `*SubmissionJourney*`, `*Revise*` with `RUN_EXTRA_TESTS=true`
    (so including the 100k-sequence submission) and `./gradlew ktlintCheck`, both green.
  - The timings and query plans above were measured directly against PostgreSQL 16 with a
    standalone benchmark script.

---

This is an alternative to #7074, which batches the same update using a generated `VALUES` list with
chunking. See "Why arrays rather than a generated `VALUES` list" above for the comparison.

Co-Authored-By: stacy <48458444+clonder@users.noreply.github.com>

🚀 Preview: https://perf-accession-update-unn.loculus.org


backend/src/main/kotlin/org/loculus/backend/service/submission/UploadDatabaseService.kt
```diff
@@ -346,21 +346,32 @@ class UploadDatabaseService(
...
-                    (submissionIdColumn eq submissionId) and (uploadIdColumn eq uploadId)
-                },
-            ) {
-                it[accessionColumn] = accession
-                it[versionColumn] = 1
-            }
+        // Assign all accessions in one statement rather than one UPDATE per sequence. The two arrays
+        // are bound as single parameters, so the statement is the same shape for any upload size and
+        // there is no parameter limit to chunk around. Exposed has no DSL for updating many rows to
+        // many different values, hence raw SQL, as in associateRevisedDataWithExistingSequenceEntries.
```
corneliusroemer:
```suggestion
```

clonder (approved):
Nice, thanks for the proposition! The unnest form is better than #7074, so I've closed mine in favour of this. Looks good to me!

corneliusroemer:
Thanks @clonder! I'll merge this one here. I'll add you as a coauthor on the commit/PR as you did all the investigation etc!

corneliusroemer:
Just doing a bit of testing on the preview - already noticing a big submission is a lot faster! It took ~30s on main to submit 10k sequences and only a second on this branch so it's a noticeable improvement even in that regime!

PR #7091, opened 2026-08-18
feat(website): 'Restore this sequence' button, refactor for sequence data UI and sequence management

tombch:
resolves #6870

#### Updates for revocation handling

- Adds a 'Restore this sequence' button on the latest revoked entry, which redirects to the form revision page to revise from the latest non-revoked version.
- Does not display the 'Revoke this sequence' button on any version when the latest is a revocation (from PR #7097).

#### Updates for revision form

- Adds a banner to the revision form page when revising from a sequence which is currently revoked.

#### Refactoring

- Moves sequence management logic into a `SequenceManagement` component.
- Includes refactor for `SequenceDataUI`: now handles revoked and unrevoked sequences on sequence details page and sequence preview modal, and removed `RevocationEntryDataTable` (from PR #7099).
- Added `utils/dataUseTerms` and moved data use terms sorting directly into the `DataUseTermsHistoryModal` and `SequenceManagement` components, as those were the only two places that actually use the data use terms history.

#### Tests

- Adds spec files for new components (`SequenceManagement`) as well as existing ones (`SequenceDataUI`, `SequencesBanner`).

#### Docs

- Adds a 'How to restore a revoked sequence' section in the Loculus docs.

### Screenshot

Adds a 'Restore this sequence' button if the sequence is revoked:

image

On the revision page, now displays a banner notifying that the user is revising a revoked sequence:

image

### PR Checklist
- [ ] All necessary documentation has been adapted (docs changes required for PPX).
- [x] The implemented feature is covered by appropriate, automated tests.
- [x] Any manual testing that has been done is documented (i.e. what exactly was tested?)

🚀 Preview: https://undo-revocation.loculus.org

PR #7097, opened 2026-08-19
fix(website): hide revoke button when the latest version is a revocation

corneliusroemer-agent:
Split out of #7089, which now carries only the backend change. Stacked on #7091 (`undo-revocation`), because that PR moves the sequence-management block into `SequenceManagement.tsx` — this change belongs in the new component. **Targets `undo-revocation`, so it must merge after #7091**; GitHub will retarget it to `main` once that lands.

## Change

The website already tried to hide the revoke button on revocations, but the guard was per *displayed* version. Revoking always targets the *latest* version — `RevokeButton` posts the bare accession and the backend resolves it against `isMaxVersion` — so browsing back to an earlier, non-revocation version of a revoked accession still rendered the button.

The button is now gated on the latest version from the sequence entry history. `SequenceManagement` already receives that history (as of #7091) and it already includes `isRevocation`, so this needs no extra request and no backend call.

Scope note: only `<RevokeButton>` is gated, not the whole "Sequence management" block. Revising a revoked accession is a supported flow based on the latest non-revocation version (see `getLatestAccessionVersionForRevision`), so the revise and data-use-terms buttons must stay available. There is a test for that.

The backend half of the fix (#7089) makes the server reject the operation outright with a 422; this just stops offering it.

## Question for the reviewer

Now that #7091 offers an **Undo revocation** button when the *displayed* version is the revocation, arguably an earlier *non-revocation* version of a revoked accession should offer the same thing rather than just having Revoke hidden. This PR does the minimal thing (hide the button) and leaves that call to you.

## Tests

New `SequenceManagement.spec.tsx` with three cases: revoke button hidden when the latest version is a revocation, revise link still offered in that case, revoke button shown when the latest version is not a revocation. The first was confirmed to fail against the un-gated component, so it is not vacuous.

Ran locally: `vitest` (3/3 pass), `npm run check-types` (0 errors), eslint and prettier clean on the changed files.


website/src/components/SequenceDetailsPage/SequenceManagement.tsx
```diff
@@ -91,14 +91,19 @@ export const SequenceManagement: FC<Props> = ({
...
-                                clientConfig={clientConfig}
-                                accessionVersion={accessionVersion.split('.')[0]}
-                                accessToken={accessToken}
-                                groupId={groupId}
-                                onRevokeSuccess={onRevokeSuccess}
-                            />
+                            {/* Revoking always targets the latest version of an accession, not the
+                                version being displayed, so hide the button whenever the latest
+                                version is already a revocation. */}
+                            {!isLatestVersionRevocation(sequenceEntryHistory) && (
```
corneliusroemer:
```suggestion                  {!isLatestVersionRevocation(sequenceEntryHistory) && (
```

tombch (approved):

tombch:
Thanks! I'm going to merge this into #7091 now and prepare that as a general website revocation changes PR

Issue #6674, opened 2026-06-15
File sharing: Requesting file sharing URL for groupId that does not exist returns nothing - should return an error message


anna-parker:
```
curl -X POST "https://backend-file-sharing-nooverwrite.loculus.org/files/request-upload?groupId=2&numberFiles=1" -H "Authorization: Bearer $TOKEN"
```

anna-parker:
oh... actually my token was wrong but I got no error... very bad either way!

theosanderson:
One gets error code 401 - seems OK to me

theosanderson:
But yeah, ~maybe~ an extra human readable body could be useful

Issue #5776, opened 2025-12-29
Improve file upload error handling and UX on website


chaoran-chen:
#5775 introduced the usage of multipart upload to the website, including a progress indicator. Based on that, we can improve the error handling. In particular, we should consider internet connectivity issues. Potential improvements:

- Automated retry of the upload of a part
- Visual feedback that the internet connection/upload has been interrupted

Issue #5449, opened 2025-11-17
Make errors more user friendly by replacing JSON.stringify


corneliusroemer:
#5399 makes an error message more user friendly, specifically on edit page.

A quick `rg "JSON.stringify"` shows that there are quite a few more instances where we could make similar improvements: https://github.com/search?q=repo%3Aloculus-project%2Floculus%20JSON.stringify&type=code

https://github.com/loculus-project/loculus/pull/5399/files#diff-439227963ca7a1e992b3d8a90c95d8349105ce143c6ba4c65d3e218aa030193aR34-R43

PR #7104, opened 2026-08-19
fix(website): parse backend error responses so users see the real message

### LINKED ISSUE #6976: `Unprocessable Entity` should not be text displayed to typical website users
E.g. 

Image

### LINKED ISSUE #7103: Website does not parse and show backend errors correctly
I tried to revise a sequence that was not in the correct state, instead of showing me the error message I got a horrible warning:

Image

```
Failed to submit revision for LOC_001DSEM.1: {"message":"Request failed with status code 422","name":"AxiosError","stack":"AxiosError: Request failed with status code 422\n    at Pt (https://deacon-service.loculus.org/_astro/createAuthorizationHeader.CJcdFn5g.js:3:10808)\n    at XMLHttpRequest.g (https://deacon-service.loculus.org/_astro/createAuthorizationHeader.CJcdFn5g.js:3:17595)",
"config":{"transitional":{"silentJSONParsing":true,"forcedJSONParsing":true,"clarifyTimeoutError":false,"legacyInterceptorReqResOrdering":true,"advertiseZstdAcceptEncoding":false,"validateStatusUndefinedResolves":true},"adapter":["xhr","http","fetch"],"transformRequest":[null],"transformResponse":[null],"timeout":0,"xsrfCookieName":"XSRF-TOKEN","xsrfHeaderName":"X-XSRF-TOKEN","maxContentLength":-1,"maxBodyLength":-1,"env":{},"headers":{"Accept":"application/json, text/plain, */*","Content-Type":"application/x-www-form-urlencoded","Authorization":"Bearer <redacted>"},"baseURL":"https://backend-deacon-service.loculus.org","method":"post","url":"/west-nile/revise","data":{},"params":{},"allowAbsoluteUrls":true},"code":"ERR_BAD_REQUEST","status":422}
```

corneliusroemer-agent:
Fixes #7103.

This PR fixes bugs that limited the helpfulness/precision of error messages.

### Examples

Before:
Google Chrome 2026-08-19 18 41 07

After:
Google Chrome 2026-08-19 18 41 14

---

Before:
Google Chrome 2026-08-19 18 46 38

After:
Google Chrome 2026-08-19 18 45 41

---

Before:
Google Chrome 2026-08-19 18 49 02

After:
Google Chrome 2026-08-19 18 48 21

---

## Root cause

Not what the earlier analysis on the issue concluded. The backend does **not** serialise `"instance": null` — it omits **`type`**.

Probed live against `backend-main.loculus.org`:

```
GET /nonexistent-organism/get-released-data
{"title":"Bad Request","status":400,"detail":"…","instance":"/nonexistent-organism/get-released-data"}

GET /get-seqset?seqSetId=doesnotexist&version=1
{"title":"Not Found","status":404,"detail":"…","instance":"/get-seqset"}
```

`instance` is present as a non-null string, so it can't be the field that fails validation. `type` is absent: `ProblemDetail.forStatus()` leaves it at its `about:blank` default and Spring's `ProblemDetailJacksonMixin` is `@JsonInclude(NON_EMPTY)`, so Jackson drops it. Both probes go through the same private `responseEntity()` helper in `ExceptionHandler.kt` that `handleUnprocessableEntityException` uses, so the 422 in the issue has the same shape.

The website's `problemDetail` schema declared `type: z.string()` (required). Running `safeParse` on the observed body names the failing field exactly:

```
[{"code":"invalid_type","expected":"string","received":"undefined","path":["type"],"message":"Required"}]
```

`isErrorFromAlias` only returns true when a declared error schema `safeParse`s the response body, so it returned `false` for every backend error. `EditPage.getErrorDetail` then fell back to `JSON.stringify(error)` — and because `AxiosError` has a custom `toJSON()` that strips `response`, that dumps the request config at the user instead of the backend's message. That's the screenshot in the issue.

## Second, independent bug (and an audit of the whole class)

`submitReviewedSequenceEndpoint` declared `errors: [notAuthorizedError]` — 401 only. zodios matches errors by status (or a `status: 'default'` entry), so `isErrorFromAlias(backendApi, 'submitReviewedSequence', …)` could never match a 400/422 regardless of the schema. Fixing only the schema would have left the "Submit edits" path in `EditPage` broken.

I then audited every endpoint in every zodios API definition the website has, by synthesising a problem-detail error at statuses 400/403/404/409/422/500 and checking whether `isErrorFromPath` matches it. Three more `backendApi` endpoints had the same gap:

| Endpoint | Declared | Unmatched before |
| --- | --- | --- |
| `getDataToEdit` | 401 | 400, 403, 404, 409, 422, 500 |
| `getSequences` | 401, 404 | 400, 403, 409, 422, 500 |
| `extractUnprocessedData` | 401 | 400, 403, 404, 409, 422, 500 |

`getSequences` is the live one — `useSubmissionOperations.ts:130` calls `isErrorFromAlias` on it. All three now declare `status: 'default'`, and a test asserts the invariant across every `backendApi` endpoint so it can't recur (verified: it fails listing all three before the fix).

`groupManagementApi`, `seqSetCitationApi` and `lapisApi` have the same gap on ~30 endpoints, but **none of them is ever passed to `isErrorFrom*`** (checked — only `backendApi` is; they go through `ZodiosWrapperClient`, which parses `response.data` directly). Those gaps are latent, not live, so I've left them alone rather than churn 30 declarations for no behaviour change.

## Changes

- `website/src/types/backend.ts` — `type: z.string().default('about:blank')`. RFC 9457 §3.1.1: *"When this member is not present, its value is assumed to be `about:blank`."* So the backend is compliant here and the website was not — a zod default states the spec's semantics rather than merely tolerating the omission, and keeps `ProblemDetail.type` a required `string` for consumers.
- `website/src/services/backendApi.ts` — declare `default`/400/422 problem-detail errors on `submitReviewedSequence`, and a `default` on `getDataToEdit`, `getSequences` and `extractUnprocessedData`.
- `website/src/components/Edit/EditPage.tsx` — fall back to `stringifyMaybeAxiosError` (used consistently elsewhere in the codebase) instead of `JSON.stringify(error)`.
- `website/src/services/backendApi.spec.ts` — regression test built from the literal live-observed body.

## Blast radius

`type` being required broke strict problem-detail parsing well beyond `EditPage`: `DataUploadForm.tsx:483`, `useSubmissionOperations.ts:116/123/130`, `backendClient.ts:195`, and `zodiosWrapperClient.ts:74` for its backend-facing subclasses (`GroupManagementClient` and `SeqSetCitationClient` both pass `(axiosError) => axiosError.data` with no normalisation, and that site uses `.parse`, so it throws and degrades to "Unknown error from backend"). The `/get-seqset` 404 probed above is literally a `SeqSetCitationClient` endpoint. Making `type` optional repairs all of them.

`serviceHooks.ts:72` and the LAPIS subclass of `ZodiosWrapperClient` were **not** affected: LAPIS does send `type` (`{"error":{"type":"about:blank","title":"Bad Request","status":400,…}}`, probed live against `lapis-main.loculus.org`). This is a backend-only bug.

Related but **not** claimed fixed: #6976. `zodiosWrapperClient.createProblemDetail`'s fallback sets `title: error.message`, while the now-working path sets `title: "Unprocessable Entity"` — so this change may be what *surfaces* #6976 rather than what fixes it. Worth a look separately.

## Why not require `type` from the backend instead?

RFC 9457 makes every member optional, and `type` specifically defaults to `about:blank` when absent — which is exactly why Spring omits it (it never leaves that default). The spec's general posture is that consumers degrade around missing members: §3.1 says a member whose value type doesn't match "MUST be ignored — i.e., processing will continue as if the member had not been present." Rejecting a whole problem document over one absent optional member is the opposite of that.

Note `type` is *not* the media type. That's the `Content-Type: application/problem+json` header, which the backend already sends correctly on every error.

Strictly, `title`/`status`/`detail` are optional per the RFC too. I deliberately kept them required: the UI needs `detail`, and a body without one is better off failing validation and hitting the (now readable) fallback than rendering an empty toast.

## Verification

- The new assertions **fail on `main` and pass with this change** (checked by stashing the source fixes).
- `npm run check-types` — 0 errors (370 files).
- `npm run test` — 714 passed.
- eslint + prettier clean.

Not manually exercised in a running website; the regression test drives the exact code path with the literal live response body.

## Latent issues found, left out of scope

- `isErrorFromAlias` matches `endpoint.path` against `config.url` with an anchored regex, so it returns `false` for any absolute `config.url`. Not in play here (the issue's dump has a relative URL), but a real second failure mode.
- `stringifyMaybeAxiosError` returns `data.detail` unchecked; a non-ProblemDetail object body yields the literal string `"undefined"`.

### DIFF
--- a/website/src/components/Edit/EditPage.tsx
+++ b/website/src/components/Edit/EditPage.tsx
@@ -15,6 +15,7 @@ import { getLatestAccessionVersionForRevision, type SequenceEntryHistory } from 
 import type { ClientConfig } from '../../types/runtimeConfig.ts';
 import { createAuthorizationHeader } from '../../utils/createAuthorizationHeader.ts';
 import { getAccessionVersionString, parseAccessionVersionFromString } from '../../utils/extractAccessionVersion.ts';
+import { stringifyMaybeAxiosError } from '../../utils/stringifyMaybeAxiosError.ts';
 import { displayConfirmationDialog } from '../ConfirmationDialog.tsx';
 import { SequenceEntryHistoryMenu } from '../SequenceDetailsPage/SequenceEntryHistoryMenu.tsx';
 import { ExtraFilesUpload } from '../Submission/DataUploadForm.tsx';
@@ -36,7 +37,11 @@ type EditPageProps = {
 const logger = getClientLogger('EditPage');
 
 /**
- * Extracts the detail field from a backend error response
+ * Extracts the detail field from a backend error response.
+ *
+ * The fallback must not be `JSON.stringify(error)`: AxiosError has a custom `toJSON()` that
+ * strips `response`, so stringifying it dumps the request config at the user instead of the
+ * backend's message (see issue #7103).
  */
 function getErrorDetail(error: unknown): string {
     if (
@@ -45,7 +50,7 @@ function getErrorDetail(error: unknown): string {
     ) {
         return error.response.data.detail;
     }
-    return JSON.stringify(error);
+    return stringifyMaybeAxiosError(error);
 }
 
 const InnerEditPage: FC<EditPageProps> = ({
--- a/website/src/services/backendApi.spec.ts
+++ b/website/src/services/backendApi.spec.ts
@@ -49,6 +49,26 @@ describe('problemDetail', () => {
 });
 
 describe('backendApi error declarations', () => {
+    // zodios matches an error response against the endpoint's declared `errors` by status code.
+    // An endpoint with no `status: 'default'` entry silently fails to match anything it did not
+    // enumerate, so `isErrorFromAlias` returns false and the caller loses the backend's message —
+    // regardless of the schema. `submitReviewedSequence` declared 401 only and hit exactly that.
+    test('every endpoint declares a catch-all error response', () => {
+        // Deliberately typed loosely: zodios' types claim `alias` and `errors` are always present,
+        // but this test exists to catch an endpoint added without them.
+        const endpoints = backendApi as readonly {
+            alias?: string;
+            path: string;
+            errors?: readonly { status: number | 'default' }[];
+        }[];
+
+        const missing = endpoints
+            .filter((endpoint) => !(endpoint.errors ?? []).some((error) => error.status === 'default'))
+            .map((endpoint) => endpoint.alias ?? endpoint.path);
+
+        expect(missing).toEqual([]);
+    });
+
     test('recognises a backend error from revise', () => {
         expect(isErrorFromAlias(backendApi, 'revise', axiosErrorForUrl('/west-nile/revise'))).toBe(true);
     });
--- a/website/src/services/backendApi.spec.ts
+++ b/website/src/services/backendApi.spec.ts
@@ -1,0 +1,51 @@
+import { isErrorFromAlias } from '@zodios/core';
+import { AxiosError, AxiosHeaders } from 'axios';
+import { describe, expect, test } from 'vitest';
+
+import { backendApi } from './backendApi.ts';
+import { problemDetail } from '../types/backend.ts';
+
+/**
+ * The exact shape the backend serialises for an error, as observed live against
+ * backend-main.loculus.org. Spring's Jackson mixin omits `type` (it never leaves its
+ * `about:blank` default) — see ExceptionHandler.responseEntity in the backend.
+ *
+ * Regression test for #7103: the website used to require `type`, so no backend error ever
+ * validated and users were shown a raw AxiosError dump instead of the message.
+ */
+const backendErrorBody = {
+    title: 'Unprocessable Entity',
+    status: 422,
+    detail: 'Accession version LOC_001DSEM.1 is not in state APPROVED_FOR_RELEASE',
+    instance: '/west-nile/revise',
+};
+
+const axiosErrorForUrl = (url: string, status = 422) => {
+    const config = { url, method: 'post', headers: new AxiosHeaders() };
+    return new AxiosError(`Request failed with status code ${status}`, 'ERR_BAD_REQUEST', config, {}, {
+        data: backendErrorBody,
+        status,
+        statusText: '',
+        headers: {},
+        config,
+    } as never);
+};
+
+describe('problemDetail', () => {
+    test('parses a backend error response that omits type', () => {
+        const result = problemDetail.safeParse(backendErrorBody);
+        expect(result.success).toBe(true);
+    });
+});
+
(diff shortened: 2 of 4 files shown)


corneliusroemer-agent:
Triage of adjacent open issues, since this PR touches the shared error-display path. Checked each against the code rather than by title; **this PR closes #7103 only.**

| Issue | Closed by this PR? | Why |
|---|---|---|
| #5449 — replace `JSON.stringify` in error messages | **Partially** | This removes the `JSON.stringify(error)` in `EditPage.getErrorDetail` — the exact function that issue links (added in #5399). Four user-facing instances remain: `SearchFullUI.tsx:306`, `SequenceViewer.tsx:39`, `submission/edit/[accession]/[version].astro:51`, `seqsets/[seqSetId].[version].astro`. Worth linking, not closing. |
| #6976 — `Unprocessable Entity` shown to users | **No — arguably worsens it** | The broken path set `title` to `"Request failed with status code 422"`; the repaired path returns the backend's real `title`, i.e. literally `"Unprocessable Entity"`. So this PR makes that string reachable in *more* places. #6976 needs a status→friendly-text mapping, which is a separate change. |
| #7105 — corrupt gzip never errors cleanly | **No** | Different mechanism: `DataUploadForm.tsx:142` blocks submission client-side with "Cannot submit: metadata file is still being processed" (added in #6922), so the backend is never called. This PR only affects how backend errors are displayed. |
| #6674 — file-sharing URL for nonexistent groupId returns nothing | **No — already fixed on `main`** | Verified live: `POST /files/request-upload?groupId=999999&numberFiles=1` → `404 {"title":"Not Found","detail":"Group(s) 999999 do(es) not exist."}`. Unrelated to this PR; the issue appears stale and could be closed. |
| #5776 — file upload error handling / UX | **No** | This improves the network-failure text (`Network Error; no response received`) but adds none of the retry or progress work that issue asks for. |

Also filed #7106 while verifying repro cases for this PR — unrelated to this change, but in the same area: duplicate `submissionId` returns a correct 422 whose `detail` concatenates the raw `PSQLException`, exposing table name, constraint name and an internal upload UUID.

One note on this branch: it is based on `ce1cfcc81` and `main` has since moved to `97361c073` (#6922). No conflict — that commit changed `DataUploadForm.tsx`, which this PR does not touch, and `handleError` there is unchanged, so the before/after behaviour described above still holds.



<sub>Edited: the #6674 row originally claimed the backend returns an empty response. That was wrong — I had not tested that endpoint. Corrected after verifying it live.</sub>

PR #7119, opened 2026-08-20
fix(website): Fix applying file mappings for compressed metadata files

### LINKED ISSUE #7120: If file sharing is enabled, bulk submission of compressed metadata with files is broken
Currently, if an organism has files enabled, and someone does a bulk submission with files and a compressed metadata file, the file mapping cannot be correctly applied. Full summary from Claude:

## Root cause

- Bulk submission of a compressed metadata file (`metadata.tsv.gz`) for an organism with extra files enabled. `processRawFile` wraps it in a `CompressedFile`.
- The linkage UI works correctly: `FormOrUploadWrapper`'s effect calls `text()` on the **ProcessedFile**, which decompresses, so `parseSubmissionFileMapping` sees real TSV and the file linkage displays as expected.
- At submit, `fileFactory`'s bulk branch returned `metadataFile.inner()`. `CompressedFile extends RawFile` and does **not** override `inner()` — only `text()` decompresses — so this is the raw compressed `File`.
- `handleSubmit` passes that `File` to `applyFileMappings`, whose `await metadataFile.text()` is now the *native* `File.text()`: gzip bytes decoded as UTF-8.
- `Papa.parse` on those bytes finds no `files.` headers, so `fileColumns` is empty, the resolved category has no matching column, and it returns `Encountered unknown category rawReads not present in metadata.`
- Submission is blocked with a misleading message — the column is present, it just wasn't readable.

## Why it wasn't caught

- The two features were never tested together: `submission-compressed-files.spec.ts` uses CCHFV, which has no extra files, and the file-sharing tests all use uncompressed metadata.
- A column mapping masks it — `applyTo` decompresses and returns a plain `File`, so that path was always fine.
- Everything looked right up to the moment of submit, because the parse step used the decompressed path.

## Scope

- Requires bulk mode + compressed metadata + extra files enabled + at least one linked file; no column mapping.
- Introduced by #6922, which added the client-side metadata rewrite. Before that the browser never needed to read the metadata, so `inner()` returning compressed bytes was harmless.

tombch:
resolves #7120

### Summary

- Adjacent issue found from reviewing #7108.
- For organisms with files enabled, and for compressed metadata files with file uploads, the decompressed file was not passed by the file factory, leading to `applyFileMappings` to fail at writing to the file (as it cannot see the metadata headers) and the submission unable to continue.
- This PR updates the file factory logic to also check if files are enabled, and if so, decompresses the file for `applyFileMappings` to later read correctly.
- I also added an integration test for compressed metadata with files.

### Notes
- This is a simple fix but I feel like these functions need further refactoring. Ideally, the preparation of the metadata and sequence files should be in one place, rather than have some processing occur in the files factory in `FormOrUploadWrapper`, and further processing occur in the `DataUploadForm`'s `handleSubmit` function, which silently depends on actions like making sure the file is decompressed in the files factory. So I'd be happy to expand on this if we don't think this is a sufficient fix for now.

### PR Checklist


integration-tests/tests/specs/features/file-sharing.spec.ts
```diff
@@ -170,6 +170,35 @@ test('bulk submit 2 seqs with 1 & 2 FASTQ files respectively', async ({
...
+        [
+            [
+                ID_1,
+                COUNTRY_1,
+                '2022-12-02',
+                SEQUENCING_INSTRUMENT,
+                filesColumnCell(Object.keys(FILES_SINGLE), ID_1),
+            ],
+        ],
+        isGzipped,
```
corneliusroemer: You could potentially parametrize this test over gzipped? (If it's a 1-1 of an existing non-gzipped test case)

tombch: Yeah I removed the new test and parameterised two of the existing ones

tombch:
@claude review (please)

tombch:
Closing in favour of #7299

PR #7122, opened 2026-08-20
fix!(website): Restrict upload of xz compressed metadata TSVs in the website

### LINKED ISSUE #7121: xz-compressed metadata cannot be used with extra file submission
Uploading `metadata.tsv.xz` in bulk mode fails for organisms with extra files enabled.

When extra files are submitted, the browser has to read the metadata file and rewrite the `files.<category>` cells with the uploaded file IDs. `CompressedFile.text()` only supports `zst`, `gz` and `zip`, and throws for `xz` — so the read fails before submission. Without extra files the file is never opened in the browser, it goes straight to the backend, which decompresses xz fine.

Current behaviour: submission is rejected with

```
Could not read metadata.tsv.xz: xz files cannot be opened for editing.
```

Accurate, but it doesn't say the restriction comes from submitting extra files, so it isn't obvious that recompressing as `.gz` fixes it.

**Workaround:** use `.gz`, `.zst`, or an uncompressed TSV.

**Possible fixes:**
- Detect it up front and give a message naming the cause and the workaround.
- Support xz decompression in the browser (needs an LZMA/wasm dependency).
- Document the limitation alongside the other filename/format restrictions.

tombch:
resolves #7121 

### Breaking changes

- **This PR prevents xz-compressed metadata files of any form (not just excel) from being submitted via the website**. Users must either use a supported compression format for metadata, submit uncompressed metadata, or they can still submit xz-compressed metadata via the API. Sequence files can still be xz-compressed. 

### Summary

- The website currently allows xz-compressed metadata TSVs, as long as the user doesn't do any actions that involve reading the metadata file within the website (such as column mapping). For this reason, excel metadata files do not support xz compression as they require parsing by the website.
- However, with the updates to bulk submission with extra files brought in by #6922 , when extra files are enabled then the metadata TSV is _always_ parsed - to check for and display file linkage. Even if the user doesn't add any extra files, it still parses to display any reuse of file IDs.
- This means currently if a user uploads an xz-compressed metadata file with file sharing enabled for the organism, they get a 'xz files cannot be opened for editing' message. This message could be more informative, but IMO it would still be confusing that some organisms without file sharing support xz and others with file sharing do not.
- Therefore, this PR also restricts providing xz-compressed metadata TSVs, in line with the existing behaviour for metadata xlsx files. 

### PR Checklist

### DIFF
--- a/docs/src/content/docs/for-users/submit-sequences.md
+++ b/docs/src/content/docs/for-users/submit-sequences.md
@@ -87,4 +87,4 @@ As with the website, data will now be processed, and you will have to approve yo
 
 ### Compressing files
 
-For both sequence and metadata files, compression is supported. The supported formats are: `zip`, `gz` (gzip), `zst` (ZStandard) and `xz` (LZMA). (Note that Excel file uploads with `xz` compression are currently not supported.)
+For both sequence and metadata files, compression is supported. The supported formats are: `zip`, `gz` (gzip) and `zst` (ZStandard). `xz` (LZMA) is additionally supported for sequence files. Metadata files with `xz` compression are currently not supported when submitting through the website.
--- a/docs/src/content/docs/for-users/submit-sequences.md
+++ b/docs/src/content/docs/for-users/submit-sequences.md
@@ -87,4 +87,4 @@ As with the website, data will now be processed, and you will have to approve yo
 
 ### Compressing files
 
-For both sequence and metadata files, compression is supported. The supported formats are: `zip`, `gz` (gzip) and `zst` (ZStandard). `xz` (LZMA) is additionally supported for sequence files. Metadata files with `xz` compression are currently not supported when submitting through the website.
+For both sequence and metadata files, compression is supported. The supported formats are: `zip`, `gz` (gzip) and `zst` (ZStandard). `xz` (LZMA) is additionally supported for sequence files. Metadata files with `xz` compression are not supported when submitting through the website, but can be submitted through the API.
--- a/website/src/components/Submission/FileUpload/fileProcessing.ts
+++ b/website/src/components/Submission/FileUpload/fileProcessing.ts
@@ -31,7 +31,17 @@ export const METADATA_FILE_KIND: FileKind<ProcessedFile> = {
         const dataExtension = isCompressed ? fileNameParts[fileNameParts.length - 2] : extension;
         const compressionExtension = isCompressed ? extension : null;
         if (dataExtension === 'tsv' && !isCompressed) return ok(new RawFile(file));
-        if (dataExtension === 'tsv' && isCompressed) return ok(new CompressedFile(file));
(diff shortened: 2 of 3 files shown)


website/src/components/Submission/FileUpload/fileProcessing.ts
```diff
@@ -31,7 +31,17 @@ export const METADATA_FILE_KIND: FileKind<ProcessedFile> = {
         const dataExtension = isCompressed ? fileNameParts[fileNameParts.length - 2] : extension;
         const compressionExtension = isCompressed ? extension : null;
         if (dataExtension === 'tsv' && !isCompressed) return ok(new RawFile(file));
-        if (dataExtension === 'tsv' && isCompressed) return ok(new CompressedFile(file));
+        if (dataExtension === 'tsv' && isCompressed) {
+            if (compressionExtension === 'xz') {
+                return err(
+                    new Error(
+                        'LZMA compression (.xz files) is not supported with TSV files yet. ' +
```
theosanderson:
```suggestion
                        'LZMA compression (.xz files) is not supported for metadata files yet. ' +
```

IMO metadata is the relevant thing, not the fact that its a TSV

website/src/components/Submission/FileUpload/fileProcessing.ts
```diff
@@ -31,7 +31,17 @@ export const METADATA_FILE_KIND: FileKind<ProcessedFile> = {
...
+                            'Please use a different compression format for TSV files.',
```
theosanderson:
```suggestion
                        'LZMA compression (.xz files) is not supported for metadata files. '
```

IMO the relevant thing is that it's the metadata file (it presumably would be true for excel too). And the user doesn't need to use a different compression format, they could leave it uncompressed.

theosanderson (approved):
lgtm

docs/src/content/docs/for-users/submit-sequences.md
```diff
@@ -87,4 +87,4 @@ As with the website, data will now be processed, and you will have to approve yo
 
 ### Compressing files
 
-For both sequence and metadata files, compression is supported. The supported formats are: `zip`, `gz` (gzip), `zst` (ZStandard) and `xz` (LZMA). (Note that Excel file uploads with `xz` compression are currently not supported.)
+For both sequence and metadata files, compression is supported. The supported formats are: `zip`, `gz` (gzip) and `zst` (ZStandard). `xz` (LZMA) is additionally supported for sequence files. Metadata files with `xz` compression are currently not supported when submitting through the website.
```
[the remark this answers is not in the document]
tombch: I dont think it needs to explain the reason but I have updated the wording

website/src/components/Submission/FileUpload/fileProcessing.ts
tombch: Yeah I updated the wording for this!

PR #7124, opened 2026-08-20
ci: migrate setup-java from deprecated 'adopt' to 'temurin'

corneliusroemer-agent:
`actions/setup-java@v5` warns that the legacy `adopt` distribution is deprecated and will be removed in v6:

> Warning: The 'adopt' legacy AdoptOpenJDK distribution is deprecated and will be removed in setup-java v6. Please migrate to the 'temurin' distribution.

Switching to `temurin` also makes the step faster and less flaky. Temurin 21 is pre-installed on the `ubuntu-latest` runners, so setup-java gets a toolcache hit instead of resolving and downloading a JDK from a remote API. That remote lookup is what occasionally fails, e.g. [this run](https://github.com/loculus-project/loculus/actions/runs/32382027328/job/96467315728):

```
Trying to resolve the latest version from remote
Error: Java setup failed due to network or configuration error(s)
Error: connect ETIMEDOUT 104.17.75.81:443
```

Same vendor and Java version as before — Temurin is the continuation of AdoptOpenJDK — so no change to what the build actually runs on.

### DIFF
--- a/.github/workflows/backend-image.yml
+++ b/.github/workflows/backend-image.yml
@@ -81,7 +81,7 @@ jobs:
         uses: actions/setup-java@v5
         with:
           java-version: "21"
-          distribution: "adopt"
+          distribution: "temurin"
       - name: Setup Gradle
         if: env.CACHE_HIT == 'false'
         uses: gradle/actions/setup-gradle@v6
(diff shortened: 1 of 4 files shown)


maverbiest (approved):
This looks alright to me but to be honest I'm not very knowledgeable about the java ecosystem/distributions so maybe someone else wants to weight in.

corneliusroemer:
> This looks alright to me but to be honest I'm not very knowledgeable about the java ecosystem/distributions so maybe someone else wants to weight in.

Neither am I but Temurin is the continuation of Adopt and those distributions are pretty interoperable AFAIK

PR #7127, opened 2026-08-21
feat(raw-reads-processing): update deacon index to mask high quality PPX consensus sequence k-mers

### LINKED ISSUE #6759: MVP: Validate dehosting tool using example raw reads of ppx organisms and create pipeline for creating deacon index
ensure we have a pipeline that if we add more organisms we can check if the host validation would potentially start flagging virus data as human data and that we can update the dehosting library dictionary with this new virus

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/6759

builds on work I did in https://github.com/loculus-project/loculus/pull/6916/changes/968e314e8f88e80477b6c53bfe6cf802596f89d6, but now also filters out high quality PPX k-mers - claude helped me convert this to a python script as the bash script got too complex when I added in running nextclade and filtering out poor QC sequences. Additionally (see comments in https://loculus.slack.com/archives/C0ALH6CB9NY/p1787513362831529 an mpox clade 1b is added as clade 1bs almost all include a commonly incorrectly insertion while nextclade annotates them as poor quality - probably as we use a very general dataset that works for all sequences)

### Screenshot

Plot of k-mers from PPX consensus sequences flagged with our current deacon index (a good approximate for how many reads we may accidentally flag as human although they are in fact viral):
image

Results with this index (note organisms without mapped kmers are dropped)

image

### PR Checklist

### DIFF
--- a/.github/workflows/deacon-index.yml
+++ b/.github/workflows/deacon-index.yml
@@ -42,8 +42,7 @@ jobs:
       - name: Create deacon index
         shell: bash -l {0}
         run: |
-          curl -fL https://objectstorage.uk-london-1.oraclecloud.com/n/lrbvkel2wjot/b/human-genome-bucket/o/deacon/misc/panhuman-1.k31w15c8.idx -o "${INDEX_VERSIONED}"
-          ls -lh "${INDEX_VERSIONED}"
+          python3 ./raw-reads-processing/build-index.py -o "${INDEX_VERSIONED}"
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v4
--- a/.github/workflows/deacon-index.yml
+++ b/.github/workflows/deacon-index.yml
@@ -2,6 +2,9 @@ name: Create and Upload Custom Deacon Index
 
 on:
   workflow_dispatch: # Only manual trigger for now, since this is a rare operation
+  push:
+    branches:
+      - filter_low_qc
 
 env:
   DEACON_INDEX_IMAGE: ghcr.io/loculus-project/deacon-index
@@ -92,8 +95,8 @@ jobs:
         run: |
           s3cmd put "${INDEX_VERSIONED}" s3://loculus-public/deacon-index/"${INDEX_VERSIONED}"
 
-          s3cmd cp s3://loculus-public/deacon-index/"${INDEX_VERSIONED}" \
-            s3://loculus-public/deacon-index/"${INDEX_LATEST}"
+          # s3cmd cp s3://loculus-public/deacon-index/"${INDEX_VERSIONED}" \
+          #  s3://loculus-public/deacon-index/"${INDEX_LATEST}"
 
   notify_on_success:
     needs: build-and-upload
--- a/raw-reads-processing/build-index.py
+++ b/raw-reads-processing/build-index.py
@@ -17,7 +17,9 @@ from pathlib import Path
 
 import requests
 
-logging.basicConfig(level=logging.INFO, format="%(asctime)s [%(levelname)s] %(message)s")
+logging.basicConfig(
+    level=logging.INFO, format="%(asctime)s [%(levelname)s] %(message)s"
+)
 logger = logging.getLogger(__name__)
 
 PATHOPLEXUS_URL = "https://pathoplexus.org"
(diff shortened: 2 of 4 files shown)


raw-reads-processing/build-index.py
```diff
@@ -0,0 +1,215 @@
...
+                )
+
+                run_nextclade(organism_fasta, dataset_name, dataset_server, tmp_dir)
+
+                logger.info("Filtering FASTA for good sequences...")
+                good_indices = good_sequence_indices(tmp_dir / "nextclade.tsv")
+                logger.info(
+                    "Found %d good sequences for %s", len(good_indices), organism
+                )
+                filter_fasta_by_index(organism_fasta, good_indices, combined_fasta)
```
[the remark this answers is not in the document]
anna-parker: I dont think this really matters

raw-reads-processing/build-index.py
```diff
@@ -0,0 +1,215 @@
...
+        with dest.open("wb") as f:
+            for chunk in response.iter_content(chunk_size=1024 * 1024):
+                f.write(chunk)
+
+
+def parse_dataset_params(url: str) -> tuple[str, str]:
+    query = urllib.parse.parse_qs(urllib.parse.urlparse(url).query)
+    dataset_name = query.get("dataset-name", [""])[0]
+    dataset_server = query.get("dataset-server", [""])[0]
+    return dataset_name, dataset_server
```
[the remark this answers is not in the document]
anna-parker: dataset name is never empty

raw-reads-processing/build-index.py
corneliusroemer, line 1: I would just not filter out anything by QC, I don't see why this is necessary and it increases complexity. The script would be half length if we just use all sequences.

corneliusroemer (commented):
Why do we still get hits after removing PPX fasta kmers from the index? Do I misunderstand this plot? I understand it as those are the kmers that would still be flagged even after filtering out PPX consensus sequence kmers?

image

anna-parker:
@corneliusroemer the plots show what happens when only filtering kmers from high quality consensus sequences, so yes there are still some hits. We can filter out everything but then we are more likely to not flag human reads

raw-reads-processing/build-index.py
anna-parker: I had this before so its easy to get back, I just didnt think we would want to filter out all kmers as there are low quality sequences and also some patent sequences in the mix

raw-reads-processing/build-index.py
```diff
@@ -0,0 +1,215 @@
...
+    subprocess.run(args, check=True)  # noqa: S603, S607
+
+
+def good_sequence_indices(tsv_path: Path) -> set[str]:
+    """FASTA headers with qc.overallStatus == "good"."""
+    with tsv_path.open(newline="") as f:
+        reader = csv.DictReader(f, delimiter="\t")
+        return {
+            row["seqName"] for row in reader if row.get("qc.overallStatus") == "good"
+        }
```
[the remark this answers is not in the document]
anna-parker: the reverse complement stuff is invalid (I tested on nextclade output - there is just a column in the tsv with a boolean `isReverseComplement`.

.github/workflows/deacon-index.yml
anna-parker, line 1: revert changes in this file before merging

raw-reads-processing/build-index.py
corneliusroemer: I think we'd remove so few extra kmers that are genuinely not in viruses that it's not worth the extra complexity but I guess the code is there. 

The downside of filtering out some is that as you discovered it's sensitive to parameters - at first the QC filter dropped genuine kmers. Parameters have to be tuned for each organism which is a hassle.

What happens if we have a reference only dataset with an organism that's got large divergence?

anna-parker: Yes the extra complexity of having to tune is definitely an issue. I do think based on my findings with mpox just masking the reference is not sufficient. So we do want to mask more. As you say currently the additional few accidentally masked kmers is low enough that we dont really have to care, however I do worry that the more viruses we add this might start having an impact.

I think its easier to switch from masking less to masking more as we will end up flagging less instead of more. So I am a bit hesitant to start with masking everything. But end of the day probably not a big difference

anna-parker:
closing in favour of https://github.com/loculus-project/loculus/pull/7196

PR #7131, opened 2026-08-21
fix(website): pre-gate the submit button instead of toasting on click

corneliusroemer-agent:
For non-submittable states that the website knows about, it might be more user friendly to disable the "Submit" button with a message indicating why it's disabled instead of allowing the click and then toasting with the error message.

We still need toasts for errors that come from backend, but for those known by websites we don't need to toast.

Google Chrome 2026-08-21 20 30 50

Google Chrome 2026-08-21 20 31 27

## Summary

Fixes the `file-sharing.spec.ts` flake introduced in #6922 (see https://github.com/loculus-project/loculus/pull/6922#issuecomment-5372812467) that took out **both** integration-test jobs on main in [run 32494192105](https://github.com/loculus-project/loculus/actions/runs/32494192105) (chromium and firefox, same assertion), and generalises the underlying UX problem: we let the user click submit and then answer with a toast, when we already knew up front that submission was blocked.

## Root cause

`expectExtraFileUploaded` is a **no-op for a replaced file**. The test-id is keyed on the file path, and `getStatusIcon` renders the same `✓` for `previousUpload` as for `uploaded`:

```tsx
case 'previousUpload':
case 'uploaded':         return <span …>✓</span>;
```

So when `file2.fastq` is replaced, the row for `file2.fastq` still satisfies the assertion from its carried-over previous upload — it only flips to the `pending` spinner once `requestFileUploads` resolves, which is *after* the assertion has already passed. In the firefox run the assertion passed **14 ms** after clicking `Replace`, long before the upload had even been assigned a URL.

The test therefore clicked submit while the replacement was still in flight. #7050's new `validateFileUploadStates` correctly refused, so `'Do you really want to submit?'` never rendered. The toast it showed instead is visible in the failure snapshot:

```
- alert:
    - text: Please wait for all files to finish uploading before submitting.
```

Aligning the Playwright action timeline against the network log:

| job | submit click finished | upload finished | margin |
|---|---|---|---|
| chromium | t=459899 | t≈459905 | missed by **6 ms** |
| firefox | 15:03:22.14 | 15:03:22.541 | missed by **~400 ms** |
 
#7050's own PR CI passed all three integration-test jobs, and the next main commit (identical website code) passed too. Nothing was missed at review time; it is a genuine race that will keep recurring until fixed.

### Why it passed before #7050

Pre-#7050 the submit button went straight to `displayConfirmationDialog`, and the file ids were read from `fileMapping` only later, inside `submitEditedDataForAccessionVersion` — i.e. after `Confirm` was clicked, by which time the upload had long landed. The test was accidentally correct, and latently unsafe: clicking `Confirm` fast enough would have submitted the *old* `fileId` for the replaced file, silently. #7050 is right to block this; it just surfaced a bad test.

### 1. Pre-gate the submit button, and always say why

Rather than only fixing the assertion, make the blocked state unclickable. Playwright auto-waits for a button to become enabled, so no assertion can outrun the upload state machine — this closes the whole family of races, not just this instance.

Every blocker now disables the button **and** renders its reason next to it (`data-testid='submit-blocked-reason'`). An inline line rather than a tooltip, deliberately: native `title` is suppressed on disabled controls in Chrome, `HoverTooltip` needs a non-empty `content` so a conditional tooltip means conditional wrapping, and an always-visible line works on touch and for screen readers — which matters when the user cannot act on the blocker anyway.

`DataUploadForm` builds an ordered `submitBlockers: string[]`, shows the first, and disables on `length > 0`:

| condition | reason shown |
|---|---|
| any file category not `uploadCompleted` | Please wait for all files to finish uploading. |
| bulk mode, no metadata file yet | Please provide a metadata file. |
| bulk mode, metadata file still parsing | Please wait for the metadata file to finish processing. |
| bulk mode, metadata file failed to parse | the parse error verbatim |
| `!confirmedNoPII` | (existing PII message) |
| `!agreedToINSDCUploadTerms` | (existing INSDC message) |

`EditPage` gets the uploads gate only; the rest do not apply there.

**Scope of what is pre-gated:** everything derivable from state we already hold. Checks that can only run at submit time stay as toasts — anything behind the async `fileFactory()` (e.g. `'Please specify an ID.'`), and `applyFileMappings`. Linkage errors are also left as toasts even though `fileLinkage` is already memoised: `getLinkageErrors` returns a joined, multi-sentence string listing file paths per category, which would be a wall of text above the button, and it is user-actionable rather than "system busy". Happy to pre-gate that too if reviewers prefer.

Every pre-gate is still re-checked in `handleSubmit` as a safety net.

### 2. `submissionFileMapping` gains a `'processing'` state

Required to make the metadata-file reason honest. `undefined` previously meant *both* "no metadata file chosen" and "parsing in flight", so any single message would be wrong for half of users — and the pre-existing toast did exactly that, claiming the file was *"still being processed"* even when none had been selected. One extra `setSubmissionFileMapping('processing')` in the parsing effect, plus an exported `SubmissionFileMappingState` type.

### 3. Make `expectExtraFileUploaded` mean what it says

`data-upload-status={file.type}` on the status element, and the page object asserts on it:

- `expectExtraFileUploaded` → `'uploaded'` (uploaded *in this session*)
- new `expectExtraFileReused` → `'previousUpload'`, used for the reused-file assertion at `file-sharing.spec.ts:481` which legitimately wants a carried-over file

Note this alone would not have closed the race — a one-render gap remains between `uploaded` being painted and the effect promoting the category to `uploadCompleted` — which is why change 1 is the actual fix. But the assertion was lying and should not stay that way.

### 4. Single source of truth for the uploads message

`uploadsInProgressMessage` is exported from `fileUpload.ts` and used by both the button label and `validateFileUploadStates`' error, so the toast and the label cannot drift. The toast loses its trailing `"before submitting"`, which was redundant.

## Testing

One existing test changed: `SubmissionForm.spec.tsx` → *"should allow submission only after agreeing to terms of INSDC submission"* previously clicked submit and asserted a toast. It now asserts the button is disabled with the reason visible, then enabled after ticking the box — a closer reading of the test's own name.

## PR Checklist
- [x] Any manual testing that has been done is documented (i.e. what exactly was tested?) — see Testing above; browser verification is explicitly *not* done.

🚀 Preview: https://fix-gate-submit-button-on.loculus.org


corneliusroemer-agent:
Depends on #7132, which fixes the underlying UI bug this PR's flake story is built on: `addFiles` leaves a replaced file rendering its previous upload (tick included) until the request for upload urls returns.

That matters for the gate added here, not just for the test. During that window the category is still `uploadCompleted`, so `hasUploadsInProgress` is false and the submit button added in this PR stays enabled — a click there would submit the stale `fileId` with no error at all. #7132 closes that hole by entering `uploadInProgress` the moment the user confirms the replacement.

Will rebase this branch on top of #7132 once it lands; both touch `addFiles` and `fileUpload.ts`. The `data-upload-status` change here also drops from load-bearing to nice-to-have with #7132 in place — `previousUpload` still renders the same tick as `uploaded`, so `expectExtraFileUploaded` remains imprecise, just no longer racy.

PR #7132, opened 2026-08-21
fix(website): stop reporting a replaced file as uploaded while it's requesting upload url

corneliusroemer-agent:
Followup to #7050 which blocked submission while upload to s3 is in process but missed the (small) window between file drop and upload url being returned from backend.

While a replaced file is uploading, the UI keeps showing the *previous* file's upload as complete — green tick and `(uploaded)` label — for as long as the request for upload urls takes. This causes e2e test flakiness as we use the tick mark as proof that upload completed, eg [this run](https://github.com/loculus-project/loculus/actions/runs/32494192105).

<details>

`addFiles` does not touch the upload state until the round trip returns:

```js
const addFiles = async () => {
    const existingFiles = fileUploadState.files.filter((file) => !filePaths.has(file.path));
    const pendingFiles = await requestFileUploads(awaiting);   // ← nothing reflects the user's choice yet
    setFileUploadState({ type: 'uploadInProgress', files: [...existingFiles, ...pendingFiles] });
```

Between confirming **Replace** and the urls arriving, the category is still `uploadCompleted` and the row for the replaced path still holds its `previousUpload` entry. Consequences:

- The user is told the replacement is uploaded when it is not. Navigate away in that window and the old file is silently kept.
- Anything deriving "is this category settled?" from the state — including the submission checks added in #7050 — sees `uploadCompleted` and will let a submission through against a stale `fileId`.

## Changes

11 lines of production code across two files, five of them comment:

- `fileUpload.ts` — admit `Awaiting` into `SingleFileUpload`, and `'awaiting'` into `UploadStatus`.
- `FolderUploadComponent.tsx` — set `uploadInProgress` with the awaiting files as soon as the user confirms, before requesting urls; render `'awaiting'` with the spinner already used for `pending`; small `fileSizeOf` helper, since `Awaiting` carries `file: File` rather than a `size` field.

## Also fixes an integration-test flake

`expectExtraFileUploaded` asserts on the tick:

```ts
await expect(this.page.getByTestId(`status_${fileCategory}_${fileName}`)).toHaveText('✓');
```

and `getStatusIcon` renders the same `✓` for `previousUpload` as for `uploaded`. So for a file being *replaced*, that assertion is satisfied by the carried-over previous upload and returns immediately — **14 ms** after clicking Replace in one recent firefox run, long before the replacement had a url. The test then submitted mid-upload, #7050's `validateFileUploadStates` correctly refused, and the confirmation dialog never appeared.

</details>

## Testing

- New unit test in `FolderUploadComponent.spec.tsx` holds `requestMultipartUpload` pending and asserts only the untouched file is ticked mid-replacement, then that both are ticked once the urls arrive. Verified red without the fix (`Unable to find an element with the text: (7 B)`) and green with it.

### DIFF
--- a/website/src/components/Submission/FileUpload/FolderUploadComponent.tsx
+++ b/website/src/components/Submission/FileUpload/FolderUploadComponent.tsx
@@ -324,9 +324,6 @@ export const FolderUploadComponent: FC<FolderUploadComponentProps> = ({
             // Updates the state of file uploads and triggers the upload of the new files
             const addFiles = async () => {
                 const existingFiles = fileUploadState.files.filter((file) => !filePaths.has(file.path));
-                // Show the newly selected files straight away. Without this, a file being replaced
-                // keeps rendering its previous upload -- complete with tick -- for as long as the
-                // request for upload urls takes, claiming an upload that has not happened yet.
                 setFileUploadState({ type: 'uploadInProgress', files: [...existingFiles, ...awaiting] });
                 const pendingFiles = await requestFileUploads(awaiting);
                 setFileUploadState({
--- a/website/src/components/Submission/FileUpload/FolderUploadComponent.tsx
+++ b/website/src/components/Submission/FileUpload/FolderUploadComponent.tsx
@@ -173,7 +173,7 @@ export const FolderUploadComponent: FC<FolderUploadComponentProps> = ({
     async function requestFileUploads(filesAwaitingUrls: Awaiting[]): Promise<Pending[]> {
         const pendingFiles: Pending[] = [];
         for (const file of filesAwaitingUrls) {
-            const { partCount, partSize } = calculatePartSizeAndCount(file.file.size);
+            const { partCount, partSize } = calculatePartSizeAndCount(file.size);
             const result = await backendClient.requestMultipartUpload(accessToken, groupId, 1, partCount);
             result.match(
                 (data) => {
@@ -181,7 +181,7 @@ export const FolderUploadComponent: FC<FolderUploadComponentProps> = ({
                         type: 'pending',
                         file: file.file,
                         path: file.path,
-                        size: file.file.size,
+                        size: file.size,
                         fileId: data[0].fileId,
                         urls: data[0].urls,
                         uploadedParts: 0,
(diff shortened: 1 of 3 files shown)


corneliusroemer:
The flake fixed her just happened again in an integration test run.

website/src/components/Submission/FileUpload/FolderUploadComponent.tsx
```diff
@@ -324,6 +324,10 @@ export const FolderUploadComponent: FC<FolderUploadComponentProps> = ({
             // Updates the state of file uploads and triggers the upload of the new files
             const addFiles = async () => {
                 const existingFiles = fileUploadState.files.filter((file) => !filePaths.has(file.path));
+                // Show the newly selected files straight away. Without this, a file being replaced
+                // keeps rendering its previous upload -- complete with tick -- for as long as the
+                // request for upload urls takes, claiming an upload that has not happened yet.
```
corneliusroemer:
```suggestion
```

website/src/components/Submission/FileUpload/fileUpload.ts
```diff
@@ -44,9 +45,9 @@ type FileError = {
     msg: string;
 };
 
-export type SingleFileUpload = Pending | Uploaded | PreviousUpload | FileError;
+export type SingleFileUpload = Awaiting | Pending | Uploaded | PreviousUpload | FileError;
```
tombch, line 48: Not a blocking issue for this PR, but we now have `AwaitingUrlState` with all `awaiting` and then `UploadInProgressState` which can also include `awaiting`. We can also ignore the `AwaitingUrlState` entirely if we go straight to uploading files with `handleFileSelect`. I guess this confusing state already existed with `handleFileSelect`, it just wasn't shown because it never explicitly set any state for awaiting.

So to me this means we could remove the `AwaitingUrlState` and have two states - `UploadInProgress` and `UploadCompleted`. I can see if I can make a follow-up PR for this

tombch (approved):
Looks good to me

website/src/components/Submission/FileUpload/fileUpload.ts
corneliusroemer: I don't understand the file enough to comment 😬 sounds good if you want to make a follow-up, this here is mostly to avoid CI flakes

PR #7133, opened 2026-08-21
perf(backend): claim and store preprocessing results one statement per batch

clonder:
splits the cheap half out of #7075.

claiming work was one insert per record and storing a result one update per record,
and each statement fired a per statement tracker trigger that upserts the same
single row of `table_update_tracker`. with many workers everyone queues on that one
row lock.

both paths now run one statement per batch with unnest arrays. the triggers are per
statement, so they fire once per batch, which means no migration and no trigger
changes.

measured on 100k rows with 8 concurrent sessions, local postgres 15 with the real
trigger from v1.28:

| | wall |
|---|---:|
| one update per row | 41847 ms |
| one update per batch | 322 ms |

the per row variant ran its loop server side in plpgsql, so it doesn't even pay the
jdbc round trips. the 500k cluster runs inside #7075 (drain 2145 s to 832 s) are
contextual evidence only, that implementation also had the lease queue and dropped
the triggers instead of keeping them.

one behavior change: a request that contains the same accession version twice now
gets a 400 with a clear message. before, the second occurrence failed with a 422
"does not exist or is not awaiting any processing results".

### PR Checklist
  - no user facing change, the one api behavior change (400 on duplicates) is
    described above
  - existing submit and extract suites, a duplicate in one request test, and a
    cross batch test class running with stream batch size 1
  - the 100k measurement above, and the full backend suite locally on jdk 25,
    583 tests, 0 failures, 1 skipped


clonder:
@claude review the latest changes

backend/src/main/resources/application.properties
```diff
@@ -31,6 +31,8 @@ loculus.cleanup.task.run-every-seconds=60
 loculus.pipeline-version-upgrade-check.interval-seconds=10
 loculus.seqset-citations.run-every-minutes=360
 loculus.maintenance.clean-up-aux-table.run-every-hours=1
+# Also the number of processed results held in memory per submit request: a whole batch is
+# serialized before it is sent, so raising this multiplies peak heap by sequence size.
```
corneliusroemer:
```suggestion
```

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -459,37 +481,104 @@ class SubmissionDatabaseService(
...
+    /**
+     * One statement per batch updates a repeated accession version only once, and from an
+     * arbitrary one of the duplicates. [seen] spans the whole request, so a duplicate is
+     * rejected the same way wherever it falls relative to a batch boundary.
+     */
+    private fun rejectDuplicates(storedResults: List<StoredProcessedResult>, seen: MutableSet<AccessionVersion>) {
+        for (storedResult in storedResults) {
+            val accessionVersion = AccessionVersion(storedResult.submitted.accession, storedResult.submitted.version)
+            if (!seen.add(accessionVersion)) {
+                throw BadRequestException(
```
corneliusroemer: We throw `UnprocessableEntity` not `BadRequest` for data that's violating a requirement

backend/src/test/kotlin/org/loculus/backend/controller/submission/SubmitProcessedDataEndpointTest.kt
```diff
@@ -95,6 +95,21 @@ class SubmitProcessedDataEndpointTest(
         }
     }
 
+    @Test
+    fun `WHEN I submit the same accession version twice in one request THEN returns bad request`() {
```
corneliusroemer: Isn't this test exactly the same as ` fun `WHEN a duplicate accession version spans two batches THEN returns bad request`() {` above?

backend/src/test/kotlin/org/loculus/backend/controller/submission/SubmitProcessedDataAcrossBatchesTest.kt
corneliusroemer, line 1: Could those tests just live in `backend/src/test/kotlin/org/loculus/backend/controller/submission/SubmitProcessedDataEndpointTest.kt`

Also, they could be covered by existing tests there - by just setting batch size to 1 there? That would keep the coverage while reducing test time and number of things we need to maintain.

corneliusroemer (commented):
Looks good. I have a few minor comments.

fhennig:
I also had a look, nothing to add from my side.

backend/src/main/resources/application.properties
clonder: done, kept a one liner on the streamBatchSize param in the code instead so the heap tie in stays documented somewhere

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
clonder: good call, changed. turns out the old per row path also returned 422 for the second occurrence, so the status code now doesn't change at all, only the message gets clearer. updated the tests

backend/src/test/kotlin/org/loculus/backend/controller/submission/SubmitProcessedDataEndpointTest.kt
clonder: they ran under different batch sizes, this class uses the default 1000 so both copies landed in the same chunk, the other class used batch size 1 so the duplicate crossed a chunk boundary and only passed if the seen set survives across chunks. with the merge from your other comment both now live here, the 2 record one within a batch and the 3 record one across batches

backend/src/test/kotlin/org/loculus/backend/controller/submission/SubmitProcessedDataAcrossBatchesTest.kt
clonder: good idea, done with a twist: batch size 1 would make every store here a single row statement, so i set the whole class to 2 like ExtractUnprocessedDataEndpointTest does and made the duplicate test 3 records so it still spans a boundary. the extra class is gone

fhennig (approved):
Everything adressed, thank you!

PR #7136, opened 2026-08-23
fix(backend): drop stale compression_migration_checked_at declaration

corneliusroemer-agent:
Small fix removing a landminde: `compression_migration_checked_at` was added by `V1.20` and dropped again by `V1.21` (#5461), but the column was never removed from the Exposed declaration in `SequenceEntriesTable`

This would cause errors when using `SequenceEntriesTable.selectAll()` (that's how Claude noticed it).

I got Claude to audit all table declarations and this was the only table that was out of sync with `docs/db/schema.sql.`


anna-parker (approved):
great catch! I recall once looking into how we can keep our db table definitions in code in sync with the actual db tables (as defined by flyway), jpa has a validation command for this as we use exposed I think we would need sth like https://www.jetbrains.com/help/exposed/migrations.html#validating-the-database-schema

PR #7137, opened 2026-08-23
fix(backend): discard processed data whose sequence entry was edited

### LINKED ISSUE #7128: editing a sequence entry breaks in-flight preprocessing for it
editing a sequence entry drops its preprocessed data for all pipeline versions. so
any preprocessing thats already running for that sequence cant submit anymore, the
row it was working on is gone.

the preprocessing version isnt part of `editedSequenceEntryData`, so theres no
obvious way to scope the reset to the version being replaced.

not urgent. the sequence gets picked up and reprocessed anyway, it just makes the
worker fail. but a wart worth fixing.

found by @fhennig while reviewing #7075

corneliusroemer-agent:
Rethinking this PR I'm now not sure we need it - it only affects reprocessing where a 10min delay on a few sequences isn't a big deal.

Fixes #7128.

When a user edits a sequence entry (which is only allowed when the _current_ prepro version's entry is `PROCESSED`) we delet sequence_entries_preprocessed_data rows all pipeline versions of that accessionVersion. Preprocessing could already be underway for a a _higher_ prepro version than current, when that delete happens. When that pipeline then submits it would get a 422 for the whole batch which would only get reprocessed once the cleanup job clears them (10min for PPX).

This isn't so bad as it only happens during reprocessing - so not sure we need to make this change here.

### Changes

A result whose claim no longer exists is now discarded: WARN log, counted separately, and the summary line reads `Updated N sequences to PROCESSED, discarded M results whose claim was revoked`. The entry is handed out again and reprocessed, as it already was. Unknown accession versions and rows in an unexpected state still fail with 422.

`Updated 0, discarded 100` repeating would indicate a lease threshold set below batch duration.

Not addressed: the update has no fencing token, so two pods of the same pipeline version can still race an edit and land pre-edit data on a fresh claim. That needs a claim id in the protocol. However, this should be a very rare issue and it would resolve on the next reprocessing so it's not permanent.

### Alternatives considered

We could block editing a sequence when any entry greater or equal to current is `IN_PROCESSING` - however that would mean people could not edit while a reprocessing is in process which could be hours and is not a good user experience.

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
@@ -289,6 +289,7 @@ class SubmissionDatabaseService(
         log.info { "updating processed data" }
 
         val processedAccessionVersions = mutableListOf<String>()
+        val discardedAccessionVersions = mutableListOf<String>()
         val processedFiles = mutableMapOf<AccessionVersion, Set<FileId>>()
         val processingResultCounts = mutableMapOf<ProcessingResult, Int>()
         BufferedReader(InputStreamReader(inputStream)).use { reader ->
@@ -306,9 +307,12 @@ class SubmissionDatabaseService(
                 submittedProcessedDataBatch.forEach { submittedProcessedData ->
                     val processingResult = submittedProcessedData.processingResult()
 
-                    insertProcessedData(submittedProcessedData, organism, pipelineVersion)
-                    processedAccessionVersions.add(submittedProcessedData.displayAccessionVersion())
-                    processingResultCounts.merge(processingResult, 1, Int::plus)
+                    if (insertProcessedData(submittedProcessedData, organism, pipelineVersion)) {
+                        processedAccessionVersions.add(submittedProcessedData.displayAccessionVersion())
+                        processingResultCounts.merge(processingResult, 1, Int::plus)
+                    } else {
+                        discardedAccessionVersions.add(submittedProcessedData.displayAccessionVersion())
+                    }
                 }
             }
         }
@@ -327,7 +331,8 @@ class SubmissionDatabaseService(
         }
 
         log.info {
-            "Updated ${processedAccessionVersions.size} sequences to $PROCESSED. " +
+            "Updated ${processedAccessionVersions.size} sequences to $PROCESSED, " +
+                "discarded ${discardedAccessionVersions.size} results whose claim was revoked. " +
                 "Processing result counts: " +
                 processingResultCounts.entries.joinToString { "${it.key}=${it.value}" }
         }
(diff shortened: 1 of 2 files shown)


backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -459,11 +464,15 @@ class SubmissionDatabaseService(
         }
     }
 
+    /**
+     * Stores the result of one processing job. Returns false if the result was discarded because the
+     * pipeline's claim on the entry had been revoked while it was working on it.
+     */
```
corneliusroemer:
```suggestion
```

backend/src/main/kotlin/org/loculus/backend/service/submission/SubmissionDatabaseService.kt
```diff
@@ -489,8 +494,44 @@ class SubmissionDatabaseService(
...
+        return true
+    }
+
+    /**
+     * True if the sequence entry still exists but has no preprocessed data row for the given pipeline
+     * version, i.e. the claim was revoked while the pipeline was working on the entry: it was edited
+     * (which resets preprocessed data for all pipeline versions), the stale-in-processing cleanup ran,
+     * or outdated pipeline data was pruned. Results of a revoked claim are obsolete and can be dropped;
+     * the entry is picked up again by [fetchUnprocessedEntriesAndUpdateToInProcessing].
+     */
```
corneliusroemer:
```suggestion
```

backend/src/test/kotlin/org/loculus/backend/controller/submission/SubmitProcessedDataEndpointTest.kt
```diff
@@ -334,6 +336,63 @@ class SubmitProcessedDataEndpointTest(
...
+        )
+            .andExpect(status().isNoContent)
+
+        convenienceClient.getSequenceEntry(accession = accessionsInProcessing.first(), version = 1)
+            .assertStatusIs(Status.PROCESSED)
+        convenienceClient.getSequenceEntry(accession = accessionsWithoutClaim.first(), version = 1)
+            .assertStatusIs(Status.RECEIVED)
+    }
+
+    // A discarded result must not publish its files, they end up referenced by nothing.
```
corneliusroemer, line 358:
```suggestion
```

backend/src/test/kotlin/org/loculus/backend/controller/submission/SubmitProcessedDataEndpointTest.kt
```diff
@@ -334,6 +336,63 @@ class SubmitProcessedDataEndpointTest(
...
+        submissionControllerClient.submitProcessedData(
+            PreparedProcessedData.withFiles(
+                accession,
+                mapOf("myFileCategory" to listOf(FileIdAndName(fileV1.fileId, "v1.txt"))),
+            ),
+            pipelineVersion = 1,
+        ).andExpect(status().isNoContent)
+        convenienceClient.approveProcessedSequenceEntries(listOf(AccessionVersion(accession, 1)))
+
+        // Pipeline version 2 submits a result for the released entry without holding a claim on it.
```
corneliusroemer, line 378:
```suggestion
```

PR #7138, opened 2026-08-24
feat!(backend): prevent db state divergence between truth and code

### LINKED ISSUE #5497: submission_id is text in postgres but varchar 255 in kotlin
Similar to https://github.com/loculus-project/loculus/issues/5493, pointed out by @corneliusroemer in https://github.com/loculus-project/loculus/pull/5398#discussion_r2545396865

These should all be text not varchar 255 - the db supports full 255 and we don't validate the length is <255

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/5497, https://github.com/loculus-project/loculus/issues/1517

Raised when reviewing https://github.com/loculus-project/loculus/pull/7136 - we should add automated checks to ensure flyway migration (true db state) and backend ORM (exposed) defined table state do not get out of sync.

So many things are already out of sync!!!!

This is terrible... exposed cant handle postgres Types so we cant actually check that the column types are the same in code and postgres without hand-rolling a check: https://www.jetbrains.com/help/exposed/migrations.html

### PR Checklist
If this validation fails the backend will no longer build
I confirmed that if I add a new column to the flyway migration and not to the code (or the reverse) the tests all fail due to this new validation.


anna-parker:
@claude review

backend/src/main/resources/db/migration/V1.35__add_missing_not_null_constraints.sql
anna-parker, line 1: The code set these as NOT NULL but this wasnt enforced by the db - we should probably check on PPX if these columns are really all not null before adding this.

anna-parker: Tested on staging (in a transaction that I later rolled back) that this is successful!

backend/src/main/kotlin/org/loculus/backend/config/BackendSpringConfig.kt
```diff
@@ -161,10 +161,12 @@ class FlywayInit(
 
         flyway.migrate()
 
-        // Since migration V1.10 we need to initialize the CurrentProcessingPipelineTable
-        // in code, because the configured organisms are not known in the SQL table definitions.
-        logger.info("Initializing CurrentProcessingPipelineTable")
         transaction {
+            checkExposedSchemaMatchesDatabase()
```
corneliusroemer, line 165: You want to do this not just in ci/tests but error the whole backend?

backend/src/main/kotlin/org/loculus/backend/service/submission/dbtables/SequenceEntriesView.kt
```diff
@@ -20,11 +20,12 @@ import org.loculus.backend.api.ProcessingResult
 import org.loculus.backend.api.Status
 import org.loculus.backend.api.SubmittedData
 import org.loculus.backend.api.toPairs
+import org.loculus.backend.config.NotFlywayManaged
 import org.loculus.backend.service.jacksonSerializableJsonb
 
 const val SEQUENCE_ENTRIES_VIEW_NAME = "sequence_entries_view"
 
-object SequenceEntriesView : Table(SEQUENCE_ENTRIES_VIEW_NAME) {
+object SequenceEntriesView : Table(SEQUENCE_ENTRIES_VIEW_NAME), NotFlywayManaged {
```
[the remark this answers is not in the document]
corneliusroemer: Yeah we should check this view too

backend/src/main/kotlin/org/loculus/backend/config/BackendSpringConfig.kt
corneliusroemer: Should we maybe (also or only) run this as part of the test suite?

corneliusroemer: Ah apparently it _is_ run in tests as well (on every test actually):
> One correction to my earlier review: the check is exercised by the backend test suite. @EndpointTest activates profile with-database, not test, so FlywayInit (@Profile("!test")) runs — that's how the test schema gets created in the first place. Every Spring context in backend-tests therefore runs checkExposedSchemaMatchesDatabase(). That's decent implicit coverage; a focused negative test would still be nice but isn't the gap I claimed it was.

Might be worth adding a test that checks the check itself, adding a column that doesn't exist in db, remove a column that exists in db etc

PR #7139, opened 2026-08-24
test(backend): assert Exposed declarations match the migrated schema

corneliusroemer-agent:
Follow-up to #7136, which removed a stale Exposed column declaration that had gone unnoticed since the migration dropping the column. This adds the test that would have caught it — and, as of the latest commit, catches type and nullability drift too.

## What it checks

It discovers every Exposed `Table` object on the classpath (19 today, including `sequence_entries_view`) and compares each against the Flyway-migrated test database on four axes: relation exists, column exists (both directions), column type, and nullability.

Types are compared by materialising the Exposed declarations as real tables in a throwaway `exposed_expected` schema and diffing the two schemas' catalogs, rather than by hand-mapping Exposed's SQL type strings onto what `information_schema` reports. Postgres does the normalisation, which is the only way `character varying(255)` vs `text` and `text[]` vs `integer[]` stay comparable without a mapping table that rots.

`format_type(atttypid, atttypmod)` is used rather than `information_schema.data_type`, which drops the type modifier and collapses every array type to the literal string `ARRAY` — `fasta_ids` would otherwise be unverifiable.

Not compared: default values (too fragile to match as SQL text), indices, and constraint names.

## Two traps worth recording

**Views are always nullable.** Postgres reports `is_nullable = YES` for every column of a view regardless of what the underlying table declares, so nullability is masked for view-backed relations. Types are still compared, so `sequence_entries_view` — redefined by seven migrations — stays covered rather than exempted.

**`descriptionDdl()` omits `NOT NULL` for primary-key columns**, and only inlines `PRIMARY KEY` for single-column keys (Exposed 1.4.0, `Column.kt:169-184`). For a composite key — `sequence_entries`, `external_metadata`, `metadata_upload_aux_table`, `sequence_entries_preprocessed_data`, `seqsets` — that means neither, so the constraint has to be restated or every key column is created nullable in the shadow schema and the nullability comparison passes vacuously on exactly the tables that matter most.

The expected table list is now asserted exactly rather than by count. The classpath scan only picks up tables declared as a Kotlin `object`; anything else is skipped silently, and a table quietly dropping out of the check is the failure this test exists to prevent.

## The 48 pre-existing mismatches

Turning on type and nullability comparison surfaces 48 disagreements between the Kotlin and the database. They are recorded in `KNOWN_DRIFT` as a baseline to burn down, so the test can start guarding against *new* drift without blocking on clearing the backlog first. It is a baseline, not a list of approved exceptions.

By group:

- **32** — Exposed declares `varchar(255)` where the migrations created `text`. Mostly harmless, but an over-255-character insert would be rejected by Exposed rather than by the database.
- **9** — the mirror image, all in `groups_table`: Exposed says `text`, the database says `varchar(255)`.
- **3** — nullability only, type agrees: `external_metadata.updated_metadata_at`, `metadata_upload_aux_table.metadata`, `sequence_entries_preprocessed_data.started_processing_at`.
- **4** — genuine type mismatches that no varchar/text sweep would surface:
  - `compression_dictionaries.created_at` — Exposed `timestamp without time zone`, database `timestamp with time zone`
  - `data_use_terms_table.restricted_until` — Exposed `date`, database `timestamp without time zone`
  - `sequence_upload_aux_table.compressed_sequence_data` — Exposed `jsonb`, database `text`
  - `table_update_tracker.last_time_updated` — Exposed `text`, database `timestamp without time zone`

#7138 fixes a large share of the first three groups. The four genuine type mismatches are not addressed by any open PR.

## Relationship to #7138

#7138 solves the same problem as a startup gate, using Exposed's `SchemaUtils.addMissingColumnsStatements` + `MigrationUtils.dropUnmappedColumnsStatements`. Two differences worth weighing:

- On Postgres those helpers do **no** type comparison at all. `SchemaUtilityApi.kt:371` only computes `incorrectType` when `currentDialect.supportsColumnTypeChange`, which is `false` in the base dialect and overridden to `true` only by H2. What surfaces there is presence, nullability, and varchar-length-when-Exposed-declares-one (via `isIncorrectSizeOrScale`) — which is why the `varchar(255)`→`text` direction is caught and the `groups_table` direction is not.
- The two approaches are complementary rather than competing: #7138 carries the cleanup, this carries the stronger comparison and covers the view. Happy to fold this query into #7138 instead of landing it separately if that's preferred.

## Verification

- Verified by re-adding the declaration #7136 removed: the test fails with `["sequence_entries.compression_migration_checked_at"]`.
- Verified that dropping the composite-PK handling makes the nullability half pass vacuously, and that masking view nullability is required (a view over a `NOT NULL` column reports `is_nullable = YES`).
- Full backend suite green: **582 tests, 75 classes, 0 failures**. `ktlintCheck` clean.

Note on the earlier revision of this description: `SchemaUtils.statementsRequiredToActualizeScheme` was ruled out because it emits 57 statements against a consistent schema (varchar width vs text, defaults, FK constraint naming). The shadow-schema diff replaces that reasoning — it gets the same normalisation for free by asking Postgres.

### DIFF
--- a/backend/src/test/kotlin/org/loculus/backend/service/SchemaConsistencyTest.kt
+++ b/backend/src/test/kotlin/org/loculus/backend/service/SchemaConsistencyTest.kt
@@ -2,69 +2,106 @@ package org.loculus.backend.service
 
 import org.assertj.core.api.Assertions.assertThat
 import org.jetbrains.exposed.v1.core.Table
+import org.jetbrains.exposed.v1.core.statements.StatementType
 import org.jetbrains.exposed.v1.jdbc.transactions.transaction
 import org.junit.jupiter.api.Test
 import org.loculus.backend.controller.EndpointTest
 import org.springframework.core.io.support.PathMatchingResourcePatternResolver
 import org.springframework.core.type.classreading.CachingMetadataReaderFactory
 
+/**
+ * Asserts that the Exposed [Table] declarations agree with the schema the Flyway migrations actually
+ * produce: same relations, same columns, same types, same nullability.
+ *
+ * Types are compared by materialising the Exposed declarations as real tables in a throwaway schema
+ * and diffing the two schemas' catalogs, rather than by mapping Exposed's SQL type strings onto what
+ * `information_schema` reports. That keeps Postgres in charge of normalisation, which is what makes
+ * `character varying(255)` vs `text` and `text[]` vs `integer[]` comparable at all.
+ *
+ * Not compared: default values (too fragile to match as SQL text), indices, and constraint names.
+ */
 @EndpointTest
 class SchemaConsistencyTest {
 
     @Test
-    fun `GIVEN the migrated database THEN every declared column exists in it`() {
-        val existingColumns = existingColumns()
+    fun `GIVEN the migrated database THEN every declared relation exists in it`() {
+        val existingRelations = transaction {
+            val names = mutableSetOf<String>()
+            exec(
+                "SELECT table_name FROM information_schema.tables WHERE table_schema = current_schema()",
+            ) { resultSet ->
+                while (resultSet.next()) {
+                    names.add(resultSet.getString("table_name"))
+                }
+            }
+            names
+        }!!
 
-        val missing = declaredTables().flatMap { table ->
-            val existing = existingColumns[table.tableName]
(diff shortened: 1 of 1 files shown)


anna-parker:
alternative in https://github.com/loculus-project/loculus/pull/7138

Issue #6096, opened 2026-03-05
Integration tests are flaky


anna-parker:
I am seeing the file sharing and bulk revision tests failing often failing again due to flake this is very tedious to have to constantly retry: 
`bulk revise 2 seqs with files`
[firefox-without-dep] › tests/specs/features/revise-sequence.spec.ts:84:14 › Bulk sequence revision › can revise multiple sequences via file upload see: https://github.com/loculus-project/loculus/actions/runs/22687998132/job/65787349094

PR #7155, opened 2026-08-25
fix(website): derive the file mapping instead of maintaining a copy of it

corneliusroemer-agent:
Follow-up to #7132, and related to the umbrella flake issue #6096.

I've noticed a integration test failure that looked like this (there were actually 5 instances of this in the past few days):

```
1) [chromium-without-dep] › tests/specs/features/file-sharing.spec.ts:451:5 › single revise seq via edit page reuses, replaces, discards and adds files 

    Error: expect(received).toBe(expected) // Object.is equality

    - Expected  - 1
    + Received  + 1

    - @SRR123456.1 1/1
    + @SRR123456.1 1/2
      ATGGATAAACGGGTGAGAGGTTCATGGGCCCTGGGAGGACAATCTGAAGTTGATCTTGACTACCACAAAA
      +
      IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII
      ↵

       at utils/link-helpers.ts:20

      18 |     expect(response.status()).toBe(200);
      19 |     const content = await response.text();
    > 20 |     expect(content).toBe(expectedContent);
```

(https://github.com/loculus-project/loculus/actions/runs/32844888050/job/97792281964)

This refactor here should fix the flakes (by fixing the underlying bug) and also how the `fileMapping` is derived, simplifying flow.

One can probably drop the unit tests - not sure how much they add. I guess they're good to have for Claude to check its work but not sure they are necessary.

<details>
<summary>Claude PR desc</summary>

## Why

Submitting is gated on `validateFileUploadStates(fileUploadStates)`, but the file mapping that gets submitted alongside it was a *different* value: `fileMapping`, which `FolderUploadComponent` kept in sync from an effect. An effect runs after the render that triggered it, so the mapping was always written a render after the state it describes, and the two could disagree at the moment someone pressed submit. Nothing needed it to be state — both parents already hold `fileUploadStates`, and the mapping is a pure function of it.

Three places where they could disagree:

- The `uploadInProgress` to `uploadCompleted` promotion and the mapping write happen in two separate effect passes, so there is a render with the new state and the previous mapping. Submitting there submits a replaced file under the id of the file it replaced — the user publishes the old file and nothing reports an error.
- On the edit page `fileUploadStates` is seeded from previous uploads by the `useState` initialiser, while `fileMapping` started as `undefined`. Submitting before the first effect took the "no files" branch and dropped every file from the revision.
- Discarding a file wrote the new state a render before the mapping caught up, so a discarded file could still be submitted.

Making it worse, the confirmation dialog captures the submit function at click time, so a mapping that was stale when the button was pressed stayed stale however long the user then took to press Confirm.

## What you may want to argue with

**The bulk linkage display changes.** While files are being added to a category that was already complete, that category's declared files now count as `missing`, so the status line reads "referenced in metadata but not uploaded" until the upload finishes. Previously it kept showing the previous upload's links throughout. I think showing the current truth is right, and it is the same class of stale affordance #7132 removed from the status tick, but it is a visible change and it is a judgement call — say so if you would prefer the old display and I will keep a stored copy for that one consumer. Submit-time linkage errors are unaffected, since validation already refuses to submit mid-upload.

## Deliberately not done

- A test that submits from `EditPage` or `DataUploadForm` with files and asserts the resulting ids. None exists, so this wiring was uncovered before this PR and still is; the new unit tests cover the derivation only. I checked equivalence by reading instead: the effect wrote `new Map(files.map(f => [f.path, f.fileId]))` under `fileCategory.name`, and `fileUploadStates` is keyed by that same name.
- `expectExtraFileUploaded` still asserts on a `✓`. #7132 made that assertion meaningful again by rendering a spinner for a file being replaced, so it no longer passes vacuously, but it would be sturdier asserting on the upload state directly.

### Screenshot

n/a, no visual change outside the bulk linkage status line described above.

### PR Checklist

Unit tests only: four new cases for the derivation, and `FolderUploadComponent.spec.tsx` reworked so its harness derives the mapping the way a parent does, keeping its existing assertions meaningful. Full website suite passes (752 tests), types and format clean. No manual browser testing.

</details>

🚀 Preview: https://fix-derive-file-mapping-a.loculus.org

### DIFF
--- a/website/src/components/Edit/EditPage.tsx
+++ b/website/src/components/Edit/EditPage.tsx
@@ -1,5 +1,5 @@
 import { isErrorFromAlias } from '@zodios/core';
-import { type FC, useState } from 'react';
+import { type FC, useMemo, useState } from 'react';
 import { toast } from 'react-toastify';
 
 import { EditableSequences } from './EditableSequences.ts';
@@ -18,12 +18,9 @@ import { getAccessionVersionString, parseAccessionVersionFromString } from '../.
 import { displayConfirmationDialog } from '../ConfirmationDialog.tsx';
 import { SequenceEntryHistoryMenu } from '../SequenceDetailsPage/SequenceEntryHistoryMenu.tsx';
 import { ExtraFilesUpload } from '../Submission/DataUploadForm.tsx';
+import { applyFileMappings, getSingleSubmissionFileMapping } from '../Submission/FileUpload/fileMapping.ts';
 import {
-    applyFileMappings,
-    getSingleSubmissionFileMapping,
-    type FileMapping,
-} from '../Submission/FileUpload/fileMapping.ts';
-import {
+    deriveFileMapping,
     getPreviousFileUploadStates,
     validateFileUploadStates,
     type FileUploadState,
@@ -76,7 +73,8 @@ const InnerEditPage: FC<EditPageProps> = ({
     const [fileUploadStates, setFileUploadStates] = useState<Map<string, FileUploadState>>(() =>
         dataToEdit.submittedData.files ? getPreviousFileUploadStates(dataToEdit.submittedData.files) : new Map(),
     );
-    const [fileMapping, setFileMapping] = useState<FileMapping | undefined>(undefined);
+
+    const fileMapping = useMemo(() => deriveFileMapping(fileUploadStates), [fileUploadStates]);
 
     const isCreatingRevision = dataToEdit.status === approvedForReleaseStatus;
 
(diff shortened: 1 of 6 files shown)


tombch:
One thing to mention is this would change how the file linkage is displayed to users - currently, with file mapping in state it only gets updated each time the file upload state is in `uploadCompleted`. So e.g. if a user uploads some files, it will tell them theyre linked, then if they start uploading more, it will keep the previous linkage until it gets updated. Now it will temporarily go back to 'all files unlinked' every time the file upload state is in progress.

I do agree though it would be nice to not have two separate bits of state, so maybe there's a way to make both work. Because the existing ux is also not ideal (ideally it would live-update an increasing number of files being linked)

tombch (approved):
Looks good to me!

PR #7169, opened 2026-08-26
fix(deployment): don't CPU limit raw-reads-processing saving 5min and reducing flakes

corneliusroemer:
We got integration test flakes because e2e values.yaml limited CPU of raw reads processing to 20m (because it was using default resources - there was no entry for it in default values.yaml

This PR fixes that situation by:
- explicitly setting resource requests directly for raw-reads-processing in values.yaml (sized to work with the big deacon index)
- giving e2e values resources with reduced memory requests sized to the smaller index used there
- Upping requests to 50m CPU as it's a central service that shouldn't be CPU limited so we don't get stuck/slow

The cpu limit caused every readtools invocation to take 30s rather than 0.2s. So this PR will speed up integration tests by quite a bit.

CI flakes caused by this 20m CPU limit looked like this:

```
 1) [chromium-without-dep] › tests/specs/features/file-sharing.spec.ts:451:5 › single revise seq via edit page reuses, replaces, discards and adds files

    Error: expect(locator).toContainText(expected) failed

    Locator: locator('[data-testid="review-page-control-panel"]')
    Expected substring: "0 awaiting processing"
    Received string:    "0 of 1 sequences processed1  awaiting processing0  no issues0  with warnings0  with errors"
    Timeout: 90000ms

    Call log:
      - Expect "toContainText" with timeout 90000ms
      - waiting for locator('[data-testid="review-page-control-panel"]')
        180 × locator resolved to <div class="flex flex-col" data-testid="review-page-control-panel">…</div>
            - unexpected value "0 of 1 sequences processed1  awaiting processing0  no issues0  with warnings0  with errors"
```

Example run with this failure: https://github.com/loculus-project/loculus/actions/runs/32958977145/job/98146909327)


anna-parker (approved):

corneliusroemer:
Will wait with merge until I have CI runs - this requires Github to be healthy again

PR #7181, opened 2026-08-26
chore(deployment): raise default CPU limit to 500m

corneliusroemer-agent:
Fix the footgun of small default cpu limits by default, discovered through https://github.com/loculus-project/loculus/pull/7169


theosanderson (approved):
yeah, happy with removing or relaxing - either is great!

corneliusroemer:
I've switched to 500m as discussed - the pr draft came from before you suggested 500m which I agree is a good default for now.

PR #6712, opened 2026-06-18
feat(ingest): allow for muting of specific hashes

maverbiest:
This PR adds functionality to ingest to download a tsv with manually specified hashes per accession that should not trigger a revision and should not cause slack alerts to be sent.

A new ingest configuration field `muted_hashes_url` points to where the tsv should be downloaded from. 

## Manual testing

The extended tests cover the muting behaviour. I also ran the new snakemake rule to download the file and confirmed it works:

```sh
(loculus-ingest) ➜  ingest git:(ingest-ignore-hashes) snakemake download_muted_hashes --config muted_hashes_url="https://raw.githubusercontent.com/maverbiest/curation_data_max/refs/heads/add-muted-hashes/muted_hashes/muted_hashes.tsv"

(loculus-ingest) ➜  ingest git:(ingest-ignore-hashes) cat results/muted_hashes.tsv
accession       hash_digest     date    comment
PP_006VZ20      662aca579d3e82278d03298cc24dcd45        2026-06-18
PP_006VZ3Y      1f1537fe97fd4ac9e49f88b0dc256c96        2026-06-18
PP_006VZ4W      c687c18c1af9ebdb0b08f2c2fc81f05f        2026-06-18
PP_006VZ5U      f06b643a7ff71141b26eb83acc5448dd        2026-06-18
PP_006VZ6S      48c419a1f4536dfd11dd33cf7a204573        2026-06-18
PP_006VZ7Q      4925d516a076a9964b28b2c874690f16        2026-06-18
```

### PR Checklist
- [x] All necessary documentation has been adapted.
- [x] The implemented feature is covered by appropriate, automated tests.
- [x] Any manual testing that has been done is documented (i.e. what exactly was tested?)

🚀 Preview: https://ingest-ignore-hashes.loculus.org

PR #7182, opened 2026-08-26
fix(config): make the default ingest configFile actually reach organisms

corneliusroemer-agent:
The `muted_hashes_url` added in #6712 has never applied to any organism. It sits on the `&ingest` anchor's `configFile`, and every one of the five organisms that run ingest declares its own `configFile`, which replaces the anchor's rather than extending it — a YAML merge key does not recurse. So the hash muting feature has been configured but inert since it merged on 2026-06-18.

The fix names the inner map `&defaultIngestConfigFile` and merges it a level down in each organism, which is what the Pathoplexus values already do for thirteen of their fifteen organisms.

The same bug exists in the Pathoplexus values for dengue and marburg, fixed separately in pathoplexus/pathoplexus#1123.


anna-parker (approved):
amazing thanks!

PR #7189, opened 2026-08-26
fix(deployment): make CI path filters match what the workflows actually exercise

corneliusroemer-agent:
I noticed that loculus-project/loculus#7188 changes only `raw-reads-processing/`, and got no integration test run. It turns out that various Github workflows have not had their path filters updated for various new services. This PR fixes that.


.github/workflows/docs-test.yml
```diff
@@ -3,7 +3,8 @@ on:
   pull_request:
     paths:
       - "docs/**"
-      - .github/workflows/docs-build-check.yml
+      - "kubernetes/loculus/values.schema.json"
```
maverbiest, line 6: Ah nice, I never looked at the loculus docs page that has all the values from the schema

maverbiest (approved):

PR #7191, opened 2026-08-27
fix(website): prevent permanent cursor-waiting in BoxWithTabsTab

maverbiest:
The new  `RawReadsMetadataTable` component has a  `BoxWithTabsTab` without `client:*` so astro server-side renders it to static HTML.

However, this tab will still contain a `Button`, which computes `hydrating`:

https://github.com/loculus-project/loculus/blob/ec240f0763fc50c23ae1bfb5bf22e3aac8a59316/website/src/components/common/Button.tsx#L53-L56

And adds `cursor-wait` as long as hydrating is true:

https://github.com/loculus-project/loculus/blob/ec240f0763fc50c23ae1bfb5bf22e3aac8a59316/website/src/components/common/Button.tsx#L67

Hydrating can never become `false` since there's no JS, just HTML, so `cursor-wait` stays forever.

You can see the issue by hovering over the 'Raw Reads Metadata Fields' tab here: https://preview-raw-reads-feature.pathoplexus.org/docs/how-to/upload-raw-reads 
(or see @tombch 's video capture in a comment below)

## Fix

This PR adds `client:load` to `BoxWithTabsTab` in `RawReadsMetadataTable.astro` so it becomes an interactive island and `hydrating` will become false

## Alternative approach

I initially made it so `BoxWithTabsTab` check wether it is passed an `onClick` in its props (hence Claude's comments below). If the `onClick` was present, it rendered a `Button`, otherwise a simple `div` with the label. However, the diff for that approach was quite a bit bigger than the current approach.

## Manual testing

I bumped the raw reads feature brach https://github.com/pathoplexus/pathoplexus/pull/1105 to the loculus commit in this PR locally and ran the website locally to confirm the `cursor-wait` is gone.

https://github.com/user-attachments/assets/56813832-0322-4b84-a3c3-0421d0c1726e

### PR Checklist


tombch:
https://github.com/user-attachments/assets/18a96c2c-4580-4bd7-bf56-2e912133c3cf

I just went to this page on firefox and I can see a 'loading' cursor when I hover over? Is this the same issue?

maverbiest:
@tombch which page is that, sorry? If it's https://preview-raw-reads-feature.pathoplexus.org/docs/how-to/upload-raw-reads page then yes that's the issue this PR is aimed at addressing. (we'd need to bump the loculus version of the PPX preview before the fix would be visible there, that's why it's still live there)

anna-parker (approved):

PR #7200, opened 2026-08-28
feat(website): error toast when a metadata file is dropped into the sequence field

### LINKED ISSUE #7197: Error when someone drops a .tsv file (or similar non fasta-like) into sequence box
Just noticed that one can drop a .tsv file into the sequences box and it doesn't error:

Image

On this page: https://pathoplexus.org/mpox/submission/232/revise?inputMode=bulk

corneliusroemer-agent:
Closes #7197

Report error when dropping a `.tsv` or `xls(x)` into the sequence file field.

Error is only thrown on file-types we accept as metadata. So one can still put pretty much anything into sequence field.

Left out on purpose:

- `.csv` is still accepted in the sequence field. It is not in the metadata field's accepted list, so under the "only reject what metadata takes" rule it falls outside, even though a dropped CSV is just as wrong.
- The "Add a segment" box on the edit page is untouched. A `.tsv` dropped there still quietly becomes one nonsense segment, but there is no metadata box next to it to confuse it with, so the swap this issue is about cannot happen.

image


website/src/components/Submission/FileUpload/fileProcessing.ts
```diff
@@ -70,7 +77,21 @@ export const FASTA_FILE_KIND: FileKind<ProcessedFile> = {
     type: 'fasta',
     icon: PhDnaLight,
     supportedExtensions: ['fasta'],
-    processRawFile: (file) => Promise.resolve(ok(new RawFile(file))),
+    processRawFile: (file) => {
+        // Permissive about file names, but a metadata-only extension means the upload fields were swapped.
```
corneliusroemer:
```suggestion
```

theosanderson (approved):
lgtm

PR #7198, opened 2026-08-28
fix(e2e): Eliminate race condition by asserting row-count=1, refactor shared logic

corneliusroemer:
Fixes some integration test flakes that have come up where we click without checking that the filter has been applied yet.

Flakes look like this:

```
 1) [firefox-without-dep] › tests/specs/features/sequence-version-banners.spec.ts:39:9 › Sequence version banners › shows correct banners for revoked, revocation, deprecated, and revised entries

     Error: locator.textContent: Error: strict mode violation: getByRole('link', { name: /LOC_[A-Z0-9]+\.1/ }) resolved to 2 elements:
         1) <a href="/seq/LOC_00001E5.1" class="text-primary-900 hover:text-primary-800 hover:no-underline">LOC_00001E5.1</a> aka getByRole('link', { name: 'LOC_00001E5.1' })
         2) <a href="/seq/LOC_00001D7.1" class="text-primary-900 hover:text-primary-800 hover:no-underline">LOC_00001D7.1</a> aka getByRole('link', { name: 'LOC_00001D7.1' })

     Call log:
       - waiting for getByRole('link', { name: /LOC_[A-Z0-9]+\.1/ })


       89 |         // Get the accession of the sequence to revise
       90 |         const toReviseLink = page.getByRole('link', { name: /LOC_[A-Z0-9]+\.1/ });
     > 91 |         const toReviseId = await toReviseLink.textContent();
          |                                               ^
       92 |         const toReviseAccession = toReviseId.split('.')[0];
       93 |
       94 |         // Click on the sequence and revise it
         at /home/runner/work/loculus/loculus/integration-tests/tests/specs/features/sequence-version-banners.spec.ts:91:47
```

🚀 Preview: Add `preview` label to enable

PR #7201, opened 2026-08-28
fix(e2e): one helper for the reload-until-appears waits, and don't let a transient reload error abort them

corneliusroemer-agent:
Fixes an integration test flake where a reload issued moments after a navigation gets rejected by the browser, which then aborts the whole wait it was part of.

Also refactors similar polling into a shared helper to reduce verbosity and improve maintainability. Lots of line number reductions in the tests!

Flake looks like this ([run 33173217408](https://github.com/loculus-project/loculus/actions/runs/33173217408/job/98855236275), chromium, on main):

```
  1) [chromium-with-dep] › tests/specs/features/sequence-fasta.dependent.spec.ts:6:9 › Sequence FASTA endpoint › returns valid FASTA with CORS headers 

    Error: page.reload: Protocol error (Page.reload): Not attached to an active page
    Call log:
      - waiting for navigation until "load"


       at pages/search.page.ts:307

      305 |             .poll(
      306 |                 async () => {
    > 307 |                     await this.page.reload();
          |                                     ^
      308 |                     accessions = await this.getAccessionVersions();
      309 |                     return accessions.length;
      310 |                 },
        at _test.expect.poll.message (/home/runner/work/loculus/loculus/integration-tests/tests/pages/search.page.ts:307:37)
        at SearchPage.waitForSequencesInSearch (/home/runner/work/loculus/loculus/integration-tests/tests/pages/search.page.ts:304:9)
        at /home/runner/work/loculus/loculus/integration-tests/tests/specs/features/sequence-fasta.dependent.spec.ts:9:35
```

<details>

## What is actually going on

The message is misleading twice over: nothing was slow, and nothing was missing from the search results.

The reload was rejected 2.6 ms after it was sent. The line before it had just clicked the "Ebola Sudan" link, which navigates to `/ebola-sudan` and gets redirected to `/ebola-sudan/search`; the reload arrived about 2 ms after that new page's response finished, while the browser was still swapping in the new frame. Chrome refuses a reload during that window, and the phrase it uses is "Not attached to an active page". The page was fine before and after — the trace still has video frames from it after the error, and the sibling test three lines down in the same file did the exact same thing successfully 3.5 seconds later.

The reason a 2.6 ms hiccup ends a 60 second wait is the part worth knowing: **`expect.poll` retries while the matcher fails, but if the function being polled *throws*, the error escapes immediately.** The `timeout` and `intervals` never come into play, and the poll's own message is never printed. I measured this against the exact Playwright version we use — a function that throws on its first call ends the poll in 3 ms. So any polled function that *does* something, rather than just reading a value, needs to handle its own errors.

That is also why the failure is so hard to read: the reported error is the reload, not the wait, so the log looks like a page that died rather than a wait that was interrupted. The giveaway is in the line above it in the job log — the test is recorded as taking `(1.0s)`, where a genuine timeout would say `(60.0s)`.

Every `page.reload()` in the integration tests (15 of them) sits inside a loop that reloads until something appears, so all of them can be killed this way. One of them is in `readonly.setup.ts`, where an abort fails every dependent test in the run.

The waits were also written out longhand fifteen times, in two shapes: an `expect.poll` block, and a bare `while (!visible) { reload; wait 2s }`. They are now `reloadAndPoll(page, read, { message })` and `reloadUntilVisible(page, locator, { message })`.

## How rare is it

Once, in the 179 failed test jobs I could still get logs for (9 days, 359 finished runs). Logs expire after two to three weeks so that is as far back as it goes.

It is worth fixing anyway because the cost of the fix is one extra loop iteration, and because the failure it produces is actively misleading — it points at a page that has died rather than at whatever the test was waiting for.

## Things you might want to decide differently

I match on the error text rather than catching everything. Catching everything is shorter, but a real failure like a refused connection would then retry silently until the wait's timeout with its cause thrown away, instead of failing straight away and saying why. Matching text means we may meet a fourth wording eventually and have to add it; happy to swap it for a blanket catch if you would rather have that.

The `console.warn` is there so a burst of these still shows up in the job log rather than vanishing. It is Node-side, so it does not reach the `console-warnings` fixture, which only listens to messages logged inside the browser page — `pages/CliPage.ts` already logs this way. Easy to drop if it looks noisy.

The nine `while` loops previously had no timeout at all, so they could only fail by using up the enclosing test's whole budget. They now get the helper's default of 90 seconds, which is the more generous of the two values the existing polls of this same "wait for a released sequence to show up" kind already use. **This is a behaviour change, not a free win, for two of them:** `sequence-version-banners` has a 300 second test budget and `override-hidden-fields` has 200, so a wait that legitimately took 150 seconds used to pass and would now fail at 90. I think 90 is right — nothing in this suite should need two and a half minutes for a sequence to appear in search, and if something does I would rather see it named than see a bare test timeout — but say the word and I will raise it. For `lineage-field` (95 second budget) and `submission-flow` (120) the test timeout fires first regardless, so bounding them changes only the error message. The ninth is in `search.page.ts`, whose callers set their own budgets.

## Left alone deliberately

Separately, and not fixed here: `waitForSequencesInSearch` polls for up to 60 seconds while the per-test timeout is also 60 seconds, so the test always runs out of time first and the poll's message ("Expected at least N sequences to appear in search results") can never actually be printed. That means a genuinely slow indexing run cannot report itself as slow indexing — it reports as a bare test timeout. `waitForAccessionVersionInSearch` has the same problem. Worth a follow-up; it did not cause this failure.

This does not overlap with #7198, which fixes a different flake in the same file and does not touch any reload.

</details>


theosanderson (approved):
lgtm

PR #7218, opened 2026-09-02
feat(website): Mirror backend filename validation in website

### LINKED ISSUE #5779: MVP: Restrict allowed characters in file name (website)
Let's mirror the restrictions on filenames as implemented in #3999 in the website to provide better and immediate feedback.

tombch:
resolves #5779 

### Summary
- Mirrors the backend filename restrictions updated in #7205 in the website, adding validation for filenames in the submission file mapping, before they are submitted to the backend.
- Unlike the backend validation which returns errors one at a time for file names, the website validates all provided file names, and displays the first five errors in a toast. 

### Screenshot

Invalid file names raise a toast that blocks submission:

image

### PR Checklist


website/src/components/Submission/FileUpload/FolderUploadComponent.spec.tsx
```diff
@@ -258,7 +259,8 @@ describe('FolderUploadComponent', () => {
 
             await userEvent.upload(screen.getByTestId('extraFiles'), file);
 
-            expect(mockOnError).toHaveBeenCalledWith('File names cannot contain whitespace.');
+            expect(mockOnError).toHaveBeenCalledWith(expect.stringContaining('File'));
+            expect(mockOnError).toHaveBeenCalledWith(expect.stringContaining('cannot contain whitespace'));
```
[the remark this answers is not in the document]
tombch: I don't think tests should exactly pin error messages. it makes changing anything tedious

website/src/components/Submission/FileUpload/FolderUploadComponent.tsx
```diff
@@ -524,15 +527,23 @@ const filterDotFiles = (files: File[]): File[] => {
...
     const folderNames = files.flatMap((f) => f.webkitRelativePath.split('/').slice(1, -1));
+    const fileNames = files.map((f) => f.name);
 
-    if (fileNames.some((n) => /\s/.test(n))) return 'File names cannot contain whitespace.';
     if (folderNames.some((p) => /\s/.test(p))) return 'Folder names cannot contain whitespace.';
+
+    const fileNameValidationResult = validateFileNames(fileNames, fileSharingConfig);
+    if (fileNameValidationResult.isErr()) {
+        return 'Encountered errors in uploaded files: ' + getFileNameErrorMessage(fileNameValidationResult.error);
+    }
```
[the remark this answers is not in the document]
tombch: The situation described here although valid, is basically never going to happen and it'd be more confusing if the folder component sometimes validated and sometimes didn't

backend/src/main/kotlin/org/loculus/backend/service/submission/FileMappingPreconditionValidator.kt
```diff
@@ -151,6 +151,8 @@ class FileMappingPreconditionValidator(
      * - Trailing periods: Windows silently strips these, and single or double period names break path normalisation
      * - Whitespace characters
      *
+     * The website mirrors this validation in fileNameValidation.ts - any changes here must also be made there.
```
anna-parker, line 154: nice!

website/src/types/config.ts
```diff
@@ -236,6 +236,9 @@ const sequenceFlaggingConfig = z.object({
 });
 export type SequenceFlaggingConfig = z.infer<typeof sequenceFlaggingConfig>;
 
+export const fileSharingConfig = z.object({ disableStrictFilenameValidation: z.boolean().default(false) });
```
anna-parker, line 239: the config can also have the field `outputFileUrlType` - will zod have any issues if this field is in the input?

website/src/components/Submission/SubmissionForm.spec.tsx
```diff
@@ -74,6 +74,7 @@ function renderSubmissionForm({
                 maxSequencesPerEntry: 1,
             }}
             dataUseTermsEnabled={dataUseTermsEnabled}
+            fileSharingConfig={{ disableStrictFilenameValidation: false }}
```
anna-parker, line 77: why is it hardcoded to strict only here?

website/src/components/Submission/FormOrUploadWrapper.spec.tsx
```diff
@@ -67,6 +67,7 @@ const MockSaveWrapper = ({
                 }}
                 setSubmissionFileMapping={vi.fn()}
                 onError={vi.fn()}
+                fileSharingConfig={{ disableStrictFilenameValidation: false }}
```
anna-parker: ah also here

website/src/components/Edit/EditPage.spec.tsx
```diff
@@ -79,6 +79,7 @@ function renderEditPage({
                     maxSequencesPerEntry: 1,
                 }}
                 sequenceEntryHistory={sequenceEntryHistory}
+                fileSharingConfig={{ disableStrictFilenameValidation: false }}
```
anna-parker, line 82: oh and here lol - if we need strict validation on the website then Im not sure we actually need to go through the hassle of passing through the fileSharingConfig

website/src/components/Submission/SubmissionForm.spec.tsx
anna-parker: omg its for testing, I somehow missed this - ignore me!

website/src/components/Submission/FileUpload/FolderUploadComponent.tsx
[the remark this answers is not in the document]
anna-parker: Agree we should not check the folder names, I think it would also be consistent to only check the defined fileNames that are sent to the backend and not the "true" fileNames - but I also think this is an uncommon situation and is more complicated as this comes from parsing the metadata file and not the actual file names - so we could also do that in a later PR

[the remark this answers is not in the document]
anna-parker: or actually, could we just remove the validation here any only do the validation on the file mapping? that seems more correct, less code and potentially less passing down of the FileSharingConfig config?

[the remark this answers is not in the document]
tombch: Yeah could do it just on the file mapping! the only thing you'd lose is the rejecting of files with invalid names before they can even be uploaded. But I guess that's not really a huge benefit and only makes total sense for the form upload anyway?

website/src/types/config.ts
tombch: No zod will just ignore it as it's not declared in the `fileSharingConfig`. We could have it just pass the `disableStrictFilenameValidation` flag only in the website, but I thought to do `fileSharing` as I think it looks nicer and more extensible if we wanted other props in the future?

anna-parker: thats great!

website/src/components/Submission/FileUpload/FolderUploadComponent.tsx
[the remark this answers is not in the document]
anna-parker: yeah Id probably do that, also as I could see someone using the advanced features to rename their non-ASCII files into ASCII files and this would error that use case

website/src/components/Submission/FileUpload/FolderUploadComponent.tsx
```diff
@@ -2,14 +2,14 @@ import { produce } from 'immer';
...
 
-import type {
-    Awaiting,
-    FileUploadState,
-    Pending,
-    PreviousUpload,
-    SingleFileUpload,
-    Uploaded,
-    UploadStatus,
+import {
```
anna-parker: these formating changes surprise me, are they actually required - is this branch using the same npm format version?

anna-parker (approved):
awesome, thanks!

website/src/components/Submission/FileUpload/FolderUploadComponent.tsx
tombch: No they aren't required - i think that was leftover from when I had filename validation in the folder upload component. I just reverted it!

PR #7220, opened 2026-09-02
fix(deposition): use e-utilities for nuccore visibility checks

anna-parker:
We no longer run ena integration tests on main due to the high number of ENA dev flakes, but I found that the liveness check tests are now failing: https://github.com/loculus-project/loculus/actions/runs/33619762911/job/100213831404 when adding a new feature (they are run on PRs with any changes to the deposition code).

The nuccore web page is behind a bot-check (RecaptchaChallengePageUi) that returns HTTP 200 for any request regardless of whether the accession exists, so query the E-utilities esummary API instead, which returns a real JSON result. 

E-utilities rate-limits unauthenticated requests to 3/second per IP, so retry on timeouts and transient (429/5xx) errors.

### Implications
We falsely marked INSDC nuccore accessions live on Genbank without actually checking... as this is an new bug we could potentially delete all liveness statuses for nuccore Genbank from the last month.

### PR Checklist
Ran curl to reproduce the test failure and see that the current nuccore status page that is downloaded is in fact a recaptcha challenge:
```
curl https://www.ncbi.nlm.nih.gov/nuccore/XY999999/
<!doctype html><html lang="en-US" dir="ltr"><head><base href="https://www.google.com/recaptcha/challengepage/"><link rel="preconnect" href="//www.gstatic.com"><meta name="referrer" content="origin"><script nonce="fDDXoBYzc-4DzTrYcL1ZiQ">window['ppConfig'] = {productName: 'RecaptchaChallengePageUi'...
```

### DIFF
--- a/ena-submission/src/ena_deposition/check_external_visibility.py
+++ b/ena-submission/src/ena_deposition/check_external_visibility.py
@@ -82,33 +82,82 @@ class TransientNCBIError(Exception):
 class NCBIVisibilityChecker(VisibilityChecker):
     """Checker for NCBI visibility"""
 
-    def check_visibility(self, config: Config, accession: str) -> datetime | None:
-        if accession.startswith("PRJ"):
-            path = "bioproject"
-        elif accession.startswith("SAM"):
-            path = "biosample"
-        else:
-            return self._check_nuccore_visibility(config, accession)
-        response = requests.get(
-            f"https://www.ncbi.nlm.nih.gov/{path}/{accession}/",
-            allow_redirects=False,
-            timeout=config.ncbi_public_search_timeout_seconds,
-        )
-        if response.status_code == HTTPStatus.OK:
-            return datetime.now(pytz.UTC)
-        return None
+    # Accession prefix -> E-utilities database name. Anything else is a nucleotide accession.
+    _DB_BY_PREFIX = (("PRJ", "bioproject"), ("SAM", "biosample"))
 
-    def _check_nuccore_visibility(self, config: Config, accession: str) -> datetime | None:
-        """The nuccore web page is behind a bot-check that returns HTTP 200 for any
-        request regardless of whether the accession exists, so query the E-utilities
-        esummary API instead, which returns a real JSON result. E-utilities rate-limits
-        unauthenticated requests to 3/second per IP, so retry on timeouts and transient
-        (429/5xx) errors."""
+    # A nuccore record can resolve to a UID while no longer being publicly available.
+    # Treat only these esummary statuses as live (empty/"live"); anything else
+    # (e.g. "suppressed", "withdrawn", "removed", "replaced") counts as not visible.
+    _LIVE_NUCCORE_STATUSES = frozenset({"", "live"})
+
+    def check_visibility(self, config: Config, accession: str) -> datetime | None:
+        db = next(
+            (db for prefix, db in self._DB_BY_PREFIX if accession.startswith(prefix)),
+            "nuccore",
+        )
+        return self._check_eutils_visibility(config, accession, db)
+
+    def _check_eutils_visibility(
+        self, config: Config, accession: str, db: str
+    ) -> datetime | None:
+        """The NCBI web pages (nuccore, bioproject, biosample) are behind a bot-check that
+        returns HTTP 200 for any request regardless of whether the accession exists, so
+        query the E-utilities API instead, which returns a real JSON result.
+
+        esearch resolves the accession to internal UID(s); an empty result means the
+        accession is not (yet) in NCBI. For nuccore we additionally fetch esummary to
+        make sure the record has not been suppressed/withdrawn after being assigned a UID.
+        """
+
+        esearch = self._eutils_get_json(
+            config,
+            "esearch.fcgi",
+            {"db": db, "term": accession, "retmode": "json"},
+            accession,
+            "esearch",
+        )
+        if esearch is None:
+            return None
+        uids = esearch.get("esearchresult", {}).get("idlist", [])
+        if not uids:
+            return None
+        if db != "nuccore":
+            return datetime.now(pytz.UTC)
+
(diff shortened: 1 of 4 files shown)


ena-submission/src/ena_deposition/check_external_visibility.py
```diff
@@ -92,6 +98,48 @@ def check_visibility(self, config: Config, accession: str) -> datetime | None:
...
+            return response
+
+        retryer = Retrying(
+            stop=stop_after_attempt(config.ena_http_get_retry_attempts),
+            wait=wait_fixed(2),
+            retry=retry_if_exception_type((requests.exceptions.Timeout, TransientNCBIError)),
+            reraise=True,
+            before_sleep=log_before_retry,
+        )
+        response = retryer(_do_get)
```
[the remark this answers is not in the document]
anna-parker: I think this is ok

ena-submission/src/ena_deposition/check_external_visibility.py
```diff
@@ -82,7 +88,7 @@ def check_visibility(self, config: Config, accession: str) -> datetime | None:
         elif accession.startswith("SAM"):
             path = "biosample"
         else:
-            path = "nuccore"
+            return self._check_nuccore_visibility(config, accession)
```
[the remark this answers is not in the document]
anna-parker: true - I will just use it for everything

ena-submission/src/ena_deposition/check_external_visibility.py
```diff
@@ -256,9 +329,6 @@ def check_and_update_visibility_for_column(
                     f"Failed to update {column_name} for {entity_type.value} {entity_id}"
                 )
         else:
-            visible_count = sum(
```
anna-parker, line 259: this additional call was not required and should be avoided now we are rate limited

anna-parker:
workflow tests were due to ENA dev's common FTP authentication error flake

ena-submission/test/test_ena_submission_integration.py
```diff
@@ -619,12 +619,12 @@ class TestFirstPublicUpdate(TestSubmission):
...
-            "insdc_accession_full_seg1": "XY999999",
-            "insdc_accession_full_seg2": "XY999998",
+            "insdc_accession_full_segment1": "XY999999.1",
+            "insdc_accession_full_segment2": "XY999998.1",
         },
         "valid_result": {
-            "insdc_accession_full_seg1": "OZ271453",
-            "insdc_accession_full_seg2": "OZ271454",
+            "insdc_accession_full_segment1": "OZ271453.1",
+            "insdc_accession_full_segment2": "OZ271454.1",
```
anna-parker:
```suggestion
            "insdc_accession_full_seg1": "XY999999.1",
            "insdc_accession_full_seg2": "XY999998.1",
},
"valid_result": {
            "insdc_accession_full_seg1": "OZ271453.1",
            "insdc_accession_full_seg2": "OZ271454.1",
```

ena-submission/src/ena_deposition/check_external_visibility.py
```diff
@@ -73,24 +75,89 @@ def check_visibility(self, config: Config, accession: str) -> datetime | None:
...
+                raise TransientNCBIError(msg)
+            return response
+
+        retryer = Retrying(
+            stop=stop_after_attempt(config.ena_http_get_retry_attempts),
+            wait=wait_fixed(2),
+            retry=retry_if_exception_type((requests.exceptions.Timeout, TransientNCBIError)),
+            reraise=True,
+            before_sleep=log_before_retry,
+        )
```
maverbiest, line 146: Not blocking but wanted to flag that there's now 4 sites where we make a `Retrying` that are almost identical. If we wanted to have one central place to configure things like `wait` we could extract this into `def http_retryer(attempts int, retry_on ...) -> Retrying`

ena-submission/src/ena_deposition/check_external_visibility.py
```diff
@@ -73,24 +75,89 @@ def check_visibility(self, config: Config, accession: str) -> datetime | None:
...
+                logger.info(
+                    f"NCBI esearch request failed for {accession}: "
+                    f"HTTP {response.status_code} - {response.text}"
+                )
+                msg = f"Transient NCBI esearch error {response.status_code} for {accession}"
+                raise TransientNCBIError(msg)
+            return response
+
+        retryer = Retrying(
+            stop=stop_after_attempt(config.ena_http_get_retry_attempts),
```
maverbiest: Would this be a good time to add `config.ncbi_http_get_retry_attempts`?

ena-submission/src/ena_deposition/check_external_visibility.py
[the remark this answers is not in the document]
maverbiest: About claude's first point regarding `reraise=True` potentially killing a whole batch of assembly checks, I'm assuming this means we'll just recheck if those assemblies are open in a later cycle, right? If so it indeed doesn't really seem like a big deal

maverbiest (approved):
Had a look at this and seems good to me! Feel free to resolve the minor comments I made if you don't want to address them

ena-submission/src/ena_deposition/check_external_visibility.py
anna-parker: I felt like we had so many different timeout configs I didnt want to add another one but I can do that if necessary

ena-submission/src/ena_deposition/check_external_visibility.py
anna-parker: I think its ok for now, I dont think we would end up saving that many lines of code with a new function Retrying is already pretty concise

ena-submission/src/ena_deposition/check_external_visibility.py
anna-parker: added!

PR #7246, opened 2026-09-06
fix(deposition): revise assembly when linked run accession (RUN_REF) changes

### LINKED ISSUE #7245: deposition: correctly handle case where only raw reads are updated not consensus sequence
**[P2] Replacing only the raw reads leaves the ENA assembly linked to the old run**

I looked into the revision path and reproduced an edge case here. `RUN_REF` now comes from `raw_reads_table`, but `has_assembly_data_changed` still compares the consensus, flat-file metadata, and mapped fields in `seq_metadata`. It doesn't see changes to the generated run accession.

For example:

1. Version 1 deposits reads as `ERR111` and an assembly referencing `ERR111`.
2. Version 2 replaces the read files but keeps the consensus and assembly metadata unchanged, producing `ERR222`.
3. Assembly creation retrieves `ERR222` here, then takes the unchanged-assembly shortcut at lines 525–528 and copies the previous assembly result without submitting an updated manifest.

Loculus consequently reports the new run, while the ENA assembly remains linked to `ERR111`. The raw-read workflow can also request suppression of that old run. Adding reads to an existing assembly without otherwise changing it has the same missing-link-update problem.

I reproduced the shortcut using the actual PR functions with database calls mocked: the new run was present, but the previous assembly was copied and no new manifest was generated. The existing raw-read revision test misses this because `get_revisions` defaults to `modify_assembly=True`, so it changes assembly metadata too.

It would be good to compare the previous and current effective run references before reusing an assembly result, and add a test with `modify_raw_reads=True, modify_assembly=False`. If updating the link needs manual handling, that could be flagged explicitly instead.

This could be turned into a follow-up issue if you'd prefer to handle it separately.

_Originally posted by @theosanderson-agent in https://github.com/loculus-project/loculus/pull/6816#discussion_r3937506176_

theosanderson-agent:
Closes #7245. Targets the `raw_reads_ena` branch of #6816, since the affected code (RUN_REF from `raw_reads_table`) only exists there.

## Problem

`assembly_table_create` now takes the `RUN_REF` for the assembly manifest from `raw_reads_table` (the run accession deposited for the same accession version). But the revision checks (`can_be_revised` / `has_assembly_data_changed`) only compared the consensus sequence, flat-file metadata and the mapped manifest fields in `seq_metadata`. A revision that only replaces or adds raw read files produces a new run accession (e.g. `ERR111` → `ERR222`) without changing anything those checks look at, so the pipeline took the "unchanged assembly" shortcut, copied the previous assembly result and never submitted a manifest with the new `RUN_REF`. Loculus then reports the new run while the ENA assembly stays linked to the old run (which the raw reads workflow may also ask to suppress).

## Fix

- New `get_run_ref(db_engine, seq_key)` reads the effective run accession for an accession version from `raw_reads_table` (also used for the current version in `assembly_table_create`, replacing the inline lookup).
- New `run_ref_diff(db_engine, last_version_entry, run_ref)` compares the previous version's effective `RUN_REF` with the one about to be written to the manifest.
- `has_assembly_data_changed` now returns `True` when the `RUN_REF` differs, so the assembly is resubmitted with the new link instead of reusing the previous result. This also covers the "reads added to an existing assembly" case (`None` → `ERRxxx`). The `TODO` about this is removed.
- `manifest_fields_changed` (used by `can_be_revised` when `allow_revision_with_manifest_changes=False`) includes the `run_ref` diff in its error, so in that mode a changed `RUN_REF` is flagged as `HAS_ERRORS` for manual revision, consistent with how other manifest field changes are handled.
- `ENA_submission.md`: documented the behaviour in the "Revising Assemblies" section.

## Tests

- **Integration test** (`TestRevisionRawReadsOnlyModificationTests`, as suggested in the issue): full submission with raw reads, then a revision with `modify_raw_reads=True, modify_assembly=False, with_raw_reads=True`. It asserts the v2 run accession differs from v1, that the assembly is resubmitted (goes through the WAITING path rather than the copy shortcut), and that exactly one new manifest is created whose `run_ref` equals the new run accession. `create_manifest` is wrapped with `autospec` + `side_effect` so the real manifest is still written. **Not run locally** (needs the ENA dev endpoint credentials and a Postgres instance), so please let CI exercise it.

`ruff check` and `ruff format` (v0.14.2, as pinned in pre-commit) pass.


ena-submission/test/test_ena_submission.py
```diff
@@ -542,6 +544,114 @@ def test_diff_records_error_when_field_resolution_fails(self):
         self.assertIn("Error resolving field", differing_fields["authors"])
 
 
+class AssemblyRevisionRunRefTests(unittest.TestCase):
```
anna-parker: I might opt just to have the integration tests - these unit tests mock so much they dont seem worth having IMO

anna-parker: I guess it doesnt hurt to have them

ena-submission/src/ena_deposition/create_assembly.py
```diff
@@ -327,15 +327,55 @@ def update_assembly_error(
...
+    """
+    Compare the RUN_REF that would be written to the manifest of the new version with the one
+    used for the previous version. The RUN_REF is taken from raw_reads_table (not from
+    seq_metadata), e.g. it changes when raw read files are replaced or added, so it is not
+    covered by manifest_fields_diff.
+    """
+    previous_run_ref = get_run_ref(db_engine, last_version_entry.pkey)
+    if previous_run_ref != run_ref:
+        return {"run_ref": f"Last: {previous_run_ref}, New: {run_ref}"}
+    return {}
```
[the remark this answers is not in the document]
anna-parker: if the run is remove it should be unlinked this is correct

ena-submission/test/test_ena_submission.py
anna-parker: ah cool claude agreed with me that these tests dont really test anything - I will remove

ena-submission/src/ena_deposition/create_assembly.py
```diff
@@ -499,12 +558,7 @@ def assembly_table_create(db_engine: Engine, config: Config):
             RawReadsTableEntry,
             conditions=asdict(seq_key),
         )
-
-        run_ref = (
-            cast(str, corresponding_raw_reads[0].result.get(EnaResultField.RUN))
-            if corresponding_raw_reads and corresponding_raw_reads[0].result
-            else None
-        )
+        run_ref = get_run_ref(db_engine, seq_key)
```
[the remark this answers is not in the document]
anna-parker: good point!

anna-parker (approved):

anna-parker:
@corneliusroemer if you are ok with this we can also already merge it into the `raw_reads_ena` branch

corneliusroemer (approved):

PR #7255, opened 2026-09-08
fix(backend): Make SeqSet description optional for `/update-seqset`

tombch:
### Summary

The `description` parameter was required for `/update-seqset`, despite being optional for `/create-seqset` - updated to be optional for update as well.

Note this was/is a backend issue only - the website always provides a string for the SeqSet description.

### PR Checklist

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCitationsDatabaseService.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCitationsDatabaseService.kt
@@ -699,7 +699,7 @@ class SeqSetCitationsDatabaseService(
         }
 
         if (oldSeqSet.name == newSeqSetName &&
-            oldSeqSet.description == newSeqSetDescription &&
+            oldSeqSet.description.orEmpty() == newSeqSetDescription.orEmpty() &&
             oldSubmittedSeqSetRecords == newSeqSetRecords
         ) {
             throw UnprocessableEntityException("SeqSet update must contain at least one change")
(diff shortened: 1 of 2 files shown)


maverbiest (approved):
Looks good to me!

backend/src/main/kotlin/org/loculus/backend/service/seqsetcitations/SeqSetCitationsDatabaseService.kt
```diff
@@ -699,7 +699,7 @@ class SeqSetCitationsDatabaseService(
         }
 
         if (oldSeqSet.name == newSeqSetName &&
-            oldSeqSet.description == newSeqSetDescription &&
+            oldSeqSet.description.orEmpty() == newSeqSetDescription.orEmpty() &&
```
maverbiest, line 702: I think it's fine to do like this, but do I understand correctly that `oldSeqSet.description` can actually never be `null` since `createSeqSet()` does `it[description] = seqSetDescription ?: ""` ?

tombch: I just had a look and the seqsets description column is just `text` in the sql schema, so technically could have null values as well? So I guess we can't just switch the Kotlin type without a migration

tombch: Although yeah in practice now they all get set to strings

maverbiest: Yeah sorry didn't mean to derail this PR, it's beyond the scope here so happy to have this merged as is! I just noticed the schema is slightly out of sync with how we really use the column if we always fall back to a string, but nbd.

tombch: No I agree it is out of sync, and imo text fields shouldn't really be nullable because it gives two different ways for something to be empty. Yeah we can follow up on this!

maverbiest: Looks like the column is non-nullable at the Exposed level (`SeqSetsTable.description`) but nullable in the DB itself 😅

PR #7259, opened 2026-09-08
fix(backend): reject reusing one file ID within a file category

corneliusroemer-agent:
Found while reviewing #6816 (raw reads to ENA), but it stands on its own and is not specific to that PR.

## Why

Duplicate file *names* within a category are already rejected, in two places. Duplicate file *IDs* were not rejected anywhere — I checked, and there was no such validation in the backend at all. So this was accepted:

```
files.raw_reads = reads_1.fq:<uuid-X>  reads_2.fq:<uuid-X>
```

I don't think allowing duplicate file ids is intentional. When using the normal website submission each file gets a separate id, so the normal flow doesn't expose duplicate ids.

The only way it can happen is if one does a revision or manually creates the file column. I can't see why giving the same file 2 names makes sense.

We can prohibit it and relax it again later if there's a good reason for it - but going from lax -> strict later is harder.

## If not merged

For raw reads we need to ensure that the 2 files are distinct, so if the backend doesn't prohibit we need to add logic there.

## Deliberately not included

A matching guard in the ENA deposition code, which would make this impossible regardless of which backend version is deployed. It seemed redundant once the backend enforces it, but it is cheap defence in depth if you want it — happy to add it here or separately.

The website side is also not done here. It has a `name:fileId` form for reusing an already-uploaded file, and it rejects duplicate file *names* inline before upload, but it has no equivalent check for duplicate ids — so a typo in a hand-written file column gets a backend 422 where every neighbouring mistake gives an inline error. Might be worth mirroring the check client-side in a separate website PR.

## Unsure

I'm not sure if we should apply the same prohibition on duplicate fileIds to prepro submissions as well. This would make it easier for downstream users but would be a breaking change that could cause errors on reprocessing (but PPX doesn't use files so this would only affect users who have duplicate ids - unlikely to be many). For now I don't enforce distinct ids for prepro to be less breaking.

https://claude.ai/code/session_018VRAmMxXE9u1EksE87X6Mo


backend/src/main/kotlin/org/loculus/backend/api/SubmissionTypes.kt
```diff
@@ -479,3 +479,19 @@ fun FileCategoryFilesMap.getDuplicateFileNames(category: FileCategory): Set<Stri
...
+
+/**
+ * File IDs used more than once within [category].
+ *
+ * Reusing one file ID under several names is rejected: consumers key their local copies by file ID
+ * (the raw reads processing service and the ENA deposition pipeline both do), so the entries would
+ * collapse onto a single downloaded object. For paired-end reads that is silently accepted by
+ * readtools/webin-cli - the file pairs with itself at 100% - and would archive mate 1 twice.
+ */
+fun FileCategoryFilesMap.getDuplicateFileIds(category: FileCategory): Set<FileId> {
```
anna-parker, line 489: I feel like this was intentional, so maybe instead of doing this check in the backend it should just be done in the raw-reads-processing pod

corneliusroemer: I don't think allowing duplicate file ids is intentional? What's the point of having the same thing twice with different names?

We can prohibit it and relax later if there's a good reason for it - but going from lax -> strict later is harder.

There's no way you can submit duplicate file ids via the website normal flow - only if you pass in file ids manually with our new columns is this possible - but most likely a typo then rather than intentional.

anna-parker: discussed async we should add this validation

anna-parker: we should add enforce within one submission/entry, however of course fileIds can be reused across entries

backend/src/main/kotlin/org/loculus/backend/service/submission/FileMappingPreconditionValidator.kt
```diff
@@ -55,6 +56,20 @@ class FileMappingPreconditionValidator(
...
+            val duplicateFileIds = fileCategoriesFilesMap.getDuplicateFileIds(category)
+            if (duplicateFileIds.isNotEmpty()) {
+                throw UnprocessableEntityException(
+                    "The files in category $category reuse the same file ID more than once: " +
+                        "${duplicateFileIds.joinToString()}. Each file must have its own file ID.",
+                )
+            }
+        }
+        return this
+    }
```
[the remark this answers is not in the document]
anna-parker: done!

anna-parker (approved):
awesome!

PR #7260, opened 2026-09-08
test(deposition): run get_ena_submission_list test in a tmp dir

corneliusroemer-agent:
Running the ena-submission unit tests leaves two untracked JSON files in the repo: `ena-submission/cchf_ena_submission_list.json` and `ena-submission/cchf_with_ena_fields_ena_submission_list.json`.

The fix is to run the test with the working directory set to a pytest `tmp_path`, which means the three repo-relative constants at the top of the module have to become absolute first.

https://claude.ai/code/session_019qnUGUTdCU5PpUb9Lp78Vc

### DIFF
--- a/ena-submission/test/test_get_ena_submission_list.py
+++ b/ena-submission/test/test_get_ena_submission_list.py
@@ -2,17 +2,22 @@
 import json
 import unittest
 from collections.abc import Iterator
+from pathlib import Path
 from typing import Any
 from unittest.mock import ANY, Mock, patch
 
 import get_ena_submission_list
 import orjsonl
+import pytest
 from deepdiff import DeepDiff
 
-CONFIG_FILE = "./test/test_config.yaml"
-GET_RELEASED_ENTRIES_PATH = "./test/data/get-released-data.ndjson"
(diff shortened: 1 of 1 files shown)


ena-submission/test/test_get_ena_submission_list.py
```diff
@@ -35,6 +40,16 @@ def fake_fetch_released_entries(config, organism) -> Iterator[dict[str, Any]]:
 
 
 class GetSubmissionListTests(unittest.TestCase):
+    @pytest.fixture(autouse=True)
```
anna-parker, line 43: I wonder if it would be easier to just update the  `get_ena_submission_list` script to also accept an output_dir as input and then we wouldn't need to do this

anna-parker: @corneliusroemer I did this here: https://github.com/loculus-project/loculus/pull/7271 - I found this fixture almost a bit magic and thought it might be nicer to be more explicit in the code

corneliusroemer: Sounds good!

corneliusroemer:
Close in favor of #7271

PR #7265, opened 2026-09-08
fix(deposition): normalise raw-read file extensions to the validated set

corneliusroemer-agent:
Make ENA submission work with what prior raw reads validation has checked and guarantees:
file ending is one of `(".fastq.gz", ".fq.gz", ".fastq", ".fq")` with arbitrary casing.

### Fix double compression bug

`download_fastq_files` decided whether to compress with `file_path.endswith((".gz", ".bz2"))`, which is case-sensitive. The upstream check in raw-reads-processing is not: `_has_extension` lower-cases before comparing, so `READS.FASTQ.GZ` is a perfectly legal submission today. It reached this code, failed the `.gz` test, and got gzipped a second time.

### Make extensions lowercase so that webin is happy

ENA submission needs to lowercase extensions so that webin-cli accepts it (it compares suffixes with `String.endsWith`, so `r.FASTQ.GZ` is rejected outright with "Invalid FASTQ file suffix".

### Misnamed extension should be rejected in raw reads validation not here

A genuinely gzipped file *named* `reads.fastq` still gets compressed twice. The right place to catch it is at submission: separate PR #7266

### Worth deciding separately

If raw-reads-processing were narrowed to gzip-only the code would be simplified a lot. We would also save on storage, raw reads on Loculus/PPX would be more uniform and faster to download for everyone (and make upload for submitter faster). The trade-off is that submitters would have to create gzipped raw reads - but that's probably not an issue (and ENA requires gzip (or bz2) anyways so it's a normal requirement.


ena-submission/src/ena_deposition/call_loculus.py
```diff
@@ -219,6 +218,29 @@ def fetch_released_entries(config: Config, organism: str) -> Iterator[dict[str,
             }
 
 
+# Kept in sync with ACCEPTED_FASTQ_EXTENSIONS in
+# raw-reads-processing/src/raw_reads_processing/file_format_validation.py, which rejects
+# anything else at submission time. Separate deployables, hence the duplication.
+ACCEPTED_FASTQ_EXTENSIONS = (".fastq.gz", ".fq.gz", ".fastq", ".fq")
```
anna-parker, line 224: lets actually also accept bz2 compression actually (as ena accepts this)

anna-parker: discussed not needed

anna-parker:
@claude review

anna-parker (approved):

Issue #5463, opened 2025-11-18
Flaky firefox integration test: `file-sharing.spec.ts:6:5 › submit a single sequence with two files `


corneliusroemer:
Somehow playwright thinks files aren't successfully uploaded even though they are

```
  1) [firefox-without-dep] › tests/specs/features/file-sharing.spec.ts:6:5 › submit a single sequence with two files 

    Test timeout of 180000ms exceeded.

    Error: expect(locator).toBeVisible() failed

    Locator: getByText('✓').first()
    Expected: visible
    Timeout: 5000ms
    Error: element(s) not found

    Call log:
      - Expect "toBeVisible" with timeout 5000ms
      - waiting for getByText('✓').first()
        - waiting for" http://localhost:3000/user" navigation to finish...
        - navigated to "http://localhost:3000/user"
        - waiting for" http://10.1.0.152:8083/realms/loculus/protocol/openid-connect/logout/logout-confirm?client_id=backend-client&tab_id=WkupjIk2lf4" navigation to finish...
        - navigated to "http://localhost:3000/logout"


      21 |         });
      22 |
    > 23 |         await expect(page.getByText('✓').first()).toBeVisible();
         |                                                   ^
      24 |         await expect(page.getByText('✓').nth(1)).toBeVisible();
      25 |     } finally {
      26 |         if (cleanup) {
        at /home/runner/work/loculus/loculus/integration-tests/tests/specs/features/file-sharing.spec.ts:23:51

    Error Context: test-results/specs-features-file-sharin-e8bcf-gle-sequence-with-two-files-firefox-without-dep/error-context.md

```

https://github.com/loculus-project/loculus/actions/runs/19475041875/job/55732543617
https://github.com/loculus-project/loculus/actions/runs/19461908991/job/55705644929
https://github.com/loculus-project/loculus/actions/runs/19348542495/job/55354887030

PR #7267, opened 2026-09-08
test(integration): wait for uploads to complete, not for N checkmarks

corneliusroemer-agent:
File upload completion was inferred from number of tick marks on page. A new component unrelated to file upload added a new tick so completion was inferred too quickly causing an error when submitting.

We now scope the tick count just to the file upload section.

How this flake looked in logs:

```
  1) [firefox-without-dep] › tests/specs/features/file-sharing.spec.ts:303:5 › bulk submit blocks a submission with an invalid file name

    Error: expect(locator).toBeVisible() failed

    Locator: getByText('Invalid filename \'CON.fastq\'').first()
    Expected: visible
    Timeout: 5000ms
    Error: element(s) not found
```

```ts
// before: satisfied once fileCount - 1 files have uploaded
page.getByText('✓').nth(i).waitFor({ state: 'visible' })

// after
expect(page.getByTestId(new RegExp(`^status_${fileCategory}_`)).filter({ hasText: '✓' }))
    .toHaveCount(fileCount, { timeout })
expect(page.getByTestId(`add_button_${fileCategory}`)).toBeEnabled({ timeout })
```

<details>

The helper waits for `fileCount` elements containing `✓` anywhere on the page (from #5462). Since [#6922](https://github.com/loculus-project/loculus/pull/6922) bulk mode also renders `✓ N files uploaded and linked to metadata!` as soon as the first file is linked, so the page holds one tick per uploaded file **plus one for that line** and the wait for N ticks is satisfied at N−1. In the failing trace the second file finished uploading 215 ms after submit was clicked.

`status_<category>_<path>` holds only the status icon, so the summary line cannot inflate the count, and the button is enabled only while the category is `uploadCompleted`, which is what submitting requires. `setInputFiles` stays un-awaited: that is #5462's fix for #5463, and since we set no `actionTimeout` awaiting it would be bounded only by the test timeout.

Also improve error informativeness when review page never loads: alert text now ends up in the playwright error message so it's fast to debug.

</details>

### DIFF
--- a/integration-tests/tests/specs/features/revise-sequence.spec.ts
+++ b/integration-tests/tests/specs/features/revise-sequence.spec.ts
@@ -6,6 +6,7 @@ import { ReviewPage } from '../../pages/review.page';
 import { RevisionPage } from '../../pages/revision.page';
 import { NavigationPage } from '../../pages/navigation.page';
 import { SingleSequenceSubmissionPage } from '../../pages/submission.page';
+import { waitForUrlReportingAlerts } from '../../utils/navigation-helpers';
 import {
     CCHF_S_SEGMENT_FULL_SEQUENCE,
     createFastaContent,
@@ -124,7 +125,7 @@ groupTest.describe('Bulk sequence revision', () => {
         await revisionPage.acceptTerms();
         await revisionPage.clickSubmit();
 
-        await expect(page).toHaveURL(/\/review/);
+        await waitForUrlReportingAlerts(page, /\/review/, { timeout: 15_000 });
         await reviewPage.waitForZeroProcessing();
 
         const overview = await reviewPage.getReviewPageOverview();
--- a/integration-tests/tests/specs/features/revise-sequence.spec.ts
+++ b/integration-tests/tests/specs/features/revise-sequence.spec.ts
@@ -125,7 +125,7 @@ groupTest.describe('Bulk sequence revision', () => {
         await revisionPage.acceptTerms();
         await revisionPage.clickSubmit();
 
-        await waitForUrlReportingAlerts(page, /\/review/, { timeout: 15_000 });
+        await waitForUrlReportingAlerts(page, '**/review', { timeout: 15_000 });
         await reviewPage.waitForZeroProcessing();
 
         const overview = await reviewPage.getReviewPageOverview();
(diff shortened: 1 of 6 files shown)


integration-tests/tests/utils/file-upload-helpers.ts
```diff
@@ -52,16 +52,19 @@ export async function prepareTmpDirForSingleUpload(
...
 ) {
     await page.getByRole('heading', { name: 'Extra files' }).scrollIntoViewIfNeeded();
-    // Trigger file upload (don't await) and wait for checkmarks to appear (indicates success)
-    void page.getByTestId(testId).setInputFiles(tmpDir);
-    return Promise.all(
-        Array.from({ length: fileCount }, (_, i) =>
-            page.getByText('✓').nth(i).waitFor({ state: 'visible' }),
-        ),
-    );
+    // Not awaited: playwright's own idea of when the upload is done was the flake fixed in #5462.
```
corneliusroemer:
```suggestion
    // Not awaited: playwright sometimes never returns, deflaked in #5462.
```

integration-tests/tests/specs/features/revise-sequence.spec.ts
```diff
@@ -124,7 +125,7 @@ groupTest.describe('Bulk sequence revision', () => {
         await revisionPage.acceptTerms();
         await revisionPage.clickSubmit();
 
-        await expect(page).toHaveURL(/\/review/);
+        await waitForUrlReportingAlerts(page, /\/review/, { timeout: 15_000 });
```
corneliusroemer:
```suggestion
        await waitForUrlReportingAlerts(page, '**/review', { timeout: 15_000 });
```

integration-tests/tests/utils/navigation-helpers.ts
```diff
@@ -0,0 +1,24 @@
...
+                .getByRole('alert')
+                .allInnerTexts()
+                .catch((): string[] => [])
+        )
+            .map((text) => text.replace(/\s+/g, ' ').trim())
+            .filter((text) => text.length > 0);
+        if (alerts.length === 0) {
+            throw error;
+        }
+        throw new Error(`Did not navigate to ${url.toString()}: ${alerts.join(' | ')}`);
```
tombch: Claude suggestion: could also include the originating error in this error - as now we just have the alerts

tombch (approved):
Looks good to me!

integration-tests/tests/utils/navigation-helpers.ts
corneliusroemer: Implemented better error message, thanks!

PR #7269, opened 2026-09-08
fix(website): disable submitting while the metadata file is being read

corneliusroemer-agent:
Seen on [this job](https://github.com/loculus-project/loculus/actions/runs/34096648264/job/101661701168) (firefox, 2026-09-07):

```
  1) [firefox-without-dep] › tests/specs/features/submission-flow.spec.ts:21:9 › Submission flow › basic file upload submission flow works

    Test timeout of 120000ms exceeded.

    Error: locator.click: Test timeout of 120000ms exceeded.
    Call log:
      - waiting for getByRole('button', { name: 'Continue under Open terms' })

      38 |         await page.getByRole('button', { name: 'Upload and proceed to Approval' }).click();
    > 39 |         await page.getByRole('button', { name: 'Continue under Open terms' }).click();
```

The data use terms dialog never opened because the submission never happened. Its page snapshot shows why:

```yaml
- region "Notifications Alt+T":
  - alert:
    - text: "Cannot submit: metadata file is still being processed."
```

Selecting a metadata file starts an async effect that reads it and parses the file mapping out of it. Submitting before that finishes hits the guard in `handleSubmit` and produces only that toast, so the test waited out its whole 120 s for a dialog that was never coming. A person who clicks quickly gets the same dead end, which is why this is a website change rather than a wait added to the test.

The mapping state now carries a `'processing'` value while the file is being read, and the submit button is disabled while it does — so playwright's click auto-waits and the test needs no change. The guard in `handleSubmit` stays, because a form can also be submitted by pressing enter and no disabled button prevents that.

The unit test defers `metadataFile.text()` and asserts the button is disabled until it resolves. I checked it fails without the change (`expect(element).toBeDisabled()`) rather than passing vacuously.

### DIFF
--- a/website/src/components/Submission/DataUploadForm.tsx
+++ b/website/src/components/Submission/DataUploadForm.tsx
@@ -1,7 +1,6 @@
 import { isErrorFromAlias } from '@zodios/core';
 import type { AxiosError } from 'axios';
 import { DateTime } from 'luxon';
-import type { Result } from 'neverthrow';
 import { type FormEvent, useState, type Dispatch, type SetStateAction, useMemo } from 'react';
 
 import { type FileFactory, FormOrUploadWrapper, type InputMode } from './FormOrUploadWrapper.tsx';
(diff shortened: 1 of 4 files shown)


corneliusroemer-agent:
Superseded by #7270. Disabling the submit button was the wrong fix: it removes the explanation of why submitting did nothing, and a disabled control is poor accessibility. The toast is the better feedback, so #7270 leaves the website alone and retries the submit in the test instead.

PR #7270, opened 2026-09-08
fix(website): wait for the metadata file when submitting, don't reject

corneliusroemer-agent:
Improve UX and deflake by having the submit handler await metadata file parsing rather than toasting "not yet parsed". The original motivation was that deflaking integration tests on this is hard: logic would have to depend on whether the click happened before parsing was done or not.

Turns out that the solution that's easy to test is also one that's good for general ux - I think this is clearly better and (according to Claude) also avoids a confusing and unrecoverable edge case error state when the file can't be read.

Code changes are fairly minimal, the majority is tests and indentation of an existing block.

Fixes a flake seen on [this job](https://github.com/loculus-project/loculus/actions/runs/34096648264/job/101661701168) (firefox, 2026-09-07):

```
  1) [firefox-without-dep] › tests/specs/features/submission-flow.spec.ts:21:9 › Submission flow › basic file upload submission flow works

    Test timeout of 120000ms exceeded.

    Error: locator.click: Test timeout of 120000ms exceeded.
    Call log:
      - waiting for getByRole('button', { name: 'Continue under Open terms' })

      38 |         await page.getByRole('button', { name: 'Upload and proceed to Approval' }).click();
    > 39 |         await page.getByRole('button', { name: 'Continue under Open terms' }).click();
```

The data use terms dialog never opened because the submission never happened. The archived page snapshot says why:

```yaml
- region "Notifications Alt+T":
  - alert:
    - text: "Cannot submit: metadata file is still being processed."
```

Apparent bug fix according to Claude:

> A read that fails now reports itself, which it never did: the exception disappeared into a `void (async () => …)()`, the mapping stayed undefined forever, and every later submit claimed the file was still being processed — a state nothing could get out of.

### DIFF
--- a/website/src/components/Submission/DataUploadForm.tsx
+++ b/website/src/components/Submission/DataUploadForm.tsx
@@ -2,7 +2,7 @@ import { isErrorFromAlias } from '@zodios/core';
 import type { AxiosError } from 'axios';
 import { DateTime } from 'luxon';
 import type { Result } from 'neverthrow';
-import { type FormEvent, useState, type Dispatch, type SetStateAction, useMemo } from 'react';
+import { type FormEvent, useEffect, useState, type Dispatch, type SetStateAction, useMemo } from 'react';
 
 import { type FileFactory, FormOrUploadWrapper, type InputMode } from './FormOrUploadWrapper.tsx';
 import { getClientLogger } from '../../clientLogger.ts';
@@ -39,6 +39,7 @@ import {
     type CategoryLinkage,
     type FileLinkage,
     type SubmissionFileMapping,
+    type SubmissionFileMappingPromise,
     validateSubmissionFileMapping,
 } from './FileUpload/fileMapping.ts';
 import { extraFilesUploadDocsUrl } from './extraFilesUploadDocsUrl.ts';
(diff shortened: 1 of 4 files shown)


website/src/components/Submission/FileUpload/fileMapping.ts
```diff
@@ -520,6 +520,9 @@ export async function applyFileMappings(
     return ok(new File([content], 'metadata.tsv', { type: 'text/tab-separated-values' }));
 }
 
+/** The parse of the uploaded metadata file, still in flight; it resolves rather than rejects. */
```
corneliusroemer: ### 
```suggestion
```

website/src/components/Submission/FormOrUploadWrapper.tsx
```diff
@@ -80,26 +80,33 @@ export const FormOrUploadWrapper: FC<FormOrUploadWrapperProps> = ({
...
-
-            const submissionFileMapping = parseSubmissionFileMapping(
-                text,
-                submissionDataTypes.files?.categories?.map((category) => category.name) ?? [],
-            );
-            setSubmissionFileMapping(submissionFileMapping);
-            if (submissionFileMapping.isErr()) onError(submissionFileMapping.error.message);
-        })();
+        setSubmissionFileMapping(
+            (async () => {
```
corneliusroemer, line 90: This function body is the same, just indented differently

website/src/components/Submission/DataUploadForm.tsx
```diff
@@ -161,24 +178,27 @@ const InnerDataUploadForm = ({
                     finalMetadataFile = finalMetadataFileResult.value;
                 }
             } else {
-                if (submissionFileMapping === undefined) {
-                    onError('Cannot submit: metadata file is still being processed.');
+                if (submissionFileMappingPromise === undefined) {
+                    onError('Cannot submit: no metadata file has been selected.');
```
tombch, line 182: We discussed in person, and think a better approach would be to call `parseSubmissionFileMapping` again at submit time. That way handle submit does everything synchronously and we shouldn't get these flakes

tombch:
I'm working on an alternate PR for this now, should also address the issue with applying file mappings for compressed metadata files

corneliusroemer:
Overcomplicated, @tombch had better ideas for how to solve this (potentially duplicating parse when doing handleSubmit)

PR #7288, opened 2026-09-09
refactor(deposition): return the previous raw-reads row instead of a raise_on_empty flag

corneliusroemer-agent:
Small refactor - it's better to handle errors in the caller rather than passing in a boolean `raise_on_empty`. Also make function names clearer (we're looking for previous version not last version - last can mean latest)

<details>

Stacked on `ena-test-improvements`, so review that one first — the diff here is just the top commit.

`last_raw_reads_entry` grew a `raise_on_empty` boolean because its two callers want different things. That flag is really carrying a piece of domain knowledge that is easy to lose: a revision is *not* guaranteed to have a previous raw_reads_table row, because the previous version may simply not have had raw reads attached. Meanwhile `update_raw_reads_results_with_latest_version` genuinely cannot proceed without one, since it copies that row's `result` forward.

Rather than a parameter that switches whether the function raises, the lookup is now total — it returns the row or `None` — and the one caller that needs an error raises it. The fact that an empty result is legitimate moves into a docstring where it can be read.

The knock-on effects are the actual readability win: both call sites stop indexing `[0]` into a list that can only ever hold one row, and the `old_run_accession` expression drops from a three-part conditional to a two-part one.

I renamed it to `previous_version_raw_reads_entry` for two reasons — it is the entry for `previous_version(...)` rather than the last entry in any general sense, and `last_raw_reads_entry` sitting next to the existing `get_last_entry` invited exactly the confusion between the submission_table and raw_reads_table lookups that this area already has. Happy to drop the rename if you would rather keep the diff smaller.

One unrelated one-character fix rode along, since it is inside the function being edited: the log line read "as there wasno change in raw read data".

Deliberately not touched here: mypy already reports three errors in this file for `library_source`, `library_selection` and `library_strategy` being passed as possibly-`None` to fields typed non-optional. That is a separate question about the removed `from_value` defaults, not something this refactor should decide, and the error count is identical before and after this commit.

</details>

https://claude.ai/code/session_01XVDB6dSsq7JHadkgyTkq7d

### DIFF
--- a/ena-submission/src/ena_deposition/create_raw_reads.py
+++ b/ena-submission/src/ena_deposition/create_raw_reads.py
@@ -328,9 +328,14 @@ def has_raw_reads_changed(
     return False
 
 
-def last_raw_reads_entry(
-    db_engine: Engine, seq_key: AccessionVersion, raise_on_empty: bool = True
-) -> list[RawReadsTableEntry]:
+def previous_version_raw_reads_entry(
+    db_engine: Engine, seq_key: AccessionVersion
+) -> RawReadsTableEntry | None:
+    """The raw_reads_table row of the version this one revises, or None if there is none.
+
+    A previous raw_reads_table row is not guaranteed to exist even for a revision: the
+    previous version may simply not have had raw reads attached.
+    """
     version_to_revise = previous_version(db_engine, seq_key)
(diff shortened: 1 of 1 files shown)


anna-parker (approved):
looks good!

PR #7289, opened 2026-09-09
fix(website): Fix `outputFileUrlType: website` file URLs resulting in 401 for logged out users

tombch:
### Summary

While looking into the different `outputFileUrlType` options, Claude flagged to me this issue here. _Website_ file URLs (e.g. `http://main.loculus.org/seq/LOC_000001Y.1/rawReads/test_a_raw_reads_1.fastq`) would result in a `401` for logged out users, even if the file was now set to be public. This is because the browser would send `Authorization: Bearer undefined` to the backend when it should not send this header at all if logged out.

This is a minor issue for our intended setup with `outputFileUrlType: backend`, because the backend file links work fine - someone could navigate directly to the website link and get this issue but unlikely. However, this would be quite a noticeable problem if someone set up Loculus with `outputFileUrlType: website`, as public files would 401 when users are logged out.

This PR fixes the creation of the authorization header, and also adds a few tests for the website redirect!

### Side note

This also made me wonder do we even need `outputFileUrlType: website` as a config option, but not in the scope of this PR.

### Screenshot

Before:

In a local Loculus instance, I released a test submission with dummy `rawReads` file `test_a_raw_reads_1.fastq`. The link `http://localhost:3000/seq/LOC_000001Y.1/rawReads/test_a_raw_reads_1.fastq` should result in file download, but instead:

Screenshot 2026-09-09 at 16 13 03

After: navigating to this link while logged out results in download of the file.

### PR Checklist


theosanderson (approved):
lgtm codewise

theosanderson:
@codex review

website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].spec.ts
```diff
@@ -0,0 +1,51 @@
...
+    test('sends an Authorization header with bearer token when logged in', async () => {
+        await callRoute({ isLoggedIn: true, token: { accessToken: 'my-token', refreshToken: 'my-refresh-token' } });
+        expect(authHeader()).toBe('Bearer my-token');
+    });
+
+    test('hands the S3 redirect to the browser rather than downloading the file itself', async () => {
+        const response = await callRoute({ isLoggedIn: false });
+        expect(response.status).toBe(307);
+        expect(response.headers.get('Location')).toBe(s3Url);
+        // Without 'manual' the file would be streamed through the website instead of from S3
```
Comment (anna-parker) [inline website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].spec.ts:48]: oh good to know!

anna-parker (approved):
looks great- as @theosanderson mentioned we want to start using this option do you want to enable it in a follow up PR after merging this?

tombch:
Yeah I can do that!

PR #7290, opened 2026-09-09
Fix enum handling

anna-parker:
resolves #

### PR Checklist

### DIFF
--- a/ena-submission/src/ena_deposition/ena_types.py
+++ b/ena-submission/src/ena_deposition/ena_types.py
@@ -10,7 +10,10 @@ class XmlNone(UserString):
 
 
 class LookupStrEnum(StrEnum):
-    """StrEnum with case-insensitive lookup by ENA's controlled-vocabulary value."""
+    """StrEnum with case-insensitive lookup by ENA's controlled-vocabulary value.
+    A raw value that is present but not in the vocabulary always raises; `required`
+    only governs whether a missing/empty value is an error or yields `None`.
+    """
 
     @classmethod
     def from_value(cls, raw_value: str | None, required: bool = False) -> Self | None:
--- a/ena-submission/src/ena_deposition/ena_types.py
+++ b/ena-submission/src/ena_deposition/ena_types.py
@@ -2,7 +2,7 @@ import dataclasses
 from collections import UserString
 from dataclasses import dataclass
 from enum import StrEnum
-from typing import Self, overload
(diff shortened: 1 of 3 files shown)


ena-submission/src/ena_deposition/ena_types.py
```diff
@@ -12,22 +12,19 @@ class XmlNone(UserString):
 class LookupStrEnum(StrEnum):
     """StrEnum with case-insensitive lookup by ENA's controlled-vocabulary value."""
 
-    @overload
     @classmethod
-    def from_value(cls, raw_value: str | None) -> Self | None: ...
```
anna-parker, line 17: defaults are not required, they should only be supplied via the manifest in the resolved from manifest function

ena-submission/src/ena_deposition/ena_types.py
```diff
@@ -12,22 +12,19 @@ class XmlNone(UserString):
...
+            if required:
+                msg = f"{cls.__name__} is required but no value was provided"
+                raise ValueError(msg)
+            return None
         normalized = raw_value.strip().lower()
         for member in cls:
             if member.value.lower() == normalized:
                 return member
-        return default
+        msg = f"{raw_value!r} is not a valid {cls.__name__}"
```
anna-parker, line 29: I guess the only thing that is debatable is if for not-required fields we can just return None if the raw value isnt a valid enum

ena-submission/src/ena_deposition/ena_types.py
```diff
@@ -12,22 +12,19 @@ class XmlNone(UserString):
...
-    @overload
     @classmethod
-    def from_value(cls, raw_value: str | None) -> Self | None: ...
-    @overload
-    @classmethod
-    def from_value(cls, raw_value: str | None, default: Self) -> Self: ...
-
-    @classmethod
-    def from_value(cls, raw_value: str | None, default: Self | None = None) -> Self | None:
+    def from_value(cls, raw_value: str | None, required: bool = False) -> Self | None:
```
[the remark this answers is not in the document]
anna-parker: I dont see any type errors so I would opt to not do this

ena-submission/src/ena_deposition/ena_types.py
anna-parker: this actually never happens, all enum fields are required

Issue #7247, opened 2026-09-07
Prohibit seqs with INSDC accessions in prepro (except for ingest)


corneliusroemer:
New prepro function/prohibition for INSDC accession field except for ingest user

PR #7302, opened 2026-09-10
experiment: refactor prepro datatypes

maverbiest:
**Not fully polished**

I was preparing to work on https://github.com/loculus-project/loculus/issues/7247 and thought the way we get the `group_id` from UnprocessedData and UnprocessedAfterNextclade in get_output_metadata currently is quite messy. 

I somehow completely blanked that I'd reviewed this PR https://github.com/loculus-project/loculus/pull/6980 and ended up with a similar concept (probably unconsciously plagiarising that PR!), with some differences:

- https://github.com/loculus-project/loculus/pull/6980 defines a ProcessingContext, which is very similar to the SubmissionContext I added here.
- Main difference is that I create SubmissionContext during parse_ndjson and add it to both UnprocessedEntry and UnprocessedAfterNextclade (so they have the same interface for getting group_id, submitter, accessionVersion etc.). The existing PR computes it per-call in _call_processing_function
- Existing PR also moves taxonomy_service into the context, out of the args, which is much better!

I'd argue we should merge https://github.com/loculus-project/loculus/pull/6980 rather than this PR: it's a smaller diff with less churn. But I wanted to post this as a draft since there are some different ideas on the data model that we could discuss or come back to in the future.

IMO UnprocessedEntry and UnprocessedAfterNextclade are quite natural places for the Context to move in the long run: it's context that remains the same from the time the entry is parsed until it is submitted back to the backend, and things like the accessionVersion and group_id are accessed in different places along the way (currently via dict["key"]). It would also get rid of the inconsistency where for UnprocessedAfterNextclade we need to do this dance to get the group_id and submitted_at:

https://github.com/loculus-project/loculus/blob/44f78e752e35f0f5aa7d56a037695f600eca86b6/preprocessing/nextclade/src/loculus_preprocessing/prepro.py#L392-L397

but not for UnprocessedData:

https://github.com/loculus-project/loculus/blob/44f78e752e35f0f5aa7d56a037695f600eca86b6/preprocessing/nextclade/src/loculus_preprocessing/prepro.py#L405-L406

### PR Checklist

### DIFF
--- a/preprocessing/nextclade/src/loculus_preprocessing/nextclade.py
+++ b/preprocessing/nextclade/src/loculus_preprocessing/nextclade.py
@@ -359,7 +359,7 @@ def write_nextclade_input_fasta(
     with open(input_file, "w", encoding="utf-8") as f:
         for entry in unprocessed:
             accession_version = entry.accessionVersion
-            for fasta_id, seq in entry.data.unalignedNucleotideSequences.items():
+            for fasta_id, seq in entry.unalignedNucleotideSequences.items():
                 id = f"{accession_version}__{fasta_id}"
                 id_map[accession_version, fasta_id] = id
                 f.write(f">{id}\n")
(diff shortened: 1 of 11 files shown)


maverbiest:
@claude review

preprocessing/nextclade/src/loculus_preprocessing/nextclade.py
```diff
@@ -793,19 +794,15 @@ def enrich_with_nextclade(  # noqa: PLR0914
             sequenceNameToFastaId: dict[SegmentName, str]
     )` object.
     """
+    submission_context: dict[AccessionVersion, SubmissionContext] = {
```
anna-parker, line 797: This is a good change!

maverbiest:
Closing as this it's going in through https://github.com/loculus-project/loculus/pull/7326

PR #7301, opened 2026-09-10
feat(config): start using outputFileUrlType website

anna-parker:
follow up to https://github.com/loculus-project/loculus/pull/7289

🚀 Preview: https://outputfileurltype-website.loculus.org

PR #7304, opened 2026-09-10
fix(website): Add `Access control allow origin` header for website file requests

tombch:
### Summary

Switching to `outputFileUrlType: website` in #7301 showed these errors in the CI:

```
Error: Unexpected console error: Access to fetch at 'http://10.1.0.60:3000/seq/LOC_000001Y.1/annotations/LOC_000001Y.1.embl' from origin 'http://localhost:3000' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.
```

this only appears for `chromium` because of logic in `console-warnings.fixture.ts`:

```
                if (!isHarmless && browserName === 'chromium') {
                    expect(false, `Unexpected console ${msg.type()}: ${messageText}`).toBe(true);
                }
            } 
```

CORS always hurts my head but through some discussions with Claude the issue is essentially:

- To get the file size, the website uses `fetch` to do a `HEAD` request to whatever the absolute file URL is.
- Before, with URL type `backend`, this was fine because the backend has `allowedOrigins('*')` in `WebConfig.kt`.
- However with URL type `website`, this hits `[fileName].ts` which never returned an `Allow origins: *` header to the browser. Therefore, any cross origin request here gets blocked by the browser.
- This only appears as an issue in CI because the playwright browser is at `localhost` but the fetch request goes to `10.1.0.60` which is the GitHub Actions runner's own private network address (therefore a cross origin request).

The fix is then to do what the backend does and `Allow origins: *` for this website endpoint. As far as I understand, this is not an issue security wise as with the current setup, `Allow origins: *` means a cross-origin request with user cookies would get a response but the browser will refuse to read it. If we also enabled `Access-Control-Allow-Credentials: true` that could cause problems with a malicious site making authenticated requests.

### Notes

I would think a more straightforward thing in future is to have the file size not need to be requested separately, and remove the separate `HEAD` request entirely, I think this is a bigger change though.

### PR Checklist

### DIFF
--- a/website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].spec.ts
+++ b/website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].spec.ts
@@ -48,4 +48,9 @@ describe('file download proxy route', () => {
         // Without 'manual' the file would be streamed through the website instead of from S3
         expect(backendRequest().redirect).toBe('manual');
     });
+
+    test('allows any origin on the S3 redirect', async () => {
+        const response = await callRoute({ isLoggedIn: false });
+        expect(response.headers.get('Access-Control-Allow-Origin')).toBe('*');
+    });
 });
--- a/website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts
+++ b/website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts
@@ -5,6 +5,13 @@ import { createAuthorizationHeader } from '../../../../utils/createAuthorization
 import { parseAccessionVersionFromString } from '../../../../utils/extractAccessionVersion';
 import { getAccessToken } from '../../../../utils/getAccessToken';
 
+// Allow requests from any origin
+// Same as the backend allows on /files/get, which this route proxies
+function withCorsHeader(headers: Headers): Headers {
+    headers.set('Access-Control-Allow-Origin', '*');
+    return headers;
+}
+
 async function proxyToBackend({ params, locals }: Parameters<APIRoute>[0], method: 'GET' | 'HEAD'): Promise<Response> {
     const runtimeConfig = getRuntimeConfig();
     const { accessionVersion, fileCategory, fileName } = params;
@@ -23,16 +30,22 @@ async function proxyToBackend({ params, locals }: Parameters<APIRoute>[0], metho
     if (response.status === 307 || response.status === 302) {
         const s3Url = response.headers.get('Location');
         if (!s3Url) {
-            return new Response('Backend redirect missing Location header', { status: 500 });
+            return new Response('Backend redirect missing Location header', {
+                status: 500,
+                headers: withCorsHeader(new Headers()),
+            });
         }
         return new Response(null, {
             status: response.status,
             // eslint-disable-next-line @typescript-eslint/naming-convention
-            headers: { Location: s3Url },
+            headers: withCorsHeader(new Headers({ Location: s3Url })),
         });
     }
 
-    return new Response(response.body, { status: response.status, headers: new Headers(response.headers) });
+    return new Response(response.body, {
+        status: response.status,
+        headers: withCorsHeader(new Headers(response.headers)),
+    });
 }
 
 export const GET: APIRoute = (ctx) => proxyToBackend(ctx, 'GET');


website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts
```diff
@@ -5,6 +5,13 @@ import { createAuthorizationHeader } from '../../../../utils/createAuthorization
 import { parseAccessionVersionFromString } from '../../../../utils/extractAccessionVersion';
 import { getAccessToken } from '../../../../utils/getAccessToken';
 
+// Allow requests from any origin
+// Same as the backend allows on /files/get, which this route proxies
```
Comment (theosanderson) [inline website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts:9]:
```suggestion
```

Just an idea - (A) maybe it's not needed (B) proxied would initially make me think we're passing through the whole file, which we're not (though I accept it's accurate)

Comment (theosanderson) [reply website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts:9]: [but we say proxy below so ignore B]

website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts
```diff
@@ -23,16 +30,22 @@ async function proxyToBackend({ params, locals }: Parameters<APIRoute>[0], metho
...
+                headers: withCorsHeader(new Headers()),
+            });
         }
         return new Response(null, {
             status: response.status,
             // eslint-disable-next-line @typescript-eslint/naming-convention
-            headers: { Location: s3Url },
+            headers: withCorsHeader(new Headers({ Location: s3Url })),
         });
     }
```
Comment (theosanderson) [inline website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts:43]: Reading this code I don't actually see that we need this branch at all?

Comment (theosanderson) [reply website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts:43]: [sorry I know that's not on you/ this PR]

Comment (theosanderson) [reply website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts:43]: @theosanderson-agent do you see a reason for the 307/302 branch?

Comment (theosanderson-agent) [reply website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts:43]: No, I don't think it's needed anymore. It looks like a leftover from history:

- The branch was added in #4298, when the fallthrough was `new Response(response.body, { status })` with **no headers**, so `Location` had to be forwarded explicitly.
- #5077 then changed the fallthrough to copy all backend headers (`new Headers(response.headers)`), which already includes `Location`. Since then the branch has been redundant: for a 307 the fallthrough would return the same status, the same `Location`, and now the same CORS header.
- The `302` check has never matched anything: the backend always returns `SC_TEMPORARY_REDIRECT` (307). The 302 seems to date from the pre-#4298 version of this route, which itself issued a 302.
- Node's `fetch` with `redirect: 'manual'` exposes the real 307 response (no opaque-redirect filtering like in browsers), so `Location` is readable in the fallthrough. The branch already relied on that anyway.

The only behavioural differences if it's removed: the 307 would also pass through any other backend headers (e.g. `Content-Length: 0`, `Vary`), which is harmless, and the "missing Location" 500 case would go away, which would just surface as a 307 without `Location` instead.

I checked locally: with the whole `if` block deleted, `[fileName].spec.ts` still passes (4/4), including the 307 + `Location` assertions.

website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts
```diff
@@ -5,6 +5,13 @@ import { createAuthorizationHeader } from '../../../../utils/createAuthorization
 import { parseAccessionVersionFromString } from '../../../../utils/extractAccessionVersion';
 import { getAccessToken } from '../../../../utils/getAccessToken';
 
+// Allow requests from any origin
+// Same as the backend allows on /files/get, which this route proxies
+function withCorsHeader(headers: Headers): Headers {
+    headers.set('Access-Control-Allow-Origin', '*');
+    return headers;
+}
```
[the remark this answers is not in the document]
Comment (anna-parker) [reply website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts:13]: I think adding a comment is sufficient for now

website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts
Comment (tombch) [reply website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts:43]: Yeah I guess it would be losing a guard that the location is there, but that is something the backend explicitly sets so should always be there really

website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts
Comment (tombch) [reply website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts]: Yeah I think the other approach would be for `FileSizeComponent` to call the website files endpoint with no host like the `FilesDialog` does, and remove it ever being a cross-origin request entirely. That seems better tbh!

Comment (anna-parker) [reply website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts:9]: claude actually thinks it is good to make these changes even if we were to refactor `FileSizeComponent`:
```
This is the URL the backend embeds in its public released-data API response (what getReleasedData/LAPIS consumers pull metadata
  from) for every file attached to a public sequence. Any third party building on top of that public API — a genome browser, an
  analysis notebook, some downstream tool — pulls that URL out of the JSON and does its own fetch() on it, from its own origin.
  That's the real cross-origin scripted consumer this header is for.
```

Comment (tombch) [reply website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts:9]: Yeah I can change the wording! In terms of the header, I think it makes sense out of the options, mainly as that's why the backend URLs already worked for the file size fetch. 

There's also the option to do a relative URL in the file size component (like how `FilesDialog` does) but I just tried this and its a lot of prop threading of accession version and file category. There's also the option to include file size in released data but not sure we want that?

Comment (anna-parker) [reply website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts:9]: > There's also the option to include file size in released data but not sure we want that?

I feel like this would make sense (but is probably not possible as released data is consumed by SILO so it needs a specific structure)

I think we should add this header either way for third party consumers

Comment (tombch) [reply website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts:9]: Yeah I agree

anna-parker (approved):
approving as I think the suggestions are small and the 302/307 branch can also be removed in a follow up PR if needbe

website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts
Comment (tombch) [reply website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts:43]: I opened an issue for this if we want to remove in a follow-up: #7317

website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts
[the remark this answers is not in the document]
Comment (tombch) [reply website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].ts:13]: I don't think we need this - if we ever want to support OPTIONS requests we can add when its needed

website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].spec.ts
```diff
@@ -48,4 +48,9 @@ describe('file download proxy route', () => {
         // Without 'manual' the file would be streamed through the website instead of from S3
         expect(backendRequest().redirect).toBe('manual');
     });
+
+    test('allows any origin on the S3 redirect', async () => {
+        const response = await callRoute({ isLoggedIn: false });
+        expect(response.headers.get('Access-Control-Allow-Origin')).toBe('*');
+    });
```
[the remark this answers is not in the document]
Comment (tombch) [reply website/src/pages/seq/[accessionVersion]/[fileCategory]/[fileName].spec.ts:55]: Added

PR #7309, opened 2026-09-11
fix(backend): improve duplicate ID error messages, exclude include postgres internals

### LINKED ISSUE #7106: Duplicate `submissionId` error shows users the raw Postgres exception with table name, constraint and internal UUID
When a metadata file has a duplicate `submissionId` we correctly return a 422, but we paste the raw Postgres exception into `detail`, so the user gets shown our table name, constraint name and an internal upload UUID:

```
Metadata file contains at least one duplicate submissionId: org.postgresql.util.PSQLException: ERROR: duplicate key value violates unique constraint "metadata_upload_aux_table_pkey"
  Detail: Key (upload_id, submission_id)=(308f5438-a27e-41e4-b507-fe18543a3547, SAMPLE_C) already exists.
```

Same for duplicate fasta ids in the sequence file, via `sequence_upload_aux_table_pkey`.

Slightly annoyingly the one bit that's actually useful to the submitter - which id is duplicated - is only there inside the Postgres `Detail:` line.

Comes from `${e.cause?.cause}` in `SubmitModel.kt` (L341 and L362). We already catch the constraint violation on purpose and map it to 422, so probably only the message needs fixing: drop the exception and name the ids instead, e.g. `Metadata file contains duplicate submissionId(s): SAMPLE_C`.

To reproduce, submit metadata with the same `submissionId` on two rows (or the same id twice in the fasta). Also reachable from the bulk upload UI, we don't check for duplicates client-side.

Not really a security issue but there's no reason to show submitters our schema.

Related:
- #6795
- #1882
- #4602
- #6976
- #5449

clonder:
Shows duplicate IDs instead of raw Postgres errors, using one helper for submission and revision.

Closes #7106

### PR Checklist

  - no docs changes needed
  - submission and revision endpoint tests
  - ran `./gradlew test ktlintCheck` locally, 638 tests passed and 1 skipped

### DIFF
--- a/backend/src/main/kotlin/org/loculus/backend/model/SubmitModel.kt
+++ b/backend/src/main/kotlin/org/loculus/backend/model/SubmitModel.kt
@@ -30,6 +30,7 @@ import org.loculus.backend.service.submission.SubmissionIdFilesMappingPreconditi
 import org.loculus.backend.service.submission.UploadDatabaseService
 import org.loculus.backend.utils.DateProvider
 import org.loculus.backend.utils.FastaReader
+import org.loculus.backend.utils.extractDuplicateRecordId
 import org.loculus.backend.utils.metadataEntryStreamAsSequence
 import org.loculus.backend.utils.revisionEntryStreamAsSequence
 import org.springframework.stereotype.Service
@@ -336,8 +337,10 @@ class SubmitModel(
             }
         } catch (e: ExposedSQLException) {
             if (e.sqlState == UNIQUE_CONSTRAINT_VIOLATION_SQL_STATE) {
+                val duplicateId = e.extractDuplicateRecordId(uploadId)?.value
                 throw DuplicateKeyException(
-                    "Metadata file contains at least one duplicate submissionId: ${e.cause?.cause}",
+                    "Metadata file contains at least one duplicate submissionId" +
+                        (duplicateId?.let { ": $it" } ?: ""),
                 )
             }
             throw e
@@ -357,8 +360,10 @@ class SubmitModel(
                 )
             } catch (e: ExposedSQLException) {
                 if (e.sqlState == UNIQUE_CONSTRAINT_VIOLATION_SQL_STATE) {
+                    val duplicateId = e.extractDuplicateRecordId(uploadId)?.value
                     throw DuplicateKeyException(
-                        "Sequence file contains at least one duplicate submissionId: ${e.cause?.cause}",
+                        "Sequence file contains at least one duplicate FASTA ID" +
+                            (duplicateId?.let { ": $it" } ?: ""),
                     )
                 }
                 throw e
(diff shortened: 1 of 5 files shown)


backend/src/test/kotlin/org/loculus/backend/controller/submission/SubmitEndpointTest.kt
```diff
@@ -213,6 +214,32 @@ class SubmitEndpointTest(
             .andExpect(jsonPath("\$.detail", containsString(expectedMessage)))
```
corneliusroemer-agent, line 214: `containsString` is how #7106 stayed hidden: the deleted cases asserted a prefix of the leaky message. Worth adding a shared matcher in `TestHelpers.kt` that asserts no `org.postgresql` / `PSQLException` / `violates unique constraint` / `Detail: Key` in the body, applied once here and in `ReviseEndpointTest.kt:486` so it covers every case without extra `Arguments` fields.

```suggestion
            .andExpect(jsonPath("\$.detail", containsString(expectedMessage)))
            .andExpect(containsNoDatabaseInternals())
```

Could later go global via `alwaysExpect` in a `MockMvcBuilderCustomizer` on `@EndpointTest`.

corneliusroemer:
Something's up with the backend tests - maybe flakiness, something didn't come up?

backend/src/test/kotlin/org/loculus/backend/controller/submission/SubmitEndpointTest.kt
corneliusroemer-agent: Roughly what I had in mind, in `TestHelpers.kt` (which already has `assertThat`, `not` and `ResultMatcher` imported):

```kotlin
private val databaseInternals = listOf(
    "org.postgresql",
    "PSQLException",
    "violates unique constraint",
    "Detail: Key",
    "_aux_table",
)

fun containsNoDatabaseInternals(): ResultMatcher = ResultMatcher { result ->
    val body = result.response.contentAsString
    databaseInternals.forEach { marker ->
        assertThat("Response leaks database internals ($marker): $body", body, not(containsString(marker)))
    }
}
```

I left "already exists" out of the list deliberately even though it appears in the Postgres detail — `FilesController.kt:116` uses that phrase legitimately.

backend/src/main/kotlin/org/loculus/backend/model/SubmitModel.kt
```diff
@@ -336,8 +337,10 @@ class SubmitModel(
             }
         } catch (e: ExposedSQLException) {
             if (e.sqlState == UNIQUE_CONSTRAINT_VIOLATION_SQL_STATE) {
+                val duplicateId = e.extractDuplicateRecordId(uploadId)?.value
```
corneliusroemer-agent, line 337: When extraction fails here there's no `log` call at all, and `DuplicateKeyException` takes no cause, so the exception is dropped — the only trace left is `ExceptionHandler`'s `log.info` of the sanitised message: no ID, no constraint, not even the SQL state. The revise path (`UploadDatabaseService.kt:117`) does log and says "contact an administrator" instead, so the same condition has two different contracts.

At minimum `log.error(e) { ... }` in the extraction-failed branch at all three sites — that branch only fires on an error shape we didn't anticipate, which is exactly when the stack trace is worth keeping. Same applies at `:363`.

corneliusroemer (commented):
Thanks, good stuff! I like that you centralize the postgres sql error extraction and reuse it.

Two things:
1. Apparently now certain errors don't get logged - when the error can't be parsed, see inline comment by @corneliusroemer-agent - would be good to fix
2. To prevent this kind of issue, it would be good to add a general test helper at appropriate places to assert on us not including raw postgres errors in user messages, again see inline comment suggestions

PR #7311, opened 2026-09-11
fix(ci): build docs image on native runners to avoid qemu hangs

### LINKED ISSUE #5642: Build docs docker image natively on arm then merge - like for website
Docs also hung just now at `npm ci` running on arm. We can do the same native build and merge thing as for website in #5514

Got stuck twice in a row actually

corneliusroemer-agent:
Resolves https://github.com/loculus-project/loculus/issues/5642

QEMU often hangs in docker builds, switch also docs image build to native dual build and merge - we did this for website image a while back in #5514 due to same failure mode. We're just copying the pattern from website image to docs - so low risk.

Docs images are used only by previews so we can see docs changes on previews. In prod docs are running via github pages.

More reliable and big speed up with arm native runner rather than QEMU: 10s of seconds rather than 5min (when it worked - when it doesn't it's 15min).

Still works:
image

<details>

The docs image has been failing on `main` roughly every other time it actually rebuilds. The arm64 half of the build dies inside `npm ci` under emulation and then sits there doing nothing until the job's 15 minute limit kills it:

```
#15 [linux/arm64 4/7] RUN npm ci
#15 23.90 qemu: uncaught target signal 4 (Illegal instruction) - core dumped
##[error]The operation was canceled.        <- 14 minutes later, nothing in between
```

This is the same failure [#5514](https://github.com/loculus-project/loculus/pull/5514) fixed for the website image in November ("Website ARM image builds hang about 1 in 10 times during `npm ci`"), so this is that fix applied to docs: build each architecture on a native runner, then merge the two manifests. Docs was the only npm-based image still building arm64 through emulation. That's the ecosystem that has this QEMU hang.

It is a hang and not a slow build, so raising the timeout would not have helped: green builds finish in 216-419 s and every failure sits at 883-888 s, which is the limit. Instances, all with the identical log line above: [33863848031](https://github.com/loculus-project/loculus/actions/runs/33863848031), [33867555784](https://github.com/loculus-project/loculus/actions/runs/33867555784), [34374633309](https://github.com/loculus-project/loculus/actions/runs/34374633309), [34589119390](https://github.com/loculus-project/loculus/actions/runs/34589119390) and [32502545323](https://github.com/loculus-project/loculus/actions/runs/32502545323) back in August. That is 4 of the last 18 builds that were not cache hits. Two were re-run and went green on the identical commit, so it is not triggered by any particular change.

</details>

🚀 Preview of docs using image built from this refactored workflow: https://docs-ci-docs-image-native-arm.loculus.org/

🚀 Preview: https://ci-docs-image-native-arm.loculus.org

### DIFF
--- a/.github/workflows/docs-image.yml
+++ b/.github/workflows/docs-image.yml
@@ -5,56 +5,34 @@ on:
     branches:
       - main
   workflow_dispatch:
-    inputs:
-      build_arm:
-        type: boolean
-        description: "Build for ARM as well"
-        default: false
-        required: false
   workflow_call:
-    inputs:
-      build_arm:
-        type: boolean
-        description: "Build for ARM as well"
-        default: false
-        required: false
 env:
   DOCKER_IMAGE_NAME: ghcr.io/loculus-project/docs
   BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
-  BUILD_ARM: ${{ github.event.inputs.build_arm || inputs.build_arm || github.ref == 'refs/heads/main' }}
   sha: ${{ github.event.pull_request.head.sha || github.sha }}
 concurrency:
-  group: ci-${{ github.ref == 'refs/heads/main' && github.run_id || github.ref }}-docs-${{github.event.inputs.build_arm}}
+  group: ci-${{ github.ref == 'refs/heads/main' && github.run_id || github.ref }}-docs
   cancel-in-progress: true
 jobs:
-  docs-docker:
-    name: Build Docs Docker Image
+  check-cache:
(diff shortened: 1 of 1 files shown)


maverbiest (approved):
Seems like a win! I read through the diff and also compared to https://github.com/loculus-project/loculus/pull/5514. As you said it's the same approach and the action is green here so should be good

maverbiest (approved):
Seems like a win! I read through the diff and also compared to https://github.com/loculus-project/loculus/pull/5514. As you said it's the same approach and the action is green here so should be good

PR #7314, opened 2026-09-13
feat(website): compare metadata before approving a revision

### LINKED ISSUE #7171: Add diff view for revisions on review page
While making some revisions on PPX I noticed how it was hard to check whether the changes were what I wanted them to be.

It would be very useful to have the submission review/approval page allow viewing a diff, similar to our released sequence version diff.

i.e. this page: https://pathoplexus.org/ebola-bdbv/submission/232/review should have something that has similar functionality to this page: https://pathoplexus.org/seq/PP_0079G2P.2/versions?compare=1%2C2

clonder:
Resolves #7171.

Revisions on the review page now show what changed compared to the previously released version: the metadata fields that differ (unchanged fields hidden by default) and whether the nucleotide sequence changed.

Review card with the metadata diff and nucleotide sequence status

### PR Checklist

- [x] Any manual testing that has been done is documented (i.e. what exactly was tested?) Tested in Chrome with local sample data.

🚀 Preview: https://feat-7171-review-metadata.loculus.org


corneliusroemer:
Would be great if you could include a screenshot/screencast of what your solution looks like :) That's super useful for reviewers so they don't need to click through things themselves

corneliusroemer:
One thing I was thinking that could be even better is not require one to go on a new page to see difference - we could show changes inline in the review card.

website/src/components/ReviewPage/ReviewCard.tsx
```diff
@@ -171,6 +180,34 @@ const ButtonBar: FC<ButtonBarProps> = ({
...
+                        ) : (
+                            <Button
+                                as='a'
+                                className={buttonBarClass(false)}
+                                href={metadataDiffUrl}
+                                aria-label={`View metadata changes for ${getAccessionVersionString(sequenceEntryStatus)}`}
+                                data-tooltip-id={`metadata-diff-tooltip-${sequenceEntryStatus.accession}`}
+                            >
+                                <GitCompare />
+                            </Button>
```
tombch: Not a blocking thing but could these two buttons be collapsed into a single one, without `as='a'`? And then instead of `metadataDiffUrl` prop, there could be e.g. a `diffAccessionVersion` handler passed down alongside the other handlers `editAccessionVersion`, `deleteAccessionVersion` etc

PR #7320, opened 2026-09-14
test(raw-reads-processing): fail instead of skip when readtools or deacon not found, allow skipping deliberately

corneliusroemer:
I noticed that a lot of Pytest tests were skipped when looking at a Claude session. It turns out we skip tests that require readtools and deacon respectively if they are missing.

I think it's better to fail by default and allow skipping deliberately

```
pytest -m "not needs_external_tools"
```

The marker is applied in conftest by fixture use rather than by hand, so a new test cannot forget it, and deselecting reports "13 deselected" rather than "skipped" — the summary now distinguishes "not asked for" from "tool missing".

Claude-Session: https://claude.ai/code/session_01CC9mVH9Wx9D7yoPsvXgE6k


raw-reads-processing/README.md
```diff
@@ -17,6 +17,26 @@ curl -L -o readtools.jar \
...
+
+Most tests are pure Python, but some run the real `readtools` jar and the real `deacon`
+binary — those are the only tests covering either tool, so they **fail** rather than skip
+when the tool is missing. The service image ships both, so this only comes up locally:
+download the jar as above (`_find_jar` also honours `READTOOLS_JAR` and the image path
+`/opt/app/lib/readtools.jar`), and install deacon via `environment.yml`.
+
+Those tests carry the `needs_external_tools` marker (applied automatically to anything
+using the `readtools_jar` or `deacon_server` fixtures). To run everything else without
+either tool installed:
```
corneliusroemer:
```suggestion
You can skip tests requiring `readtools` and `deacon` with:
```

raw-reads-processing/test/conftest.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def require_external_tool(what: str, how_to_get_it: str) -> None:
+    """Fail, rather than skip, when a tool the tests exercise is unavailable.
+
+    These are the only tests that run the real readtools/deacon binaries; the
+    rest is pure Python. A silent skip therefore makes a green local run look
+    like it covered them when it did not. The service image always has both, so
+    this only fires locally, where `-m "not needs_external_tools"` is the
+    deliberate way to leave them out.
```
anna-parker, line 26: this felt very LLM-ish
```suggestion
    """Fail, when a tool the tests exercise is unavailable.

    The service image should always have all required services, so
    this only fires locally. To skip instead run tests with `-m "not needs_external_tools"`
```

raw-reads-processing/pyproject.toml
```diff
@@ -19,6 +19,9 @@ packages = ["src/raw_reads_processing"]
 "test/**" = ["S101"]
 
 [tool.pytest.ini_options]
+# Tests that drive the real readtools jar / deacon binary. Applied automatically
```
anna-parker, line 22: Im not sure we really need to add a marker for this, i got claude to create https://github.com/loculus-project/loculus/pull/7356 as an alternative which touches less files - let me know what you think

clonder: compared this with #7356: the flag there skips only the tool that's actually missing, and the marker deselects both regardless. I think we maybe could go with #7356 and close this one?

anna-parker:
I merged the alternative PR, @corneliusroemer feel free to make a new PR if you still see any problems :-)

PR #7323, opened 2026-09-14
test(integration-tests): retry tooltip hover to hover not just once before React hydrates

corneliusroemer-agent:
Fix flake where hover over tooltip didn't happen when the browser was so fast that react hadn't hydrated yet. Hover fires only once when the mouse moves over the icon, if react hydrates after that, the tooltip never shows. Fix is to make the mouse move on it and back and on it again.

Flake looked like this in logs: `single-submit.spec.ts:45` with `#field-tooltipsampleCollectionDate`

The info icon is server-rendered, so it is visible immediately and the test hovers it; react-tooltip only mounts the tooltip element when React hydrates. Once the pointer is parked inside the anchor, attaching a `mouseenter` listener afterwards never sees an event.

Note: this is hard to test in practice as the flake happens only sometimes and the only signal we get is that this test no longer fails. At the very least it's not making things worse as CI has been passing reliably.

https://claude.ai/code/session_01G2SBL1Jp6B9XJ4AQfFnN68


integration-tests/tests/utils/hover-helpers.ts
```diff
@@ -0,0 +1,26 @@
...
+const defaultTimeout = 15_000;
+
+type HoverOptions = { timeout?: number };
+
+/**
+ * Hovers `anchor` until `target` becomes visible. A hover that lands before the anchor's island
+ * has hydrated is lost for good, because the pointer is already inside it when the listener is
+ * attached; moving away and back re-delivers it.
+ */
+export async function hoverUntilVisible(
```
maverbiest: nit but to me the name `hoverUntilVisible` doesn't capture that it moves back and forth. Maybe `retryHoverUntilVisible` or similar?

maverbiest (approved):
One nit but looks good!

integration-tests/tests/utils/hover-helpers.ts
corneliusroemer:
```suggestion
export async function hoverRepeatedlyUntilVisible(
```

integration-tests/tests/specs/features/single-submit.spec.ts
```diff
@@ -51,10 +52,7 @@ test('field description tooltip appears on hover', async ({ page, groupId }) =>
     const infoIcon = page.locator('[data-tooltip-id="field-tooltipsampleCollectionDate"]');
     await expect(infoIcon).toBeVisible();
 
-    await infoIcon.hover();
-
-    // The tooltip should become visible after the hover delay
     const tooltip = page.locator('#field-tooltipsampleCollectionDate');
-    await expect(tooltip).toBeVisible({ timeout: 2000 });
+    await hoverUntilVisible(infoIcon, tooltip);
```
corneliusroemer:
```suggestion
    await hoverRepeatedlyUntilVisible(infoIcon, tooltip);
```

integration-tests/tests/specs/features/single-submit.spec.ts
```diff
@@ -1,6 +1,7 @@
 import { expect } from '@playwright/test';
 import { test } from '../../fixtures/group.fixture';
 import { SingleSequenceSubmissionPage } from '../../pages/submission.page';
+import { hoverUntilVisible } from '../../utils/hover-helpers';
```
corneliusroemer:
```suggestion
import { hoverRepeatedlyUntilVisible } from '../../utils/hover-helpers';
```

Issue #3998, opened 2025-04-15
Limit max total size of files


chaoran-chen:
A maintainer should be able to limit the maximum total size of files uploaded to a Loculus instance. This is relevant for instances that allow untrusted users to upload files if they want to control costs. As far as I know (but I'm not 100% sure), S3 instances (like AWS) don't support imposing a hard size limit to a bucket, so we have to implement it in our backend.

The backend should stop providing new pre-signed URLs to upload files if they could lead to exceeding the limit. #3997 should be done first.

anna-parker:
In my understanding this can be done with a content header added by the backend when creating the pre-signed URL: 
For presigned POST, your backend creates a POST policy with:

```
["content-length-range", 0, 10485760]
```

That example limits uploads to 10 MiB. AWS documents content-length-range as a valid presigned POST: https://docs.aws.amazon.com/boto3/latest/reference/services/s3/client/generate_presigned_post.html?utm_source=chatgpt.comcondition.

maverbiest:
Starting to look at this now. Unfortunately it seems like the AWS Java SDK we use in Loculus does not support pre-signed POST requests with a `content-length-range` (https://github.com/aws/aws-sdk-java-v2/issues/1493, https://github.com/aws/aws-sdk-java-v2/issues/6577). This is a shame, means the approach will probably be to check the file size after the upload completes and reject it (and delete) if it's too big. The obvious downside is that the file will then briefly be in our S3 even if it's too large.

theosanderson-agent:
Thanks for digging into this @maverbiest — I looked into your finding and I can confirm the SDK limitation is exactly as you describe: aws-sdk-java-v2 has no presigned POST / POST-policy support (#1493 open since 2019, #6577 still open), so `content-length-range` isn't reachable through the SDK. Agreed that @anna-parker's suggestion can't be implemented as written.

I did want to flag an alternative that might avoid the "file briefly in S3" downside, since it works with the SDK we already have.

### Signing `Content-Length` into the presigned PUT

Setting `.contentLength(n)` on the `PutObjectRequest` / `UploadPartRequest` before presigning makes `content-length` a **signed** header (the URL comes out with `X-Amz-SignedHeaders=content-length;host;if-none-match`), so S3/MinIO rejects any request whose length differs — before storing anything.

I tested this against MinIO using our exact SDK version (2.47.1) and the same presigner configuration as `S3Service`:

**Single PUT**, signed for 100 bytes:

| Request | Result |
[...]

PR #7325, opened 2026-09-14
feat(website): add a file size limit to website (enforced before requesting presigned urls from backend)

anna-parker:
working towards https://github.com/loculus-project/loculus/issues/3998

This will not affect submissions made via the backend API (that will need to be handled in the backend)

To use this new feature add the limit to your values.yaml as follows: 
```
fileSharing:
  outputFileUrlType: website
  maxFileSizeBytes: 2147483648 # 2 GiB
```

### PR Checklist
Submitting an over 2GB file is blocked by the website without starting submission:
image


kubernetes/loculus/values.schema.json
```diff
@@ -1857,6 +1857,12 @@
           "groups": ["general"],
           "default": false,
           "description": "If true, only base filename restrictions are enforced; the stricter character allowlist (alphanumerics, underscores, periods, hyphens) is not applied. Not officially supported and may result in unexpected behaviour."
+        },
+        "maxFileSizeBytes": {
```
tombch, line 1861: I'm wondering, should this be configurable per file category rather than a single value across files? Although I guess then it would also need configuring per organism which is a bigger change

website/src/components/Edit/EditPage.tsx
```diff
@@ -279,6 +279,7 @@ const InnerEditPage: FC<EditPageProps> = ({
                         fileUploadStates={fileUploadStates}
                         setFileUploadStates={setFileUploadStates}
                         onError={(msg) => toast.error(msg, { position: 'top-center', autoClose: false })}
+                        maxFileSizeBytes={fileSharingConfig.maxFileSizeBytes}
```
tombch: Maybe the prop passed around could be `fileSharingConfig: FileSharingConfig`, e.g. similar to `clientConfig`, and means any future file sharing props would be available

kubernetes/loculus/values.schema.json
tombch: Actually the more I think about this a single cap for now sounds better, individual configurable file sizes isn't really needed yet

anna-parker: I think its fine to start with one hard limit and if users would like it more configurable we can add later!

website/src/components/Edit/EditPage.tsx
anna-parker: done :-)

website/src/components/Submission/DataUploadForm.tsx
```diff
@@ -471,6 +473,7 @@ export const ExtraFilesUpload = ({
     setFileUploadStates: Dispatch<SetStateAction<Map<string, FileUploadState>>>;
     fileLinkage?: FileLinkage;
     onError: (message: string) => void;
+    fileSharingConfig?: FileSharingConfig;
```
tombch: I think this doesn't need to have a `?` as its always provided? Same for `FolderUploadComponent`

tombch (approved):
Other comment is not blocking, looks good to me!

website/src/components/Submission/DataUploadForm.tsx
anna-parker: sorry I did push this but it seems like Github is having issues, I will wait until it shows up!

PR #7312, opened 2026-09-11
feat!(prepro): restrict preprocessing inputs for non-INSDC users

maverbiest:
resolves #7247 #7118

## Breaking change

This PR is breaking for deployments that use the default nextclade preprocessing pipeline. Previously, values provided by users for metadata fields marked as `noInput` in the `values.yaml` were silently ignored. Now, this leads to an error, meaning submission providing `noInput` values will be rejected by preprocessing.

Please ensure to communicate this new behaviour to users that may be affected by this, and update any internal workflows that currently submit entries with `noInput` metadata.

## Summary

This PR addresses https://github.com/loculus-project/loculus/issues/7118, which asks to error when (non-INSDC) submitters provide `noInput` metadata fields in their metadata files, and to add a warning when submitters provide metadata fields that are not in the schema. In doing so, it also resolves https://github.com/loculus-project/loculus/issues/7247 since INSDC accessions are all `noInput` in the config.

**Note:** I had to update the integration tests (the fixture `integration-tests/tests/test-data/cchfv_test_metadata.tsv`, inline generation for the CLI tests) because it still had `hostNameScientific` and `hostTaxonId`. These fields have been `noInput` since we added the taxonomy service, so they were silently discarded because no spec lists these fields as inputs (only `processed.hostTaxonId`). I intentionally kept some of these column around (just without values) because it tests that the error only appears if there is an actual value provided in the `noInput` column.

## Screenshot

Made a test group on the preview, submitted west nile sequence with `hostNameScientific` (which is `noInput`) in the metadata. It errors in preprocessing (I've updated the error message since this screenshot):

grafik

Then, I submitted the same sequence, same metadata as the INSDC ingest group. Here, it passes preprocessing with no issues:

grafik

This is a different submission, which shows that a warning is shown for the `Geo_Location` that was in the metadata file, but is not in our schema:

grafik

### PR Checklist
- [x] All necessary documentation has been adapted.
- [x] The implemented feature is covered by appropriate, automated tests.
- [x] Any manual testing that has been done is documented (i.e. what exactly was tested?)

🚀 Preview: Add `preview` label to enable

PR #7326, opened 2026-09-14
refactor(prepro): split ExternalServices out of ProcessingContext object and add all custom metadata fields to ProcessingContext 

anna-parker:
builds on work started in https://github.com/loculus-project/loculus/pull/6980 and https://github.com/loculus-project/loculus/pull/7302

- adds all custom metadata fields for an accession to the ProcessingContext field and stops overwriting the `inputMetadata` with these fields
- moves ExternalServices out of the ProcessingContext into its own object and updates tests (this is logical but also prevents circular imports now that context is used earlier on
- remove dead code (submissionId is not used in tests - same info comes from test name)
- simplify the date formatting test functions (majority of LOC is this change, overall it massively decreases the LOC) -> I can move this to its own PR if needbe

Includes changes from https://github.com/loculus-project/loculus/pull/7330
- `ProcessingContext` is now created during `parse_ndjson`
- `UnprocessedData` datatype is merged into `UnprocessedEntry`
- use `UnprocessedEntryFactory.create_unprocessed_entry` in the tests

### PR Checklist
- [ ] ~All necessary documentation has been adapted.~


preprocessing/nextclade/src/loculus_preprocessing/processing_functions.py
```diff
@@ -39,19 +39,6 @@
 options_cache: dict[str, dict[str, str]] = {}
 
 
-@dataclass(frozen=True)
```
anna-parker, line 42: moved to datatypes

preprocessing/nextclade/src/loculus_preprocessing/datatypes.py
```diff
@@ -112,9 +114,25 @@ class UnprocessedEntry:
...
+    """Runtime context that is the same for every processing function call for a given
+    accession, as opposed to `FunctionArgs` which holds the literal, per-function arguments
+    declared in the organism's YAML config.
+    """
+
+    accession_version: AccessionVersion
+    submissionId: str  # noqa: N815
+    submitter: str
+    is_insdc_ingest_group: bool
+    group_id: int
```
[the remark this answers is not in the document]
anna-parker: not sure this is really an improvement but I guess we can do it

preprocessing/nextclade/src/loculus_preprocessing/datatypes.py
maverbiest, line 196:
```suggestion
submitter: str
group_id: int
```

I believe we can do this now because we now get this from the properly typed context

maverbiest:
Nice! I agree that this makes prepro a lot more readable.

I wonder if it makes sense to add the ExternalServices to the ProcessingContext as well? Atm we only have the taxonomy service for which the URL is configured at the Config level (so same across the board), and it would save an argument for the processing functions. But maybe this would be an issue in the future?

There were also some things I thought about when working on ~https://github.com/loculus-project/loculus/pull/7312~ https://github.com/loculus-project/loculus/pull/7302 that I personally think are worth considering. We don't need to add them if you feel differently but maybe we can discuss:
1. Since all values we're putting in the context are known from the start we can construct the ProcessingContext right in `parse_ndjson()`
2. This means that `UnprocessedData` can be simplified: it currently carries `submitter`, `group_id`, `submittedAt`, `submissionId`. All of this would be folded into one `context` field.
3. `UnprocessedEntry` is just an `accessionVersion` and `UnprocessedData`. After **2**, `UnprocessedData` already carries the `accessionVersion`. This means we can simplify by combining the two types into:
```python
@dataclass
class UnprocessedEntry:
    context: ProcessingContext  # noqa: N815
    metadata: InputMetadata
    unalignedNucleotideSequences: dict[SequenceName, NucleotideSequence | None]  # noqa: N815
    files: dict[FileCategory, list[FileIdAndNameAndReadUrl]] | None

    @property
    def accessionVersion(self) -> AccessionVersion:  # noqa: N802
        return self.context.accessionVersion
```

anna-parker:
> I wonder if it makes sense to add the ExternalServices to the ProcessingContext as well?

I initially wanted to do this but wasnt able to do it withoyt a circular dependency, but I think it is ok to split them up as they are different things

anna-parker:
> Since all values we're putting in the context are known from the start we can construct the ProcessingContext right in parse_ndjson()

Hmm... good idea this might be possible! if you want to make a PR onto this with the changes today all ready that would be great as I might not get to it soon :-)

maverbiest:
> make a PR onto this with the changes

Sure! I got started on this here: https://github.com/loculus-project/loculus/pull/7330

anna-parker:
@claude review

maverbiest (approved):
I did another pass through the diff, this looks really good to me!

I also ran the regression test pipeline to compare the preview of this branch to main's preview, which showed no differences:

```sh
(pp-integrity) ➜  regression-testing git:(main) ✗ for diff in results/*.diff; do
        echo "Differences in $diff:"
        grep -v '^[[:space:]]*$' "$diff" | wc -l
    done
Differences in results/cchf-multi-ref.meta.main.prepro-refactor-context.diff:
0
Differences in results/cchf-multi-ref.seq.main.prepro-refactor-context.diff:
0
Differences in results/cchf.meta.main.prepro-refactor-context.diff:
0
Differences in results/cchf.seq.main.prepro-refactor-context.diff:
0
Differences in results/ebola-sudan.meta.main.prepro-refactor-context.diff:
0
Differences in results/ebola-sudan.seq.main.prepro-refactor-context.diff:
0
Differences in results/enteroviruses.meta.main.prepro-refactor-context.diff:
0
Differences in results/enteroviruses.seq.main.prepro-refactor-context.diff:
0
Differences in results/west-nile.meta.main.prepro-refactor-context.diff:
0
Differences in results/west-nile.seq.main.prepro-refactor-context.diff:
0
```

PR #7313, opened 2026-09-11
feat(docs): update the loculus docs with AWS specific configuration information

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/4926

Additionally fixes the values.schema.json: Protocol prefix is actually required and backend will not be able to talk to S3 without it.

### PR Checklist
- [x] All necessary documentation has been adapted.
- [ ] ~The implemented feature is covered by appropriate, automated tests.~
- [ ] ~Any manual testing that has been done is documented (i.e. what exactly was tested?)~

🚀 Preview: Add `preview` label to enable

PR #7327, opened 2026-09-15
feat!(backend): use abortMultiPartUpload to limit file size in backend

### LINKED ISSUE #3997: Limit max. size of a file
The maintainer should be able to limit the max size of a file.

To do that, we need to move away from supporting PUT-based uploads and use POST-based uploads only.

anna-parker:
claude's attempt at https://github.com/loculus-project/loculus/issues/3998

resolves https://github.com/loculus-project/loculus/issues/3997, https://github.com/loculus-project/loculus/issues/3998

AWS sets a size limit of 5GB for files uploaded with single presigned URLs, adding a stricter limit there only seems possible by forcing users to add a `content-length` - which seems too tedious, so we will leave the 5GB limit as is. 

Multi part uploads however are more risky as the maximum object size is 48.8 TiB, which we never want to accidentally accept. This PR enforces a size limit by aborting uploads when `complete-multi-part-upload` is called if the size of all parts is over a preconfigured limit.

`getFileSize/headObject` can't see the file until after CompleteMultipartUpload — S3 only assembles the object at completion time; before that, the key doesn't exist, only loose uploaded parts do. Instead we check size before completing, using S3's ListParts API (which reports each uploaded part's size while the upload is still in progress), and aborting via [AbortMultipartUploadRequest](https://docs.aws.amazon.com/de_de/AmazonS3/latest/API/API_AbortMultipartUpload.html) if it's too big.

This means an oversized file is never assembled into a stored S3 object (there will still storage cost until it is aborted sadly)

### BREAKING CHANGES

This sets the file size limit for all file uploads as 5GB, this can be increased or decreased for multi-part uploads and submissions via the webpage (for single part uploads via the API the default will remain 5GB), this can be modified using 

```
fileSharing:
  outputFileUrlType: website
  maxFileSizeBytes: 2147483648 # 2 GiB
```

### PR Checklist

#### @tombch tested locally with a custom limit of 12000 bytes, and commented out the website validation: 

I uploaded and re-uploaded the same 5kb file, and it passed under the limit (assigned file ID `FILE_0001` and `FILE_0003`):

image

I uploaded and re-uploaded a file that was too big (assigned file ID `FILE_0002` and `FILE_0004`):

image

image

And minio only has the objects under the limit:
```
$ kubectl exec deploy/minio -- mc ls --recursive local/
[2026-09-15 11:42:40 UTC] 5.1KiB STANDARD loculus-preview-private/files/FILE_0001
[2026-09-15 11:47:30 UTC] 5.1KiB STANDARD loculus-preview-private/files/FILE_0003
```

Also tested locally with the `maxFileSizeBytes` removed from `values.yaml` and no file size validation was applied.

#### On preview submitted 3 parts, 2 over 2.52GB via the API

Via the API I requested 3 presigned URLs, I then uploaded large files to each, stored the etags and then called 

```
curl -X 'POST' \
  'https://backend-file-size-backend.loculus.org/files/complete-multipart-upload' \
  -H 'accept: */*' \
  -H 'Authorization: Bearer <REDACTED>' \
  -H 'Content-Type: application/json' \
  -d '[
  {
    "fileId": "FILE_009F",
    "etags": [
      "5a38c8408b492d57ae206e467e4a0e21", "d7f9172456756a35fac366631955ebb3", "d7f9172456756a35fac366631955ebb3"
    ]
  }
]'
```

```
{
  "title": "Unprocessable Content",
  "status": 422,
  "detail": "File FILE_009F exceeds the maximum allowed file size of 2147483648 bytes (actual size: 5611111762 bytes). The upload has been aborted.",
  "instance": "/files/complete-multipart-upload"
}
```

🚀 Preview: https://file-size-backend.loculus.org


backend/src/test/resources/backend_config_s3_max_file_size.json
anna-parker, line 1: is this new config actually needed?

backend/src/main/kotlin/org/loculus/backend/service/files/S3Service.kt
```diff
@@ -114,6 +116,42 @@ class S3Service(private val s3Config: S3Config) {
...
+                    .key(getFileIdPath(fileId))
+                    .uploadId(uploadId)
+                    .partNumberMarker(partNumberMarker)
+                    .build(),
+            )
+            totalSize += response.parts().sumOf { it.size() }
+            partNumberMarker = if (response.isTruncated) response.nextPartNumberMarker() else null
+        } while (partNumberMarker != null)
+        totalSize
+    }
```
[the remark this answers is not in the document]
anna-parker: I think this is ok, its extra cautious and if a useer is uploading this much we can safely abort IMO

backend/src/main/kotlin/org/loculus/backend/controller/FilesController.kt
```diff
@@ -203,11 +207,26 @@ class FilesController(
...
+                    s3Service.abortMultipartUpload(fileId, uploadId)
+                    throw UnprocessableEntityException(
+                        "File $fileId exceeds the maximum allowed file size of $maxFileSizeBytes bytes " +
+                            "(actual size: $fileSize bytes). The upload has been aborted.",
+                    )
+                }
+            }
+
             s3Service.completeMultipartUpload(fileId, uploadId, etags)
             filesDatabaseService.completeMultipartUpload(fileId)
```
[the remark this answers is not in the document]
anna-parker: but thats fine, you should remove the large files from the batch (I guess the only issue is that the error message is a bit confusing maybe as those submissions havent been completed but aborted)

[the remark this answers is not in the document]
anna-parker: I improved the error message, not sure its really necessary but oh well

kubernetes/loculus/values.yaml
```diff
@@ -62,6 +62,7 @@ dataUseTerms:
     restricted: https://#TODO-MVP/restricted
 fileSharing:
   outputFileUrlType: website
+  maxFileSizeBytes: 5368709120 # 5 GiB
```
tombch, line 65: I think the max file size in `values_preview_server` should be removed / updated to 5GB

backend/src/main/kotlin/org/loculus/backend/service/files/S3Service.kt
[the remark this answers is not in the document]
tombch: I agree - I don't think we need to worry about users doing something like this. If it happened its also a sign they've messed up their upload somehow

backend/src/test/kotlin/org/loculus/backend/controller/files/CompleteMultipartUploadFileSizeLimitEndpointTest.kt
```diff
@@ -0,0 +1,142 @@
...
+
+    @BeforeEach
+    fun prepareNewGroup() {
+        groupId = groupManagementClient
+            .createNewGroup(group = DEFAULT_GROUP, jwt = jwtForDefaultUser)
+            .andGetGroupId()
+    }
+
+    @Test
+    fun `GIVEN a file exceeding maxFileSizeBytes THEN completion is rejected and the upload is aborted`() {
```
tombch, line 65: not blocking but can these tests be folded into `CompleteMultipartUploadEndpointTest.kt` ? Imo would make sense to be there

tombch: ah these tests use a different file size. Not worth changing!

backend/src/main/kotlin/org/loculus/backend/controller/FilesController.kt
```diff
@@ -197,17 +204,56 @@ class FilesController(
...
         val multipartUploadIds = filesDatabaseService.getUncompletedMultipartUploadIds(fileIdsAndEtags.keys)
-        val alreadyCompleted = fileIdsAndEtags.keys - multipartUploadIds.map { it.first }.toSet()
-        if (alreadyCompleted.isNotEmpty()) {
-            throw UnprocessableEntityException(
-                "The following files have already been completed: " + alreadyCompleted.joinToString(),
-            )
+        val notPending = fileIdsAndEtags.keys - multipartUploadIds.map { it.first }.toSet()
+        if (notPending.isNotEmpty()) {
+            val nonExistent = filesDatabaseService.getNonExistentFileIds(notPending)
+            val alreadyCompleted = notPending - nonExistent
```
tombch, line 210: Ah so does this mean previously, an unknown file id would have come back with a 'the following have already been completed' message?

tombch: Definitely good to now have the error distinguish between completed vs. not found

tombch:
I tested locally with a custom limit of 12000 bytes, and commented out the website validation: 

I uploaded and re-uploaded the same 5kb file, and it passed under the limit (assigned file ID `FILE_0001` and `FILE_0003`):

image

I uploaded and re-uploaded a file that was too big (assigned file ID `FILE_0002` and `FILE_0004`):

image

image


And minio only has the objects under the limit:
```
$ kubectl exec deploy/minio -- mc ls --recursive local/
[2026-09-15 11:42:40 UTC] 5.1KiB STANDARD loculus-preview-private/files/FILE_0001
[2026-09-15 11:47:30 UTC] 5.1KiB STANDARD loculus-preview-private/files/FILE_0003
```

Also tested locally with the `maxFileSizeBytes` removed from `values.yaml` and no file size validation was applied.

backend/src/main/kotlin/org/loculus/backend/controller/FilesController.kt
anna-parker: > Ah so does this mean previously, an unknown file id would have come back with a 'the following have already been completed' message?

I think so!

backend/src/test/kotlin/org/loculus/backend/controller/files/CompleteMultipartUploadFileSizeLimitEndpointTest.kt
anna-parker: these 2 tests can be combined though as they basically share the start

kubernetes/loculus/values.schema.json
```diff
@@ -1862,7 +1862,7 @@
           "type": "integer",
           "minimum": 1,
           "groups": ["general"],
-          "description": "The maximum allowed size, in bytes, of an individual uploaded file. Files exceeding this size are rejected by the website. If unset, no limit is enforced."
+          "description": "The maximum allowed size, in bytes, of an individual uploaded file. Files exceeding this size are rejected by the website and by the backend for multipart uploads (note for single part uploads the default limit of 5GB cannot be modified). If unset, no limit is enforced."
```
anna-parker:
```suggestion
          "description": "The maximum allowed size, in bytes, of an individual uploaded file. Files exceeding this size are rejected by the website and by the backend for multipart uploads (note for single part uploads the default maximum size limit of 5GB cannot be modified and even if a lower `maxFileSizeBytes` threshold is configured it cannot be enforced for single part uploads). If unset, no limit is enforced."
```

kubernetes/loculus/values.schema.json
```diff
@@ -1862,7 +1862,7 @@
           "type": "integer",
           "minimum": 1,
           "groups": ["general"],
-          "description": "The maximum allowed size, in bytes, of an individual uploaded file. Files exceeding this size are rejected by the website. If unset, no limit is enforced."
+          "description": "The maximum allowed size, in bytes, of an individual uploaded file. Files exceeding this size are rejected by the website and by the backend for multipart uploads (note for single part uploads the default maximum size limit of 5GB cannot be modified and even if a lower `maxFileSizeBytes` threshold is configured it cannot be enforced for single part uploads). If unset, no limit is enforced."
```
anna-parker:
```suggestion
          "description": "The maximum allowed size, in bytes, of an individual uploaded file. Files exceeding this size are rejected by the website and by the backend for multipart uploads (note for single part uploads the default maximum size limit of 5GB cannot be modified and even if a lower `maxFileSizeBytes` threshold is configured it cannot be enforced for single part uploads). If unset, the website does not enforce a limit and multi-part uploads use the configured S3 service limit."
```

kubernetes/loculus/values.yaml
anna-parker: we discussed this is fine for now :-)

docs/src/content/docs/for-administrators/configuring-extra-files.md
```diff
@@ -85,6 +87,8 @@ The example above configures the `rawReads` file category.
 
 If a user submits these files, they will be passed along to the processing pipeline as well, and the pipeline can read them, pass them through as output files, or generate additional fields or process them in any other way.
 
+You can also set a maximum accepted file size for your loculus instance in `fileSharing.maxFileSizeBytes`.
```
tombch: Not blocking but a note here as well about the single part uploads being unconditionally max 5GB might be good

tombch (approved):
Looks good to me!

docs/src/content/docs/for-administrators/configuring-extra-files.md
tombch:
```suggestion
You can also set a maximum accepted file size for your loculus instance in `fileSharing.maxFileSizeBytes`. Files exceeding this size are rejected by the website and by the backend for multipart uploads (note for single part uploads the default maximum size limit of 5GB cannot be modified and even if a lower `maxFileSizeBytes` threshold is configured it cannot be enforced for single part uploads). If unset, the website does not enforce a limit and multi-part uploads use the configured S3 service limit."
```

backend/src/main/kotlin/org/loculus/backend/controller/FilesController.kt
```diff
@@ -197,17 +204,56 @@ class FilesController(
...
                 throw UnprocessableEntityException("No etags provided for file ID $fileId.")
             }
+
+            if (maxFileSizeBytes != null) {
+                val fileSize = s3Service.getMultipartUploadSize(fileId, uploadId)
+                if (fileSize > maxFileSizeBytes) {
+                    // Abort so the oversized file is never assembled into an object in storage.
+                    s3Service.abortMultipartUpload(fileId, uploadId)
+                    log.warn {
+                        "Aborted multipart upload for file $fileId: actual size $fileSize bytes exceeds the " +
```
corneliusroemer, line 238: Might be worth mentioning the uploadId as well to allow better log correlation

backend/src/main/kotlin/org/loculus/backend/service/files/S3Service.kt
```diff
@@ -114,6 +116,42 @@ class S3Service(private val s3Config: S3Config) {
...
+
+    fun abortMultipartUpload(fileId: FileId, uploadId: String) = s3ErrorMapping {
```
corneliusroemer, line 143: Does this delete the parts from s3? If not, we might want to keep track of aborted multipart uploads in our db so we can clean up ourselves later - and also to be able to track who uploaded etc.

backend/src/test/kotlin/org/loculus/backend/controller/files/CompleteMultipartUploadFileSizeLimitEndpointTest.kt
```diff
@@ -0,0 +1,127 @@
...
+
+        filesClient.completeMultipartUploads(listOf(fileIdAndEtags))
+            .andExpect(status().isUnprocessableContent)
+            .andExpect(
+                content().string(containsString("exceeds the maximum allowed file size of $MAX_FILE_SIZE_BYTES bytes")),
+            )
+            .andExpect(content().string(containsString(fileIdAndUrls.fileId)))
+
+        // The file should be discarded from S3 and the database.
+        assertNull(s3Service.getFileSize(fileIdAndUrls.fileId))
```
corneliusroemer, line 84: Does it check it's gone from s3 or just from our db files table?

docs/src/content/docs/for-administrators/configuring-extra-files.md
```diff
@@ -85,6 +87,8 @@ The example above configures the `rawReads` file category.
 
 If a user submits these files, they will be passed along to the processing pipeline as well, and the pipeline can read them, pass them through as output files, or generate additional fields or process them in any other way.
 
+You can also set a maximum accepted file size for your loculus instance in `fileSharing.maxFileSizeBytes`. Files exceeding this size are rejected by the website and by the backend for multipart uploads (note for single part uploads the default maximum size limit of 5GB cannot be modified and even if a lower `maxFileSizeBytes` threshold is configured it cannot be enforced for single part uploads). If unset, the website does not enforce a limit and multi-part uploads use the configured S3 service limit."
```
corneliusroemer, line 90: Why can't we enforce limit for single-part? Can't we, at time of submission (rather than file upload), reject the submission if it's got a too large file irrespective of whether it was single or multi-uploaded?

I'm not sure I fully grok why we can't do the size rejection at time of submission as opposed to file upload. Yes, the files would materialize if upload is completed but we would garbage collect them as they won't be pointed at by any entries?

I might miss something here - but IIUC the code here would still leave the multipart parts undeleted even after rejecting the multipart completion?

corneliusroemer:
Some comments after reading through, nothing blocking but some things might potentially be worth a follow up if I didn't misunderstand the code.

Update: Ah I see this path isn't supposed to happen on the website, you said you commented out website validation. Still I guess it might be nice to show problem detail to website users in general.

Not super important but just wondering reading through PR desc: @tombch do you know why we don't give the more accurate error here (file too large):

image

Was this screenshot maybe produced with an earlier version of this PR that didn't yet have the new error messages?

`Failed to make request` is also not quite right, it should be `Request failed` as the request was made it just failed with 422.

backend/src/main/kotlin/org/loculus/backend/controller/FilesController.kt
corneliusroemer-agent: Correct — `uploadId` is destructured in the enclosing `forEach`, so it's in scope right here. Group and uploader would be worth adding too: `deleteFileEntry` three lines down removes the only record of who this was.

backend/src/main/kotlin/org/loculus/backend/service/files/S3Service.kt
corneliusroemer-agent: Yes — `AbortMultipartUpload` frees the storage of the previously uploaded parts, that's what it's for. So this path needs no extra cleanup.

The DB half of your comment stands though. `deleteFileEntry` drops the row outright and this branch only `log.warn`s, with no `auditLogger.log` — unlike `S3GarbageCollectionTask`, which audit-logs its deletions. So there's no durable record that a given group tried to push an oversized file.

There is a leak here, but a different one: multipart uploads that are **never completed at all**. Nothing ever aborts those —

- no `ListMultipartUploads` sweep anywhere in `backend/`
- no `AbortIncompleteMultipartUpload` lifecycle rule; `minio-deployment.yaml`'s `postStart` does only `mc mb` + `mc anonymous set-json`
- `S3GarbageCollectionTask` → `s3Service.deleteFile` → `deleteObject`, which doesn't touch pending parts, and `getDeletionCandidateFiles` works off the `files` table, so the row gets deleted while the parts stay

A user can request presigned part URLs, upload 40 TB of parts, never call complete, and nothing reclaims it. Worth an issue plus a bucket lifecycle rule.

Separately: `abortMultipartUpload` isn't wrapped in try/catch while `deleteFileEntry` is. If the abort throws, `s3ErrorMapping` maps it and the client gets an S3 error rather than the size message, and the DB row survives.

backend/src/test/kotlin/org/loculus/backend/controller/files/CompleteMultipartUploadFileSizeLimitEndpointTest.kt
corneliusroemer-agent: S3 — `getFileSize` is a `headObject` (`S3Service.kt:272`); the DB is checked separately on the next line via `getNonExistentFileIds`.

But the assertion is vacuous as a check that the abort worked: the object doesn't exist until `CompleteMultipartUpload` runs, so `assertNull(getFileSize(...))` passes even if `s3Service.abortMultipartUpload(...)` were deleted from the controller. Nothing in this test covers the abort. `listParts` returning empty parts, or throwing `NoSuchUpload`, is what would.

docs/src/content/docs/for-administrators/configuring-extra-files.md
corneliusroemer-agent: Taking the parts in order.

**Single-part at upload time** — the doc is right. A presigned PUT can't bind `Content-Length`; enforcing a size at upload would mean a presigned POST policy with `content-length-range`, which is a different upload mechanism.

**Size rejection at submission time** — this works, and the hook already exists. `FileMappingPreconditionValidator.validateFilesUploaded` already calls `s3Service.getFileSize(fileId)` for every file and stores it via `setFileSize`. Adding `if (fileSize > maxFileSizeBytes) throw UnprocessableEntityException(...)` there is a few lines and covers single-part and multipart uniformly. One caveat: `getUncheckedFileIds` filters to `size IS NULL`, so a file ID that's already been sized is skipped — the check would want to read the stored size too, not just newly-measured ones.

**"we would garbage collect them"** — correct, `S3GarbageCollectionTask` deletes files not referenced by any submission. But the default `gcOrphanRetentionPeriodMinutes` is 10080 (7 days) and it's two-phase (mark on one run, delete on a later one, task runs daily). So an oversized single-part upload sits in the bucket for at least a week, and someone can park an arbitrary number of 5 GB objects for that long. That's the argument for keeping the immediate multipart abort *and* adding the submission-time check, rather than swapping one for the other.

**"would still leave the multipart parts undeleted"** — this one is wrong: `AbortMultipartUpload` does free the parts. The parts that genuinely leak are from multipart uploads that are never completed at all; details in the `S3Service.kt` thread.

corneliusroemer-agent:
On the `Failed to make request` screenshot — not an old build. The website can't display *any* backend problem detail.

`website/src/types/backend.ts:278` has `type: z.string()` as required, and the backend never sends `type`. The curl output in this PR description shows it: `{"title","status","detail","instance"}`. Live main agrees:

```
$ curl 'https://backend-main.loculus.org/ebola-sudan/get-released-data?compression=bogus'
{"title":"Bad Request","status":400,"detail":"Failed to convert 'compression' with value: 'bogus'","instance":"/ebola-sudan/get-released-data"}
```

So `problemDetail.safeParse(...)` fails at `backendClient.ts:195` and every error falls through to the generic `Failed to make request: ${e.message}` at `:204`. The backend's 422 here is fine — `FolderUploadComponent` surfaces `err.detail` correctly, it just never gets a parsed detail to surface. This affects every backend error shown in the website, not just this endpoint.

`backendClient.spec.ts:23` is why it went unnoticed: the fixture in "returns a ProblemDetail received from the backend" includes `type`, so the test passes against a response shape the backend doesn't produce.

Making `type` optional in the zod schema (and fixing the fixture) should restore the real messages everywhere. Agreed on `Request failed` over `Failed to make request` while touching that line.

backend/src/main/kotlin/org/loculus/backend/service/files/S3Service.kt
anna-parker: we delete multipart uploads that are never completed with a lifecycle policy in the S3 bucket

anna-parker: so this is handled

anna-parker:
> Was this screenshot maybe produced with an earlier version of this PR that didn't yet have the new error messages?

@corneliusroemer we have nice website error messages, @tombch intentionally removed them to check the backend alone will prevent large file uploads

see https://github.com/loculus-project/loculus/pull/7325 with a picture

@corneliusroemer I addressed all your comments, now I dont think any of the issues you raised are a problem as we have lifecycle policies on s3 (see https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpu-abort-incomplete-mpu-lifecycle-config.html) to handle incomplete uploads, aborting the upload removes parts immedietly and garbage collection will clean up any db entries unlinked to submissions

This is covered in https://github.com/loculus-project/loculus/pull/7313 - I now added you as a reviwer!

docs/src/content/docs/for-administrators/configuring-extra-files.md
anna-parker: exactly the only way to enforce limits in single part uploads is via `content-length` (@theosanderson was against this in the past) unless we switch to presigned POST requests where we could add `content-length-range` (much more complicated and out of scope).

anna-parker: > the code here would still leave the multipart parts undeleted

claude is correct, this is wrong `AbortMultipartUpload` frees up the parts (additionally we have a lifecycle policy to delete unfinished multi part uploads after 1day)

backend/src/main/kotlin/org/loculus/backend/controller/FilesController.kt
anna-parker: we could do this

backend/src/test/kotlin/org/loculus/backend/controller/files/CompleteMultipartUploadFileSizeLimitEndpointTest.kt
anna-parker: ah this is maybe a testing gap we could add in a follow up

PR #7341, opened 2026-09-17
fix(preprocessing): make EMBL features valid INSDC on truncated and minus-strand CDSes

corneliusroemer-agent:
Fixing 3 additional things that the base branch doesn't yet:

- ID line said `RNA` in our golden file which is invalid per ENA flatfile spec.
- Add strand to gene - this is required per GFF spec, nextclade doesn't add it (yet) but it's easy to infer from first segment
- Flatfile spec wants truncation markers on truncated translations. This is not too hard as nextclade already includes truncation info in the JSON

### Related

Biopython has a bug in that it emits `BiopythonWarning: Non-standard molecule type: genomic RNA` on the ID line change. Its check is written against GenBank's shorter list, and `genomic RNA` is what ENA asks for, so the warning is wrong - will report upstream

https://claude.ai/code/session_01CrcBoGWBYP6uWL1w9j5xnV


preprocessing/nextclade/src/loculus_preprocessing/embl.py
```diff
@@ -1,9 +1,37 @@
...
+  keeps those entries correct.
+- `strand`, `phase` and `truncation` sit on a CDS's *segments*, never in `attributes`.
+  Taking `phase` from the attributes instead yields codon_start 1 for every CDS and
+  mistranslates any 5'-truncated one. Use `phase`, not `frame`; they differ.
+- Every `attributes` value is a `list[str]` -- GFF3 attributes are multi-valued. Values
+  used as numbers need converting, and a repeated value becomes a repeated qualifier.
+- Segments are listed in GFF row order, which is not necessarily transcription order: a
+  coordinate-sorted GFF splices a minus-strand multi-exon CDS back to front. Nextclade's
+  own translations share the bug, so such a dataset must list minus-strand exons 5'->3'.
+- Translations always use the standard genetic code; /transl_table is not honoured.
```
corneliusroemer:
```suggestion
```

preprocessing/nextclade/src/loculus_preprocessing/embl.py
```diff
@@ -1,9 +1,19 @@
+"""Build EMBL features from a Nextclade annotation.
+"""
```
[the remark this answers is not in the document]
corneliusroemer-agent: Fixed in d4696dfa.

preprocessing/nextclade/src/loculus_preprocessing/embl.py
```diff
@@ -181,21 +191,51 @@ def _build_gene_feature(gene: dict[str, Any]) -> SeqFeature:
...
+    # The annotation carries a strand only on a CDS's segments, so a gene takes its CDSes'.
+    cdses = gene.get("cdses", [])
+    strand = None
+    if cdses and cdses[0].get("segments"):
+        strand = -1 if cdses[0]["segments"][0].get("strand") == "-" else 1
     # In FeatureLocation start and end are zero based, exclusive end.
     # thus an embl entry of 123..150 (one based counting) becomes a location of [122:150]
     return SeqFeature(
-        FeatureLocation(start=gene_range["begin"], end=gene_range["end"]),
+        FeatureLocation(start=gene_range["begin"], end=gene_range["end"], strand=strand),
```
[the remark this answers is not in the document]
corneliusroemer-agent: Checked this against real INSDC records and it is right, so fixed in bd215c024: the gene now takes its CDSes' truncation the same way it already took their strand, and the golden file in `annotations.dependent.spec.ts` becomes `gene 1..>910` beside `CDS 1..>910`.

Verification is [QB011581.1](https://www.ncbi.nlm.nih.gov/nuccore/QB011581.1), a SARS-CoV-2 genome with a gap in it, which carries the marker on both features at both ends:

```
     gene            27103..>27250
                     /gene="ORF6"
     CDS             27103..>27250
...
     gene            <27762..27788
                     /gene="ORF7b"
     CDS             <27762..27788
```

The two smaller points stand and are deliberately left alone: a gene with no CDSes still gets neither strand nor markers, because the annotation offers no other place to read them from, and `cdses[0]` is still authoritative for both. Both would need a rule for what a gene with CDSes on opposite strands even means, which Nextclade does not produce here.

preprocessing/nextclade/src/loculus_preprocessing/embl.py
```diff
@@ -181,21 +191,51 @@ def _build_gene_feature(gene: dict[str, Any]) -> SeqFeature:
...
 
+def _segment_truncation(segment: dict[str, Any]) -> tuple[int, int]:
+    """Nucleotides of this segment (5', 3') that the sequence does not show."""
+    truncation = segment.get("truncation")
+    if not isinstance(truncation, dict):
+        return 0, 0
+    if "both" in truncation:
+        five_prime, three_prime = truncation["both"]
+        return int(five_prime), int(three_prime)
+    return int(truncation.get("fivePrime", 0)), int(truncation.get("threePrime", 0))
```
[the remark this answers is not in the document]
corneliusroemer-agent: Fixed in #7342, which replaces this function with a pydantic model of the annotation. The three arms are now parsed explicitly and anything else raises with the offending value named, so a surprise costs a clear validation error rather than the entry's whole annotation file.

The key spelling was the real worry and it is now pinned to a source: the model is written against `packages/nextclade-schemas/output-json.schema.json` in nextclade 3.23.0, where `Truncation` is the Rust enum serialised as `"none"`, `{"fivePrime": n}`, `{"threePrime": n}` or `{"both": [n, m]}`. The single-ended spellings were inferred here and the schema confirms them.

preprocessing/nextclade/src/loculus_preprocessing/embl.py
```diff
@@ -181,21 +191,51 @@ def _build_gene_feature(gene: dict[str, Any]) -> SeqFeature:
...
-    locations = [
-        FeatureLocation(start=segment["range"]["begin"], end=segment["range"]["end"], strand=strand)
-        for segment, strand in zip(segments, strands, strict=False)
-    ]
+    # Only the outer segments can run past the sequence; a splice junction is a known boundary.
+    five_truncated = _segment_truncation(segments[0])[0] > 0
+    three_truncated = _segment_truncation(segments[-1])[1] > 0
+    last = len(segments) - 1
+    locations = []
+    for index, (segment, strand) in enumerate(zip(segments, strands, strict=False)):
```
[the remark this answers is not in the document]
corneliusroemer-agent: Both fixed in #7342. `NextcladeCds.segments` is declared `Field(min_length=1)`, so an empty list is rejected at parse time with a named error and the indexing here cannot be reached with one; the caller's `if segments` guard is gone with it. The dead `zip` is gone too — the strand is computed inside the loop now, so neither the `strands` list nor the zip survives.

preprocessing/nextclade/tests/test_nextclade_preprocessing.py
```diff
@@ -1495,7 +1495,7 @@ def test_get_seq_features_handles_real_reverse_complemented_cchf_annotation():
     cds_features = [feature for feature in features if feature.type == "CDS"]
     assert len(cds_features) == 1
     cds_feature = cds_features[0]
-    assert str(cds_feature.location) == "[0:466](-)"
+    assert str(cds_feature.location) == "[<0:>466](-)"
```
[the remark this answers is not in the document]
corneliusroemer-agent: Agreed that the symmetric `both` fixture could not catch an inverted swap. #7342 adds `test_get_seq_features_marks_the_truncated_end_by_its_coordinate_not_its_strand`, which runs a 5'-only truncation on each strand and asserts `[<0:9](+)` against `[0:>9](-)` — it fails if the mapping is inverted, and it is also what pins the `fivePrime` spelling. The gene case is covered by two more: one for the strand, one for the markers.

The minus-strand multi-segment CDS is not covered, so the `complement(join(...))` part order is still only exercised by the single-segment CCHF record.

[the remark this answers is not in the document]
corneliusroemer-agent: Correcting my reply above on one point: the gene marker test I pointed at ran on the plus strand only, so it passed whether the minus-strand swap was present, absent or inverted — the same weakness this comment identified in the `both` fixture. bf7922e5e runs it on both strands and asserts the gene and CDS locations are identical; deleting the swap or inverting it now fails it, which I checked by making both mutations.

integration-tests/tests/specs/features/annotations.dependent.spec.ts
```diff
@@ -56,10 +56,10 @@ FT                   /mol_type="genomic RNA"
 FT                   /organism="Sudan ebolavirus"
 FT                   /geo_loc_name="France"
 FT                   /collection_date="2021-05-12"
-FT   gene            1..910
+FT   gene            1..>910
```
anna-parker, line 59: are you sure this `>` is correct? I dont recall ever seeing it in an embl flatfile, see e.g. https://www.ebi.ac.uk/ena/browser/api/embl/CY114383.1?lineLimit=1000

anna-parker: I only see `>` and `<` in the complement range: https://www.ebi.ac.uk/ena/browser/api/embl/AY049081.1?lineLimit=1000

preprocessing/nextclade/src/loculus_preprocessing/datatypes.py
```diff
@@ -265,15 +265,6 @@ class MoleculeType(StrEnum):
     GENOMIC_RNA = "genomic RNA"
     VIRAL_CRNA = "viral cRNA"
 
-    @property
```
anna-parker, line 268: ah yeah it does look like this was wrong, not sure why I ever did that - good catch!

integration-tests/tests/specs/features/annotations.dependent.spec.ts
anna-parker: and here I dont see it at all: https://www.ebi.ac.uk/ena/browser/api/embl/PZ884253.1?lineLimit=1000 - I would opt to not add `>`/`<` personally

preprocessing/nextclade/src/loculus_preprocessing/embl.py
```diff
@@ -1,9 +1,18 @@
+"""Build EMBL features from a Nextclade annotation."""
+
 import logging
 from dataclasses import dataclass
 from typing import Any
 
 from Bio.Seq import Seq
-from Bio.SeqFeature import CompoundLocation, FeatureLocation, Reference, SeqFeature
+from Bio.SeqFeature import (
+    AfterPosition,
```
anna-parker, line 9: this means: `AfterPosition - Specify the position as being found after some base` (https://biopython.org/docs/dev/api/Bio.SeqFeature.html) - I thought nextclade gives exact positions so this doesnt seem to be needed?

anna-parker: ah sorry now looking at https://www.insdc.org/submitting-standards/feature-table/ and yes this is required if the sequence is truncated
```
<1..888                   The feature starts before the first sequenced base and 
                          continues to and includes base 888

1..>888                   The feature starts at the first sequenced base and 
                          continues beyond base 888
```

integration-tests/tests/specs/features/annotations.dependent.spec.ts
anna-parker: sorry I realize this is required if the sequence is truncated (which I assume it is for this example)

anna-parker: Im a bit confused... it seems like the sequence goes to position 910... am I making a 0/1-indexing error?

anna-parker: ah sorry its because the gene would in theory continue, so if its truncated you write`>POS` where POS is the last position before truncation

preprocessing/nextclade/src/loculus_preprocessing/embl.py
```diff
@@ -181,21 +190,58 @@ def _build_gene_feature(gene: dict[str, Any]) -> SeqFeature:
...
     return SeqFeature(
-        FeatureLocation(start=gene_range["begin"], end=gene_range["end"]),
+        FeatureLocation(start=start, end=end, strand=strand),
         type="gene",
         qualifiers=qualifiers,
     )
 
 
+def _segment_truncation(segment: dict[str, Any]) -> tuple[int, int]:
+    """Nucleotides of this segment (5', 3') that the sequence does not show."""
```
anna-parker: I wonder if we could improve the explanation here a bit, maybe
```suggestion
    """If a (5' to 3') segment is truncated (i.e. the gene would extend outside of the sequenced portion) we need to add `>` or `<` to the side where the truncation occured."""
```

anna-parker (approved):
thanks, sorry it took me a bit to understand this but looks great!

preprocessing/nextclade/src/loculus_preprocessing/datatypes.py
corneliusroemer: I think Biopython warned you wrongly about it - they use the wrong strings!

anna-parker: ah that would explain it!

corneliusroemer:
I've made an issue for the Biopython false positive warning, also a PR:
- https://github.com/biopython/biopython/issues/5319
- https://github.com/biopython/biopython/pull/5320

PR #7345, opened 2026-09-17
fix(website): parse backend problem details that omit type

### LINKED ISSUE #7103: Website does not parse and show backend errors correctly

corneliusroemer-agent:
Fixes #7103.

Properly render backend error messages. The declared problem detail schema required a field the backend didn't emit, resulting in access tokens spilling over looking horrible (as well as unsafe).

I've added an integration test checking rerevision doesn't work and the error message is correct and not a fallback. This is better than a unit test which would not catch the backend's emitted json drifting away. It's cheap because we're just tagging it onto an existing revision test. In fact, we had a website test but the input was mismatched with what the backend actually returns.

A follow-up branch handles the fallback itself, which leaked credentials: #7346 - it's a bigger refactor that's harder to review hence doing the simple one first.

### PR Checklist

https://claude.ai/code/session_01JwYjh8SH7BDoe5VaneW68j

🚀 Preview: https://fix-website-parse-backend.loculus.org


integration-tests/tests/specs/features/revise-sequence.spec.ts
```diff
@@ -129,6 +129,24 @@ groupTest.describe('Bulk sequence revision', () => {
 
         const overview = await reviewPage.getReviewPageOverview();
         expect(overview.total).toBeGreaterThanOrEqual(SEQUENCES_TO_REVISE);
+
+        // Keeps the problem-detail contract under test: if the shape the backend sends drifts, the
+        // website falls back to a generic wrapper and this fails.
```
corneliusroemer:
```suggestion
```

integration-tests/tests/specs/features/revise-sequence.spec.ts
```diff
@@ -129,6 +129,24 @@ groupTest.describe('Bulk sequence revision', () => {
...
+        await groupTest.step(
+            "a second revision is rejected in the backend's own words",
```
corneliusroemer:
```suggestion
            "reject second revision before approval, error message displayed correctly",
```

website/src/types/backend.ts
```diff
@@ -275,12 +275,15 @@ export const submitFiles = uploadFiles.merge(
     }),
 );
 
+// Not a full RFC 9457 model but the predicate for "can the website display this error": the RFC
```
anna-parker: Im not sure we need this comment

integration-tests/tests/specs/features/revise-sequence.spec.ts
```diff
@@ -130,6 +131,29 @@ groupTest.describe('Bulk sequence revision', () => {
...
+
+                await revisionPage.goto(TEST_ORGANISM, groupId);
+                await revisionPage.uploadMetadataFile('revision_metadata.tsv', revisionMetadata);
+                await revisionPage.uploadSequenceFile('revised_sequences.fasta', fastaContent);
+                await revisionPage.acceptTerms();
+                await revisionPage.clickSubmit();
+
+                await expect(
+                    page.getByText(
+                        /The submitted file content was invalid: Accession versions are not in one of the states/,
```
anna-parker, line 149: awesome!

anna-parker (approved):
I confirmed just to revise a sequence that has a pending revision fails with a nicely formatted error message now 

Image

website/src/types/backend.ts
```diff
@@ -275,12 +275,15 @@ export const submitFiles = uploadFiles.merge(
...
+// makes every member optional, so requiring `type` or `instance`, which are never shown, only
+// rejects valid responses. The three that are shown stay required so callers never render undefined.
```
corneliusroemer:
```suggestion
// Technically RFC 9457 also doesn't guarantee title/status/detail
// to be non-null, but in practice our backend emits those
// so this allows us to avoid some verbose null handling.
```

PR #7347, opened 2026-09-17
refactor(website): one mapping from an error to a problem detail

corneliusroemer-agent:
Stacked on #7346, which is stacked on #7345. Review those first, and this diff against `fix/website-error-message-leak`.

The same mapping from an error to a problem detail existed in three places, because the React hooks path never had one. `ZodiosWrapperClient.call()` returns `Result<T, ProblemDetail>`, but `backendClientHooks` is a bare `ZodiosHooks` with no error mapping, so nine component call sites received a raw `AxiosError` and each re-derived what the backend had said. `lapisClientHooks.useGetSequence` hand-rolled a third, partial copy — which is why `SequenceViewer` was dumping.

`asProblemDetail` is now the single conversion, and `formatErrorMessage` is `asProblemDetail(error).detail`, so there is one implementation rather than a fourth. The error a caller gets carries both halves — `status` to switch on, `title`/`detail` to show — so nothing needs the axios error:

- `DataUploadForm.handleError` switches on `problem.status` instead of `error.response.status`, and no longer takes an `action` parameter just to call `isErrorFromAlias`.
- The three `isErrorFromAlias` branches in `useSubmissionOperations` are deleted. They were dead: `formatErrorMessage` already returned exactly `error.response.data.detail` for those inputs.
- `serviceHooks.getSequenceHook` uses `problemDetailFromResponse` instead of its own `safeParse`, keeping the `data.error ?? data` unwrapping LAPIS needs.

`problemDetailFromResponse` exists because "the response carried a problem detail" is a real distinction, not something to infer from a status of 0: an unparseable body still reports `Received unexpected message from backend`. An existing test caught that when I first inferred it from status — behaviour is unchanged.

`ZodiosWrapperClient` deliberately keeps its own policy and is commented as such: it reports a body it cannot parse as coming from the service rather than surfacing it, and it owns the 401 `www-authenticate` branch and the `x-request-id` logging.

### PR Checklist

`CI=1 npm run test` in `website/`: 899 tests passing. `npm run check-types`, `eslint` and `prettier` clean. Tests cover the LAPIS `error` nesting, `status` alongside `detail`, and non-error inputs — `null` in particular, which crashed the first version of this and is now a regression test.

https://claude.ai/code/session_01JwYjh8SH7BDoe5VaneW68j


corneliusroemer-agent:
Folded into #7346. Keeping them apart meant reviewing `formatErrorMessage.ts` in an intermediate state that this PR then rewrote (33 lines, then 64 more changed on the same file), and the stack gave no parallelism since this could not merge before #7346 anyway.

PR #7352, opened 2026-09-18
fix(website): remove presigned URL from toast error

tombch:
In my opinion we shouldn't directly expose presigned URLs to users like this - replaced the presigned URL with 'S3'. This error shouldn't appear anyway unless something has gone wrong with S3.

### PR Checklist


website/src/utils/multipartUpload.ts
```diff
@@ -34,13 +34,11 @@ export async function uploadPart(presignedUrl: string, part: Blob): Promise<stri
         body: part,
     });
     if (!response.ok) {
-        throw new Error(
-            `Failed to upload part to ${presignedUrl}: response status ${response.status}: ${response.statusText}`,
-        );
+        throw new Error(`Failed to upload part to S3: response status ${response.status}: ${response.statusText}`);
```
anna-parker, line 37: I wonder if we could maybe add website logs that have the presigned url, so we can debug this?

tombch: Yeah I was wondering about that, this is a client side error though so can't add to the astro server-side website logs. We could add some console logging for the user but that's something we have not done before. I think with file ID and part number in the message like suggested above, and if it can be reproduced the presigned url is visible in network tab, then all of that should help with debugging?

PR #7353, opened 2026-09-18
fix(preprocessing): clarify invalid host errors and show them when editing

### LINKED ISSUE #7351: Confusing errors when supplying invalid host taxon
Image

This should simply say "Goose" is not a valid option, please refer to X for the list of valid options.

In addition unlike other fields the error does not appear on the edit page:

Image

theosanderson-agent:
Submitting an unknown host such as `Goose` currently produces a generic validation failure and, when the derived host taxon ID is required, an additional misleading missing-field error. Neither message appears beside the Host input when editing because the form matches annotations against processed output fields.

For taxonomy lookup responses indicating an unknown host (404), return a message naming the invalid value, explaining that an NCBI taxon ID or scientific name is expected, and pointing to NCBI Taxonomy. Skip additional required/conditionally-required checks when processing that field has already produced an error; genuinely missing required values still receive the existing requirement error. INSDC ingested invalid hosts continue to generate warnings.

The edit form now associates errors and warnings with their unprocessed input fields, so annotations for derived outputs such as `hostTaxonId` are displayed beside the submitted `host` input. Regression tests cover optional, required, and conditionally required hosts, a missing required host, and error/warning visibility in both the edit-page tooltip and expanded field.

Fixes #7351.

Validation:

- Full preprocessing suite: 152 passed, using Nextclade 3.23.0 and DIAMOND 2.2.6.
- Full website suite: 887 passed across 74 files.
- Website `npm run check-types`: passed (Astro and TypeScript).
- Website `npm run format`: passed.
- Preprocessing mypy: passed (17 source files).
- Ruff format check and auto-fixable lint diff for changed Python files: passed; strict lint adds no findings relative to the base.
- Regression tests reproduced the reported failures before the implementation changes.

🚀 Preview: https://fix-7351-host-taxon-error.loculus.org


theosanderson-agent (commented):
I tested this on the preview (Ebola Sudan, single-entry form) and read the diff. The host part works well. There's one regression on the edit form, plus a couple of smaller notes.

## Screenshots from the preview

**Review page:** host `Goose` now gets a single, actionable error, with no extra "is required" message:

review page with Goose host

**Edit page:** the error now shows next to the submitted **Host** input (it comes from the derived `hostTaxonId` output):

edit page host error

## Regression: duplicated errors on the edit form for fields with several derived outputs

Switching `mapErrorsAndWarnings` from `processedFields` to `unprocessedFields` means *every* output computed from an input now attaches its errors to that input's row. `sampleCollectionDate` feeds three outputs (`sampleCollectionDate`, `sampleCollectionDateRangeLower`, `sampleCollectionDateRangeUpper`), so an invalid date (`2023-13-45`) now shows three near-identical errors under *Collection date*. Going by the old matching code, main shows only the `sampleCollectionDate` one:

edit page with triplicated date error

This will happen for any input that's used by more than one processing spec.

One option that keeps the host fix: match on `processedFields` first, and fall back to `unprocessedFields` only when nothing matched for that row:

```ts
const annotationsFor = (annotations: ProcessingAnnotation[]) => {
    const byProcessed = annotations.filter((a) => a.processedFields.some((f) => f.name === key && f.type === type));
    const matching =
        byProcessed.length > 0
            ? byProcessed
            : annotations.filter((a) => a.unprocessedFields.some((f) => f.name === key && f.type === type));
    return matching.map((a) => a.message);
};
```

`host` has no processed field of its own, so it falls back to the `hostTaxonId` annotation. `sampleCollectionDate` matches its own processed field and keeps showing just one error. Deduplicating by message wouldn't help here, because each message contains the output field name. A test with one input feeding two outputs would lock this in.

## Other notes

- **Overlap with #7390:** the `prepro.py` change (skip requirement checks when `processing_result.errors` is non-empty) is the same change as #7390, which fixes #7389. Whichever merges second will conflict. The fixes are equivalent, so the second one can just take the other's version (the comments differ slightly). The tests are complementary: #7390 covers a generic invalid required date, and this PR covers host with required/conditional/missing.
- **Nit:** the message ends `…NCBI Taxonomy: https://www.ncbi.nlm.nih.gov/taxonomy.`. That trailing period will be included when someone copies the URL, and when it's auto-linked, e.g. in emails or the CLI. Consider dropping it, or wrapping the URL: `(https://www.ncbi.nlm.nih.gov/taxonomy)`.
- Only 404s get the new wording; other status codes keep "Host validation for … failed.", which seems right since those are service problems rather than bad input.

PR #7356, opened 2026-09-18
test(raw-reads-processing): fail instead of skip when readtools or deacon not found, allow skipping deliberately

anna-parker:
Pytest tests are skipped require readtools and deacon respectively if they are missing.

It's better to fail by default and allow skipping deliberately.

alternative to https://github.com/loculus-project/loculus/pull/7320

### PR Checklist
I confirmed locally this is WAI.


raw-reads-processing/test/conftest.py
```diff
@@ -0,0 +1,8 @@
+def pytest_addoption(parser):
+    parser.addoption(
+        "--skip-missing-deps",
+        action="store_true",
+        default=False,
+        help="Skip tests that need external tools (deacon binary, readtools "
+        "jar) instead of failing them when those tools aren't available.",
+    )
```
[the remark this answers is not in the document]
anna-parker: I confirmed that claude is wrong and the commands do in fact work from the raw-reads-processing repo as documented in the readme

raw-reads-processing/test/conftest.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def missing_dependency(request, message: str) -> NoReturn:
+    if request.config.getoption("--skip-missing-deps"):
+        pytest.skip(message)
+    pytest.fail(message)
```
clonder: pytrace=False? then a missing tool prints one line instead of a traceback into conftest

clonder (approved):

raw-reads-processing/test/conftest.py
anna-parker: yes that is nicer! added!

PR #7366, opened 2026-09-21
feat(website): Show single file category as section heading

theosanderson:
image

### Motivation
- Avoid showing redundant headings and make the guidance text more specific when there is only a single extra-file category.
- Improve reliability of Playwright folder uploads by using `data-testid` for scrolling and file input targeting.

### Description
- Change `ExtraFilesUpload` to display the single file category `displayName` as the main section heading and to interpolate the category name into the bulk-upload guidance and documentation link text when only one category is present by computing `singleFileCategoryDisplayName`.
- Add a `showCategoryHeading` prop to `FolderUploadComponent` (default `true`) and conditionally render the per-category `<h3>` so callers can hide the duplicate heading when the section already names the category.
- Update `FolderUploadComponent` tests to assert the new behavior and add `DataUploadForm.spec.tsx` to cover single vs multiple category heading and guidance text scenarios.
- Update the Playwright file-upload helper `uploadFilesFromTmpDir` to use `getByTestId(fileCategory)` for scrolling and calling `setInputFiles` instead of relying on a static heading.

### Testing
- Ran the component unit tests under `website` with `vitest`, including the new `DataUploadForm.spec.tsx` and the updated `FolderUploadComponent.spec.tsx`, and they passed.
- Ran the integration/Playwright tests that exercise folder uploads after updating the helper to use `data-testid`, and they passed.
- Confirmed mocked multipart upload helpers used in tests behaved as before with the new prop and conditional rendering, with no regressions observed in the test suite.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6ab0ea2e538c8325aa0b5b954393a18b)

🚀 Preview: https://codex-display-file-catego.loculus.org


website/src/components/Submission/DataUploadForm.tsx
```diff
@@ -428,14 +431,14 @@ export const ExtraFilesUpload = ({
     return (
         <div className='grid sm:grid-cols-3 gap-x-16 gap-y-4'>
             <div>
-                <h2 className='font-medium text-lg'>Extra files</h2>
+                <h2 className='font-medium text-lg'>{singleFileCategoryDisplayName ?? 'Extra files'}</h2>
                 <p className='text-gray-500 text-sm'>
                     {inputMode === 'bulk'
-                        ? 'Upload a folder of files or individual files for your sequences. Each file must be referenced by its name in the corresponding file category column of your metadata.'
+                        ? `Upload a folder of files or individual files for your sequences. Each file must be referenced by its name in the ${singleFileCategory?.name ?? 'corresponding file category'} column of your metadata.`
```
anna-parker: I wonder if we should add the `files.` prefix to the column mentioned here

website/src/components/Submission/DataUploadForm.tsx
```diff
@@ -428,14 +431,14 @@ export const ExtraFilesUpload = ({
...
                         : 'Upload a folder of files or individual files for this sequence.'}{' '}
                     For more information please refer to the{' '}
                     <a href={extraFilesUploadDocsUrl} target='_blank' className='text-primary-600 hover:underline'>
-                        extra files documentation
+                        {singleFileCategoryDisplayName?.toLowerCase() ?? 'extra files'} documentation
```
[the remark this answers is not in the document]
theosanderson: yes this is a hack but we can change this later when we need to

website/src/components/Submission/DataUploadForm.tsx
theosanderson: yes! if it doesn't show up we defs should - sorry still wiating for the preview but I agree it won't

website/src/components/Submission/DataUploadForm.spec.tsx
```diff
@@ -0,0 +1,63 @@
...
+    it('uses a single file category as the section heading without repeating it', () => {
+        render(
+            <ExtraFilesUpload
+                {...defaultProps}
+                fileCategories={[{ name: 'images', displayName: 'Microscopy images' }]}
+            />,
+        );
+
+        expect(screen.getByRole('heading', { level: 2, name: 'Microscopy images' })).toBeInTheDocument();
+        expect(screen.queryByRole('heading', { level: 3, name: 'Microscopy images' })).not.toBeInTheDocument();
```
anna-parker, line 27: Im not sure we need this
```suggestion
```

anna-parker: ah ok, I see its to confirm there aren't duplicate headers - then I guess we can keep!

anna-parker (approved):
lets add the `files.` prefix - otherwise looks great!

Issue #6795, opened 2026-06-29
Null byte in submission id causes 504 at database insertion - not caught by backend before


corneliusroemer:
I've been testing edge cases for submission id and string fields.

A null byte in the submission id is only rejected by the database causing a 500 error, should be caught earlier by backend as part of validation.

We might want to restrict submission id to a subset of ascii, also excluding control sequences etc. Right now one can even put emojis and right to left text in there.

Submission id longer than 255 characters gets internal server error rather than HTTP 422 - should catch in validation:

```
{
   "type":"about:blank",
   "title":"Internal Server Error",
   "status":500,
   "detail":"Value can't be stored to database column because exceeds length (500 > 255)",
   "instance":"/ebola-sudan/submit"
}
```

Similar for null byte:

```
{
   "type":"about:blank",
   "title":"Internal Server Error",
   "status":500,
   "detail":"java.sql.BatchUpdateException: Batch entry 0 INSERT INTO metadata_upload_aux_table (submitter, group_id, uploaded_at, submission_id, fasta_ids, metadata, files, organism, upload_id) VALUES (('test_user_da440fce'), ('108'::int4), ('2026-06-29 18:50:05.695364+00'), ('null\u0000'), ('{\"null\u0000'), ('{\"geoLocCountry\":\"France\",\"sampleCollectionDate\":\"2026-06-29\",\"authorAffiliations\":\"Edge Case Institute\",\"authors\":\"EdgeCase, Test\",\"geoLocAdmin1\":\"null byte in submissionId\"}'), (NULL), ('ebola-sudan'), ('ec8e496f-0a27-453b-ad42-2a0aaf3d3c68')) was aborted: ERROR: invalid byte sequence for encoding \"UTF8\": 0x00\n  Where: unnamed portal parameter $4  Call getNextException to see other errors in the batch.",
   "instance":"/ebola-sudan/submit"
}
```

Meanwhile we accept very weird strings as submission ids (and also in string fields):

- Bidi controls — RTL override U+202E
- Invisible chars — zero-width space U+200B, BOM / ZWNBSP U+FEFF
- Raw invalid UTF-8 bytes (0xFF 0xFE 0x80 0xC0)

We should probably prohibit all of them by validating/parsing in backend. For submissionId we already prohibit whitespace but that's not enough.

Questionable for submission id but might be ok for general strings: non-ASCII, e.g. Chinese/Arabic characters

PR #7369, opened 2026-09-21
fix(raw-reads-validation): sanitize null bytes out of raw reads validation errors

### LINKED ISSUE #7358: A null byte anywhere in a raw reads validation error wedges the whole preprocessing batch forever
If the plaintext inside a .fastq.gz contains null bytes it causes the backend to never accept the batch that prepro submits (postgres doesn't accept null byte in jsonb) - so up to 100 sequences never get processed, stuck forever until an admin intervenes with db surgery.

This is 2-3 levels of failure (could have been caught at each level, Swiss cheese style):
- Raw reads processing should not return null bytes verbatim in its responses
- Prepro should not pass through null bytes verbatim
- Backend should notice null bytes and not hand them to postgres (though at that point it's already too late, but the error message in logs is better)

If a `.fq.gz` decompresses to something whose first offending line contains a null byte, readtools quotes that line back verbatim in its error, we serialize it into the result json as `\u0000`, and Postgres refuses it:

```
org.postgresql.util.PSQLException: ERROR: unsupported Unicode escape sequence
POST /west-nile/submit-processed-data - Responding with status 500
```

`SubmissionDatabaseService` is `@Transactional` at class level, so the whole POST rolls back and every entry in it loses its results. With `batch_size: 100` that is up to 100 entries, from any group submitting that organism, stuck in `IN_PROCESSING` forever. They re-fail every retry cycle, `get-data-to-edit` 500s on them, and they can't even be deleted — `delete-sequence-entry-versions` 422s on `IN_PROCESSING`. So it needs db surgery.

Minimal trigger:

```sh
printf 'notes\x00more\n' | gzip > reads.fq.gz    # submit as files.rawReads
```

A genuine FASTQ that picked up a stray trailing null or a run of padding does it too, because readtools then quotes *that* line — an interrupted upload or a bad concatenation would be enough, and the submitter sees nothing wrong with their file. A BAM renamed to `.fastq.gz` is the other easy way in, since BGZF is already gzip and sails through our double-gzip check.

In general, we should validate for null bytes at all places we touch postgres.

Here's a little (a bit bad but better than none) video that GLM-5.3-Flash made for me with Playwright:

https://github.com/user-attachments/assets/abc10ef5-a7ab-4b61-bbd3-f4f4dc99aea1

corneliusroemer-agent:
Fixes #7358.

## Problem this PR fixes

A null byte anywhere in a submitted `.fastq.gz` currently wedges up to a whole batch of sequence entries in `IN_PROCESSING` until someone does database surgery.

<details>
readtools quotes the offending line of the file back verbatim in its error, that error becomes the message of an annotation, and Postgres refuses the resulting `jsonb`. Because `SubmissionDatabaseService` is `@Transactional` at class level, the whole `submit-processed-data` POST rolls back, so every entry in the batch loses its results — from any group that happened to submit that organism. They re-fail every retry cycle, `get-data-to-edit` returns 500 for them, and they cannot even be deleted, since `delete-sequence-entry-versions` rejects `IN_PROCESSING`.
</details>

JSONB characters that we need to sanitize per [Postgres docs](https://www.postgresql.org/docs/16/datatype-json.html): `jsonb` "rejects `U+0000, Unicode surrogate pairs must be correct.

## Implementation

JSON sanitization is done on the `Annotation` model itself to catch all possible sources of problematic bytes. Also on `ProcessingFailure` as that's a separate object that gets serialized for the HTTP response.

A null byte is replaced with a visible `<NUL>` to be still visible to submitters.

## Tests

New tests cover the sanitization function. I've also added a fixture to check we don't have any problematic bytes on all test output, this is cheap and is extra insurance in case we refactor and make test false negative somehow.

Also add 2 integration test cases to our parametrized raw reads error test to verify this all goes through end to end and acts as a good regression prevention (null in read name and renamed bam).

## Not addressed

This is the proximal fix only, in the raw reads processing service. Preprocessing and the backend should each reject null bytes on their own account; the backend one would also cover #6795.

In backend, class-level `@Transactional` means any per-entry failure rolls back the whole batch; null bytes are one trigger of that, and this PR closes only that trigger. Some way to unstick entries already wedged in `IN_PROCESSING` might be worth its own issue. Potentially also not refusing whole batch when single sequence has issues - but that's a judgment call.

🚀 Preview: https://fix-raw-reads-null-byte-s.loculus.org


corneliusroemer:
@anna-parker PR should be ready, I'll do tests on previews with this PR and deacon once merged.

Might still add an integration/e2e test separately - those are cheap now with parametrization.

raw-reads-processing/test/test_file_validation.py
```diff
@@ -123,32 +122,6 @@
 FIXTURES_DIR = Path(__file__).parent / "fixtures"
 
 
-def _find_jar() -> str | None:
```
corneliusroemer, line 126: Moved to conftest so it can be shared

corneliusroemer:
I've adopted @anna-parker's control character replacement so that logs in terminals don't end up mangled.

integration-tests/tests/specs/features/file-sharing.spec.ts
```diff
@@ -128,6 +128,16 @@ const RAW_READS_FAILURES: { id: string; files: Record<string, string | Buffer>;
             files: { 'reads.fastq.gz': contaminatedReads() },
             error: /high proportion of human reads/i,
         },
+        {
+            id: 'bam-renamed-as-fastq',
```
anna-parker, line 132: great test!

corneliusroemer: Yeah I really like the parametrization here! In a followup-PR we should have a parallel test that checks valid edge cases don't cause errors. Needs a separate test but could be equally parametrized.

raw-reads-processing/src/raw_reads_processing/datatypes.py
```diff
@@ -20,9 +21,37 @@ class RequestWithFiles(BaseModel):
...
+
+_SANITIZE_TABLE = {0x00: NULL_BYTE_PLACEHOLDER} | dict.fromkeys(
+    _CONTROL_CHARS, UNDECODABLE_PLACEHOLDER
+)
+
+
+def sanitize_for_json(text: str) -> str:
+    """Replace what Postgres rejects inside jsonb (U+0000, lone surrogates) and
+    control characters, which reach a terminal through the logs. Tab, newline
+    and carriage return are kept."""
```
corneliusroemer:
```suggestion
    control characters"""
```

raw-reads-processing/src/raw_reads_processing/deacon.py
```diff
@@ -128,7 +128,8 @@ def run_deacon_filter(
             args,
             check=True,
             capture_output=True,
-            text=True,
+            encoding="utf-8",
+            errors="replace",
```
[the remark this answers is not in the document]
corneliusroemer: I think we've fixed this now - we catch unidecode errors?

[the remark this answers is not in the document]
corneliusroemer: Ah this is now fixed on #7370

raw-reads-processing/AGENTS.md
```diff
@@ -0,0 +1,13 @@
...
+```sh
+cd raw-reads-processing && ruff check . && ruff format --diff .
+```
+
+`pyproject.toml` here replaces the root `ruff.toml` rather than extending it, so a different
+set of rules applies, and the two disagree in both directions. CI (`raw-reads-processing-unit-tests.yaml`)
+runs from this directory with the ruff version pinned in `environment.yml`, so a newer
+ruff will report findings CI does not.
+
+The readtools tests are skipped unless `READTOOLS_JAR` points at a downloaded jar (see
```
anna-parker: I guess this will change with https://github.com/loculus-project/loculus/pull/7320 so Im not sure we should add it

raw-reads-processing/AGENTS.md
```diff
@@ -0,0 +1,13 @@
...
+README); the deacon tests are skipped unless `deacon` is on PATH.
```
corneliusroemer:
```suggestion
```

raw-reads-processing/test/test_file_validation.py
```diff
@@ -539,3 +513,37 @@ def test_too_many_bam_files_are_rejected():
...
+
+    def fake_run(args, **kwargs):
+        return real_run([sys.executable, "-c", script], **kwargs)
+
+    monkeypatch.setattr(file_format_validation.subprocess, "run", fake_run)
+    with pytest.raises(InvalidSubmission) as exc_info:
+        validate_with_readtools({"reads.fastq": Path(reads)}, FileFormat.FASTQ)
+
+    assert "bad byte \ufffd in header" in exc_info.value.error.message
+    assert "\\ud" not in exc_info.value.error.model_dump_json()
```
anna-parker, line 549: shouldnt we assert \xff is not in the exec info? maybe Im confused

anna-parker (approved):
lots of tests - very nice!

raw-reads-processing/test/test_file_validation.py
anna-parker: or what does this final assertion check?

corneliusroemer: It checks that we've escaped it in the error message and that there's nothing unescaped/raw in the json

corneliusroemer:
> lots of tests - very nice!

I hope the tests are also useful not just lines :D

raw-reads-processing/test/test_file_validation.py
corneliusroemer: Ah the last one checks we're not escaping it differently from how we intend, not really important as is - could also drop but let's just leave for now.

anna-parker: I feel like this should be covered by asserting on the exact error message, can we drop it in https://github.com/loculus-project/loculus/pull/7370 maybe?

Issue #7110, opened 2026-08-19
raw-reads-processing: send slack notifications if deacon server panics/crashes and the pod needs to be restarted


anna-parker:
Identified during https://github.com/loculus-project/loculus/pull/6916 - it can happen that deacon panics and we need to restart the pod, we should be notified of these failures.

Additionally we might want to add additional server logs

PR #7370, opened 2026-09-21
fix(raw-reads-validation): upgrade deacon to 0.17.1 to fix server crash and CRLF handling, also catch FASTQ reader errors

### LINKED ISSUE #7359: A FASTQ with CRLF line endings crashes the deacon server, every entry with files processed in the ~30s window errors with "Internal Error"
`readtools` accepts CRLF (Windows line endings) FASTQ happily, then deacon aborts on it — the `\r` ends up joined to the sequence, so sequence and quality lengths differ by one ("Sequence length (151) and quality length (150) do not match" locally):

```
Internal Error. Raw reads processing service failed: Deacon filter failed with exit code -6.
```

The CRLF entry itself records its own error normally and isn't retried, so at least we don't crash-loop.

Until deacon has a fix, we can catch CRLF in our own validation (or in readtools) before deacon ever sees the file, and stop one bad input taking the shared server down with it. A reprocess endpoint would clear the stale verdicts here and would help with the null-byte batch-wedge issue too.

Upstream issue: https://github.com/bede/deacon/issues/101

corneliusroemer:
resolves #7359, helps with #7110 by making crashes rarer

Upgrade to deacon v0.17.1 to fix 2 bugs that affected us (bede/deacon#101 and bede/deacon#100)

- CRLF results in miscounted bases or outright crash/error 
- server mode crashes after a single malformed request: now server stays up, no collateral damage

Adds regression tests for both.

Also fixes another bug noticed while working on this: our own `median_read_length` function would fail on invalid unicode in read names with InternalError blaming ourselves instead of providing a helpful error message to submitter.

Adds integration tests to catch regressions.


raw-reads-processing/src/raw_reads_processing/deacon.py
```diff
@@ -69,13 +70,36 @@ def median_read_length(
...
+            lengths = [
+                len(seq)
+                for _, seq, _ in itertools.islice(FastqGeneralIterator(fh), sample_size)
+            ]
+    # A damaged file rather than a malformed record; subclasses ValueError.
+    except UnicodeDecodeError as error:
+        # The position is an offset into the decoder's chunk, not into the file.
+        message = (
+            f"File '{file_name}' contains invalid Unicode "
+            f"(byte 0x{error.object[error.start]:02x}: {error.reason}). "
```
anna-parker, line 88: in the follow up PR Im not sure if this will get encoded as `UNDECODABLE_PLACEHOLDER`

anna-parker: ah it wont, it is now safe ascii which is good

raw-reads-processing/src/raw_reads_processing/deacon.py
```diff
@@ -69,13 +70,36 @@ def median_read_length(
...
+        raise InvalidSubmission(
+            Annotation(fileNames=[file_name], message=message)
+        ) from error
+    except ValueError as error:
+        # Otherwise the submitter gets an "Internal error" that blames us for their file.
+        message = f"Failed to parse file '{file_name}': {error} {FALSE_POSITIVE_HINT}"
+        logger.error(message)
+        raise InvalidSubmission(
+            Annotation(fileNames=[file_name], message=message)
+        ) from error
```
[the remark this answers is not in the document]
anna-parker: I guess this is ok as we should have validated the gzip before calling this function

raw-reads-processing/src/raw_reads_processing/deacon.py
```diff
@@ -12,6 +12,7 @@
 from raw_reads_processing.config import Config
 from raw_reads_processing.datatypes import Annotation, DeaconSummary, FileName
 from raw_reads_processing.errors import InvalidSubmission, ProcessingFailure
+from raw_reads_processing.file_format_validation import FALSE_POSITIVE_HINT
```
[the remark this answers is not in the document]
anna-parker: yeah this seems like sth to move to errors

anna-parker (dismissed):
claude has some good suggestions! already approving :-)

raw-reads-processing/src/raw_reads_processing/deacon.py
```diff
@@ -69,13 +70,36 @@ def median_read_length(
...
+            f"(byte 0x{error.object[error.start]:02x}: {error.reason}). "
+            f"{FALSE_POSITIVE_HINT}"
+        )
+        logger.error(message)
+        raise InvalidSubmission(
+            Annotation(fileNames=[file_name], message=message)
+        ) from error
+    except ValueError as error:
+        # Otherwise the submitter gets an "Internal error" that blames us for their file.
+        message = f"Failed to parse file '{file_name}': {error} {FALSE_POSITIVE_HINT}"
```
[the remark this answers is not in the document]
corneliusroemer: Let's do separately #7372

raw-reads-processing/src/raw_reads_processing/deacon.py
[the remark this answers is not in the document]
corneliusroemer: Yeah, but I've added an extra branch defensively, just a few lines - could also remove again

corneliusroemer:
The new integration tests have teeth - they fail on deacon 0.17.0! See failures in https://github.com/loculus-project/loculus/pull/7371

```
 1) [firefox-without-dep] › tests/specs/features/file-sharing.spec.ts:140:5 › reject invalid raw_reads submissions with helpful errors 

Error: expect(locator).toContainText(expected) failed
```

Logs show what happened:

```
1. raw-reads-processing returned 200 for each file — it did its job and produced annotations.
2. readtools quoted the offending line back verbatim, null byte included. From the preprocessing pod:
Sequence header must start with @: notes\u0000...
3. Backend rejected the insert:
PSQLException: ERROR: unsupported Unicode escape sequence
  Detail: \u0000 cannot be converted to text.
POST /ebola-sudan/submit-processed-data - Responding with status 500
4. Preprocessing raised RuntimeError and retried — two 500s, ten minutes apart.
5. Playwright saw 0 of 9 sequences processed / 9 awaiting processing for the full 300 s (577 polls).
```

raw-reads-processing/src/raw_reads_processing/deacon.py
```diff
@@ -69,14 +74,46 @@ def median_read_length(
     Read length should be homogeneous within a sequencing run, so a small sample from
     the start of the file is representative and costs only a few milliseconds.
     """
-    with xopen(path, "rt", threads=0) as fh:
-        lengths = [
-            len(seq)
-            for _, seq, _ in itertools.islice(FastqGeneralIterator(fh), sample_size)
-        ]
+    try:
+        with xopen(path, "rt", threads=0) as fh:
```
anna-parker, line 78: IMO we can add:
```suggestion
        with xopen(path, "rt", threads=0, errors="replace") as fh:
```

raw-reads-processing/test/test_deacon.py
```diff
@@ -176,3 +177,92 @@ def test_host_reads_at_or_below_threshold_passes(tmp_path):
...
+    reads.write_bytes(b"@read\xff0\n" + b"A" * 150 + b"\n+\n" + b"I" * 150 + b"\n")
+
+    with pytest.raises(InvalidSubmission) as exc_info:
+        deacon_module.median_read_length(reads, "reads.fastq")
+    message = exc_info.value.error.message
+    assert exc_info.value.error.fileNames == ["reads.fastq"]
+    assert "invalid Unicode" in message
+    assert "0xff" in message
+    assert FALSE_POSITIVE_HINT in message
+    # Named, never echoed, so the message cannot carry submitter bytes.
```
corneliusroemer:
```suggestion
```

raw-reads-processing/src/raw_reads_processing/deacon.py
corneliusroemer: This would mean readnames with garbage unicode would pass through - I think it's ok to error on mangled read names, unlikely that these are supposed to be invalid.

corneliusroemer: We could add errors replace, but then we don't need the unicode decode error exception catch.

corneliusroemer:
Thanks so much @bede for fixing both bugs and releasing a new version! Great that CRLF is now fully supported (even if rare, you never know!)

raw-reads-processing/src/raw_reads_processing/deacon.py
```diff
@@ -69,13 +70,36 @@ def median_read_length(
...
+        raise InvalidSubmission(
+            Annotation(fileNames=[file_name], message=message)
+        ) from error
     if not lengths:
-        message = f"Failed to determine median read length for file '{file_name}'. File may be empty or corrupted."
+        message = (
+            f"Failed to determine median read length for file '{file_name}'. "
+            f"File may be empty or corrupted. {FALSE_POSITIVE_HINT}"
+        )
         logging.error(message)
```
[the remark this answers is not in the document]
corneliusroemer: done

raw-reads-processing/src/raw_reads_processing/deacon.py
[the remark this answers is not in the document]
corneliusroemer: done

raw-reads-processing/src/raw_reads_processing/deacon.py
anna-parker: ah true this is nicer as is

anna-parker (approved):
looks good thanks!

corneliusroemer:
Chromium E2E passed so merging

PR #7379, opened 2026-09-21
feat(prepro): handle exceptions thrown in enrich_with_nextclade gracefully

### LINKED ISSUE #7378: Prepro crashes if enrich_with_nextclade throws an exception, leading to sequences crash looping
```
DEBUG:charset_normalizer:Encoding detection: ascii is most likely the one.
DEBUG:loculus_preprocessing.prepro:Processing 1 unprocessed sequences
ERROR:loculus_preprocessing.prepro:Processing failed. Traceback : . Unprocessed data: [UnprocessedEntry(accessionVersion=...., data=UnprocessedData(..., unalignedNucleotideSequences={}, files=None))]
Traceback (most recent call last):
  File "/opt/conda/lib/python3.14/site-packages/loculus_preprocessing/prepro.py", line 865, in run
    processed = process_all(unprocessed, dataset_dir, config)
  File "/opt/conda/lib/python3.14/site-packages/loculus_preprocessing/prepro.py", line 773, in process_all
    nextclade_results = enrich_with_nextclade(unprocessed, dataset_dir, config)
  File "/opt/conda/lib/python3.14/site-packages/loculus_preprocessing/nextclade.py", line 810, in enrich_with_nextclade
    batch = assign_segment_for_alignment(unprocessed, config=config, dataset_dir=dataset_dir)
  File "/opt/conda/lib/python3.14/site-packages/loculus_preprocessing/nextclade.py", line 754, in assign_segment_for_alignment
    batch = assign_all_single_segments(unprocessed, config=config)
  File "/opt/conda/lib/python3.14/site-packages/loculus_preprocessing/nextclade.py", line 626, in assign_all_single_segments
    sequence_assignment = assign_single_segment(
        entry.data.unalignedNucleotideSequences,
        config=config,
    )
  File "/opt/conda/lib/python3.14/site-packages/loculus_preprocessing/nextclade.py", line 615, in assign_single_segment
    unalignedNucleotideSequences={"main": next(iter(input_unaligned_sequences.values()))},
                                          ~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
StopIteration
DEBUG:loculus_preprocessing.prepro:Fetching unprocessed sequences
```
This leads all sequences in the batch to fail, prepro doesnt return any errors so the backend will keep resending in a loop.

anna-parker:
resolves https://github.com/loculus-project/loculus/issues/7378

Currently if `enrich_with_nextclade` throws an exception prepro will crash without sending the backend a response, leading the backend to resend the data that caused the crash again and again and never show the submitters what is wrong. This will now lead prepro to return errors for the whole batch, so at least users know what is wrong.

code by me, tests by claude

### PR Checklist


preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -770,7 +770,13 @@ def process_all(
     processed_results = []
     logger.debug(f"Processing {len(unprocessed)} unprocessed sequences")
     if config.alignment_requirement != AlignmentRequirement.NONE:
-        nextclade_results = enrich_with_nextclade(unprocessed, dataset_dir, config)
+        try:
+            nextclade_results = enrich_with_nextclade(unprocessed, dataset_dir, config)
+        except Exception as e:
+            logger.error(f"Error occurred while enriching with nextclade: {e}")
+            for entry in unprocessed:
+                processed_results.append(processed_entry_with_errors(entry.accessionVersion))
```
theosanderson, line 778: Am I understanding correctly that this applies to a whole batch although the error may relate to only one sequence (and the batch may span multiple submitters).

If so I think in this case we should display a clear message saying something like:

"This sequence failed due to a technical error. This problem may be entirely unrelated to your specific sequence. Please contact an administrator if this persists."

(As an aside from just reading the code, IMO `processed_entry_with_generic_error` would be a better name for the function)

theosanderson: but actually this may be worse than the current state - in the current state e.g. when we clear out the failing seq other sequences will now pass - I think we should rethink this

theosanderson: Also potentially these errors might sometimes be stochastic?

theosanderson: To me the right approach to something like this probably involves counting how many times a seq has been processed and marking it failed after repeated failures (while acknowledging that this could still be unconnected to the seq in Q)

anna-parker: > but actually this may be worse than the current state - in the current state e.g. when we clear out the failing seq other sequences will now pass - I think we should rethink this

Its very likely we would be retrying all sequences in the same batch again and again. This is exactly the situation we have with ENA, you submit a sequence, you have no info what went wrong and cant retry the submission yourself. Here users can resubmit intentionally and are prompted to contact us. I think this is a much better state to be in. Again, we should in theory never hit this case unless an assumption we make doesnt hold in which case submitters should contact us. 

I do agree we should have a class of errors that we retry instead of returning an internal error to users, e.g. raw reads processing or host validation service unavailable - but I think we can do that separately and I actually wouldnt necessarily retry these types of unexpected errors.

anna-parker: > this probably involves counting how many times a seq has been processed and marking it failed after repeated failures

This would have to be done in the backend I guess, Im not against it, probably good to have generally

anna-parker: I would propose in a follow up PR adding a distinction between FAILED and RETRYABLE_FAILURE, and then the backend can retry submissions in RETRYABLE_FAILURE for a given amount of tries until it fully fails.

I dont think user input should ever cause the prepro pod to crash.

maverbiest: Could we add a fallback for enrich_with_nextclade that we only run when a batch errors? The fallback would iterate over the entries one at a time so it can pinpoint the problematic one(s) and attach a proper error.

Kind of like covid testing: run the whole batch first, if there is a hit, check individually

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
```diff
@@ -770,7 +770,13 @@ def process_all(
     processed_results = []
     logger.debug(f"Processing {len(unprocessed)} unprocessed sequences")
     if config.alignment_requirement != AlignmentRequirement.NONE:
-        nextclade_results = enrich_with_nextclade(unprocessed, dataset_dir, config)
+        try:
+            nextclade_results = enrich_with_nextclade(unprocessed, dataset_dir, config)
+        except Exception as e:
```
corneliusroemer, line 775: We might want to distinguish potentially recoverable errors from permanent ones. Maybe a dataset download failed, in that case we might want to retry a few times.

preprocessing/nextclade/src/loculus_preprocessing/prepro.py
corneliusroemer: > Am I understanding correctly that this applies to a whole batch although the error may relate to only one sequence (and the batch may span multiple submitters).

Yes, although that is also true of the status quo. The exception kills the whole pod and everything is stuck until it gets retried after 11min. Submitters can't even delete their sequences (except for a ~1min window which they don't know about). These entries can also contaminate others on retry (though unlikely).

> If so I think in this case we should display a clear message saying something like: "This sequence failed due to a technical error. This problem may be entirely unrelated to your specific sequence. Please contact an administrator if this persists."

Yes definitely, and send a Slack alert (as we already do from other processes as well). Although this PR shouldn't be blocked on Slack, I think as it's an improvement already.

> (As an aside from just reading the code, IMO processed_entry_with_generic_error would be a better name for the function)

Agreed

> but actually this may be worse than the current state - in the current state e.g. when we clear out the failing seq other sequences will now pass - I think we should rethink this

Not sure what you mean by clearing out - deleting? If we know which single sequence is the culprit (it could also be multiple or not a problem with any particular sequence at all) we can delete processed rows for the others and they will process correctly.

> Again, we should in theory never hit this case unless an assumption we make doesnt hold in which case submitters should contact us.

IIUC there might also be temporary/recoverable problems - though I'd have to look into what's in the function, maybe dataset download isn't part of it. But there could also be things like failure to write to disk etc.

> I do agree we should have a class of errors that we retry instead of returning an internal error to users, e.g. raw reads processing or host validation service unavailable - but I think we can do that separately and I actually wouldnt necessarily retry these types of unexpected errors.
> I would propose in a follow up PR adding a distinction between FAILED and RETRYABLE_FAILURE, and then the backend can retry submissions in RETRYABLE_FAILURE for a given amount of tries until it fully fails.

Or `INTERNAL_ERROR` as these are unexpected failures that we didn't expect.

> I dont think user input should ever cause the prepro pod to crash.

Yes, if the exception is due to a particular shape of a sequence, we shouldn't crash. However, if something permanently goes wrong in the pod that causes it to error every time irrespective of particular sequences then it's better to crash.

> Could we add a fallback for enrich_with_nextclade that we only run when a batch errors? The fallback would iterate over the entries one at a time so it can pinpoint the problematic one(s) and attach a proper error.
Kind of like covid testing: run the whole batch first, if there is a hit, check individually

Yes, this is a good backstop until we handle errors more gracefully. We'd just still want to send some kind of alert when this fallback path is hit. But it would mean that errors are contained to sequences that trigger it, reducing blast radius Kind of like covid testing: run the whole batch first, if there is a hit, check individually

maverbiest: > Yes, this is a good backstop until we handle errors more gracefully.

I had Claude vibe something up here https://github.com/loculus-project/loculus/pull/7385
It added some things I don't really like but happy to get it into a better state if we like the direction

PR #7382, opened 2026-09-22
fix(website): check for absent sequences on edit page

### LINKED ISSUE #7375: Edit form does not validate sequences are present when required in contrast to revise form
Website counterpart to #7374 - we only hit this in production because neither website nor backend threw an error.

`EditPage.tsx` only validates required sequences are present on the revision branch but not on the edit branch.

## Details

In `submitEditedDataForAccessionVersion`, the revision branch checks the sequence and
aborts with a message:

https://github.com/loculus-project/loculus/blob/2bfdee60594f02702a9fafa52ff7d3240e8aac89/website/src/components/Edit/EditPage.tsx#L147-L154

The edit branch has no equivalent guard and passes the record through unchecked:

https://github.com/loculus-project/loculus/blob/2bfdee60594f02702a9fafa52ff7d3240e8aac89/website/src/components/Edit/EditPage.tsx#L178-L186

`getSequenceRecord` filters out rows whose `value` is `null`, returning `{}` when no row
remains:

https://github.com/loculus-project/loculus/blob/2bfdee60594f02702a9fafa52ff7d3240e8aac89/website/src/components/Edit/EditableSequences.ts#L159-L174

Rows are file-backed, so removing the uploaded file sets the value to `null`:

https://github.com/loculus-project/loculus/blob/2bfdee60594f02702a9fafa52ff7d3240e8aac89/website/src/components/Edit/SequencesForm.tsx#L49

and `update` then drops the row:

https://github.com/loculus-project/loculus/blob/2bfdee60594f02702a9fafa52ff7d3240e8aac89/website/src/components/Edit/EditableSequences.ts#L107-L108

The result is an empty record, submitted without complaint.

## Suggested fix

Apply the same emptiness check to the edit branch, conditional on
`submissionDataTypes.consensusSequences` as the revision branch already is.

maverbiest:
resolves #7375

This adds a similar check that is already in place for revision: if `submissionDataTypes.consensusSequences` is true, there must be sequences or the submission will be refused and a toast with error message is shown.

There is also a new tests that asserts the toast appears if an edit without a sequence is attempted (had to add an import for `ToastContainer` to the spec file for this).

### Screenshot

On the preview edit page, trying to submit after removing the consensus sequence:

grafik

### PR Checklist

🚀 Preview: https://fix-edit-page-requires-se.loculus.org


website/src/components/Edit/EditPage.tsx
```diff
@@ -175,12 +175,20 @@ const InnerEditPage: FC<EditPageProps> = ({
...
                 );
             }
+            const sequenceRecord = editableSequences.getSequenceRecord();
+            if (submissionDataTypes.consensusSequences && Object.keys(sequenceRecord).length === 0) {
+                toast.error('Please enter a sequence.', {
+                    position: 'top-center',
+                    autoClose: false,
+                });
+                return;
+            }
```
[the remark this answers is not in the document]
maverbiest: I checked on the `getSequenceRecord()` because that's what gets passed to the `submitEdit` endpoint.

But I quite like the suggestion to put this helper on `EditableSequences` and check it in both branches

tombch (approved):
Also tested and looks good to me!

website/src/components/Edit/EditPage.tsx
```diff
@@ -175,6 +174,13 @@ const InnerEditPage: FC<EditPageProps> = ({
                     ]),
                 );
             }
+            if (submissionDataTypes.consensusSequences && editableSequences.hasNoSequences()) {
+                toast.error('Please enter a sequence.', {
```
anna-parker: I would write consensus sequence.

I also wonder if the backend error message is a bit clearer e.g. `Submissions for organism {} must contain at least one consensus sequence` or sth to that effect (with the organism name)

anna-parker (approved):
some comments on the error message but otherwise looks good!

website/src/components/Edit/EditPage.tsx
maverbiest: Yeah agree that that's better, I updated the message

anna-parker: nice! feel free to merge :-)

corneliusroemer:
Thanks! Might be worth still adding e2e coverage through an existing integration test - not doing a full new one, just extending an existing edit one. I'm having a look at how to do that.

PR #7384, opened 2026-09-22
fix(website): Center edit page form

### LINKED ISSUE #7376: Single seq edit form expands to full screen width, in contrast to single seq revise form
Single seq edit form is very wide on desktop:

Image

while revise is more reasonably narrow:

Image

We should be consistent - I think the narrower is better.

tombch:
resolves #7376

### Screenshot

Before:

image

After:

image

### PR Checklist


corneliusroemer (approved):
Thanks! Wondering what your opinion is about wide vs narrow - if you think it's better to be consistently wide than narrow I'm happy with either, just the inconsistency is not great, I think.

Maybe this point from Claude worth doing if you think, but also not that important if you'd rather merge:

> The error branch ([version].astro:49-57) still renders full width, so a fetch failure jumps to a different layout than the success case. Worth wrapping it in the same container if you want the page to look stable either way.

tombch:
I agree its good to be consistent, and overall I prefer the narrow view for forms. I also added an error box for the error case.

Before:

image

After:

image

PR #7387, opened 2026-09-22
test(website): integration coverage for the edit-page missing-sequence error

corneliusroemer-agent:
Integration coverage for the toast added in #7382, which is currently covered by a unit test only.

The single-segment edit flow already discards the submitted sequence and adds a new one, so the empty state exists mid-test for free: it now attempts a submit there, asserts the toast, dismisses it and carries on as before. One extra button click, no backend round trip.


maverbiest (approved):
This looks good to me, thanks for adding it!

corneliusroemer:
We generally don't have that much non-happy path coverage for integration tests - worth adding especially when it's as cheap as here, this costs max a few seconds.

PR #7391, opened 2026-09-23
feat(website): format raw reads upload guidance like the sequences and metadata section

theosanderson-agent:
## What

On the submission upload form, the guidance text in the left column of the raw reads / extra files section was one paragraph in `text-sm text-gray-500`. The "Sequences and metadata" section above it uses a different layout: a short, slightly larger lead line, then longer detail text in smaller, lighter type. This PR gives the raw reads section the same layout. **The wording is unchanged.**

Theo: The inconsistency made raw reads loom larger than Sequences and Metadata when if anything the opposite should be true

## Screenshots

"Test organism (with files)" submission page. Before = main.loculus.org, after = this PR's preview.

**Bulk upload, with the Metadata section above for comparison**

| Before | After |
| --- | --- |
| before | after |

**Single-entry form**

| Before | After |
| --- | --- |
| before | after |

🚀 Preview: https://feat-raw-reads-guidance-f.loculus.org


website/src/components/Submission/DataUploadForm.tsx
```diff
@@ -438,11 +438,23 @@ export const ExtraFilesUpload = ({
...
+                                <i>{`${FILES_HEADER_PREFIX}${singleFileCategory.name}`}</i>
+                            ) : (
+                                'corresponding file category'
+                            )}{' '}
+                            column of your metadata.{' '}
+                        </>
+                    )}
                     For more information please refer to the{' '}
-                    <a href={extraFilesUploadDocsUrl} target='_blank' className='text-primary-600 hover:underline'>
+                    <a href={extraFilesUploadDocsUrl} target='_blank' className='text-primary-700 opacity-90'>
```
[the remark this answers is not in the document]
theosanderson-agent: Kept `target='_blank'` on purpose: this link sits next to the raw reads upload, and those uploads can be large and still in progress when someone opens the docs, so leaving the page in the same tab would lose them. Added `rel='noreferrer'` in 88e0033e9.

corneliusroemer (commented):
Consistency improvement is good. While we touch this, it might be good to make the font of the gray text slightly darker to make it more readable. My gut feeling is that contrast is below what's recommended for accessibility.

website/src/components/Submission/DataUploadForm.tsx
```diff
@@ -438,11 +438,28 @@ export const ExtraFilesUpload = ({
                         : 'Extra files'}
                 </h2>
                 <p className='text-gray-500 text-sm'>
-                    {inputMode === 'bulk'
-                        ? `Upload a folder of files or individual files for your sequences. Each file must be referenced by its name in the ${singleFileCategory ? `${FILES_HEADER_PREFIX}${singleFileCategory.name}` : 'corresponding file category'} column of your metadata.`
-                        : 'Upload a folder of files or individual files for this sequence.'}{' '}
+                    Upload a folder of files or individual files for{' '}
+                    {inputMode === 'bulk' ? 'your sequences' : 'this sequence'}.
```
tombch: We could also remove the full stop here for consistency?

tombch:
```suggestion
                    {inputMode === 'bulk' ? 'your sequences' : 'this sequence'}
```

theosanderson:
I made an issue for Cornelius's point. Can someone approve?

(I'm personally undecided on the contrast point and whether we should change it, so don't want to hold up to figure that out)

tombch (approved):

PR #7393, opened 2026-09-23
ci(integration-tests): run Playwright with 3 workers instead of 2, allowlist unused keycloak font 404

corneliusroemer-agent:
Speed up integration tests by using 3 Playwright workers up from 2. On chromium the test step drops from ~11 min to ~7 min, Firefox ~15min to ~14min.

## Measurements

Benchmarked for different numbers of workers:

| workers | run | chromium | firefox | cli |
|---|---|---|---|---|
| 2 | [35846770807](https://github.com/loculus-project/loculus/actions/runs/35846770807) | 656 s | 911 s | 89 s |
| **3** | [35846907451](https://github.com/loculus-project/loculus/actions/runs/35846907451) | **425 s** | 818 s | 90 s |
| 4 | [35847931624](https://github.com/loculus-project/loculus/actions/runs/35847931624) | 441 s | 822 s | 99 s |
| 5 | [35848066764](https://github.com/loculus-project/loculus/actions/runs/35848066764) | 434 s | 889 s | 91 s |

## Why it was lowered to 2, and why that reason no longer applies

#6370 (May 2026) went from 4 to 2 workers because of intermittent `ERR_EMPTY_RESPONSE` / dropped-connection failures, suspected to be caused by an overloaded runner (CPU was near 100% even at 2 workers). Since then those failures have been traced to other causes.

<details>

- In the runs traced in August, `ERR_EMPTY_RESPONSE` and its firefox form `NS_ERROR_NET_RESET` came from the runner **running out of disk** (`94% of 71.6 GiB used`, then kubelet evicting hundreds of pods mid-run), not from CPU load. Those eviction storms were in 20 of 396 failed runs before 2026-07-25 and 0 of 180 after. They stopped around the k3s upgrade, for reasons not established, and a normal run still sits at about 90% disk.
- Most of the remaining recurring flakes were test bugs and have been mostly fixed: the upload wait that counted a `✓` one early (#7267), the stale search results table (#7198), the tooltip hover that landed before hydration (#7323), and the 90 s processing waits that `raw-reads-processing`'s 20m CPU limit in CI had pushed over the edge (the limit is now gone).

None of today's 8 browser jobs, at any worker count, had a pod eviction, a browser crash or an OOM kill.

The one class that might still be load-sensitive is Firefox occasionally dying mid-test ("Target page, context or browser has been closed"). It is rare (n≈2 in recent surveys) and not understood, and more workers means more browser processes on the runner. If that or other load-sensitive failures come back after this change, the fix is either to deflake the specific test or to go back to 2.

</details>

Also ignores a 404 on `resources-common/node_modules/patternfly/dist/fonts/OpenSans-Light-webfont.*` in the chromium console check. Keycloak's login theme requests that font and Keycloak doesn't serve it. It was the only failure in 8 runs at 3 workers (7/8 chromium, 8/8 firefox, 7/8 cli; [5 dispatched rounds](https://github.com/loculus-project/loculus/actions/runs/35861224003) plus the PR's own runs), and an unidentified Keycloak 404 had failed `edit-account.spec.ts:6` at 2 workers before.

If it turns out flakes go up again, we can just lower back down to 2 - it's cheap.

https://claude.ai/code/session_01LkfZS1emsiUHd4P4pK8sra


corneliusroemer:
I'll do a few more runs to see whether it really is stable enoughl.

